PR #11, opened 2016-11-18
Fix the interpolation function for dense output

dawbarton:
The interpolation function for dense output can silently change the input
parameters (using sort) and provides the output in a (potentially) different
order from the input. This commit now returns the output values in the same
order as the input and does not modify the input variables.

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -9,24 +9,21 @@ Get the value at tvals where the solution is known at the
 times ts (sorted), with values timeseries and derivatives ks
 """
 function ode_interpolation(tvals,ts,timeseries,ks,alg,f)
-  if typeof(tvals) <: StepRange
-  else
-    sort!(tvals) # Make sure it's sorted
-  end
+  idx = sortperm(tvals)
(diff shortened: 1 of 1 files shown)


src/dense.jl
```diff
@@ -9,24 +9,21 @@ Get the value at tvals where the solution is known at the
 times ts (sorted), with values timeseries and derivatives ks
```
ChrisRackauckas: Mind changing the docstring as well? LGTM

ChrisRackauckas:
Ehh that's small enough that I'll handle it. Thanks!

ChrisRackauckas:
Actually the docstring probably doesn't need to be changed. Cool!

dawbarton:
Thanks. The commit also means you can use LinSpaces to generate the time points (and avoids the assumption that UnitRanges are always ascending, which they aren't).

ChrisRackauckas:
oh cool. The plot recipe should probably be fixed then. It collects a LinSpace to avoid the previous issue: https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/solutions/solution_interface.jl#L53

PR #40, opened 2017-04-05
WIP: SSPRK

ranocha:
This is a first implementation of the second-order, two-stage SSP Runge-Kutta method. Since this is my first time contributing to OrdinaryDiffEq.jl, I would like to know whether these changes are acceptable. If everything is okay, I will continue implementing other SSP Runge-Kutta methods in the same way.

What tests should I add for the new methods?


ChrisRackauckas:
Looks good!

Create a new file for the SSPRK methods in test/ode (a refactor/renaming needs to happen, but that'll be later). You can follow the template of the Rosenbrock methods and have it run some convergence tests:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/test/ode/ode_rosenbrock_tests.jl#L10

```julia
dts = 1.//2.^(8:-1:4)
testTol = 0.2

prob = prob_ode_linear

sim = test_convergence(dts,prob,SSPRK22())
@test abs(sim.𝒪est[:final]-2) < testTol

prob = prob_ode_2Dlinear

sim = test_convergence(dts,prob,SSPRK22())
@test abs(sim.𝒪est[:final]-2) < testTol
```

You might need to change that `dts` or something to make the convergence plotting look correct. 

I'd also `plot(sim)` to see what the convergence plot looks like and make sure it's a nice straight line. The plot isn't in the tests, but it's good to double check.

You should also put a test in here:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/test/ode/ode_dense_tests.jl#L65

for making sure that the dense output is good. Just follow what's there for `RK4` or the `Midpoint` method. You'll need to adjust the numbers on the test to match how close the fit should be (this is a regression test). Before setting the numbers, make sure you plot it with the interpolation (should be the default `plot(sol)`. Check it by eye: if it's not smooth then there's an issue in the interpolation.

ranocha:
@ChrisRackauckas: Thank you very much for your guidance. I've added some convergence tests for SSPRK22.

test/ode/ode_dense_tests.jl
```diff
@@ -64,6 +64,24 @@ sol2 =solve(prob,Midpoint(),dt=1//2^(4),dense=true)
 
 prob = prob_ode_linear
 
+sol =solve(prob,SSPRK22(),dt=1//2^(2),dense=true)
+
+sol(interpd_1d,0:1//2^(4):1)
+
+sol2 =solve(prob,SSPRK22(),dt=1//2^(4),dense=true)
+
+@test maximum(map((x)->maximum(abs.(x)),sol2[:] - interpd_1d)) < 0.15
```
ChrisRackauckas: That number is really high in comparison to Midpoint and the SSPRK22 on systems. Can you post the plot?

ranocha: How can I post a plot?

ChrisRackauckas: If you drag and drop a file into the textbox then Github handles it.

bifplot

ranocha: And I just realised that there is a typo - it should be `1.5e-2` instead of `0.15`.

ChrisRackauckas: Oh okay, that sounds more reasonable!

ranocha: file:///home/henranoc/figure_1.png

ranocha: figure_1

ChrisRackauckas: It's okay. If it passes with that stricter tolerance then it's probably good. I'll take a look right now to double check.

ChrisRackauckas: Looks good. You'd know if the interpolation was bad because you'd see little bumps where the steps are. With the tolerance low enough, it'll be a regression test which makes sure no change causes that to happen in the future. 1.5e-2 is probably good enough considering what `Midpoint` is at. I'll wait for tests to pass and then merge.

ranocha: Great, thank you. I'll get to some other SSPRK methods.

test/ode/ode_dense_tests.jl
```diff
@@ -64,6 +64,24 @@ sol2 =solve(prob,Midpoint(),dt=1//2^(4),dense=true)
...
+prob = prob_ode_2Dlinear
+
+sol =solve(prob,SSPRK22(),dt=1//2^(2),dense=true)
+
+sol(interpd,0:1//2^(4):1)
+
+sol2 =solve(prob,SSPRK22(),dt=1//2^(4),dense=true)
+
+@test maximum(map((x)->maximum(abs.(x)),sol2[:] - interpd)) < 2.5e-2
+
```
ChrisRackauckas: `prob = prob_ode_linear`

Need to make `RK4` do its 1d check. (I know this test file is really bad 👎 )

ChrisRackauckas: I just remembered I could just edit that now. No need to bother you with that!

ranocha:
I'va added the three-stage, third-order method `SSPRK33`. However, for time integration only (i.e. simple quadrature), it is of fourth order.

ranocha:
I've added a ten-stage, fourth-order method. For now, this will be the last SSPRK method I'll add.

ChrisRackauckas:
Okay cool. It's all looking good so I'll merge when tests pass.

ranocha:
Again, thank you very much for your guidance :)

ChrisRackauckas:
Travis is failing in the changing size test for the 4th order method

https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/218960823#L676

src/caches.jl
```diff
@@ -77,6 +77,86 @@ end
...
+  u₄::uType
+  k::rateType
+  k₄::rateType
+  du::rateType
+  tmp::uType
+  fsalfirst::rateType
+end
+
+u_cache(c::SSPRK104Cache) = ()
+du_cache(c::SSPRK104Cache) = (c.k,c.du,c.fsalfirst)
```
ChrisRackauckas: need to add k4 here

src/caches.jl
```diff
@@ -77,6 +77,86 @@ end
...
+  uprev::uType
+  u₄::uType
+  k::rateType
+  k₄::rateType
+  du::rateType
+  tmp::uType
+  fsalfirst::rateType
+end
+
+u_cache(c::SSPRK104Cache) = ()
```
ChrisRackauckas: Need to add u4 here.

ChrisRackauckas:
Thank you! Let me know if the devdocs should be changed anywhere to make this easier.

ranocha:
Okay, thank you. I didn't know what these functions do, just copy & paste.

ChrisRackauckas:
Those are the iterators so that way the `resize!` function knows what all of the internal cache variables are. Otherwise some of the arrays are for different sizes. This is for quite an edge case where the number of ODEs is changing over time due to events, like adaptive spatial meshes or this kind of thing:

http://docs.juliadiffeq.org/latest/features/callback_functions.html#Example-2:-Growing-Cell-Population-1

It's an edge case, so I just went ahead and took care of it.

ranocha:
Maybe I'll sum up the work and add a pull request to the docs tomorrow.

ChrisRackauckas:
Thanks! No worries. Take your time.

PR #44, opened 2017-05-04
Compat.Iterators needs 0.18 minimum

tkelman:
should have mentioned that bit https://github.com/JuliaLang/Compat.jl/commit/a53f1023da52c5098e5ddb70ef49379fcfee1c87

remove duplicate less strict Compat entry
and import Compat in the test that uses it


ChrisRackauckas:
Thanks!

PR #52, opened 2017-05-17
`ode_interpolant` uses `@evalpoly` in `interpolants.jl`

ranocha:
`@evalpoly` uses Horner's rule to evaluate the polynomials and will be more stable and faster than the direct evaluation of monomials.


ChrisRackauckas:
Beautiful. You cut enough code that the code coverage went noticeably down. Nice work!

ranocha:
Thanks. I'd like to restructure the tests in `test/ode/ode_dense_tests.jl` a bit to make them more readable - at least for me - similarly to the tests of the SSPRK methods there. Would this be okay for you?

ChrisRackauckas:
Yes. The only issue is that there's a lot of regression tests in there, so the tolerances are different for every method. I think putting that in a giant array wouldn't be readable, but maybe there's a better way. Also there's a few oddities thrown in there, like a few tests on Bigs, some extra stuff on Euler, and Vern7 I think has a test that shows that out-of-place functions with matrix `u0` works with interpolation (this test was at first an accident, but is very useful!). So just make sure the weird stuff is kept somehow, and fold the others in a way that's clear to match tolerances to algorithms.

PR #54, opened 2017-05-17
Add first derivatives for interpolation

gabrielgellner:
Give definitions for the dense outputs that have a interpolation coefficient tables.


gabrielgellner:
Ugh ... don't know why I didn't get the updated test code ... Man that is a lot of work to merge :(

PR #61, opened 2017-06-04
Internally broadcast for every method

ChrisRackauckas:
This is a huge change. It's expected that it won't be as efficient until Julia stack-allocates views, but local tests go from 350 seconds to 500 seconds. That seems like quite a bit. I am not sure how much of that is compile-time, but compile-time does seem significantly longer.

@ScottPJones can you help me profile this when you have time?

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.6.0-pre
-DiffEqBase 1.5.0
+DiffEqBase 1.10.0
 Parameters 0.5.0
 ForwardDiff 0.5.0
 GenericSVD 0.0.2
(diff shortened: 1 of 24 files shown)


ChrisRackauckas:
Actually looks like it was 400 seconds to 500 seconds. If half of that is compilation time, then the rest is non-stack-allocated views which would be fine.

ChrisRackauckas:
There is a tiny bit of an extra slowdown, and it looks to be because of an interaction with `@muladd`:

https://github.com/JuliaLang/julia/issues/22217

ChrisRackauckas:
Here's the total damage due to switching to broadcast. Before:

capture2

After:

capture

Most of the performance loss is because `@.` accidentally turns off `@muladd` though. That's mentioned here:

https://github.com/JuliaLang/julia/issues/22217

and will likely be fixed in the move of `@muladd`'s functionality to Base. In addition, non-stack-allocated views are the other part that hurts, since that builds some painful GC usage. However, from the tests like 80% of that regression was due to Julia v0.5 -> v0.6, using a different computer which has not rebuilt the system image, and existed before the broadcast changes. So actually... the amount that is due to us here is minimal and will be automatically fixed in Julia Base in the near future. So while it's a regression, this is a documentation that it's at least not due to us... so yay?

Issue #39, opened 2017-04-05
Strong Stability Presurving RK Methods


ChrisRackauckas:
This was brought up by @ranocha. These methods are good for discretizations of hyperbolic PDEs.

http://epubs.siam.org/doi/abs/10.1137/07070485X
http://www.worldscientific.com/worldscibooks/10.1142/7498
http://www.cfm.brown.edu/people/sg/SSPpage/sspsite/erk_nl.html

The RK methods can be implemented like any other RK method. However, the higher order RK methods can use something called a `limiter`. Since this is defined by the parts of the `ODEProblem`, it can be a toggle in the algorithm. Example:

```julia
SSPRK3(limiter=:highly_conservative)
```

These will need a little bit of extra documentation for that, but that should be fine. The limiter only needs to do things in the `perform_step` method, so nothing about the implementation needs to be specific. The RK method without the limiters should be added first, and tested, and then the limiters should be added.

A side note for making this simpler to work with. The symbol should be saved as a type parameter in the algorithm. A helper function should probably be added in alg_utils.jl that looks kind of like this:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/alg_utils.jl#L22

which is used to grab the symbol from the type information of the symbol. This will allow the branches for/not-for using the limiter to be compile-time information and compile away depending on the algorithm choice. 

From what I can see, there are not specific dense outputs for these, and so a simple Hermite polynomial should be sufficient. These do not have an error estimator from what I can see, which means they may not be able to be adaptive.

ranocha:
I have implemented a first version of the two-stage, second-order SSP Runge-Kutta method in #40. Before I continue implementing methods, I would like to know whether these changes are acceptable.

ChrisRackauckas:
I found these materials on adding dense output:

[...]

PR #65, opened 2017-06-22
Add dense SSP output for `SSPRK22` and `SSPRK33`

ranocha:
This adds dense SSP output for the SSP methods `SSPRK22` and `SSPRK33` as described by Ketcheson, Lóczi, Jangabylova, Kusmanov: Dense output for SSP RK methods (2017).

See #39.

### DIFF
--- a/src/dense/interpolants.jl
+++ b/src/dense/interpolants.jl
@@ -118,30 +118,28 @@ end
 end
 
 """
-Second order strong stability preserving (SSP) interpolant for the
-two stage, second order SSP Runge-Kutta method.
+Second order strong stability preserving (SSP) interpolant.
 
 Ketcheson, Lóczi, Jangabylova, Kusmanov: Dense output for SSP RK methods (2017).
 """
-@inline function ode_interpolant(Θ,dt,y₀,y₁,k,cache::SSPRK22ConstantCache,idxs,T::Type{Val{0}})
-  Θ1 = 1-Θ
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Is there a reason why you didn't add it to the 4th order method as well? Is the 3rd order Hermite a better idea than the 2nd order SSP interpolant? I know there is that order limit so that is the highest SSP there is. I'd pick whatever is useful for now, and then if it's really necessary add a type parameter in the algorithm constructor that can change it. But likely 4th order is just a bad idea for interpolation.

ranocha:
In section 4.4, Ketcheson et al. do not mention SSPRK(10,4). The reason may be that the resulting dense SSP output has a stricter restriction on the time step than the base method. They don't mention this explicitly but I will have a look at it.
If this is correct, we can as well have a higher order Hermite interpolation, since the dense 'SSP' output won't be SSP either.

ranocha:
`SSPRK104` does not fulfil any of the conditions of Theorem 5 of Ketcheson et al. In section 4.4, they mention that they have not been able to construct dense SSP output with SSP coefficients as large as the base method. Thus, it would require deeper investigations or may be impossible to find an acceptable 2nd order SSP dense output for `SSPRK104`.

ChrisRackauckas:
Thanks for the overview. It's good to have it documented. I have started to contact Ketcheson to see if there's stuff about this, along with error estimators for adaptivity. We'll see how it that goes.

PR #71, opened 2017-06-29
Fix for element-wise tolerances

devmotion:
Fixes use of element-wise tolerances with in-place functions, as discussed in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/70. It should cover all solvers in this package.


ChrisRackauckas:
the same change should be done to the adaptive methods in StochasticDiffEq.jl

PR #73, opened 2017-06-30
Add third-order symplectic integrator

YingboMa:


ChrisRackauckas:
I updated it. It needs:

```
RecursiveArrayTools
DiffEqBase
OrdinaryDiffEq
```

master for the convergence tests. I'll get this stuff merged so tests can start passing. It shows that VelocityVerlet converges with Order 2 🥇 , but Ruth3 does not converge. Also, none of them converge in L2, showing that something seems odd with the interpolations.

Can you work on getting Ruth3 to converge?

ChrisRackauckas:
Also for VelocityVerlet, there are two forms described here:

https://en.wikipedia.org/wiki/Verlet_integration#Velocity_Verlet

It does should have to settings: one which doesn't assume `a` only depends on position, and another which does. Or maybe two separate algorithms if the implementation gets nasty, but I think they should probably be together. Can you add the other form?

ChrisRackauckas:
The git history got a little funky and this had tests passing (with some marked `@test_broken` though) so I pulled it in. But Ruth3 and interpolations still need a fix.

PR #74, opened 2017-06-30
Make the output of a solution interpolation be a DiffEqArray

gabrielgellner:


gabrielgellner:
Strange, what does the check failure mean? I am not sure how to read the reports.

ChrisRackauckas:
it's unrelated. Other things need to be tagged right now. Thanks!

PR #76, opened 2017-07-04
Sanitize tstops

devmotion:
Hi!

This is fixes the problem with initial discontinuities occurring after the final time point, as mentioned in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/11. An example would be
```julia
julia> using DelayDiffEq, DiffEqBase, OrdinaryDiffEq, Base.Test
julia> lags = [.2]
julia> f = function (t,u,h,du)
         du[1] = -h(t-.2)[1] + u[1]
       end
julia> h = (t) -> [0.0]
julia> prob = ConstantLagDDEProblem(f,h,[1.0],lags,(0.0,0.8))
julia> alg = MethodOfSteps(Tsit5(),constrained=false)
julia> sol = solve(prob,alg)
```
which yields time steps
```julia
julia> sol.t
16-element Array{Float64,1}:
 0.0     
 0.10002 
 0.153761
 0.18164 
 0.194284
 0.198926
 0.2     
 0.200528
 0.205805
 0.258583
 0.4     
 0.48159 
 0.6     
 0.760701
 0.8     
 1.0  
```

A simpler (and quite absurd) example with ODEs where the current definition of `tstops_internal` leads to problems is shown in the test file.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -68,27 +68,18 @@ function init{algType<:OrdinaryDiffEqAlgorithm,recompile_flag}(
       error("Fixed timestep methods require a choice of dt or choosing the tstops")
   end
 
-  if tspan[1] == tspan[2]
+  if tspan[1] == tspan[end]
     error("Timespan is trivial")
   end
 
-  d_discontinuities_col = collect(d_discontinuities)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Ahh yes, I should've saw this coming. Good catch! 👍 Will merge when tests pass.

ChrisRackauckas:
Looks great! Good work man!

PR #80, opened 2017-07-13
Avoid undefined entries if k is array of arrays

devmotion:
As discussed in https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/15#issuecomment-315132474, this prevents initialization of arrays `k` with undefined elements in the case of not in-place functions with not scalar `u`.

### DIFF
--- a/src/integrators/verner_rk_integrators.jl
+++ b/src/integrators/verner_rk_integrators.jl
@@ -163,9 +163,8 @@ end
   integrator.k = k
 
   # Avoid undefined entries if k is an array of arrays
-  prototype = f(integrator.t, integrator.uprev)
   @inbounds for i in eachindex(integrator.k)
-    integrator.k[i] = zero(prototype)
+    integrator.k[i] = zero(integrator.uprev)./oneunit(integrator.t)
   end
 end
 
@@ -331,9 +330,8 @@ end
   integrator.k = k
 
   # Avoid undefined entries if k is an array of arrays
-  prototype = f(integrator.t, integrator.uprev)
   @inbounds for i in eachindex(integrator.k)
-    integrator.k[i] = zero(prototype)
+    integrator.k[i] = zero(integrator.uprev)./oneunit(integrator.t)
   end
 end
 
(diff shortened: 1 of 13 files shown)


src/integrators/verner_rk_integrators.jl
```diff
@@ -153,6 +161,12 @@ end
   integrator.kshortsize = 10
   k = eltype(integrator.sol.k)(integrator.kshortsize)
   integrator.k = k
+
+  # Avoid undefined entries if k is an array of arrays
+  prototype = f(integrator.t, integrator.uprev)
```
ChrisRackauckas: why is a function evaluation necessary here?

devmotion: Oh I guess I just missed that one could fill `k` with `zero(integrator.sol.k)`.

devmotion: If `eltype(integrator.sol.k)` is e.g. `Array{Float64,1}` we do only know the type of the elements of `k` but not their length. So I thought the safest way would be to use the result of a function evaluation as a prototype since `k` is filled with results of such function evaluations later.

ChrisRackauckas: every value would be a rate and have to match the length of `uprev`. That can be done without a function call.

devmotion: I thought about this as well, but `zeros` always creates an array, whereas `k` might contain numbers, arrays or something completely different.

ChrisRackauckas: what about just doing `uprev./t` then? That's always going to work and has essentially no cost.

devmotion: That would be problematic for `t == 0`.

ChrisRackauckas: the values don't matter, just the types.

devmotion: Now I'm a bit confused - do you mean something like `zero(uprev)./oneunit(t)`?

ChrisRackauckas: that would work too.

PR #84, opened 2017-07-14
Allow arbitrary integrator in definition of savevalues!

devmotion:
This is needed by https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/18.

### DIFF
--- a/src/integrators/integrator_utils.jl
+++ b/src/integrators/integrator_utils.jl
@@ -68,7 +68,7 @@ end
   end
 end
 
-@inline function savevalues!(integrator::ODEIntegrator,force_save=false)
+@inline function savevalues!(integrator,force_save=false)
   while !isempty(integrator.opts.saveat) && integrator.tdir*top(integrator.opts.saveat) <= integrator.tdir*integrator.t # Perform saveat
     integrator.saveiter += 1
     curt = pop!(integrator.opts.saveat)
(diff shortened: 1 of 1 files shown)


devmotion:
Ouch that's embarrasing.... I should have tested it before, but I wrongly assumed that's not necessary since https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/18 passed all tests locally.

devmotion:
I'll close this since I'm convinced that https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/18 (which would need this PR) is not the best solution to the problem that currently `save_idxs`, `saveat`, `save_everystep`, and `save_start` cannot be applied to delay differential equations. Even more since probably lines

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L103

and

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L127

would have to check for `MethodOfSteps` algorithms with underlying composite algorithms as well.

I rather suggest constructing a suitable solution at the end of the delay differential algorithm from the complete solution, similar to the approach in https://github.com/devmotion/DelayDiffEq.jl/tree/final_solution.

ChrisRackauckas:
Thanks. The final solution one does look promising. I haven't been able to review it since my computers are full right now, but I agree that's probably easiest.

PR #124, opened 2017-08-07
Fix DEDataArray

devmotion:
After applying these small changes data array tests pass with the new implementation in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/65.


ChrisRackauckas:
Not waiting for tests because they'll fail for other reasons. Great, thanks for the confirmation.

PR #128, opened 2017-08-08
Update tableaus and calculation of error estimates in low order RK methods

devmotion:
This PR changes the calculation of error estimates in low order RK methods, by introducing new coefficients `btilde` as discussed at https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/93#issue-247283211 and `calculate_residuals(!)` as discussed at https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/107#issuecomment-320476772.

I thought it is better to add these changes step by step, so the low order RK methods will be followed by other methods soon.

## Result of Linear Work Precision Benchmark

master:
figure_1_linear_master

with this PR:
figure_1_linear_next

comparison of all low order RK methods with this PR:
figure_4_linear_next_owrenzen


ChrisRackauckas:
Looks great! I'm surprised it helps the benchmarks that much. Looks like we're back to pre-elementwise tols again.

ChrisRackauckas:
Travis needs the latest DiffEqBase to release. I'll test locally and merge.

ChrisRackauckas:
Good work!

ChrisRackauckas:
BTW the new residual estimators on RK4 are missing from this, but those are currently very odd and alias a bunch. Just something to note though.

devmotion:
Just noticed that tstops tests fail since RK4 is adaptive now by default.

PR #131, opened 2017-08-09
Add forth order improved Runge-Kutta-Nyström method

YingboMa:


ChrisRackauckas:
Still has some merge conflicts.

PR #137, opened 2017-08-09
Update Rosenbrock methods and explicit RK

devmotion:
This updates Rosenbrock methods and explicit RK integrator to the new calculation of error estimates, removes the third argument of `initialize!`, and changes the third argument of `perform_step!` to `repeat_step=false`. Moreover, Rosenbrock methods prevent recalculations if a step is repeated, as discussed at https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/7.

Some plots of the ROBER benchmarks:

master
figure_2_master

PR
figure_1_rosenbrock

master
figure_4_master

PR
figure_4_rosenbrock

master
figure_5_master

PR
figure_5_rosenbrock


src/integrators/rosenbrock_integrators.jl
```diff
@@ -1,335 +1,387 @@
...
-  γ = dt*d
-
   @. linsolve_tmp = fsalfirst + γ*dT
 
+  # Jacobian
   if has_invW(f)
-    f(Val{:invW},t,uprev,γ,W) # W == inverse W
-    A_mul_B!(vectmp,W,linsolve_tmp_vec)
+    # skip calculation of inv(W) if step is repeated
+    repeat_step || f(Val{:invW}, t, uprev, γ, W) # W == inverse W
```
ChrisRackauckas: Same as `!repeat_step &&`? I find this a little non-standard and would like to have it using `&&` instead since that's easier for me to read.

devmotion: Yes, exactly. I'll change it 👍

ChrisRackauckas:
LGTM. Ready to merge when tests pass.

I don't like the mixing of the style changing (adding the spacing, etc.) with the larger PR though. Not that I don't like the changes, but it makes it harder to check the PR for correctness since there's a lot more lines that are changed.

devmotion:
Sorry for that! Usually I try to keep such changes to a minimum since I also don't like the increased number of changed lines it causes. But I needed some structure in the Rosenbrock methods in order to compare both in-place and not in-place methods and to get to know which blocks and statements I have to change for `repeat_step`.

PR #140, opened 2017-08-10
Update composite integrator and Feagin RK, exponential RK, low order RK, and Verner RK methods

devmotion:
This PR updates additional methods.


ChrisRackauckas:
Nice job.

PR #141, opened 2017-08-10
Updates of SDIRK and generic implicit integrators

devmotion:
Again this updates calculation of error estimates and function arguments. However, this PR might need a more careful inspection since I also tried to reduce the number of calculations and therefore I introduced some new variables, especially in TRBDF2. The SDIRK integrators could be further improved, I guess, by saving `κ*tol` instead of both `κ` and `tol` in the cache (at least in `perform_steps!` only the product is needed) and by saving all constants of TRPBDF2 in a tableau. This is not part of this PR yet, but could be made an additional commit. I hope I didn't introduce any floating point issues by these simplifications.

This PR should hopefully also address https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/134, however, I'm not sure whether https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/139 still remains an issue.

I compared these updated methods to the master branch with the ROBER benchmark for high tolerances:

master
figure_1_sdirk_master

PR
figure_1_sdirk_pr

The plots are very similar apart from the Trapezoid curves which I don't understand.

### DIFF
--- a/src/integrators/sdirk_integrators.jl
+++ b/src/integrators/sdirk_integrators.jl
@@ -287,7 +287,7 @@ end
 
   cache.ηold = η
   cache.newton_iters = iter
-  u = u .+ dz
+  u = @. u + (z + dz)
 
   if integrator.opts.adaptive
     if integrator.iter > 2
@@ -430,7 +430,7 @@ end
 
   cache.ηold = η
   cache.newton_iters = iter
-  u .+= dz
+  @. u += z + dz
 
   if integrator.opts.adaptive
     if integrator.iter > 2
@@ -492,12 +492,6 @@ end
   c1 = γ/4
   c2 = a2/2
 
-  # b1 = ω
-  # bhat1 = (1-ω)/3
-  # b2 = ω
-  # bhat2 = (3ω + 1)/3
-  # b3 = d
-  # bhat3 = d/3
   btilde1 = (1-sqrt(2))/3
   btilde2 = 1/3
   btilde3 = (sqrt(2)-2)/3
@@ -644,12 +638,6 @@ end
   c1 = γ/4
   c2 = a2/2
 
-  # b1 = ω
-  # bhat1 = (1-ω)/3
-  # b2 = ω
-  # bhat2 = (3ω + 1)/3
-  # b3 = d
-  # bhat3 = d/3
   btilde1 = (1-sqrt(2))/3
   btilde2 = 1/3
   btilde3 = (sqrt(2)-2)/3
--- a/src/integrators/sdirk_integrators.jl
+++ b/src/integrators/sdirk_integrators.jl
@@ -13,7 +13,7 @@ end
   @unpack t,dt,uprev,u,f = integrator
   @unpack uf,κ,tol = cache
 
-  κtol = κ*tol # save in cache instead of κ and tol?
+  κtol = κ*tol
 
   if integrator.success_iter > 0 && !integrator.u_modified && integrator.alg.extrapolant == :interpolant
     u = current_extrapolant(t+dt,integrator)
@@ -34,6 +34,7 @@ end
 
   z = u .- uprev
   iter = 1
+  u = uprev .+ z
   b = dt.*f(t+dt,u) .- z
   dz = W\b
   ndz = integrator.opts.internalnorm(dz)
@@ -45,8 +46,8 @@ end
   fail_convergence = false
   while (do_newton || iter < integrator.alg.min_newton_iter) && iter < integrator.alg.max_newton_iter
     iter += 1
-    u = u .+ dz
-    b = dt.*f(t+dt,uprev + z) .- z
+    u = uprev .+ z
+    b = dt.*f(t+dt,u) .- z
     dz = W\b
     ndzprev = ndz
     ndz = integrator.opts.internalnorm(dz)
@@ -67,18 +68,23 @@ end
 
   cache.ηold = η
   cache.newton_iters = iter
-  u = u .+ dz
+  u = uprev .+ z
 
   if integrator.opts.adaptive && integrator.success_iter > 0
-    # Use 2rd divided differences a la SPICE and Shampine
+    # local truncation error (LTE) bound by dt^2/2*max|y''(t)|
+    # use 2nd divided difference (DD) a la SPICE and Shampine
     uprev2 = integrator.uprev2
     tprev = integrator.tprev
 
-    dt1 = 1/(dt*(t+dt-tprev))
-    dt2 = 1/((t-tprev)*(t+dt-tprev))
-    dt3 = dt^2/6
+    # TODO: change recursive calculation to better algorithm
+    diff10 = t-tprev
+    diff11 = dt
+    diff20 = t+dt-tprev
+    c = 7/12 # default correction factor in SPICE (LTE overestimated by DD)
+    r = c*dt^2 # by mean value theorem 2nd DD equals y''(s)/2 for some s
 
-    dEst = @. dt3*abs(dt1*(u - uprev) + dt2*(uprev-uprev2))
+    dEst = @. r*abs(((u-uprev)/diff11-(uprev-uprev2)/diff10)/diff20)
+
     atmp = calculate_residuals(dEst, uprev, u, integrator.opts.abstol,
                                integrator.opts.reltol)
     integrator.EEst = integrator.opts.internalnorm(atmp)
@@ -108,7 +114,7 @@ end
   @unpack uf,du1,dz,z,k,J,W,jac_config,κ,tol = cache
   mass_matrix = integrator.sol.prob.mass_matrix
 
-  κtol = κ*tol # save in cache instead of κ and tol?
+  κtol = κ*tol
 
   if integrator.success_iter > 0 && !integrator.u_modified && integrator.alg.extrapolant == :interpolant
     current_extrapolant!(u,t+dt,integrator)
@@ -151,6 +157,7 @@ end
 
   @. z = u - uprev
   iter = 1
+  @. u = uprev + z
   f(t+dt,u,k)
   scale!(k,dt)
   k .-= z
@@ -168,7 +175,7 @@ end
   fail_convergence = false
   while (do_newton || iter < integrator.alg.min_newton_iter) && iter < integrator.alg.max_newton_iter
     iter += 1
-    u .+= dz
+    @. u = uprev + z
     f(t+dt,u,k)
     scale!(k,dt)
     k .-= z
@@ -196,18 +203,23 @@ end
 
   cache.ηold = η
   cache.newton_iters = iter
-  u .+= dz
+  @. u = uprev + z
 
   if integrator.opts.adaptive && integrator.success_iter > 0
-    # Use 2rd divided differences a la SPICE and Shampine
+    # local truncation error (LTE) bound by dt^2/2*max|y''(t)|
+    # use 2nd divided difference (DD) a la SPICE and Shampine
     uprev2 = integrator.uprev2
     tprev = integrator.tprev
 
-    dt1 = 1/(dt*(t+dt-tprev))
-    dt2 = 1/((t-tprev)*(t+dt-tprev))
-    dt3 = dt^2/6
+    # TODO: change recursive calculation to better algorithm
+    diff10 = t-tprev
+    diff11 = dt
+    diff20 = t+dt-tprev
+    c = 7/12 # default correction factor in SPICE (LTE overestimated by DD)
+    r = c*dt^2 # by mean value theorem 2nd DD equals y''(s)/2 for some s
 
-    @. k = dt3*abs(dt1*(u - uprev) + dt2*(uprev-uprev2))
+    @. k = r*abs(((u-uprev)/diff11-(uprev-uprev2)/diff10)/diff20)
+
     # does not work with units - additional unitless array required!
     calculate_residuals!(k, k, uprev, u, integrator.opts.abstol, integrator.opts.reltol)
     integrator.EEst = integrator.opts.internalnorm(k)
@@ -234,7 +246,7 @@ end
   @unpack uf,κ,tol = cache
   dto2 = dt/2
 
-  κtol = κ*tol # save in cache instead of κ and tol?
+  κtol = κ*tol
 
   if integrator.success_iter > 0 && !integrator.u_modified && integrator.alg.extrapolant == :interpolant
     u = current_extrapolant(t+dt,integrator)
@@ -252,8 +264,10 @@ end
     J = ForwardDiff.derivative(uf,uprev)
     W = 1 - dto2*J
   end
+
   z = u .- uprev
   iter = 1
+  u = uprev .+ z
   b = dto2.*f(t+dto2,u) .- z
   dz = W\b
   ndz = integrator.opts.internalnorm(dz)
@@ -265,7 +279,7 @@ end
   fail_convergence = false
   while (do_newton || iter < integrator.alg.min_newton_iter) && iter < integrator.alg.max_newton_iter
     iter += 1
-    u = u .+ dz
+    u = uprev .+ z
     b = dto2.*f(t+dto2,u) .- z
     dz = W\b
     ndzprev = ndz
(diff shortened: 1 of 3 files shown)


src/integrators/sdirk_integrators.jl
```diff
@@ -1,27 +1,29 @@
...
 
-@muladd function perform_step!(integrator,cache::ImplicitEulerConstantCache,f=integrator.f)
-  @unpack t,dt,uprev,u = integrator
-  @unpack uf = cache
-  uf.t = t
+@muladd function perform_step!(integrator, cache::ImplicitEulerConstantCache, repeat_step=false)
+  @unpack t,dt,uprev,u,f = integrator
+  @unpack uf,κ,tol = cache
+
+  κtol = κ*tol # save in cache instead of κ and tol?
```
ChrisRackauckas: it's fine here. No need to over optimize. This allows for making the values adaptive in the future.

src/integrators/sdirk_integrators.jl
```diff
@@ -72,15 +67,21 @@ end
 
   cache.ηold = η
   cache.newton_iters = iter
-  u = uprev + z
+  u = u .+ dz
```
ChrisRackauckas: should be `u = uprev .+ dz`

src/integrators/sdirk_integrators.jl
```diff
@@ -304,10 +287,7 @@ end
 
   cache.ηold = η
   cache.newton_iters = iter
-  u = uprev + 2z
-  integrator.fsallast = f(t+dt,u)
-  integrator.k[1] = integrator.fsalfirst
-  integrator.k[2] = integrator.fsallast
+  u = @. u + (z + dz)
```
ChrisRackauckas: This isn't right. `u = uprev + 2z`

src/integrators/sdirk_integrators.jl
```diff
@@ -405,18 +395,14 @@ end
...
-    do_newton = (η*ndz > κ*tol)
-  else
-    do_newton = true
-  end
+  do_newton = integrator.success_iter == 0 || η*ndz > κtol
 
   fail_convergence = false
   while (do_newton || iter < integrator.alg.min_newton_iter) && iter < integrator.alg.max_newton_iter
     iter += 1
+    u .+= dz
```
ChrisRackauckas: This isn't right. `@. u = uprev + z`

src/integrators/sdirk_integrators.jl
```diff
@@ -230,71 +218,66 @@ end
...
-  else
-    do_newton = true
-  end
+  do_newton = integrator.success_iter == 0 || η*ndz > κtol
 
   fail_convergence = false
   while (do_newton || iter < integrator.alg.min_newton_iter) && iter < integrator.alg.max_newton_iter
     iter += 1
-    b = -z .+ dto2.*f(t+dto2,uprev + z)
+    u = u .+ dz
```
ChrisRackauckas: This isn't right. `u = uprev .+ z`

src/integrators/sdirk_integrators.jl
```diff
@@ -209,19 +196,20 @@ end
 
   cache.ηold = η
   cache.newton_iters = iter
-  @. u = uprev + z
+  u .+= dz
```
ChrisRackauckas: This isn't right.

src/integrators/sdirk_integrators.jl
```diff
@@ -169,18 +161,14 @@ end
...
-    do_newton = (η*ndz > κ*tol)
-  else
-    do_newton = true
-  end
+  do_newton = integrator.success_iter == 0 || η*ndz > κtol
 
   fail_convergence = false
   while (do_newton || iter < integrator.alg.min_newton_iter) && iter < integrator.alg.max_newton_iter
     iter += 1
+    u .+= dz
```
ChrisRackauckas: This isn't right.

src/integrators/sdirk_integrators.jl
```diff
@@ -481,204 +465,211 @@ end
...
   # TODO: Add extrapolant initial guess
-  zprev = dt*integrator.fsalfirst
+  zprev = dt.*integrator.fsalfirst
 
   ##### Solve Trapezoid Step
 
   zᵧ = zprev
   iter = 1
-  uᵧ = @. (uprev + d*zprev) + d*zᵧ
+  uᵧ = @. uprev + γ*zᵧ
```
ChrisRackauckas: Add a comment for this.

src/integrators/sdirk_integrators.jl
devmotion: Because of floating point issues otherwise?

src/integrators/sdirk_integrators.jl
```diff
@@ -481,204 +465,211 @@ end
...
   ################################## Solve BDF2 Step
 
   ### Initial Guess From Shampine
-  z = (1.5 + sqrt(2))*zprev + (2.5 + 2sqrt(2))*zᵧ -
-      (6 + 4.5sqrt(2))*(uᵧ - uprev)
+  z = @. a1*zprev + a2*zᵧ
 
   iter = 1
-  u = @. (uprev + ω*zprev + ω*zᵧ) + d*z
+  u = @. uprev + c1*zprev + c2*zᵧ
```
ChrisRackauckas: Add a comment here.

src/integrators/sdirk_integrators.jl
```diff
@@ -687,11 +678,14 @@ end
...
+  # TODO: Add extrapolant initial guess
   @. zprev = dt*integrator.fsalfirst
+
+  ##### Solve Trapezoid Step
+
   @. zᵧ = zprev
   iter = 1
-
-  @. uᵧ = (uprev + d*zprev) + d*zᵧ
+  @. uᵧ = uprev + γ*zprev
```
ChrisRackauckas: Add a comment here.

src/integrators/sdirk_integrators.jl
```diff
@@ -687,11 +678,14 @@ end
     end
   end
 
+  # TODO: Add extrapolant initial guess
```
ChrisRackauckas: No, this guess is the smoothed aggregation form FSAL, this should not be extrapolated.

src/integrators/sdirk_integrators.jl
```diff
@@ -687,11 +678,14 @@ end
...
   @. zprev = dt*integrator.fsalfirst
+
+  ##### Solve Trapezoid Step
+
   @. zᵧ = zprev
```
ChrisRackauckas: This could possibly be changed to an extrapolation in the future.

src/integrators/sdirk_integrators.jl
ChrisRackauckas: Because this would be wrong if changed to extrapolation.

src/integrators/sdirk_integrators.jl
devmotion: Ah OK, I just copied it from the not in-place method. So I'll remove it in both the in-place and not in-place method?

src/integrators/sdirk_integrators.jl
```diff
@@ -481,204 +465,211 @@ end
...
+    W = 1 - ddt*J
   end
 
   # TODO: Add extrapolant initial guess
-  zprev = dt*integrator.fsalfirst
+  zprev = dt.*integrator.fsalfirst
 
   ##### Solve Trapezoid Step
 
   zᵧ = zprev
```
ChrisRackauckas: As with below, the extrapolant comment should be moved down here. And it shouldn't be assumed this will hold.

src/integrators/sdirk_integrators.jl
```diff
@@ -445,7 +430,7 @@ end
 
   cache.ηold = η
   cache.newton_iters = iter
-  @. u = uprev + 2*z
+  @. u += z + dz
```
ChrisRackauckas: I don't think this is right? `u` is aggregated differently here than in the stages. It's tough to understand if this is correct.

ChrisRackauckas:
In the SDIRK methods there is an aggregation by which the `u`s are defined via the `z`s in a transform. It's done like this so that way they don't incur floating point issues due to successive iterations. Then after the iterations, the recombination of the last `u` is not necessarily the same as the one used in the iterations. That is definitely the case the `Trapezoid`, and probably what caused the issue there. While the other way would give you second order, it would not always work well on stiff equations, which is what you see in the ROBER benchmark (thanks for showing that sanity check).

src/integrators/sdirk_integrators.jl
devmotion: Since mathematically we have `z <- u - uprev` initially, then an update `z <- z + dz`, at the beginning of every iteration `u <- u + dz = u + z`, at the end of every iteration `z <- z + dz`, and hence at this point (since either the loop was completely skipped or left after a full iteration - premature exit of an iteration by failed convergence is handled before) again `u <- u + dz = uprev + z`.

ChrisRackauckas: If floating point works correctly then yes. There's a good chance this adds a few small values in the end of the Newton iteration though if the equation is stiff enough and the values are large enough. In that case, repeat adding of `dz` to `u` will accumulate errors. It's fine to "accumulate errors` onto `z` because that's the iteration variable, so it will iterate more if `z` is sufficiently incorrect. But accumulating into `u` which is then called in `f` can lead to `f` being off, which will cause problems in stiff problems. In the end, this accumulation doesn't make `u` exactly satisfy the tableau which can lead to order loss.

src/integrators/sdirk_integrators.jl
devmotion: Mathematically it should be correct, I guess, but again it is apparently not a good idea because it could accumulate floating point errors.

src/integrators/sdirk_integrators.jl
devmotion: Yes, that makes sense. But we could e.g. still cache the result of `uprev + ω*zprev + ω*zᵧ` and replace the brackets in lines https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/sdirk_integrators.jl#L767 and
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/sdirk_integrators.jl#L792
with it, couldn't we? We could add `d*z` to this result in every iteration and after the loop to obtain the next `u`. But we should not just increase `u` by some `d*Δz` every time because this would accumulate errors. Is that correct?

ChrisRackauckas: Yes, that's actually precisely what Hairer does.

devmotion:
I corrected the update of `u` in these methods, but the strange result for the trapezoidal rule was not affected by this change. However, I figured out that this is due to error estimation with divided differences which I tried to simplify; but apparently I changed it in such a way that it is more vulnerable to floating point issues. It seems divided differences are always vulnerable to roundoff errors when computed in a recursive fashion, so maybe it would be good to improve it e.g. following the notes on page 191 of http://profs.scienze.univr.it/~caliari/pdf/C07.pdf or any other better algorithm?

And I was wondering why we actually divide by 12 in this case - I know the local truncation error can be bound by `h³/12*max |y'''(t)|`, but according to the [mean value theorem for divided differences](https://en.wikipedia.org/wiki/Mean_value_theorem_(divided_differences)) the 3rd divided difference that we calculate already equals `y'''(s)/6` for some `s` in the time interval from `tprev2` to `t+dt`. So why do we not just divide the resulting 3rd divided difference by 2?

ChrisRackauckas:
>And I was wondering why we actually divide by 12 in this case - I know the local truncation error can be bound by h³/12*max |y'''(t)|, but according to the mean value theorem for divided differences the 3rd divided difference that we calculate already equals y'''(s)/6 for some s in the time interval from tprev2 to t+dt. So why do we not just divide the resulting 3rd divided difference by 2?

Yeah, that looks like it's a mistake. See "The Spice Book" for the original reference. I looked again and they used 7/12, but note that the DD can overstate the truncation error by several times.

>It seems divided differences are always vulnerable to roundoff errors when computed in a recursive fashion, so maybe it would be good to improve it e.g. following the notes on page 191 of http://profs.scienze.univr.it/~caliari/pdf/C07.pdf or any other better algorithm?

Yeah, it's a problem. This should be a separate PR though.

devmotion:
I corrected the Newton iterations and added some comments. Moreover, I added a correction factor and set it to the default value `7/12` of the SPICE book. However, in my opinion the calculation of divided differences seems not only vulnerable to rounding errors but also mathematically incorrect in the master branch - according to the definition of divided differences there should only be recursive subtractions and divisions, I guess.

So I changed the calculation of 2nd and 3rd divided differences in both Euler and Trapezoid methods to a hopefully correct recursive calculation, i.e. I mainly replaced some `+` signs with `-` signs compared to the master branch. The convergence tests still pass but this leads to a really bad performance of the Trapezoid methods in the ROBER benchmarks whereas the Euler methods are almost unaffected by this change. I assume this is due to the fact that the new introduced subtractions have a much larger impact on the calculation of higher order divided differences. I tried a lot of things to improve the benchmarks again, but neither changing broadcasts to for loops nor defining `DD30` and `DD31` in the same style with only replaced `-` signs had any influence. However, since I'm far from an expert in both ODE solvers and divided differences maybe I messed up something and the original version was correct?

PR
figure_1_pr_last_default_spice

I also did some tests with changing the correction factor to 1/6 (current master) or 1 (i.e. without any additional scaling):

1/6:
figure_1_pr_1-6

1:
figure_1_pr_1

src/integrators/sdirk_integrators.jl
```diff
@@ -390,11 +407,8 @@ end
   end
 
   @. z = u - uprev
-  iter = 0
-  κ = cache.κ
-  tol = cache.tol
-
-  iter += 1
+  iter = 1
+  @. u = uprev + z
```
ChrisRackauckas: Redundent

src/integrators/sdirk_integrators.jl
```diff
@@ -725,30 +754,36 @@ end
...
 
-  @. u = (uprev + ω*zprev + ω*zᵧ) + d*z
+  #=
+  Original assignments
+    @. uᵧ = (uprev + d*zprev) + d*zᵧ
+    @. z = (1.5 + sqrt(2))*zprev + (2.5 + 2sqrt(2))*zᵧ - (6 + 4.5sqrt(2))*(uᵧ - uprev)
+  with d = (2 - sqrt(2))/2 can be simplified to
+    @. z = -sqrt(2)/2*zprev + (1 + sqrt(2)/2)*zᵧ
+  =#
+  @. z = a1*zprev + a2*zᵧ
```
ChrisRackauckas: You double checked that this did not change the predictions?

src/integrators/sdirk_integrators.jl
devmotion: I thought so in the beginning as well, but then I got confused because of lines such as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/141/files/660aa9c90743d54f2aa6c22468ea561ba3480cf9?diff=split#diff-57e331c13262e0ef1f111ed17cda714fL39 where exactly this recalculation of `u` (which might be different to the old `u` because of rounding errors, I guess?) is done before the function evaluation - but also it was only present in the not in-place method... So in the end I decided to do these recalculations in every method. But if it's not needed, I'll just remove it everywhere.

devmotion: By the way, the sam line is also present in the new SDIRK2 method: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/sdirk_integrators.jl#L855

ChrisRackauckas: Yeah it was a mistake. These implementations haven't been cleaned up yet so there's some details that I've opened issues for to investigate. This is one of those things I noticed that can probably be removed, but didn't because I'd want to regression test when doing so.

ChrisRackauckas:
I really really appreciate you doing this. However, since these algorithms are very new and in the cleaning phase I can't accept a PR without showing no regressions since I'm still only barely sure it all works correctly (for example, I'm pretty sure the extrapolation timing in Trapezoid should be dt/2, which hasn't been done yet). So I think you should split the PR to do all of the aesthetic and broadcasting fixes so that can merge quickly, and keep the divided differences change separate.

The other thing to note is that Trapezoid is in a different scaling for `z` than the other algorithms. It's using the Hairer transform. I realized later I didn't like it, but haven't gotten back around to it. But this may be throwing off the reasoning about the algorithm since it is different than the others.

src/integrators/sdirk_integrators.jl
devmotion: You mean whether the new definition leads to the same `z` as the old one? Or whether the solution stays the same? Is there a reason to prefer the old version to the mathematically equivalent new version if there would be a slight difference (due to floating point issues maybe)?

devmotion:
I can completely understand that, and I'll just transfer all other changes except of the divided differences to a new PR. Nevertheless, I think there might be both a mathematical and a numerical problem with these divided differences, so one should keep it in mind at least and have another look at it when everything else is settled down.

ChrisRackauckas:
>Nevertheless, I think there might be both a mathematical and a numerical problem with these divided differences, so one should keep it in mind at least and have another look at it when everything else is settled down.

Agreed.

ChrisRackauckas:
`sdirk_integrators.jl` changed quite substantially. Do you plan on reviving this? Or splitting the efforts to new PRs?

devmotion:
I'm already working on a new PR without the changes of divided differences, just have to do some final tests and corrections. So I'll close this and will open another PR soon which will already cover the changes in `sdirk_integrators.jl`.

PR #142, opened 2017-08-10
Update high order RK and threaded DP5 method

devmotion:
More updates. I compared the changes to master with the linear work-precision benchmarks:

master
figure_linear_master

PR
figure_linear_high


ChrisRackauckas:
LGTM.

Issue #116, opened 2017-08-06
Strang splitting


jagot:
For composed linear operators of the type `F = A + B + C + ...`, there should, in addition to the normal methods (implicit Euler, Crank–Nicolson, etc), which operate using `F*u` and `F\u`, an operator splitting method which approximate the propagator as `...exp(C/2)exp(B/2)exp(A)exp(B/2)exp(C/2)...` with automatic symmetric placement of the exponentials. My suggestion is that the first operator should be placed inner-most.

ChrisRackauckas:
I'll make the most basic version today which just does two operators to see what's required, and that skeleton would be easy to build off of.

ChrisRackauckas:
The basic version is up and just is fixed at `exp(B/2)exp(A)exp(B/2)`. It shows there's an error in the way the `expm` was setup though, so right now it just grabs the internal arrays. Something better needs to be done there.

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/commit/9ee72aecedb9d27ee73b72183f754dfc378298ac

ChrisRackauckas:
Is this all under the topic of Lie group integrators? If so, there's a gold mine of methods. Hairer III labels them as Magnus, Crouch-Grossman methods, and Munthe-Kaas

https://arxiv.org/pdf/1207.0069.pdf
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.30.7762&rep=rep1&type=pdf

And an adaptive timestepping one:

https://link.springer.com/content/pdf/10.1007%2Fs10483-008-0901-5.pdf

ChrisRackauckas:
Superseded by https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/1063

Issue #115, opened 2017-08-06
Dual linear solvers interface


jagot:
In relation to #90, and through [discussions on Gitter](https://gitter.im/JuliaDiffEq/Lobby?at=5986fad645fc670746fc69a7), there is an idea of creating a dual interface to the solvers for linear problems:

1. For standard methods such as implicit Euler and the trapezoidal rule (Crank–Nicolson), for each `DiffEqOperator` at least a `A_mul_B!` should be provided, and optionally a `\`.

   - For certain types, `\` can be inferred, e.g. for `Tridiagonal`, `lufact!` could be used. If the operator is not time-dependent (except for an overall scalar coefficient), the `lufact!` can be performed ahead of time.
   - For generic matrix-free operators (as well as sparse matrices, etc), a solver such as GMRES could be inferred instead.

2. Alternatively, the user might specify an optimized `expmv!` implementation.

The solution manner should be inferred from the operator, i.e. `has_expmv`, and it should be possible to mix-and-match, i.e. for `u' = F(u,t)u` with `F = A + B`, the user should be able to provide a `expmv!` for `B` and use a Crank–Nicolson propagator for `A`:

```julia
expmv!(tmp1, B, dt/2, u)
expmv!(tmp2, A, dt, tmp1)
expmv!(v, B, dt/2, tmp2)
```

where `expmv!(tmp2, A, dt, tmp1)` resolves through dispatch to a generic exponentiator (probably the user can set a flag in the `DiffEqOperator` that `:CrankNicolson`/`:ImplicitEuler` is to be used).

---

Side note: I am unsure whether the time-step should be absorbed into the operator or not; for Krylov approximations to the exponential, one usually computes the subspace without the time-step, and applies it in the subspace exponential (notation the same as in http://epubs.siam.org/doi/10.1137/0729014):

```julia
exp(τA)v ≈ βVₘₘ_m exp(τH_m)e₁
```

where `H_m` is the approximation of `A` in the `m`th Krylov subspace spanned by the columns of `V_m`.

ChrisRackauckas:
>Side note: I am unsure whether the time-step should be absorbed into the operator or not; for Krylov approximations to the exponential, one usually computes the subspace without the time-step, and applies it in the subspace exponential

That can be the choice of the solver. If the solver does a direct `t*A` it absorbs. But it can write the algorithm other ways of course.

ChrisRackauckas:
[...]

PR #157, opened 2017-08-15
Generalized Strang splitting

jagot:
I made some simple changes to allow for `AffineDiffEqOperator`:s composed of an arbitrary number of `AbstractDiffEqOperator`:s; this implements #116. Per [discussions on Gitter](https://gitter.im/JuliaDiffEq/Lobby?at=5992c74bee5c9a4c5fc0e877), I renamed `StrangSplitting` to `MidpointSplitting`, since that is what we are implementing here; Strang splitting is more general than this. In keeping with this, the RHS is now evaluated at the midpoint of the interval.

In `perform_step!`, I use a small `@swap!` macro, I am unsure if there is a better place to put this.

As of now, only out-of-place `expm` is supported (only suitable for small systems), but I submit this PR to see if I am going in the right direction, before starting on #115.


test/linear_method_tests.jl
```diff
@@ -30,5 +30,37 @@ function (p::typeof(L))(::Type{Val{:analytic}},t,u0)
 end
 
 prob = ODEProblem(L,u0,(0.0,1.0))
-sol = solve(prob,StrangSplitting(),dt=1/10)
+sol = solve(prob,MidpointSplitting(),dt=1/10)
 using Plots; pyplot; plot(sol)
```
ChrisRackauckas: Can't plot in tests because it'll bog down the CI which is why no plotting package is a dependency. This'll need to be commented out.

ChrisRackauckas:
Can you add a convergence test? Example:

```julia
dts = 1./2.^(7:-1:4)
sim  = test_convergence(dts,prob,MidpointSplitting())
@test abs(sim.𝒪est[:l2]-2) < 0.2
```

If the analytic solution is set this'll be all that's needed. You can check it via `plot(sim)`

jagot:
Sorry for the delay, I have implemented a simple convergence test and rebased onto master. The convergence test is trivially generalizable to the complex case, as soon as JuliaDiffEq/DiffEqDevTools.jl#4 is fixed.

src/integrators/linear_integrators.jl
```diff
@@ -118,20 +118,37 @@ function initialize!(integrator,cache::StrangSplittingCache,f=integrator.f)
   f(integrator.t,integrator.uprev,integrator.fsalfirst) # For the interpolation, needs k at the updated point
 end
 
-function perform_step!(integrator,cache::StrangSplittingCache,f=integrator.f)
+macro swap!(x,y)
```
ChrisRackauckas: This should go to utils

ChrisRackauckas:
Thanks! Note that this won't be part of the normal testing until DiffEqOperators.jl is registered, which I hope is sooner rather than later.

PR #158, opened 2017-08-16
[WIP] Add DPRKN8 a.k.a. RKN8(6)9FM algorithm

YingboMa:


jiahao:
Not sure where the references are being stored, but I would recommend referencing the [Corrigendum](https://academic.oup.com/imajna/article/11/2/297/723882/High-Order-Embedded-Runge-Kutta-Nystrom-Formulae) in the code also, since its existence wasn't well known and it was essential to correcting two coefficients

ChrisRackauckas:
Any computation which was used to transform the coefficients or help complete them I like directly referenced in the code. Stiff solvers do a lot of transformations to improve speed, like:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/tableaus/rosenbrock_tableaus.jl#L665
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/tableaus/sdirk_tableaus.jl#L639

So yeah, add a quick comment to the code with exactly what was done. 

Additionally, we will need to gather citations for everything in the citations page. 

http://juliadiffeq.org/citing.html

This needs a huge update because it's missing a lot of methods. Each method and any "implementation helpers" should get a spot on the citations page to make it easy for users to properly cite the algorithms they use. That requires a PR to the website:

https://github.com/JuliaDiffEq/juliadiffeq.github.io

@YingboMa do you think you can do citation PRs for the RKN methods?

YingboMa:
> do you think you can do citation PRs for the RKN methods?

Yes, of course.

PR #160, opened 2017-08-17
Replace Calculus.jl with DiffEqDiffTools functionality for finite differencing

dextorious:
This should only be merged after DiffEqDiffTools gets tagged and @ChrisRackauckas is happy with the changes overall.


src/integrators/rosenbrock_integrators.jl
```diff
@@ -31,7 +31,7 @@ end
       if alg_autodiff(integrator.alg)
         ForwardDiff.derivative!(dT, tf, vec(du2), t)
       else
-        dT = Calculus.finite_difference(tf, t, integrator.alg.diff_type)
+        DiffEqDiffTools.finite_difference!(dT, tf, t, Val{integrator.alg.diff_type}, vec(du2), Val{:DiffEqDerivativeWrapper})
```
ChrisRackauckas: `integrator.alg.diff_type` should already be stored as a `Val`

dextorious: Sorry about that, leftover from testing. Fixed.

src/algorithms.jl
```diff
@@ -121,15 +121,15 @@ struct StrangSplitting <: OrdinaryDiffEqAlgorithm end
 
 struct ImplicitEuler{CS,AD,F,K,T,T2,Controller} <: OrdinaryDiffEqNewtonAdaptiveAlgorithm{CS,AD,Controller}
   linsolve::F
-  diff_type::Symbol
+  diff_type::DataType
```
ChrisRackauckas: Not type-stable.

dextorious: Hopefully fixed in the latest commit via parametrization.

PR #164, opened 2017-08-20
WIP: Update of SDIRK and generic implicit integrators

devmotion:
This is still very much work in progress, however after staring at all these lines for so long I want to show at least some changes and the ROBER benchmarks I did.

First of all, my aim was to improve the code without introducing any regressions - hence so far I did not touch the implementation of divided differences which probably should be done in another PR. I almost succeeded, just `KenCarp4` shows a huge regression; I could not immediately figure out why, at least it caused no errors during benchmarking. Unfortunately I had problems to benchmark algorithms with fixed time steps, i.e. `ImplicitMidpoint` and `SSPSDIRK2`, because of `Larger maxiters is needed` - I tried to increase `maxiters` but it still did not work, so maybe there are additional adjustments (e.g. different `dt`) needed or even a completely different benchmark. Hence I do not know whether these methods show a regression.

Moreover, it seems there are major errors in the current implementation of `Trapezoid` and `SDIRK2` but I haven't tried to fix those yet.

### master

figure_sdirk_master_1-5

figure_sdirk_master_5-8

figure_sdirk_master_9-13

figure_sdirk_master_14-15

### PR

figure_sdirk_pr_1-5

figure_sdirk_pr_5-8

figure_sdirk_pr_9-13

figure_sdirk_pr_14-15

### DIFF
--- a/src/caches/sdirk_caches.jl
+++ b/src/caches/sdirk_caches.jl
@@ -1,4 +1,4 @@
-mutable struct ImplicitEulerCache{uType,rateType,J,JC,UF,uEltypeNoUnits} <: OrdinaryDiffEqMutableCache
+mutable struct ImplicitEulerCache{uType,rateType,uNoUnitsType,J,JC,UF,uEltypeNoUnits} <: OrdinaryDiffEqMutableCache
   u::uType
   uprev::uType
   uprev2::uType
@@ -7,7 +7,9 @@ mutable struct ImplicitEulerCache{uType,rateType,J,JC,UF,uEltypeNoUnits} <: Ordi
   k::rateType
   z::uType
   dz::uType
+  b::uType
   tmp::uType
+  atmp::uNoUnitsType
   J::J
   W::J
   jac_config::JC
@@ -28,9 +30,10 @@ function alg_cache(alg::ImplicitEuler,u,rate_prototype,uEltypeNoUnits,
   J = zeros(uEltypeNoUnits,length(u),length(u)) # uEltype?
   W = similar(J)
   z = similar(u)
-  dz = similar(u); tmp = similar(u)
+  dz = similar(u); tmp = similar(u); b = similar(u)
   fsalfirst = zeros(rate_prototype)
   k = zeros(rate_prototype)
+  atmp = similar(u,uEltypeNoUnits)
   vfr = VectorFReturn(f,size(u))
   uf = UJacobianWrapper(vfr,t)
   if alg_autodiff(alg)
@@ -51,7 +54,7 @@ function alg_cache(alg::ImplicitEuler,u,rate_prototype,uEltypeNoUnits,
   else
     tol = min(0.03,first(reltol)^(0.5))
   end
-  ImplicitEulerCache(u,uprev,uprev2,du1,fsalfirst,k,z,dz,tmp,J,W,jac_config,uf,ηold,κ,tol,10000)
+  ImplicitEulerCache(u,uprev,uprev2,du1,fsalfirst,k,z,dz,b,tmp,atmp,J,W,jac_config,uf,ηold,κ,tol,10000)
 end
 
 mutable struct ImplicitEulerConstantCache{F,uEltypeNoUnits} <: OrdinaryDiffEqConstantCache
@@ -81,22 +84,18 @@ function alg_cache(alg::ImplicitEuler,u,rate_prototype,uEltypeNoUnits,
   ImplicitEulerConstantCache(uf,ηold,κ,tol,100000)
 end
 
-mutable struct ImplicitMidpointConstantCache{F,uEltypeNoUnits,uType,tType} <: OrdinaryDiffEqConstantCache
+mutable struct ImplicitMidpointConstantCache{F,uEltypeNoUnits} <: OrdinaryDiffEqConstantCache
   uf::F
   ηold::uEltypeNoUnits
   κ::uEltypeNoUnits
   tol::uEltypeNoUnits
   newton_iters::Int
-  uprev3::uType
-  tprev2::tType
 end
 
 function alg_cache(alg::ImplicitMidpoint,u,rate_prototype,uEltypeNoUnits,tTypeNoUnits,
                    uprev,uprev2,f,t,dt,reltol,::Type{Val{false}})
   uf = UDerivativeWrapper(f,t)
   ηold = one(uEltypeNoUnits)
-  uprev3 = u
-  tprev2 = t
 
   if alg.κ != nothing
     κ = alg.κ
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
`SSPSDIRK` and `ImplicitMidpoint` are due to https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/147

`Trapezoid` and `SDIRK2` are probably due to my tinkering with stage predictors last night. Must've been a bad choice.

ChrisRackauckas:
Let me get `Trapezoid` and `SDIRK` benchmarking properly and then rebase to that, and we should be good to go.

ChrisRackauckas:
`SDIRK2` was just due to the default stage predictor. I changed the default to `extrapolant=:constant` and it's fine now.

For `Trapezoid`, the problem is deeper. The previous `Trapezoid` was a mixture of `ImplicitMidpoint` and the real `Trapezoid`. The result was something which was bizarrely L-stable, but that meant it wasn't symmetric like it was supposed to be. MATLAB and Shampine note that `ode23t` should fail on very stiff problems, and ROBER is the classic super stiff problem... for the fact that it worked was actually troubling haha. This means it violated its near symplecticity, which was pointed out by a user. The new `ImplicitMidpoint` is actually closer to `Trapezoid` than the new `Trapezoid`. That figures into why the truncation error was so wonky: it was the wrong method!

The new `Trapezoid` is implemented as an ESDIRK. But from Shampine's TRBDF2 paper we know that should actually be unstable on ROBER, so that's probably fine. In fact, the fact that it's broken on ROBER is probably a good sign, because the decreased stability means it's probably not L-stable which is correct (it cannot be L-stable and limit numerical dampening / be mostly symplectic). `ImplicitMidpoint` needs to have this issue as well.

When using them on a nonstiff VanDerPol I get similar results to what I'd expect, so I am going to assume for now those are in the right ballpark. The real test for them would be a symplectic benchmark measuring the energy change and dampening. 

I'll pull these changes in and of course we can always refine them in the future.

src/integrators/sdirk_integrators.jl
```diff
@@ -89,165 +98,167 @@ end
...
-  f(t+dt,u,k)
-  scale!(k,dt)
+  # initial step of Newton iteration
+  iter = 1
+  tstep = t + dt
+  @. u = uprev + z
+  f(tstep,u,k)
   if mass_matrix == I
-    k .-= z
+    @. b = dt*k - z
```
ChrisRackauckas: do we need both `k` and `b`?

src/integrators/sdirk_integrators.jl
```diff
@@ -491,25 +526,28 @@ end
...
+  integrator.fsalfirst = integrator.f(integrator.t, integrator.uprev) # Pre-start fsal
 
   # Avoid undefined entries if k is an array of arrays
   integrator.fsallast = zero(integrator.fsalfirst)
   integrator.k[1] = integrator.fsalfirst
   integrator.k[2] = integrator.fsallast
 end
 
-@muladd function perform_step!(integrator,cache::TrapezoidConstantCache,repeat_step=false)
+@muladd function perform_step!(integrator,cache::TrapezoidConstantCache,f=integrator.f)
```
ChrisRackauckas: this reverted the repeat_step change

src/integrators/sdirk_integrators.jl
```diff
@@ -518,67 +556,75 @@ end
...
-  tol = cache.tol
-
-  iter += 1
-  u = uprev + (zprev + z)/2
-  b = -z .+ dt.*f(t+dt,u)
+  # initial step of Newton iteration
+  iter = 1
+  tstep = t + dt
+  tmp = @. uprev + dto2*integrator.fsalfirst
+  u = @. tmp + 1//2*z
```
ChrisRackauckas: `/2` instead, this doesn't compile as well as I'd hope. At least it used to do some bizzare things.

src/integrators/sdirk_integrators.jl
```diff
@@ -961,155 +997,158 @@ end
...
 
-  @. uᵧ = (uprev + d*zprev) + d*zᵧ
-
   ################################## Solve BDF2 Step
 
   ### Initial Guess From Shampine
-  a1 = (1.5 + sqrt(2)); a2 = (2.5 + 2sqrt(2)); a3 = (6 + 4.5sqrt(2))
-  @. z = a1*zprev + a2*zᵧ - a3*(uᵧ - uprev)
-  iter = 1
+  z = @. α1*zprev + α2*zᵧ
```
ChrisRackauckas: I assume this calculation is recorded in a comment in the tableau?

ChrisRackauckas:
I think most of this is fine. Could we just revert `KenCarp4` and do that one in a different PR? That would be much simpler.

ChrisRackauckas:
`Kverno4` and `KenCarp4` have failures in their convergence tests. Revert those and do those separately if you can't find what's up easily.

src/integrators/sdirk_integrators.jl
devmotion: It's necessary if one wants to support units, since `k` is a rate but `dt*k - z` has the same units as `u`.

src/integrators/sdirk_integrators.jl
devmotion: That's unintended, I'll fix it 👍

src/integrators/sdirk_integrators.jl
devmotion: Yes, I documented it: https://github.com/devmotion/OrdinaryDiffEq.jl/blob/a0150ee09d8581302ad89343d9750f0e59965693/src/tableaus/sdirk_tableaus.jl#L26-L33

src/caches/sdirk_caches.jl
```diff
@@ -611,7 +603,7 @@ mutable struct Kvaerno3Cache{uType,rateType,J,JC,UF,uEltypeNoUnits,tType,Tab} <:
   tab::Tab
 end
 
-u_cache(c::Kvaerno3Cache)    = (c.uprev2,c.zᵧ,c.z,c.Δzᵧ,c.Δz)
+u_cache(c::Kvaerno3Cache)    = (c.z₁,c.z₂,c.z₃,c.z₄,c.dz)
```
devmotion: Is this (and similar other definitions of `u_cache`) correct?

src/integrators/sdirk_integrators.jl
ChrisRackauckas: I see. Okay, let's go for it.

src/caches/sdirk_caches.jl
ChrisRackauckas: no, but that needs a lot more to work. It'll have to be another PR which extends the cache tests: these are very difficult on the stiff methods.

devmotion:
I finally could fix the convergence errors:
figure_sdirk_pr_9-13

ChrisRackauckas:
cool, merging when tests pass.

PR #166, opened 2017-08-22
WIP: propagate! interface for linear operators

jagot:
This depends on JuliaDiffEq/DiffEqBase.jl#69 and JuliaDiffEq/DiffEqOperators.jl#33.

I've rewritten the midpoint splitting.

Should some kind of testing be implemented? Which/how?

### DIFF
--- a/src/integrators/linear_integrators.jl
+++ b/src/integrators/linear_integrators.jl
@@ -131,17 +131,29 @@ function perform_step!(integrator,cache::MidpointSplittingCache,f=integrator.f)
 
   copy!(tmp, uprev)
   for B in reverse(Bs)
-    u .= expm((dt/2)*B)*tmp
+    propagate!(u, B, dt/2, tmp)
     @swap!(tmp,u)
   end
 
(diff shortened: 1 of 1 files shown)


src/integrators/linear_integrators.jl
```diff
@@ -131,17 +131,29 @@ function perform_step!(integrator,cache::MidpointSplittingCache,f=integrator.f)
...
-    tmp .= expm((dt/2)*B)*u
+    propagate!(tmp, B, dt/2, u)
     @swap!(u,tmp)
   end
 
   f(t+dt,u,integrator.fsallast)
   @pack integrator = t,dt,u
 end
+
+function propagate!(v, L, t, u)
```
ChrisRackauckas: This should probably be part of the Operator's interface, not internal to the solver. So it should probably be in DiffEqBase and get tested there, and the solvers can just import the function and assume it works on whatever the given operator is.

ChrisRackauckas:
We're building this into the operators instead.

PR #169, opened 2017-08-24
ssp_coefficient

ranocha:
SSP coefficients as algorithm traits


ChrisRackauckas:
For reference this is missing `ImplicitEuler` and `SSPSDIRK2`

PR #170, opened 2017-08-25
Pull request/95457c59

ranocha:
Added `stage_limiter!` and `step_limiter!` for explicit SSP methods.
Added an adaptive `SSPRK932` method.

After adding the limiters, I had to increase the tolerance for `SSPRK432` in `ode_dense_tests.jl`. Although the new tolerances still seem to be reasonable, I have no clue why I had to increase them. Maybe some floating point rearrangements due to the limiters? However, these limiters default to `nothing`, so that should be no problem...


ChrisRackauckas:
>After adding the limiters, I had to increase the tolerance for SSPRK432 in ode_dense_tests.jl. Although the new tolerances still seem to be reasonable, I have no clue why I had to increase them. Maybe some floating point rearrangements due to the limiters? However, these limiters default to nothing, so that should be no problem...

I've seen just adding parenthesis to the error estimator cause differences of that scale. So yeah, the compiler optimizing differently can do that. But the regression test is still the same order of magnitude plus convergence tests pass so it's probably fine.

PR #177, opened 2017-08-28
Added some SSPRK methods.

ranocha:
The memory consumption of `SSPRK22`, `SSPRK33`, `SSPRK432`, and `SSPRK932` could be reduced by removing `tmp::uType` in the corresponding caches. However, this temporary storage is used in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/22b52572ebc3d03d2d42685c9cdd340a0d704511/src/callbacks.jl#L62. Thus, it might be useful to use another temporary storage for the callbacks, since memory consumption can be problematic for big problems / on the GPU.


ChrisRackauckas:
Yes, `tmp` is directly used in the callbacks. It might be better to have a `temporary_array(cache)` return the appropriate field from the cache, but until we get around to doing an overhaul like that it'll need to have a `tmp`. For now you can make `tmp` just alias something else though, unless you're willing to make the full cache change.

ranocha:
Is there some assumption on `cache.u` or could I simply use `cache.u` as temporary storage for the callbacks?

ChrisRackauckas:
>Is there some assumption on cache.u or could I simply use cache.u as temporary storage for the callbacks?

`u` is always special since that's the value that's saved and then made into `uprev`. 

We might need to pass enough so you can check if callbacks exist and choose to make a new storage based on that?

ranocha:
That might probably be a solution. However, since the whole framework is not optimised with respect to storage requirements (and I didn't have any problems up to now...), we can just leave it as it is.

ChrisRackauckas:
>However, since the whole framework is not optimised with respect to storage requirements (and I didn't have any problems up to now...)

Yes, this is a larger issue. Quite RAM I've just been using storage in order to make sure everything is pre-allocated, and not making sure it's minimal. I have opened up issues for lots of places that can be improved (Runge-Kutta methods can get a big improvement through aliasing), but it hasn't been something I've gotten back to since, at least for what I do, RAM is so cheap. But yes, we should clearly document where it needs to be decreased, especially for "optimal storage" type methods, and weed this out. Another place that should be "optional" is saving the prototypes: `rate_prototype` etc. Those are probably unnecessary after they are used to generate the cache, but I think might get saved to the integrator.

ChrisRackauckas:
Please feel free to push the memory use issue by opening and issue and summarizing a few things. It's probably time to fix it.

PR #180, opened 2017-08-28
Improve save_idxs

devmotion:
When I wrote a notebook of benchmarks for DelayDiffEq, I discovered that `save_idxs` does not seem to work completely correct. First I thought this is a problem in DelayDiffEq (and there might still be issues there), but it is at least partly also a problem in OrdinaryDiffEq: Using `save_idxs` such that `typeof(u[save_idxs]) != typeof(u)` does not work. So e.g. it is not possible to extract single components of the state vector with scalars, and maybe even worse it is not possible to extract a vector or vectorized component of a matrix. When I had fixed this issue and finally could extract arbitrary components I noticed another issue: it is not possible to plot solutions retrieved by scalar indexing since the interpolation of exactly that case is broken. Hence this PR additionally includes an improved interpolation that covers this case. Use cases and examples are given in the added test file.

### DIFF
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -322,21 +322,20 @@ function ode_addsteps!{calcVal,calcVal2,calcVal3}(k,t,uprev,u,dt,f,cache,always_
 end
 
 @inline function ode_interpolant{TI}(Θ,dt,y₀,y₁,k,cache::OrdinaryDiffEqMutableCache,idxs,T::Type{Val{TI}})
-  # determine output type
-  # required for calculation of time derivatives with autodifferentiation
-  oneunit_Θ = oneunit(Θ)
-  S = promote_type(typeof(oneunit_Θ * oneunit(eltype(y₀))), # Θ*y₀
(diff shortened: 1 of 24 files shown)


src/dense/generic_dense.jl
```diff
@@ -107,8 +107,8 @@ function ode_interpolation(tvals,id,idxs,deriv)
   tdir*tvals[idx[1]] < tdir*ts[1] && error("Solution interpolation cannot extrapolate before the first timepoint. Either start solving earlier or use the local extrapolation from the integrator interface.")
   if typeof(idxs) <: Number
     vals = Vector{eltype(first(timeseries))}(length(tvals))
-  elseif typeof(idxs) <: AbstractVector
-     vals = Vector{Vector{eltype(first(timeseries))}}(length(tvals))
+  elseif typeof(idxs) <: AbstractArray
```
ChrisRackauckas: ...! Thanks. For some reason I couldn't find out how to do this before haha

ChrisRackauckas:
Thanks. I actually just couldn't find out how to do this before.

Would you mind making similar changes to StochasticDiffEq.jl? It should be incorrect in the same ways since the code is so similar.

PR #184, opened 2017-09-01
Reduce cache size for RKN methods

YingboMa:


src/caches/rkn_caches.jl
```diff
@@ -120,24 +117,23 @@ du_cache(c::Nystrom5VelocityIndependentCache) = (c.fsalfirst,c.k₂,c.k₃,c.k
 
 function alg_cache(alg::Nystrom5VelocityIndependent,u,rate_prototype,uEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,::Type{Val{true}})
   k₁ = zeros(rate_prototype)
-  k₂ = zeros(rate_prototype)
-  k₃ = zeros(rate_prototype)
-  k₄ = zeros(rate_prototype)
+  k₂, k₃ = zeros(rate_prototype).x
```
ChrisRackauckas: Might as well get the units right on these. They will all be the same as `zeros(rate_prototype.x[2])`. If you make them all like you did, then some of them will have velocity units and some will have acceleration units.

ChrisRackauckas:
Merge when tests pass.

PR #186, opened 2017-09-06
Updates for DDEs

devmotion:
This adds support of `AbstractContinuousCallback`, a keyword `reduce_size` to `savevalues`, as proposed in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/34, a method `handle_discontinuities!` to avoid a complete copy of `apply_step!` in DelayDiffEq, support for (almost) arbitrary discontinuity types, and the possibility to skip evaluation and application of discrete callbacks by setting `force_stepfail` in continuous callbacks.

Both https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/34 and https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/35 depend on this PR.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.6.0
-DiffEqBase 1.20.0
+DiffEqBase 1.23.0
 Parameters 0.5.0
 ForwardDiff 0.5.0
 GenericSVD 0.0.2
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -1,12 +1,12 @@
 # Use Recursion to find the first callback for type-stability
 
 # Base Case: Only one callback
-function find_first_continuous_callback(integrator,callback::ContinuousCallback)
+function find_first_continuous_callback(integrator, callback::AbstractContinuousCallback)
   (find_callback_time(integrator,callback)...,1,1)
 end
 
 # Starting Case: Compute on the first callback
-function find_first_continuous_callback(integrator,callback::ContinuousCallback,args...)
+function find_first_continuous_callback(integrator, callback::AbstractContinuousCallback, args...)
   find_first_continuous_callback(integrator,find_callback_time(integrator,callback)...,1,1,args...)
 end
 
--- a/src/integrators/integrator_utils.jl
+++ b/src/integrators/integrator_utils.jl
@@ -347,7 +347,7 @@ function apply_step!(integrator)
 
   # Update fsal if needed
   if !isempty(integrator.opts.d_discontinuities) &&
-      typeof(integrator.t)(top(integrator.opts.d_discontinuities)) == integrator.t
+      top(integrator.opts.d_discontinuities) == integrator.t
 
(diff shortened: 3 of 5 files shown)


ChrisRackauckas:
?

ChrisRackauckas:
oh haha

devmotion:
I wanted to rerun tests against the new `DiffEqBase` version.

ChrisRackauckas:
It's not clear to me what's causing the issue, but that does sound like a big issue. Is it stepping more? That could mean the FSAL resets might be off.

devmotion:
I checked it and the underlying cause is the changed type of `d_discontinuities` in `DEOptions`. Changing it back to `tstopsType` solves the issues on my computer, everything else (also FSAL updates) are fine. I assume the type can not be inferred? Since a main point of this PR is to allow other types of `d_discontinuities` do you know a way around this problem?

devmotion:
Providing types of parameters when constructing `opts` improved times and allocations of tests on my computer dramatically. It was even better than on current master.

ChrisRackauckas:
That's likely another Julia bug related to https://github.com/JuliaLang/julia/issues/22370 . Is there anything with this case that you noticed may be the exact cause of the issue? I thought before that it was 16 parameters but I couldn't make an MWE. But yes, there is some kind of inference related bug when building large types that doesn't have an issue that is on display here, and I'm not quite sure what it is... and the fix is to explicitly parameterize it.

devmotion:
I'm not sure what the exact issue is... And it seems Rosenbrock tests, mass matrix tests, and dense tests still show a huge regression. The regression also persists after changing `apply_step!` (and hence FSAL updates) back to master. So far I have no idea why there's still a regression.

devmotion:
I figured out what's causing the issues - it's the additional `tType` in line
https://github.com/devmotion/OrdinaryDiffEq.jl/blob/102a8a775bd646613b0e45b8aa6a857f08050f4a/src/solve.jl#L74
But I don't understand why it is a problem. Nevertheless, I'm thinking about getting rid of all type conversions for discontinuities, in this line and also in https://github.com/devmotion/OrdinaryDiffEq.jl/blob/102a8a775bd646613b0e45b8aa6a857f08050f4a/src/integrators/integrator_utils.jl#L350 by defining `Discontinuity <: Number` with appropriate conversion and promotion rules.

devmotion:
I could make discontinuities work in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/35 without an additional dependency on `Unitful`, so it seems we can indeed get rid of the conversions completely. Can we merge this PR?

ChrisRackauckas:
All of the regressions are gone?

devmotion:
Yes, it's even an improvement to the current master. Times of tests are approximately the same, whereas memory and allocations are reduced by a factor of 2-10, depending on the test.

ChrisRackauckas:
Great. I approved earlier and took a look right now and it LGTM. Merge if you think it's ready.

PR #189, opened 2017-09-09
Fix eps() for types with units

devmotion:
This PR ensures that `eps` is called with unitless types of `t` and `u` since it is not defined for types with units. However, because of use of `eps` in SDIRK methods these methods will fail if unitless type of `t` or `u` are not subtypes of `AbstractFloat`. So maybe it would be better to completely get rid of all `eps` calls there and define minimal values before hand, similarly to `dtmin` which is either set to `10eps` or `1//(10)^10`.

Moreover, I discovered that `dtacc` should be of type `tType` instead of type `tTypeNoUnits` because of lines
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L226-L232
However, this change would break `DelayDiffEq` since it just copies `dtacc`, hence it is not included here.

### DIFF
--- a/src/integrators/sdirk_integrators.jl
+++ b/src/integrators/sdirk_integrators.jl
@@ -42,7 +42,7 @@ end
   ndz = integrator.opts.internalnorm(dz)
   z = z .+ dz
 
-  η = max(cache.ηold,eps(first(u)))^(0.8)
+  η = max(cache.ηold,eps(first(integrator.opts.reltol)))^(0.8)
   do_newton = integrator.success_iter == 0 || η*ndz > κtol
 
   # Newton iteration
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
>Moreover, I discovered that dtacc should be of type tType instead of type tTypeNoUnits because of lines
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L226-L232
However, this change would break DelayDiffEq since it just copies dtacc, hence it is not included here.

could you make an issue for this?

src/integrators/sdirk_integrators.jl
```diff
@@ -42,7 +42,7 @@ end
   ndz = integrator.opts.internalnorm(dz)
   z = z .+ dz
 
-  η = max(cache.ηold,eps(first(u)))^(0.8)
+  η = max(cache.ηold,eps(first(integrator.opts.reltol)))^(0.8)
```
ChrisRackauckas: why this change? I'm not sure this makes sense?

ChrisRackauckas: Oh I see, nevermind.

devmotion: `u` is a number or an array of some type which might contain units, `reltol` however is a number or an array of the unitless type of elements in `u`. So `first(u)` can fail, but `first(integrator.opts.reltol)` won't (or otherwise also the calculation of residuals will fail) and it will still return the same unitless value for unitless types of `u`. Moreover, at first glance, I couldn't see another field with a unitless type of `u`.

PR #195, opened 2017-09-12
Add Embedded Runge–Kutta–Nyström 5(4)M algorithm

YingboMa:
Here is the convergence plot of `ERKN5` algorithm.

newplot

`ERKN5` has the same convergence rate with `ERKN4`, which is a little bit weird. It uses less steps than `ERKN4` at high accuracy though.

```julia
julia> sim = test_convergence(dts,prob_big,ERKN5(),dense_errors=true);

julia> @test sim.𝒪est[:L2] ≈ 4 rtol = 1e-1
Test Passed


julia> @test sim.𝒪est[:l2] ≈ 5 rtol = 1e-1
Test Passed

julia> sol = solve(prob, ERKN4(),reltol=1e-8); length(sol.u)
37

julia> sol = solve(prob, ERKN5(),reltol=1e-8); length(sol.u)
28
```


YingboMa:
The coefficients of `ERKN5` embedded pair's velocity calculation can be reduced to zeros. I am skeptical about that.

```julia
julia> sol = solve(prob, ERKN4(),reltol=1e-8, dense_errors=true); length(sol.u)
37

julia> sol.errors
Dict{Symbol,Float64} with 5 entries:
  :l∞    => 5.9831e-8
  :L2    => 8.65193e-7
  :final => 3.97553e-8
  :l2    => 2.34511e-8
  :L∞    => 3.84904e-6

julia> sol = solve(prob, ERKN5(),reltol=1e-8, dense_errors=true); length(sol.u)
28

julia> sol.errors
Dict{Symbol,Float64} with 5 entries:
  :l∞    => 8.01741e-7
  :L2    => 5.17624e-6
  :final => 4.67161e-7
  :l2    => 3.71865e-7
  :L∞    => 3.40637e-5
```

ChrisRackauckas:
These results looks fine.

>The coefficients of ERKN5 embedded pair's velocity calculation can be reduced to zeros.

What do you mean?

YingboMa:
`bptilde*` have the same value with `bp*`.

ChrisRackauckas:
It still seems to track the error well enough. I don't think that's much of an issue though since the velocity updates are linear so that will have less error anyways. it's probably a small algorithm quirk without much of a practical effect.

ChrisRackauckas:
One of the rationals overflowed on the 32-bit tests. You need to `Int64` something.

ChrisRackauckas:
Thanks. Can you add the docs and the citation?

YingboMa:
Yeah, will do.

Issue #201, opened 2017-10-01
Events tests for not-in-place methods


ChrisRackauckas:
@ikirill's 3 issues (#199, #198, and #197) all deal with events for not-in-place methods, which is currently untested. As you can see here:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/test/ode/ode_event_tests.jl#L21

only one method is tested, and on only one problem. This should get expanded to for sure use the "special dense" algorithms as well to set the `add_steps!` function for #197. A Hermite interpolation like `BS3()` should get added for #198, and an extra test with a `terminate!` call should be added for #199. Of course, these other issues should then be addressed as well. 

This is a good beginner issue with the implementation for each of the fixes sketched out in the issues, so this is being labelled for Hacktoberfest. If you have more questions, please feel free to ask.

ChrisRackauckas:
@ranocha 's updates seem to have handled this.

Issue #200, opened 2017-09-28
Update for ForwardDiff v0.6.0


ranocha:
Updating to ForwardDiff.jl `v0.6.0` crashes many routines, e.g. 140 errors in "Dense Tests". Version `v0.5.0` seems to be fine.

A temporary workaround may be to set an upper version number and update to the new version of ForwardDiff later.

ChrisRackauckas:
Were upper bounds added to METADATA for this @jrevels ? I assume that JuliaCIBot would've found them necessary?

ranocha:
The adaption of DiffEqCallbacks.jl was relatively simple, only some lines had to be adapted, cf. https://github.com/JuliaDiffEq/DiffEqCallbacks.jl/pull/18/files#diff-458d65710a0dbc4df9bccbf98ac9f198. However, doing the same changes in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/f1a265b042daa8bbd867b4ecbe54c7f12f59944e/src/misc_utils.jl#L30
solves some problems but results in 
```
ERROR: LoadError: MethodError: Cannot `convert` an object of type ForwardDiff.Dual{ForwardDiff.Tag{DataType,Float64},Float64,1} to an object of type Float64
This may have arisen from a call to the constructor Float64(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] convert(::Type{ForwardDiff.Dual{ForwardDiff.Tag{OrdinaryDiffEq.##1#4{OrdinaryDiffEq.ImplicitRHS{#time_derivative,Array{Float64,1},Float64,OrdinaryDiffEq.DiffCache{Array{Float64,1},Array{ForwardDiff.Dual{ForwardDiff.Tag{DataType,Float64},Float64,1},1}}}},Float64},Float64,1}}, ::ForwardDiff.Dual{ForwardDiff.Tag{OrdinaryDiffEq.##1#4{OrdinaryDiffEq.ImplicitRHS{#time_derivative,Array{Float64,1},Float64,OrdinaryDiffEq.DiffCache{Array{Float64,1},Array{ForwardDiff.Dual{ForwardDiff.Tag{DataType,Float64},Float64,1},1}}}},Float64},ForwardDiff.Dual{ForwardDiff.Tag{DataType,Float64},Float64,1},1}) at /home/henranoc/.julia/v0.6/ForwardDiff/src/dual.jl:322
[...]

Issue #199, opened 2017-09-22
Calling terminate! in a callback produces an incorrect answer in sol with out-of-place problems


ikirill:
```
function test1()
    f(t, u) = [u[1]]
    cb = ContinuousCallback((t,x,_int) -> (@show t, x[1]; x[1] - exp(1)), int -> (println("Stop!"); DE.terminate!(int)))
    prob = ODEProblem(f, [1.0], (0.0, 2.0), callback=cb)
    sol = DE.solve(prob, DE.DP8())
    @printf "Error: %.2e\n" abs(sol[1,end]/exp(1) - 1)
    sol
end
```

Output:

```
...
omitted
...

(t, x[1]) = (0.9999934025713091, 2.718281828459046)
(t, x[1]) = (0.9999934025713086, 2.718281828459045)
Stop!
Error: 5.57e-01
Timeseries Solution with uType Array{Float64,1} and tType Float64
retcode: Success
Interpolation: specialized 7th order interpolation
t:
4-element Array{Float64,1}:
 0.0     
 0.237167
 0.999993
 0.999993
u:
4-element Array{Array{Float64,1},1}:
 [1.0]    
 [1.26765]
 [4.23201]
 [4.23201]
```

So it stops integration correctly, but probably the number that goes into sol is after the full timestep, not taking into account terminate!, or something like that.

ChrisRackauckas:
Can you check if this is specific to out-of-place functions? I.e., try using

```julia
f(t, u,du) = du.=u
```

ikirill:
Yes, it's specific, that change fixes it.

ChrisRackauckas:
I think the larger metaissue here is that we have really good test coverage for event handling with inplace functions (in fact, in general inplace test coverage is better since I'd assume >95% of actual users will be solving systems of DEs). All of your issues seem to occur with out of place functions + events, which is not well covered. We should not only fix up these issues, but extend the event and cache tests to this case.

ChrisRackauckas:
The relevant callback code is here:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/callbacks.jl#L136

with the way that time is supposed to be "pulled back" being this function:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/callbacks.jl#L138
[...]

Issue #198, opened 2017-09-22
Hermite interpolation with out of place functions


ikirill:
This works with DP8(), but not BS3():

```
function test1()
    f(t, u) = [u[1]]
    cb = ContinuousCallback((t,x,_int) -> x[1] - exp(1), int -> (println("Stop!"); DE.terminate!(int)))
    prob = ODEProblem(f, [1.0], (0.0, 2.0), callback=cb)
    sol = DE.solve(prob, DE.BS3())
    @printf "Error: %.2e\n" abs(sol[1,end]/exp(1) - 1)
    sol
end
```

```
julia> H.test1()
ERROR: type BS3ConstantCache has no field fsalfirst
Stacktrace:
 [1] ode_addsteps!(::Array{Array{Float64,1},1}, ::Float64, ::Array{Float64,1}, ::Array{Float64,1}, ::Float64, ::CHowardControl.#f#20, ::OrdinaryDiffEq.BS3ConstantCache{Float64,Float64}, ::Type{Val{true}}, ::Type{Val{false}}, ::Type{Val{false}}) at /Users/kirill/.julia/v0.6/OrdinaryDiffEq/src/dense/generic_dense.jl:314
 [2] ode_addsteps! at /Users/kirill/.julia/v0.6/OrdinaryDiffEq/src/dense/generic_dense.jl:9 [inlined]
 [3] ode_addsteps! at /Users/kirill/.julia/v0.6/OrdinaryDiffEq/src/dense/generic_dense.jl:8 [inlined]

...
omitted the rest of the stacktrace
```

ChrisRackauckas:
The problem is here:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/dense/generic_dense.jl#L314

Specifically, the problem will occur when a method with Hermite interpolation is used with an out of place function. Somehow we didn't catch that before. A fix would be to just `similar` `u`, but changing the units so that way it's the units of `typeof(oneunit(u)/oneunit(t))`. I'll get to this in a bit.

PR #204, opened 2017-10-08
fix #197: ode_addsteps! for RosenbrockXYConstantCache

### LINKED ISSUE #197: tf not defined in stiff_addsteps
It looks like here: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/f1a265b042daa8bbd867b4ecbe54c7f12f59944e/src/dense/stiff_addsteps.jl#L3 and here: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/f1a265b042daa8bbd867b4ecbe54c7f12f59944e/src/dense/stiff_addsteps.jl#L17 the variable `tf` is not defined. In the other functions, it's defined by `@unpack`, so it seems like a typo. I have no minimal testcase for this, but it got triggered somehow in one of my tests.

ranocha:
I've also prepared some new tests for #198, #199, #201. However, I will not have much time in the next few days. Thus, this may be merged so that I can solve the other problems later.

I've also reverted the changes of 7a3191ea7bfda7f7ec238f4e0021da49faa6354e in order to make the tests pass using ForwardDiff 0.5, see also #200.


ChrisRackauckas:
Thanks!

PR #207, opened 2017-10-11
fix #199

### LINKED ISSUE #199: Calling terminate! in a callback produces an incorrect answer in sol with out-of-place problems

ranocha:
There are some additional tests for events using StaticArrays.jl, needing https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/pull/21. Thus, I've commented them out.


ChrisRackauckas:
Thanks!

PR #208, opened 2017-10-11
SSPRK432, SSPRK932 are adaptive

ranocha:


ChrisRackauckas:
Approved. I can't merge from my phone because of the Mac failure.

PR #213, opened 2017-10-17
allow initial callback changes

ChrisRackauckas:
https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/194


ChrisRackauckas:
To make this not cause a performance hit I changed the default initialization: https://github.com/JuliaDiffEq/DiffEqBase.jl/commit/32d413d24cad12e27d86dd9afbdf7c38c1f22df1

PR #218, opened 2017-11-02
add reinit

ChrisRackauckas:


ChrisRackauckas:
Right now this resets without resetting `tstops` and `saveat`. Is that weird? Probably...?

Datseris:
I do not have any opinion on the weirdness, but if it helps you to know, if I used the `reinit` I would probably use both `tstops` and `saveat` arguments (not together though).

devmotion:
I think it is a bit weird. In my opinion,
```julia
integrator = init(prob,alg,timeseries,ts,ks,recompile;kwargs...)
reinit!(integrator)
solve!(integrator)
```
should give exactly the same solution as
```julia
integrator = init(prob,alg,timeseries,ts,ks,recompile;kwargs...)
solve!(integrator)
```
as far as this is possible (some initial options are probably not cached currently).

Hence I also think it's odd that one has to reset `dt` explicitly. Are all initial assignments in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L271-L287 that are not reset not relevant for `solve!`? And I guess we would need some reinitialization of callbacks to make `reinit!` work if the integrator contains callbacks.

ChrisRackauckas:
>Hence I also think it's odd that one has to reset dt explicitly.

I see what you mean. But in many cases it's not a good idea: you can cut that out and be more efficient. I added an option on this and made `auto_dt_reset!` part of the integrator interface as well so users can trigger that part of the code.

>Are all initial assignments in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L271-L287 that are not reset not relevant for solve!?

Now all of the ones which carry state are reset.

>And I guess we would need some reinitialization of callbacks to make reinit! work if the integrator contains callbacks.

Hmm, this is definitely something you wouldn't want to do in all cases, since in many cases that's for just building some caches. Should there be a toggle on this as well?

ChrisRackauckas:
`tstops`, `saveat`, and `d_discontinuities` can get reset. I'll add that next. Anything else? We'll need to cache the original forms of those though.

Datseris:
I think it is worth considering whether this can be actually fast, because the more and more stuff start to happen with `reinit!`. From my understanding, the only reason to add `reinit!` is for efficiency; you want something faster than creating and initializing a new integrator from scratch.

Reinitializing callbacks I don't think this fits the purpose of `reinit`

ChrisRackauckas:
>Reinitializing callbacks I don't think this fits the purpose of reinit

What's a case where they are a high cost? Usually it's a pass through. Could just be another option.

Datseris:
I guess I misunderstood. 
For some reason I understood that callbacks would be re-initialized on every `reinit!`

ChrisRackauckas:
This reinits it all and has options to turn each on/off. @devmotion satisfied?

devmotion:
LGTM. Maybe we should some additional tests of `reset_dt` and `reset_callbacks`?

ChrisRackauckas:
Added a test for `reset_dt`. `reinit_callbacks` is very simple and just calling the same function that is called in initialization, so I am pretty sure any tests on callback initialization would catch a bug here, and I can't think of an easy enough example right now. So I documented that it should be added (https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/219) but for now don't think that's blocking.

Datseris:
Awesome that this got merged! 
What does `reinit_cache` do?

Currently I am doing:
```
@. integ2.u = integ1.u + (integ2.u - integ1.u)/a
u_modified!(integ2, true)
set_proposed_dt!(integ2, integ1)
```
to "reinit" an integrator to a state given by what you see.

If I want my reinit to be as efficient as possible, should I make *all* keyword arguments be `false`?

ChrisRackauckas:
Uhh, that makes no sense man. A few notes:

1) You'll be surprised how little of your computing is spent re-initing. This is cheaper than a step, and you do far more than 1 step to solve the ODE.
2) `erase_sol` is essentially free since Julia lazy resizes arrays. You use it if you want to erase the current solution, don't use it if you don't want to. 
3) `reinit_callbacks` will only change the performance if you have callbacks with a [special initialization that is very costly](http://docs.juliadiffeq.org/latest/features/callback_functions.html#DiscreteCallback-1). Off the top of my head, only the projection callbacks do something like that, but even then that "super major cost" is just one vector allocation... I think it would be very rare that this would affect performance. You'd more likely turn this off because it could effect correctness. The vast majority of callbacks have no initialization at all, so this is exactly free (look at the default initialize function).
4) `reset_dt` is the only one that really effects performance. Two reasons why. For one, it allocates 3 times because it was made to only be run once. That can be fixed by https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/220, but in the grand scheme of things that's really nothing... What's more important here is the `dt` choice. This algorithm for auto-detection of `dt` is conservative. If your adaptivity has already found a good range of `dt` for your problem, re-doing this `dt` initialization can slow things down by lowering `dt` and forcing more steps. If the problem at the start of the timespan is different than the end, and the start requires smaller `dt`, if you don't manually lower it or let this estimate a lower `dt` then you'll have a few extra step rejections. However, once again, this still only has a relatively minor effect on performance.

Benchmark it and you'll see why worry too much about these details is silly. Of course we'll optimize the last part of the auto `dt` choice in the future, in reality the only thing which has a measurable impact is choosing the right `dt` to start at, and a bad choice can add like 4 steps to the solver (it'll exponentially grow to the right size anyways). 

The reason why resolving was bad before is because the `cache` had to be re-created (that's the big type that holds like 10+ copies of the vector). That is a ton of allocations. This re-init uses the same cache, so the major factor that you measured before is now being avoided (along with a Julia bug on the generation of the integrator, that's avoided since we just use the same integrator).

Datseris:
So, just to be clear, `reinit_cache = true` *does not recreate the cache* and I can "not" care about it?

ChrisRackauckas:
>EDIT: what does reinit_cache = true even does then? When should I care about it?

FSAL algorithms need to pre-calculate the first derivative call in order to be correct. If you make this false, you'll have errors in most cases. This is different than actually re-building the type and allocating the values though. For example, for `Tsit5()` it literally sets a few points (these are cheaper than a single integer addition) and then does one `f` call:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/low_order_rk_integrators.jl#L800

(and without that `f` call, the first step would be wrong anyways).

This option won't actually be documented since the user should always set this to `true` since not using it is wrong. The only reason why this option exists is because the DDE solver needs to skip this and initialize the cache using its own function (it still needs to be done in order to be correct, but in this case it needs to use the proper function with the history enclosed in it!). So just ignore it.

Datseris:
Thanks for the info man. Sorry for all the silly questions but I want to have the best possible integration of DiffEq to DynamicalSystems.jl and this requires some basic knowledge from my part as well.

ChrisRackauckas:
just ping me for review and I'll be glad to look it over.

PR #222, opened 2017-11-06
Propagate initialize_save

devmotion:
In https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/53 I discovered that `initialize_save` is not propagated to `initialize_callbacks!`. I guess we should add tests for `initialize_callbacks!` (and not only for `reinit_callbacks`, as suggested by https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/219); according to https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/194#issuecomment-337942482 there might already exist tests in a private repo?


ChrisRackauckas:
Yes, there's a research project for which I am using this. A good test that would be similar is a linear ODE with a zero initial condition that has a callback give it a new initial condition.

PR #228, opened 2017-11-21
update for difftools 2.0

ChrisRackauckas:


devmotion:
Could the derivation of `jac_config` be moved to a separate function? However, this is no substantial issue IMO.

ChrisRackauckas:
Yeah, that's a good idea.

PR #232, opened 2017-12-22
use false position method

ChrisRackauckas:


ChrisRackauckas:
@isuruf this is having SymEngine build problems?

Issue #219, opened 2017-11-04
Add a test for reinit_callbacks


ChrisRackauckas:

ChrisRackauckas:
A good test that would be similar is a linear ODE with a zero initial condition that has a callback give it a new initial condition. It should both initialize correctly and reinit correctly, and compare to an analytical solution.

rudrasohan:
@ChrisRackauckas  should I be using `cb = ManifoldProjection(f)` for the callback part ?

ChrisRackauckas:
This can use any callback with a non-trivial init. I don't think that holds for ManifoldProjection?

ChrisRackauckas:
Closed by https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/244

PR #244, opened 2018-01-22
Added Callback in tests

rudrasohan:
With reference to the issue #219, I have added the callback feature in the `reinit_test.jl` tests using `ContinuousCallback(condition,affect!)` feature of `DiffEqCallbacks` . Please do inform if further changes are required.

### DIFF
--- a/test/reinit_test.jl
+++ b/test/reinit_test.jl
@@ -58,7 +58,8 @@ solve!(integrator)
 
 u = saved_values.saveval
 t = saved_values.t
-saved_values = SavedValues(Float64, Float64)
+resize!(saved_values.t,0)
+resize!(saved_values.saveval,0)
 reinit!(integrator)
 integrator.dt = dt
 solve!(integrator)
--- a/test/reinit_test.jl
+++ b/test/reinit_test.jl
@@ -1,4 +1,4 @@
-using OrdinaryDiffEq, DiffEqProblemLibrary, Base.Test
+using OrdinaryDiffEq, DiffEqProblemLibrary, DiffEqCallbacks, Base.Test
 
 prob = prob_ode_2Dlinear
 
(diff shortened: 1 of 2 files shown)


test/reinit_test.jl
```diff
@@ -44,3 +44,29 @@ solve!(integrator)
...
+end
+cb = ContinuousCallback(condition,affect!)
+integrator = init(prob,Tsit5();dt=1//2^(4),callback = cb)
+dt = integrator.dt
+solve!(integrator)
+
+u = copy(integrator.sol.u)
+t = copy(integrator.sol.t)
+
+reinit!(integrator)
```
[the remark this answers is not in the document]
rudrasohan: But saving callback can only save the values when callback is called. It cannot change the initial condition. So we need the Continuous Callback for initializing.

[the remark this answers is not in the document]
ChrisRackauckas: `SavingCallback`'s reinit has to reset the `saveat` values, so it will only save at the `saveat` values the second time if `reinit` on a callback works as expected.

test/reinit_test.jl
```diff
@@ -46,27 +46,22 @@ solve!(integrator)
...
 integrator = init(prob,Tsit5();dt=1//2^(4),callback = cb)
 dt = integrator.dt
 solve!(integrator)
 
-u = copy(integrator.sol.u)
-t = copy(integrator.sol.t)
-
+u = saved_values.saveval
+t = saved_values.t
+saved_values = SavedValues(Float64, Float64)
```
rudrasohan: @ChrisRackauckas when I remove this line 61in the `reinit_test.jl` the tests pass. But when this line is present the tests fail. Does the `reinit()` function not calling the callback ?

ChrisRackauckas: You shouldn't `saved_values = SavedValues(Float64, Float64)`. That changes the pointer to the `saved_values` object. Instead, do `resize!(saved_values.t,0)`, etc.

rudrasohan: Thanks a lot. I have done the necessary changes. Now it's working fine.

ChrisRackauckas:
Will merge when CI passes

ChrisRackauckas:
Thanks!

rudrasohan:
@ChrisRackauckas thanks a lot the help.

PR #252, opened 2018-02-07
Add documentation string to ODEIntegrator

Datseris:
This is important for DynamicalSystems.jl due to how I create the docs! :)


ChrisRackauckas:
It's weird to just have partial docs here though. Why not just point to the page, or put the whole thing there? It's weirdly in the middle.

Datseris:
Okay. Can we at least leave `step!(integrator)` and the paragrah that shows the relevant fields there?

Simply for interactiveness purposes.

ChrisRackauckas:
Sure, a relatively short form is fine.

Datseris:
Okay, check if the string is fine, I thing it is quite minimal but still very helpful.

PR #256, opened 2018-02-18
Implement set_{t,u,ut}! for ODEIntegrator

### LINKED ISSUE #253: Support resetting time?
Consider solving ODEs with a periodic external force for long period of time.  If you choose a large `tspan`, `sin(t)` becomes inaccurate at some point.  In this case, if there is a way to reset the time tracked by the integrator without touching its caches like `dt` for adaptive methods, I can just set `tspan` to be the (integer multiple of the) period of the external oscillator and then keep calling `solve!`.  This way, `t` can be kept around 1.0 so that there is no loss in the accuracy.

Note that `eps(1e10) ≈ 1.9e-6` which is comparable to the default `abstol = 1e-6`.  For solving some crazy chaotic system with long transient or convergence properties, I think it is imaginable that some researchers want to simulate for that long amount of time.  Also, sometimes you draw so-called bifurcation diagram by slowly varying parameters.  In such case, the total simulated time can be even longer.

Just to be a bit more precise, what I want to achieve is to do something like

```julia
for _ in 1:N
    reinit!(integrator, integrator.sol[end])
    solve!(integrator)
    ... do something with the result ...
end
```

but without resetting `dt` and other algorithm-related caches (if any).

tkf:
Currently, `set_u!` and `set_ut!` just fallback to `reinit!`.

@ChrisRackauckas I just put the error check you mentioned in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/86#issuecomment-366445767 to `set_t!`.  Probably it's faster if you implement it rather than reviewing this PR. :stuck_out_tongue_winking_eye:

Really closes #253

### DIFF
--- a/src/integrators/integrator_interface.jl
+++ b/src/integrators/integrator_interface.jl
@@ -232,3 +232,35 @@ function DiffEqBase.auto_dt_reset!(integrator::ODEIntegrator)
   integrator.tdir,integrator.opts.dtmax,integrator.opts.abstol,integrator.opts.reltol,
   integrator.opts.internalnorm,integrator.sol.prob,integrator)
 end
+
+function DiffEqBase.set_t!(integrator::ODEIntegrator, t::Real)
+  if alg_extrapolates(integrator.alg) || !isdtchangeable(integrator.alg)
+    reinit!(integrator, integrator.u;
+            t0 = t,
+            reset_dt = false,
+            reinit_callbacks = false,
+            reinit_cache = false)
+  else
+    integrator.t = t
+  end
+end
+
+function DiffEqBase.set_u!(integrator::ODEIntegrator, u)
+  # TODO: define the "fast path"
+  # At the moment, fallback to reinit!:
+  reinit!(integrator, u;
+          t0 = integrator.t,
+          reset_dt = false,
+          reinit_callbacks = false,
+          reinit_cache = false)
+end
+
+function DiffEqBase.set_ut!(integrator::ODEIntegrator, u, t::Real)
+  # TODO: define the "fast path"
+  # At the moment, fallback to reinit!:
+  reinit!(integrator, u;
+          t0 = t,
+          reset_dt = false,
+          reinit_callbacks = false,
+          reinit_cache = false)
+end
--- a/src/integrators/integrator_interface.jl
+++ b/src/integrators/integrator_interface.jl
@@ -234,6 +234,10 @@ function DiffEqBase.auto_dt_reset!(integrator::ODEIntegrator)
 end
 
 function DiffEqBase.set_t!(integrator::ODEIntegrator, t::Real)
+  if integrator.opts.save_everystep
+    error("Integrator time cannot be reset unless it is initialized",
+          " with save_everystep=false")
+  end
   if alg_extrapolates(integrator.alg) || !isdtchangeable(integrator.alg)
     reinit!(integrator, integrator.u;
             t0 = t,
@@ -246,6 +250,10 @@ function DiffEqBase.set_t!(integrator::ODEIntegrator, t::Real)
 end
 
 function DiffEqBase.set_u!(integrator::ODEIntegrator, u)
+  if integrator.opts.save_everystep
+    error("Integrator state cannot be reset unless it is initialized",
+          " with save_everystep=false")
+  end
   # TODO: define the "fast path"
   # At the moment, fallback to reinit!:
   reinit!(integrator, u;
@@ -256,6 +264,10 @@ function DiffEqBase.set_u!(integrator::ODEIntegrator, u)
 end
 
 function DiffEqBase.set_ut!(integrator::ODEIntegrator, u, t::Real)
+  if integrator.opts.save_everystep
+    error("Integrator state cannot be reset unless it is initialized",
+          " with save_everystep=false")
+  end
   # TODO: define the "fast path"
   # At the moment, fallback to reinit!:
   reinit!(integrator, u;
--- a/src/integrators/integrator_interface.jl
+++ b/src/integrators/integrator_interface.jl
@@ -254,25 +254,11 @@ function DiffEqBase.set_u!(integrator::ODEIntegrator, u)
     error("Integrator state cannot be reset unless it is initialized",
           " with save_everystep=false")
   end
-  # TODO: define the "fast path"
-  # At the moment, fallback to reinit!:
-  reinit!(integrator, u;
-          t0 = integrator.t,
-          reset_dt = false,
-          reinit_callbacks = false,
-          reinit_cache = false)
+  integrator.u = u
+  u_modified!(integrator, true)
 end
 
 function DiffEqBase.set_ut!(integrator::ODEIntegrator, u, t::Real)
-  if integrator.opts.save_everystep
-    error("Integrator state cannot be reset unless it is initialized",
-          " with save_everystep=false")
-  end
-  # TODO: define the "fast path"
-  # At the moment, fallback to reinit!:
-  reinit!(integrator, u;
-          t0 = t,
-          reset_dt = false,
-          reinit_callbacks = false,
-          reinit_cache = false)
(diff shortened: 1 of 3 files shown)


test/integrator_interface_tests.jl
```diff
@@ -0,0 +1,61 @@
...
+  solve!(integrator1)
+
+  solve!(integrator2)
+
+  @test integrator1.t == integrator2.t
+  if setter in [set_u!, set_ut!] || alg === Trapezoid
+    rtol = integrator1.opts.reltol
+    atol = integrator1.opts.abstol
+    if alg === Trapezoid
+      rtol *= 100
```
tkf: I had to increase rtol of test with `Trapezoid` by 100.  It sounds awful but I don't have enough understanding of the internal at the moment to debug the source of inaccuracy (or this use of `opts.reltol` is correct).

tkf:
Looks like Travis fails because it is tested against the released version of DiffEqBase.  It would be nice to have a way to test against upstream masters.  Like checking if the PR is against `next` (or `dev` or something) branch then checkout all `*DiffEq*` packages in REQUIRE.  I think it can be done by just checking `$TRAVIS_BRANCH`.

ChrisRackauckas:
Best to just release.

tkf:
Maybe.  But since DiffEqBase does not have much test cases, I thought it'd be useful if you can test it with downstream libraries fist.  Also, if it turns out that the interface was wrong while implementing the downstream, then you have to bump up major version of DiffEqBase.  I thought it makes adding a new interface to DiffEqBase harder. But I know configuring Travis is not super fun...

ChrisRackauckas:
>Also, if it turns out that the interface was wrong while implementing the downstream, then you have to bump up major version of DiffEqBase.

We take that the documented API is the released API, not the code on the master branch. There's some undocumented unreleased code on master branches all the time. We don't expect anyone to use that and if they do it's "at your own risk" until documented. In that sense, it's not breaking to change it. What is required for master branch though is that it should be working (hopefully with tests, though our test coverage could use some work in edge cases).

>But since DiffEqBase does not have much test cases, I thought it'd be useful if you can test it with downstream libraries fist. 
>I thought it makes adding a new interface to DiffEqBase harder

Usually this can be done locally. In practice, DiffEqBase should be seldom changed so it's fine. Fixing Travis here is actually pretty easy since we can just issue a `Pkg.checkout("DiffEqBase")` in the Travis script before testing, or have a separate test run that does that. However, I prefer to just test locally to get the interface good enough, and then always release. If you rely on testing master branches, then your code ends up like Makie where users need master branches and you don't even know it some times. Instead, DiffEq does a lot of releases, which means that users are getting the new code. And again, changes to the actual documented interface are exceedingly rare. If we stick the the documented definition of released, then most of the issues go away (and the undocumented definition is not one I would want to support!)

tkf:
Just to be clear, my suggestion was not to test against DiffEqBase master all the time.  Specifically, I was suggesting something like this (untested):

```yaml
before_script:
  - julia --color=yes -e 'Pkg.clone(pwd())'
  - julia --color=yes -e 'Pkg.build("OrdinaryDiffEq")'
  -  if [ "$TRAVIS_BRANCH" = next ]; then
         julia --color=yes -e 'Pkg.checkout("DiffEqBase")'
     fi

script:
  - julia --color=yes --check-bounds=yes -e 'Pkg.test("OrdinaryDiffEq"; coverage=true)'
```

Then OrdinaryDiffEq is tested against DiffEqBase master only if contributor request PR against `next` branch.  After DiffEqBase is released, `next` can be merged to `master`.

It is a complicated workflow and the local test is the ultimate way to do flexible testing.  Also, you need to handle two branches and constantly merging `master` into `next` if you want to avoid merge hell just before releases...  OK.  I started to realize it imposes more burden on your side.  Initially, I thought it's nice if we can share test result via Travis.

src/integrators/integrator_interface.jl
```diff
@@ -232,3 +232,47 @@ function DiffEqBase.auto_dt_reset!(integrator::ODEIntegrator)
...
+end
+
+function DiffEqBase.set_u!(integrator::ODEIntegrator, u)
+  if integrator.opts.save_everystep
+    error("Integrator state cannot be reset unless it is initialized",
+          " with save_everystep=false")
+  end
+  # TODO: define the "fast path"
+  # At the moment, fallback to reinit!:
+  reinit!(integrator, u;
```
ChrisRackauckas: You should be able to just change `u` here and set do `u_modified(integrator,true)`.

tkf:
Re: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/256#discussion_r168975743
Use `u_modified!` in `set_u!` and now `set_ut!` is just an alias of `set_u!`-then-`set_t!`.

tkf:
Not sure if it helps, but I just plotted the "trivial test" with `Trapezoid` and `set_t!`:

image

This test calls `set_t!` which in turn calls `reinit!` with the current time and state. So I thought it was a no-op. But two solutions diverge...  Well, I guess I don't need a plot to conclude that. :confused: 

(Note that blue line is funny for t < 1.0 but that's just because the cache is missing from `integrator1.sol`, I suppose.  Obviously, actual steps (green hexagons) for t < 1.0 match with the one from `integrator2.sol` (orange circles).)

<details>
<summary>Julia script</summary>
```julia
using DiffEqBase: set_t!, set_u!, set_ut!
using OrdinaryDiffEq

alg = Trapezoid
setter = set_t!

f = (u,p,t) -> (2u)
prob = ODEProblem(f,1/2,(0.0,2.0))
t_half = 1.0

# save_everystep = false
save_everystep = true
integrator1 = init(prob, alg(); save_everystep=save_everystep)
integrator2 = init(prob, alg(); save_everystep=save_everystep)

step!(integrator1, t_half)  # "inexact" stepping w/o tstops
if save_everystep
    # Fool set_*!
    integrator1.opts.save_everystep = false
    u_old = deepcopy(integrator1.sol.u)
    t_old = deepcopy(integrator1.sol.t)
end
if setter === set_t!
    set_t!(integrator1, integrator1.t)
elseif setter === set_u!
    set_u!(integrator1, integrator1.u)
else
    set_ut!(integrator1, integrator1.u, integrator1.t)
end
if save_everystep
    integrator1.opts.save_everystep = true
end
# step!(integrator1, prob.tspan[2], true)
solve!(integrator1)

solve!(integrator2)

@show integrator1.u
@show integrator2.u
@show integrator1.u - integrator2.u

using Plots
plt = plot()
plot!(plt, integrator1.sol, color=1, label="1.sol")
plot!(plt, integrator2.sol, color=2, label="2.sol")
if save_everystep
    scatter!(plt, t_old, u_old,
             markershape=:hexagon, markersize=8,
             color=3, label="1.u_old")
end
scatter!(plt, integrator1.sol.t, integrator1.sol.u, color=1, label="1.u")
scatter!(plt, integrator2.sol.t, integrator2.sol.u, color=2, label="2.u")
plot!(plt, xlim=(0.0, 2.0), ylim=(prob.u0, integrator2.sol[end]),
      legend=:topleft)
plt
```
</details>

tkf:
OK, I see that `reinit!` does

```julia
integrator.uprev = integrator.u
integrator.uprev2 = integrator.uprev  # if alg_extrapolates(integrator.alg)
integrator.tprev = t0
```

so it is not a no-op even when `u = integrator.u` and `t0 = integrator.t`.

ChrisRackauckas:
`reinit!` will reset stepping behavior and the PI adapter while `set` will not.

ChrisRackauckas:
This needs in place tests.

ChrisRackauckas:
The extrapolation problem is documented in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/257 . We can leave that for now and finish up with the in place tests.

tkf:
Thanks! So I added the inplace tests.

ChrisRackauckas:
Thanks!

(We really need to fix the Travis timeout 👎 .)

PR #258, opened 2018-02-21
Fix reinit!: copy u0 and t0 to solution data

### LINKED ISSUE #246: reinit! does not clear integrator.sol.interp.timeseries
In my https://github.com/tkf/LyapunovExponents.jl/pull/1, I realized that re-using ODEIntegrator causes a problem when using different `u0`.  It seems that the integration result is correct.  Only the recording into the solution object is not happening correctly.

The hack I found is to reset `integrator.sol.interp.timeseries` manually.  I didn't go deep in the code to see if that's the good way to do it.  But if so, it's very trivial to fix `reinit!`.

Here is the code to demonstrate the issue and the hack.  The assert should fail unless you uncomment `empty!(integrator.sol.interp.timeseries)` and `push!(integrator.sol.interp.timeseries, u1)`:

```julia
using DifferentialEquations

f(t, u, du) = (du .= -u)
u0 = [10.0]
t0 = 0
t1 = 1.0
t2 = 2 * t1

# Solve the ODE in one go:
long_ode = ODEProblem(f, copy(u0), (t0, t2))
long_sol = solve(long_ode)
times = linspace(t0, t2, 100)
long_data = hcat(map(long_sol, times)...)

half_ode = ODEProblem(f, copy(u0), (t0, t1))
alg, extra_kwargs = default_algorithm(half_ode)
integrator = init(half_ode, alg)

solve!(integrator)
u1 = integrator.sol[end]
first_data = [integrator.sol(t) for t in times if t < t1]

# Continue solving the ODE for tspan = "(t1, t2)"; OK, since t2 = 2 * t1
reinit!(integrator, u1)
# Without the following lines, `@assert` below does not work:
# empty!(integrator.sol.interp.timeseries)
# push!(integrator.sol.interp.timeseries, u1)
solve!(integrator)
second_data = [integrator.sol(t - t1) for t in times if t >= t1]

cat_data = hcat(vcat(first_data, second_data)...)

# It is easy to see if you plot it (but don't use `plot(integrator)`):
# using Plots
# plot(integrator.sol)

@assert isapprox(cat_data, long_data;
                 rtol=integrator.opts.reltol,
                 atol=integrator.opts.abstol)
```

checked with
```
julia> Pkg.status("OrdinaryDiffEq")
 - OrdinaryDiffEq                2.36.0
```

tkf:
closes #246

The code I added in `reinit!` doesn't touch `integrator.sol.k` but the relevant part in `init` seems to be just allocating arrays for it. So I guessed that it was not required. But I'm not very confident about it.

### DIFF
--- a/src/integrators/integrator_interface.jl
+++ b/src/integrators/integrator_interface.jl
@@ -196,6 +196,15 @@ function DiffEqBase.reinit!(integrator::ODEIntegrator,u0 = integrator.sol.prob.u
     resize!(integrator.sol.u,resize_start)
     resize!(integrator.sol.t,resize_start)
     resize!(integrator.sol.k,resize_start)
+    if integrator.opts.save_start
+      copyat_or_push!(integrator.sol.t,1,t0)
+      if integrator.opts.save_idxs == nothing
+        copyat_or_push!(integrator.sol.u,1,u0)
+      else
+        u_initial = u0[integrator.opts.save_idxs]
+        copyat_or_push!(integrator.sol.u,1,u_initial,Val{false})
+      end
+    end
     if integrator.sol.u_analytic != nothing
       resize!(integrator.sol.u_analytic,0)
     end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Yes, the first part of `k` is essentially blank, so this is good.

PR #259, opened 2018-02-21
Separate Pkg.build and Pkg.test

tkf:
I'm just curious to see how long it takes in Pkg.build


tkf:
2 min: https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/344172422#L750
That's totally negligible.

PR #260, opened 2018-02-24
[WIP] Exponential Krylov Integrators

MSeeker1340:
This is meant to track my progress in the implementation of exponential integrators that uses Krylov methods to approximate matrix functions, as suggested by @ChrisRackauckas .

Currently I have implemented the `LawsonEulerKrylov` and `ExpEulerKrylov` methods using `LawsonEuler` as a template and the `expmv`/`phimv` functions from [Expokit.jl](https://github.com/acroy/Expokit.jl). Usage example:

```julia
using OrdinaryDiffEq, DiffEqOperators

# Need to first define norm for `DiffEqArrayOperator`
Base.norm(f::DiffEqArrayOperator, l) = norm(f.A, l)

# Solve for the test system u' = au + bu.^2
a = 1.0; b = 0.01; N = 1000

f1 = DiffEqArrayOperator(a * eye(N))
f2(u,p,t) = b * u.^2

u0 = randn(N); tspan = (0., 1.)
prob = SplitODEProblem(f1, f2, u0, tspan)
dt = 0.01

sol1 = solve(prob, LawsonEulerKrylov(); dt=dt)
sol2 = solve(prob, ExpEulerKrylov(); dt=dt)
```

I'm still having some problems with the test suites though as the convergence tests seem to fail. Also I'm not sure if this works on the nonlinear Schrödinger equation at the moment.

One problem is that the Krylov methods and the full caching methods share most of their code. It may be a good idea to restructure the code so that code reuse is familited between the two versions, while also opening up ways for us to include other approximations to matrix functions (say the Chebyshev method). For example, I can expand on the `AbstractDiffEqLinearOperator` interface (as suggested [in the docs](http://docs.juliadiffeq.org/latest/features/diffeq_operator.html)) and have a particular algorithm's `expmv`/`phimv` calls dispatched with respect to the types of its linear part. 

Finally, the documentation of `SplitODEProblem` and `DiffEqOperators` are a bit lacking at the moment, so I can also do something about that first. (By the way why is `DiffEqOperators` its own package? This makes development with split ODE problems a bit awkward.)

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -13,3 +13,4 @@ Compat 0.18.0
 Reexport
 MuladdMacro 0.0.2
 StaticArrays
+Expokit 0.0.1
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -26,6 +26,9 @@ module OrdinaryDiffEq
 
   import ForwardDiff.Dual
 
+  # Required by exponential Krylov methods
+  import Expokit: expmv, expmv!
+
   # Required by temporary fix in not in-place methods with 12+ broadcasts
   import StaticArrays: SArray
 
@@ -146,7 +149,7 @@ module OrdinaryDiffEq
 
   export GenericIIF1, GenericIIF2
 
-  export LawsonEuler, NorsettEuler, ETDRK4
+  export LawsonEuler, NorsettEuler, ETDRK4, LawsonEulerKrylov
 
   export SymplecticEuler, VelocityVerlet, VerletLeapfrog, PseudoVerletLeapfrog,
          McAte2, Ruth3, McAte3, CandyRoz4, McAte4, McAte42, McAte5,
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -27,7 +27,7 @@ module OrdinaryDiffEq
   import ForwardDiff.Dual
 
   # Required by exponential Krylov methods
-  import Expokit: expmv, expmv!
+  import Expokit: expmv, expmv!, phimv, phimv!
 
   # Required by temporary fix in not in-place methods with 12+ broadcasts
   import StaticArrays: SArray
@@ -149,7 +149,7 @@ module OrdinaryDiffEq
 
   export GenericIIF1, GenericIIF2
 
-  export LawsonEuler, NorsettEuler, ETDRK4, LawsonEulerKrylov
+  export LawsonEuler, NorsettEuler, ETDRK4, LawsonEulerKrylov, ExpEulerKrylov
 
   export SymplecticEuler, VelocityVerlet, VerletLeapfrog, PseudoVerletLeapfrog,
          McAte2, Ruth3, McAte3, CandyRoz4, McAte4, McAte42, McAte5,
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -25,6 +25,7 @@ isdtchangeable(alg::GenericIIF2) = false
 isdtchangeable(alg::LawsonEuler) = false
 isdtchangeable(alg::NorsettEuler) = false
 isdtchangeable(alg::LawsonEulerKrylov) = false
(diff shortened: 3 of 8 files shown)


MSeeker1340:
I added a test set for the exponential Krylov solvers. It should run without problem.

```julia
using Base.Test
@time @testset "Linear-Nonlinear Krylov Methods Tests" begin include("test/linear_nonlinear_Krylov_tests.jl") end

Test Summary:                         | Pass  Total
Linear-Nonlinear Krylov Methods Tests |    4      4
 32.219506 seconds (22.43 M allocations: 3.906 GiB, 2.82% gc time)
```

I haven't gone through all of `DiffEqOperator` yet so I just wrote an ad-hoc functor type in my test script. I'm not sure about the suitable problem size for the test though: it probably doesn't make sense to use a small system for Krylov methods, however for an `N = 100` system the test is may a bit too long?

There's a small bug for complex-valued system which I found by testing the Schrodinger Equation. Basically when defining an `ODEProblem` with complex variables the `abstol` and `reltol` defaults to complex numbers:

```julia
f(u,p,t) = 1.01*u
u0 = complex(rand(), rand())
prob = ODEProblem(f, u0, (0.0,1.0))
integrator = init(prob, Tsit5())
@show integrator.opts.abstol
@show integrator.opts.reltol

integrator.opts.abstol = 1.0e-6 + 0.0im
integrator.opts.reltol = 0.001 + 0.0im
```

Since I pass `abstol` to `expmv` as the tolerance parameter (I still need to look at stepsize control in the documentation to figure out how exactly to convert the tolerance from one to the other), using a complex number will fail. I fixed this temporarily by taking its absolute value.

It seems to me that such a behavior is maybe an oversight? Let me look into this to see if a simple fix can solve this problem.

ChrisRackauckas:
>I'm not sure about the suitable problem size for the test though: it probably doesn't make sense to use a small system for Krylov methods, however for an N = 100 system the test is may a bit too long?

Yeah, the issue here is that Travis is really really slow so we can't put difficult tests on CI. But we have a repo OrdinaryDiffEqExtendedTests.jl where we have developed a separate CI that can do more intensive checks.

ChrisRackauckas:
>It seems to me that such a behavior is maybe an oversight? Let me look into this to see if a simple fix can solve this problem.

Looks like that's what was reported yesterday:

https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/263

That's oversight that should be fixed.

ChrisRackauckas:
I think `MatrixOperator` is pretty much the same thing as `DiffEqArrayOperator`. At least it has the same idea:

https://github.com/JuliaDiffEq/DiffEqOperators.jl/blob/master/src/array_operator.jl

Let's make sure tests pass, then start refactoring (put this with the other algorithms via `expmv` choice, find out how to get the array operator working, etc)

ChrisRackauckas:
Windows failures are unrelated. Looks like the binary caching server is down.

ChrisRackauckas:
Looks like the test failure is just a namespacing issue. `@testset` blocks don't clear global function names afterwards, so it's best to modify the name to be test based. I need to look into put a `let` block around the tests or something like that...

ChrisRackauckas:

ChrisRackauckas:
Where's this at right now?

MSeeker1340:
I'm currently writing `expmv` and related methods for `DiffEqOperators`. The final interface for the Krylov methods should probably be built upon the existing methods (e.g. by including a parameter to indicate caching or lazy evaluation via Krylov). This will require modifying the existing methods instead of writing new ones (the approach of this PR).

Let me do a final test to make sure the _Expokit.jl_ interface work as intended, then I'll archive the PR.

ChrisRackauckas:
Sounds great! Yup, I agree it should extend the previous methods (and simply set the `expm` cache to `nothing`), and the `expmv!` should be tied to the array itself (so that way different methods can be applied to different arrays in an aggregate).

MSeeker1340:
All tests passed, closing PR.

PR #261, opened 2018-02-27
Imex methods

rudrasohan:
@ChrisRackauckas I have created the perform step for CNAB method with reference to the paper [IMEX](https://www.math.utah.edu/~vshankar/5620/IMEX.pdf) . I have not used any tableau till now . I have hard-coded the coefficients value and made a seperate file for imex but if required later I will shift it to any other file as it may seem suitable. This is a modified version of the trapezoid using ideas from KenCarp as suggested.


rudrasohan:
I have faced some problem while writing the final step. It will be really helpful if you can tell whether the steps are correct or any correction is required. 


` if typeof(integrator.f) <: SplitFunction

    u = temp + z/2
    y1 = dt*f2(u,p,t + dt)
    temp += 3/2*y1 + z/2
    u = temp
  `else
    u = tmp + z/2`

src/perform_step/imex_perform_step.jl
```diff
@@ -0,0 +1,259 @@
...
+    z = z + dz
+  end
+
+  if (iter >= integrator.alg.max_newton_iter && do_newton) || fail_convergence
+    integrator.force_stepfail = true
+    return
+  end
+
+  if typeof(integrator.f) <: SplitFunction
+    u = temp + z/2
```
ChrisRackauckas: you mean tmp?

src/perform_step/imex_perform_step.jl
```diff
@@ -0,0 +1,259 @@
...
+    y1 = dt*f2(u,p,t + dt)
```
ChrisRackauckas: This shouldn't be using the new `u`, and it shouldn't be using `t+dt = t_{n+1}`

src/perform_step/imex_perform_step.jl
```diff
@@ -0,0 +1,259 @@
...
+    temp += 3/2*y1 + z/2
```
ChrisRackauckas: `z/2` is a iteration temporary and not `f_{n-1}`. The value you need here isn't actually cached, and so it needs to be added to the cache to be carried over the used here.

src/perform_step/imex_perform_step.jl
rudrasohan: Yes, thats a typo

src/perform_step/imex_perform_step.jl
rudrasohan: @ChrisRackauckas in the simple trapezoid rule before the newton iteration.
`u = tmp + z/2` this step does the `g(tn)` right and the z we get after the newton iteration is the `g(tn+1)` 

In kencarp after the 2nd step the calculation is.
   `u = tmp + γ*z₂
    k2 = dt*f2(u,p,t + 2γ*dt)
    tmp = uprev + a31*z₁ + a32*z₂ + ea31*k1 + ea32*k2`

For my method I also went along similar lines 
The u = temp + z/2 
The 2nd line should I be using zprev instead of u right as that is the un
`y1 = dt*f2(zprev,p,t)`

src/perform_step/imex_perform_step.jl
rudrasohan: For this last step the of the 2nd step kencarp uses 
`tmp = uprev + a31*z₁ + a32*z₂ + ea31*k1 + ea32*k2`

I need to store 
`u = uprev + z1/2 + z2 /2 + 3/2*y1  -1/2*y0` 
Shouldn't the `f_{n-1}` be y0 from the lines
52-56

ChrisRackauckas: No that should be `y1 = f_{n} = f(t,u)`.

src/perform_step/imex_perform_step.jl
ChrisRackauckas: `z = dt*f(t+dt,u_{n+1})`. See line 72 to see what it's iterating. It will iterate until `b = dt*f(u, p, tstep) - z = 0`

ChrisRackauckas: The reference for this way of writing it is https://www.sciencedirect.com/science/article/pii/0168927495001158

ChrisRackauckas:

rudrasohan:
alg

rudrasohan:
After correcting the caches the Closest candidates are even not showing the CNAB method.

src/algorithms.jl
```diff
@@ -616,3 +616,31 @@ end
...
                                    ImplicitEuler,ImplicitMidpoint}
+
+
+
+
+###############################################################################
+
+### IMEX Methods
+
+struct CNAB{CS,AD,F,FDT,K,T,T2,Controller} <: OrdinaryDiffEqNewtonAdaptiveAlgorithm{CS,AD,Controller}
```
rudrasohan: @ChrisRackauckas is this definition correctly writtn for the alg_cache ?

ChrisRackauckas:
Play with it a bit. Double check the signature and make sure it's the right number of args and such. The problem doesn't immediately jump out to me.

rudrasohan:
Will it be a good idea to upload the imex tests also and see whether Travis can give a better error message to pin point it or will it be redundant ?

ChrisRackauckas:
You should add tests, yes. But Travis won't be any more helpful than local testing. Usually it's easiest to getting working locally first.

rudrasohan:
@ChrisRackauckas I have uploaded the tests. I was looking at the TravisCI build for possible signs of error but the build stopped prematurely with this error `The job exceeded the maximum time limit for jobs, and has been terminated.`

ChrisRackauckas:
The timings were about 3x different so I'm just re-running Travis to see if that was a fluke.

ChrisRackauckas:
Yeah that was Travis overloading their servers. BTW you can restart tests just by closing and re-opening a PR.

src/caches/imex_caches.jl
```diff
@@ -0,0 +1,110 @@
...
+  κ::uEltypeNoUnits
+  tol::uEltypeNoUnits
+  newton_iters::Int
+  #tab::Tab
+  uprev::uType
+  y0::rateType
+  y1::rateType
+end
+
+function alg_cache(alg::CNAB,u,rate_prototype,uEltypeNoUnits,tTypeNoUnits,uBottomEltypeNoUnits,uprev,f,t,dt,reltol,p,calck,::Type{Val{false}})
```
ChrisRackauckas: This doesn't look right. `u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,
                   tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{true}})`

rudrasohan: @ChrisRackauckas  I tried out this but the method matching error is still there. But this time the error is pointing to this line `CNABConstantCache(uf,ηold,κ,tol,10000)`  .The terms which we pass in this statement, What do they represent ?

ChrisRackauckas: They are the quasi-Newton iteration tolerances. Check it out in the algorithm to see what they are doing. Hairer Solving Ordinary Differential Equations II also discusses this a bit, but that detail shouldn't be necessary. Just make sure that your returned cache matches our new cache constructor.

rudrasohan: There are some values which we `@unpack` in the perform step while some we use directly. What is the use of that `@unpack`?

ChrisRackauckas: It's just putting the variables from the cache into scope. See https://github.com/mauro3/Parameters.jl

src/perform_step/imex_perform_step.jl
```diff
@@ -0,0 +1,248 @@
...
+  if (iter >= integrator.alg.max_newton_iter && do_newton) || fail_convergence
+    integrator.force_stepfail = true
+    return
+  end
+
+  cache.ηold = η
+  cache.newton_iters = iter
+
+  if typeof(integrator.f) <: SplitFunction
+    @ .tmp += z/2
```
YingboMa: This should probably be `@. tmp += z/2`

test/split_methods_tests.jl
```diff
@@ -185,3 +187,31 @@ sol = solve(prob,KenCarp5())
...
+function (::typeof(prob.f))(::Type{Val{false}},u0,p,t)
+    exp(2t)*u0
+end
+
+sol = solve(prob,CNAB(),dt=0.1)
+dts = 1.//2.^(8:-1:4)
+sim = test_convergence(dts,prob,CNAB())
+@test abs(sim.𝒪est[:l∞]-3) < testTol
+
+f1 = (u,p,t) -> zero(u)
```
YingboMa: Could you please add an in-place version?

PR #263, opened 2018-02-28
Complex compatibility for SDIRK solvers

MSeeker1340:
This is to fix the issue mentioned here:

https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/263

Current only the `Trapezoid` solver is changed. If this is ok I will do a sweep and fix all methods in _src/sdirk_caches.jl_.

```julia
using OrdinaryDiffEq
fun(du,u,p,t) = du .= 1.01 * u
u0 = complex.(rand(3), rand(3))
prob = ODEProblem(fun, u0, (0.,1.))
integrator = init(prob, Trapezoid(autodiff=false))
step!(integrator)
@show integrator.u
@show integrator.opts.abstol
@show integrator.cache.tol

integrator.u = Complex{Float64}[0.301987+0.34158im, 0.375765+0.213668im, 0.959003+0.567313im]
integrator.opts.abstol = 1.0e-6
integrator.cache.tol = 0.03
```

Note that the non in-place version still wouldn't work because it always uses autodiff and it seems `ForwardDiff` doesn't support complex numbers at the moment.

Any complex type should work fine, for example `Complex{BigFloat}`:

```julia
fun(du,u,p,t) = du .= BigFloat(1.01) * u
u0 = complex.(BigFloat.(rand(3)), BigFloat.(rand(3)))
prob = ODEProblem(fun, u0, (0.,1.))
integrator = init(prob, Trapezoid(autodiff=false))
step!(integrator)
@show integrator.u
@show integrator.opts.abstol
@show integrator.cache.tol

integrator.u = Complex{BigFloat}[2.679858206626884697903598767224898104940434506808948024123236895484489044748831e-01+3.921263588544868552646750965152075957478933078567815669442307422935384996729011e-01im, 1.88132689477089063573093699949069719988702691486065157103197451208253684743379e-01+2.253309322148819911393625846574983901669478764490170771497564706309733078302334e-01im, 8.244022822168234507675387805035939410329717133546315069417702181305616784803873e-01+2.47932958697165370811102887489999317237715898236108469078444600521011934967557e-01im]
integrator.opts.abstol = 1.000000000000000000000000000000000000000000000000000000000000000000000000000004e-06
integrator.cache.tol = 2.999999999999999888977697537484345957636833190917968750000000000000000000000000e-02
```

There's also a minor fix that corrects the ordering of arguments for the `alg_cache` function.

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -253,7 +253,3 @@ ssp_coefficient(alg::SSPRK104) = 6
 # We shouldn't do this probably.
 #ssp_coefficient(alg::ImplicitEuler) = Inf
 ssp_coefficient(alg::SSPSDIRK2) = 4
-
-# Compute type for tolerance and related values
-toltype(::Type{T}) where {T <: Real} = T
-toltype(::Type{Complex{T}}) where {T <: Real} = T
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -253,3 +253,7 @@ ssp_coefficient(alg::SSPRK104) = 6
 # We shouldn't do this probably.
 #ssp_coefficient(alg::ImplicitEuler) = Inf
 ssp_coefficient(alg::SSPSDIRK2) = 4
+
+# Compute type for tolerance and related values
+toltype(::Type{T}) where {T <: Real} = T
+toltype(::Type{Complex{T}}) where {T <: Real} = T
--- a/src/caches/kencarp_kvaerno_caches.jl
+++ b/src/caches/kencarp_kvaerno_caches.jl
@@ -1,38 +1,39 @@
-mutable struct KenCarp3ConstantCache{F,uEltypeNoUnits,Tab} <: OrdinaryDiffEqConstantCache
+mutable struct KenCarp3ConstantCache{F,uToltype,Tab} <: OrdinaryDiffEqConstantCache
   uf::F
-  ηold::uEltypeNoUnits
-  κ::uEltypeNoUnits
-  tol::uEltypeNoUnits
+  ηold::uToltype
+  κ::uToltype
+  tol::uToltype
   newton_iters::Int
   tab::Tab
 end
 
-function alg_cache(alg::KenCarp3,u,rate_prototype,uEltypeNoUnits,tTypeNoUnits,uBottomEltypeNoUnits,
+function alg_cache(alg::KenCarp3,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,
                    uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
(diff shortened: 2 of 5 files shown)


ChrisRackauckas:
All of the tests have a 2x regression from this PR, but I'm not entirely sure why that would be the case.

ChrisRackauckas:

src/alg_utils.jl
```diff
@@ -253,3 +253,7 @@ ssp_coefficient(alg::SSPRK104) = 6
 # We shouldn't do this probably.
 #ssp_coefficient(alg::ImplicitEuler) = Inf
 ssp_coefficient(alg::SSPSDIRK2) = 4
+
+# Compute type for tolerance and related values
```
ChrisRackauckas: This is misplaced. It should be in utils.jl

ChrisRackauckas: Actually, instead of using this you can just do `real(T)`. 

```julia
julia> real(Complex{Float64})
Float64
```

MSeeker1340:
A note for future code refactoring:

The code for `Kvaerno3` is included in _sdirk_caches.jl_, which by its name should probably be in _kencarp_kvaerno_caches.jl_. 

I'll check to see if the complex tests in #264 works now.

ChrisRackauckas:
>The code for Kvaerno3 is included in sdirk_caches.jl, which by its name should probably be in kencarp_kvaerno_caches.jl.

Good catch. Thanks!

Yeah, we really need to refactor the SDIRK methods some. It needs to be in a way that doesn't lose the GPU compatibility and allows the different stage predictors for each one though, which is what makes it difficult. But the Jacobian part at the top can definitely be refactored package-wide, and the Newton iterations could probably be made from a macro.

MSeeker1340:
Travis seems to still hang at the moment so I'll post the test result using my local machine here:

```julia
using Base.Test
cd(Pkg.dir("OrdinaryDiffEq"))
@time @testset "Complex Tests" begin
    include("test/complex_tests.jl")
end

Test Summary: | Pass  Broken  Total
Complex Tests |   19       2     21
 56.411503 seconds (41.78 M allocations: 2.122 GiB, 1.26% gc time)
```

Basically, the state for complex solvers is:

- Explicit solvers work (in-place/non in-place/non in-place + `StaticArrays`);

- Implicit solvers with autodiff don't work because `ForwardDiff` doesn't seem to support complex numbers at the moment;

- Implicit solvers without autodiff and using in-place function definition work;

- Implicit solvers using non in-place function definition don't work because they're hard-wired to use autodiff (should probably give a warning if we use in-place function with `autodiff=false`).

    - Thus `StaticArrays` also won't work because they can only be used with in-place functions.

ChrisRackauckas:
I ran the benchmarks and it's just Travis that's acting up. I'll get that fixed up. Thanks!

PR #264, opened 2018-03-01
Tests for complex ODEs

antoine-levitt:
This tests explicit and implicit solvers for a simple complex ODE (Schrödinger equation), using various combinations of inplace, out of place, static arrays, autodiff and finite differences. Ref https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/263, https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/259

I put in a random selection of ODE solvers (wanted to enumerate all subtypes of OrdinaryDiffEqAlgorithm but that had a number of tricks to it). Explicit solvers are working, implicit are not.

Btw, the Schrödinger equation i psi' = H(t) psi / eps with eps small is a pretty fun equation (the solution stays on the eigenstates of H(t) for eps small) that is highly oscillatory and therefore tricky to solve. If working, it would make for a cute showcase of DifferentialEquations's unique capabilities (complex numbers, stiff solvers with autodiff, matrix exponentials...)

### DIFF
--- a/test/complex_tests.jl
+++ b/test/complex_tests.jl
@@ -1,0 +1,32 @@
+# Solve the Landau-Zener problem i ψ' = H(t) ψ, with H(t) = [t 1;1 -t]
+
+using Base.Test
+using StaticArrays
+using OrdinaryDiffEq, DiffEqBase
+
+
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:

PR #265, opened 2018-03-01
Native Adams-Bashforth (3 steps)-Moulton (2 steps) method

sipah00:
Work on issue #4 
I have added Adams bashforth (3 steps) `AB3()` and Adams-Bashforth (3 steps) and Adams-Moulton (2 steps) `ABM32()`.
I have tested it on a example:  
```using OrdinaryDiffEq
f(u,p,t) = u-t*t+1
u0=1/2
tspan = (0.0,2.0)
h = 0.2
prob = ODEProblem(f,u0,tspan)
sol = solve(prob,ABM32(),dt=0.2)
```
output given by `Tsit5()`:  
```
u: 11-element Array{Float64,1}:
 0.5     
 0.829299
 1.21409 
 1.64894 
 2.12723 
 2.64086 
 3.17995 
 3.73241 
 4.28348 
 4.81519 
 5.30548 
```
output given by `AB3()`:  
```
u: 11-element Array{Float64,1}:
 0.5     
 0.827333
 1.20988 
 1.64405 
 2.12191 
 2.63505 
 3.17364 
 3.72569 
 4.27648 
 4.80809 
 5.2986  
```
output given by `ABM32()`: 
```
u: 11-element Array{Float64,1}:
 0.5     
 0.827333
 1.21147 
 1.64573 
 2.12328 
 2.63601 
 3.17399 
 3.72509 
 4.27451 
 4.80415 
 5.29193 
```

### DIFF
--- a/src/perform_step/adams_bashforth_moulton_perform_step.jl
+++ b/src/perform_step/adams_bashforth_moulton_perform_step.jl
@@ -64,8 +64,8 @@ end
     end
   else
     @. u  = uprev + (dt/12)*(23*k1 - 16*k2 + 5*k3)
-    cache.k3 = copy(k2)
-    cache.k2 = copy(k1)
+    cache.k2,cache.k3 = k3,k2
+    cache.k2 .= k1
   end
   f(k, u, p, t+dt)
 end
@@ -127,13 +127,13 @@ end
     @. tmp = uprev + (2/3)*dt*k1
     f(ralk2, tmp, p, ttmp)
     @. u = uprev + (dt/4)*(k1 + 3*ralk2)       #Ralston Method
-    cache.k2 = copy(k1)
+    cache.k2 .= k1
   else
     perform_step!(integrator, AB3Cache(u,uprev,fsalfirst,k2,k3,ralk2,k,tmp))
     k = integrator.fsallast
     @. u = uprev + (dt/12)*(5*k + 8*k1 - k2)
-    cache.k3 = copy(k2)
-    cache.k2 = copy(k1)
+    cache.k2,cache.k3 = k3,k2
+    cache.k2 .= k1
   end
   f(k, u, p, t+dt)
 end
--- a/src/perform_step/adams_bashforth_moulton_perform_step.jl
+++ b/src/perform_step/adams_bashforth_moulton_perform_step.jl
@@ -129,7 +129,7 @@ end
     @. u = uprev + (dt/4)*(k1 + 3*ralk2)       #Ralston Method
     cache.k2 .= k1
   else
-    perform_step!(integrator, AB3Cache(u,uprev,fsalfirst,copy(k2),k3,ralk2,k,tmp))
+    perform_step!(integrator, AB3Cache(u,uprev,fsalfirst,copy(k2),k3,ralk2,k,tmp))  #Here passing copy of k2, otherwise it will change in AB3()
     k = integrator.fsallast
     @. u = uprev + (dt/12)*(5*k + 8*k1 - k2)
     cache.k2,cache.k3 = k3,k2
--- a/src/perform_step/adams_bashforth_moulton_perform_step.jl
+++ b/src/perform_step/adams_bashforth_moulton_perform_step.jl
@@ -1,0 +1,123 @@
+function initialize!(integrator,cache::AB3ConstantCache)
+  integrator.fsalfirst = integrator.f(integrator.uprev, integrator.p, integrator.t) # Pre-start fsal
+  integrator.kshortsize = 2
+  integrator.k = typeof(integrator.k)(integrator.kshortsize)
+
+  # Avoid undefined entries if k is an array of arrays
+  integrator.fsallast = zero(integrator.fsalfirst)
+  integrator.k[1] = integrator.fsalfirst
+  integrator.k[2] = integrator.fsallast
+end
+
+@muladd function perform_step!(integrator,cache::AB3ConstantCache,repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  k1 = integrator.fsalfirst
+  cnt = length(integrator.sol.t)
+  if cnt == 1 || cnt == 2
+    ttmp = t + (2/3)*dt
+    ralk2 = f(uprev + (2/3)*dt*k1, p, ttmp)
+    u = uprev + (dt/4)*(k1 + 3*ralk2)
+  else
+    u1 = integrator.sol(t-dt)
+    u2 = integrator.sol(t-2*dt)
+    k2 = f(u1, p, t-dt)
+    k3 = f(u2, p, t-2*dt)
+    u  = uprev + (dt/12)*(23*k1 - 16*k2 + 5*k3)
+  end
+  integrator.fsallast = f(u, p, t+dt)
+  integrator.k[1] = integrator.fsalfirst
+  integrator.k[2] = integrator.fsallast
+  integrator.u = u
+end
+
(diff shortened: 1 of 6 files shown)


sipah00:
@ChrisRackauckas, This is my first try at issue #4.
Looking forward to your suggestions.

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -0,0 +1,123 @@
...
+@muladd function perform_step!(integrator,cache::AB3ConstantCache,repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  k1 = integrator.fsalfirst
+  cnt = length(integrator.sol.t)
+  if cnt == 1 || cnt == 2
+    ttmp = t + (2/3)*dt
+    ralk2 = f(uprev + (2/3)*dt*k1, p, ttmp)
+    u = uprev + (dt/4)*(k1 + 3*ralk2)
+  else
+    u1 = integrator.sol(t-dt)
```
ChrisRackauckas: If saving is disabled this won't work. It's better to setup a cache of old values.

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -0,0 +1,123 @@
...
+    u2 = integrator.sol(t-2*dt)
+    k2 = f(u1, p, t-dt)
+    k3 = f(u2, p, t-2*dt)
```
ChrisRackauckas: You don't need to calculate thi. `k3` is `k2` of last time. `k2` is `k1` of last time. And then `k1` comes from `fsalfirst` calculations. So you actually don't need any function calls other than the FSAL interpolation one.

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -0,0 +1,123 @@
...
+  integrator.k[2] = integrator.fsallast
+end
+
+@muladd function perform_step!(integrator,cache::AB3ConstantCache,repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  k1 = integrator.fsalfirst
+  cnt = length(integrator.sol.t)
+  if cnt == 1 || cnt == 2
+    ttmp = t + (2/3)*dt
+    ralk2 = f(uprev + (2/3)*dt*k1, p, ttmp)
```
ChrisRackauckas: This is the Ralston step? Comment it.

ChrisRackauckas:
I gave some comments for how to cut out a bunch of function evaluations. You'll need to take the re-use of `k`s into account to make this fast.

Additionally, this needs tests. See the `tests` folder for how they are written. It should get a convergence test to make sure it's correct. Use `plot(sim)` on a convergence simulation to see what the errors look like to debug it.

ChrisRackauckas:

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -105,19 +119,21 @@ end
...
   else
-    perform_step!(integrator,integrator.cache.onestep_cache)
-    u2 = integrator.sol(t-dt)
-    k2 = integrator.fsallast
-    f(k3, u2, p, t-dt)
-    @. u = uprev + (dt/12)*(5*k2 + 8*k1 - k3)
+    perform_step!(integrator, AB3Cache(u,uprev,fsalfirst,k2,k3,ralk2,k,tmp))
+    k = integrator.fsallast
+    @. u = uprev + (dt/12)*(5*k + 8*k1 - k2)
+    cache.k3 = copy(k2)
```
ChrisRackauckas: For these you don't need to copy. Instead, a 2-way pointer rotation is fine. `cache.k2,cache.k3 = k3,k2`. I'm not sure you can do that all in one line but you get the point.  Then you need to move `k` into `k2` via `.=`.

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -43,19 +49,23 @@ end
...
   if cnt == 1 || cnt == 2
-    @. ttmp = t + (2/3)*dt
-    f(ralk2, uprev + (2/3)*dt*k1, p, ttmp)
-    @. u = uprev + (dt/4)*(k1 + 3*k2)
+    ttmp = t + (2/3)*dt
+    @. tmp = uprev + (2/3)*dt*k1
+    f(ralk2, tmp, p, ttmp)    
+    @. u = uprev + (dt/4)*(k1 + 3*ralk2)        #Ralston Method
+    if cnt == 1
+      cache.k3 = copy(k1)
```
ChrisRackauckas: same with the mention about copying below.

src/perform_step/adams_bashforth_moulton_perform_step.jl
sipah00: @ChrisRackauckas, By doing `cache.k2,cache.k3 = k1,k2` `chache.k2` refer to `k1` and  if `k1` changes then `cache.k2` will also change, here value should not change.

ChrisRackauckas: I didn't suggest `cache.k2,cache.k3 = k1,k2` for that reason. Note that I mentioned that you can swap the last two pointers, but always need to `.=` for the `k` (`k1`) part.

sipah00: I got it.

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -0,0 +1,139 @@
...
+  @unpack tmp,fsalfirst,k2,k3,ralk2,k = cache
+  k1 = integrator.fsalfirst
+  cnt = integrator.iter
+  if cnt == 1 || cnt == 2
+    ttmp = t + (2/3)*dt
+    @. tmp = uprev + (2/3)*dt*k1
+    f(ralk2, tmp, p, ttmp)    
+    @. u = uprev + (dt/4)*(k1 + 3*ralk2)        #Ralston Method
+    if cnt == 1
+      cache.k3 = copy(k1)
```
ChrisRackauckas: You don't need to `copy` if the cache is already created. Just `.=`

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -0,0 +1,139 @@
...
+  k1 = integrator.fsalfirst
+  cnt = integrator.iter
+  if cnt == 1
+    ttmp = t + (2/3)*dt
+    @. tmp = uprev + (2/3)*dt*k1
+    f(ralk2, tmp, p, ttmp)
+    @. u = uprev + (dt/4)*(k1 + 3*ralk2)       #Ralston Method
+    cache.k2 .= k1
+  else
+    perform_step!(integrator, AB3Cache(u,uprev,fsalfirst,copy(k2),k3,ralk2,k,tmp))
```
ChrisRackauckas: is this copy needed?

sipah00: @ChrisRackauckas, yes, `k2 (ABM())` should not change in `AB()`.

ChrisRackauckas: Could you add a comment? Most `perform_step!` methods don't allocate, so it's a good thing to mention why this has to happen.

ChrisRackauckas:
I think this is ready to merge, but the allocation should get a quick comment and it's good to go.

PR #267, opened 2018-03-03
Krylov Method Swtich in Exponential Integrators

MSeeker1340:
Second attempt at adding Krylov methods support to the existing exponential integrators.

The basic idea is to add a `krylov` argument to the algorithms' constructor. If it is set to `true`, no caching will be done at the initialization step and `expmv`/`expmv!` from _Expokit_ will be called in the iteration steps. This would also allow for further development using other lazy matrix function-vector application techniques, for example Chebyshev approximation, by adding other arguments to the algorithm while retaining the basic structure of the caches/iterations.

The exponential Krylov integrators should work nicely with the large-dimension derivative operators from _DiffEqOperators_ because they naturally support the _Expokit_ interface. (There's a minor problem that the derivative operators currently won't work with the caching version because `dt*L` yields a `LinearCombinations` and `expm(dt*L)` is not defined for the type. This should be easy to fix though).

Other low-order exponential methods can be modified similarly using `expmv` and `phimv` from _Expokit_, while methods using higher-order phi functions require more work.

A note about the tests: there's a peculiar problem that if I run the Krylov tests *after* the linear-nonlinear convergence tests there will be error. This may have something to do with the analytic solution defined in the linear-nonlinear convergence tests somehow polluting the global namespace. I need to take a deeper look into this.

### DIFF
--- a/src/perform_step/exponential_rk_perform_step.jl
+++ b/src/perform_step/exponential_rk_perform_step.jl
@@ -71,7 +71,11 @@ function perform_step!(integrator, cache::NorsettEulerConstantCache, repeat_step
   @unpack t,dt,uprev,u,f,p = integrator
   rtmp = integrator.fsalfirst
   A = f.f1
-  u = uprev + ((expm(dt*A)-I)/A)*(A*uprev + rtmp)
+  if integrator.alg.krylov
+    u = phimv(dt,A,rtmp,uprev; tol=integrator.opts.reltol)
+  else
+    u = uprev + ((expm(dt*A)-I)/A)*(A*uprev + rtmp)
+  end
   rtmp = f.f2(u,p,t+dt)
   k = A*u + rtmp # For the interpolation, needs k at the updated point
   integrator.fsallast = rtmp
@@ -95,13 +99,17 @@ end
 
 function perform_step!(integrator, cache::NorsettEulerCache, repeat_step=false)
   @unpack t,dt,uprev,u,f,p = integrator
-  @unpack k,rtmp,tmp,expA,phi1 = cache
+  @unpack k,rtmp,tmp = cache
   A = f.f1
 
-  A_mul_B!(tmp,A,uprev)
-  tmp .+= rtmp
-  A_mul_B!(rtmp,phi1,tmp)
-  @. u = uprev + rtmp
+  if integrator.alg.krylov
+    phimv!(u,dt,A,rtmp,uprev; tol=integrator.opts.reltol)
+  else
+    A_mul_B!(tmp,A,uprev)
+    tmp .+= rtmp
+    A_mul_B!(rtmp,cache.phi1,tmp)
+    @. u = uprev + rtmp
+  end
   A_mul_B!(tmp,A,u)
   f.f2(rtmp,u,p,t+dt)
   @. k = tmp +  rtmp
--- a/src/perform_step/exponential_rk_perform_step.jl
+++ b/src/perform_step/exponential_rk_perform_step.jl
@@ -14,7 +14,11 @@ function perform_step!(integrator, cache::LawsonEulerConstantCache, repeat_step=
   @unpack t,dt,uprev,u,f,p = integrator
   rtmp = integrator.fsalfirst
   A = f.f1
-  @muladd u = expm(dt*A)*(uprev + dt*rtmp)
+  if integrator.alg.krylov
+    @muladd u = expmv(dt, A, uprev + dt*rtmp; tol=integrator.opts.reltol)
+  else
+    @muladd u = expm(dt*A)*(uprev + dt*rtmp)
+  end
   rtmp = f.f2(u,p,t+dt)
   k = A*u + rtmp # For the interpolation, needs k at the updated point
   integrator.fsallast = rtmp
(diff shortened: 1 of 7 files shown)


src/perform_step/exponential_rk_perform_step.jl
```diff
@@ -14,7 +14,11 @@ function perform_step!(integrator, cache::LawsonEulerConstantCache, repeat_step=
   @unpack t,dt,uprev,u,f,p = integrator
   rtmp = integrator.fsalfirst
   A = f.f1
-  @muladd u = expm(dt*A)*(uprev + dt*rtmp)
+  if integrator.alg.krylov
+    @muladd u = expmv(dt, A, uprev + dt*rtmp; tol=integrator.opts.reltol)
+  else
+    @muladd u = expm(dt*A)*(uprev + dt*rtmp)
```
ChrisRackauckas: Should use the cached `expA` here? Looks like that's an error from the original code.

ChrisRackauckas:
I like this. I assume this needs the DiffEqOperators.jl release for tests to pass? 

>There's a minor problem that the derivative operators currently won't work with the caching version because dt*L yields a LinearCombinations and expm(dt*L) is not defined for the type. This should be easy to fix though

Yeah, they need to support coefficients. The operators need quite a bit of work but for now we can ignore that and use the DiffEqArrayOperator in tests while fixing it up.

ChrisRackauckas:
Restarted for new DiffEqOpeartors

MSeeker1340:
Again, if I put the Krylov tests before the splits methods tests they will pass, but not the other way around.

I think I've isolated the error causing this issue. It does has something to do with defining analytic solutions:

```julia
using OrdinaryDiffEq
using DiffEqOperators
using Base.Test
cd(Pkg.dir("OrdinaryDiffEq"))

# From "linear_nonlinear_convergence_tests.jl"
# Not related to the Krylov tests
μ = 1.01
u0 = rand(2)
A = [2.0 -1.0; -1.0 2.0]
linnonlin_f1 = DiffEqArrayOperator(A)
linnonlin_f2 = (du,u,p,t) -> du .= μ .* u
prob = SplitODEProblem(linnonlin_f1,linnonlin_f2,u0,(0.0,1.0))

@time @testset "Linear-Nonlinear Krylov Methods Tests" begin
    include("test/linear_nonlinear_krylov_tests.jl")
end

Test Summary:                         | Pass  Total
Linear-Nonlinear Krylov Methods Tests |    2      2
 15.429666 seconds (17.75 M allocations: 931.546 MiB, 1.76% gc time)
```

However after adding analytic solution for `prob`, the subsequent Krylov tests would fail:

```julia
using OrdinaryDiffEq
using DiffEqOperators
using Base.Test
cd(Pkg.dir("OrdinaryDiffEq"))

# From "linear_nonlinear_convergence_tests.jl"
# Not related to the Krylov tests
μ = 1.01
u0 = rand(2)
A = [2.0 -1.0; -1.0 2.0]
linnonlin_f1 = DiffEqArrayOperator(A)
linnonlin_f2 = (du,u,p,t) -> du .= μ .* u
prob = SplitODEProblem(linnonlin_f1,linnonlin_f2,u0,(0.0,1.0))

# Analytic solution. Doesn't matter what it does.
(::typeof(prob.f))(::Type{Val{:analytic}},u0,p,t) = nothing

@time @testset "Linear-Nonlinear Krylov Methods Tests" begin
    include("test/linear_nonlinear_krylov_tests.jl")
end

Linear-Nonlinear Krylov Methods Tests: Error During Test
  Got an exception of type LoadError outside of a @test
  LoadError: MethodError: no method matching A_mul_B!(::Void, ::DiffEqOperators.DiffEqArrayOperator{Float64,Array{Float64,2},DiffEqOperators.DiffEqScalar{Void,Float64},DiffEqOperators.##1#2}, ::Array{Float64,1})
...
```

Putting split methods tests before the Krylov tests has a similar effect because it also defines some `SplitODEProblem` along with their analytic solutions. Somehow the combination of the two breaks things...

MSeeker1340:
Should pass now. 

After defining analytic solution for another `SplitODEProblem`, `DiffEqBase.has_analytic` somehow thinks all similar `SplitFunction`, despite having different type parameters, to all have analytic solutions. Thus when calling `solve` Julia will try to calculate an non-existent analytic solution to estimate errors, which causes an error.

The ad-hoc fix manually sets `DiffEqBase.has_analytic` for the problems tested in the Krylov script. I will need to look into _DiffEqBase_ to see what is the root of the problem though.

ChrisRackauckas:
>The ad-hoc fix manually sets DiffEqBase.has_analytic for the problems tested in the Krylov script. I will need to look into DiffEqBase to see what is the root of the problem though.

The root of the problem is the way the traits are placed. We have an whole new backwards-compatible setup that doesn't have this problem, but it just hasn't been completed yet (well, the ODE case is complete, but not all of the off-shoots).

https://github.com/JuliaDiffEq/DiffEqBase.jl/issues/52

ChrisRackauckas:
Is this still WIP? It seems like it looks good.

MSeeker1340:
I think you can merge this now, as all the low-order exponential methods (Lawson Euler/Norsett Euler) are covered. (Need to add corresponding documentation though.)

I'll have a look at the EPIRK paper in the meantime.

ChrisRackauckas:
Yup that's what I thought. Cool. Thanks for the PR. Looks great!

PR #270, opened 2018-03-06
Tidying up the exponential integrators

MSeeker1340:
1. Adds an option to specify the Krylov subspace size in the algorithms' constructor, which I forgot to include in the last version.

2. Uses the `normbound` estimate from `DiffEqOperators` for the Krylov methods so that they can work with linear combinations of operators (could use a more refined norm estimation in the future).

3. Compatibility fix for the linear operator part. In its current form the exponential integrators can only handle `DiffEqArrayOperator` as the linear operator and furthermore sometimes forget to account for the coefficient. After the update, they should work for all `AbstractDiffEqLinearOperator` defined in DiffEqOperators as well as their linear combinations. (Haven't touched the generic IIF methods though. They may require more work to have the derivative operators be acceptable by user-supplied `nlsolve` methods).

Needs the latest changes to `DiffEqOperators` be released for the tests to pass.

Documentation update for this:

https://github.com/JuliaDiffEq/DiffEqDocs.jl/pull/101

### DIFF
--- a/src/caches/linear_nonlinear_caches.jl
+++ b/src/caches/linear_nonlinear_caches.jl
@@ -127,7 +127,7 @@ function alg_cache(alg::NorsettEuler,u,rate_prototype,uEltypeNoUnits,uBottomElty
     A = f.f1
     if isa(A, DiffEqArrayOperator) && typeof(A.A) <: Diagonal
         _expA = expm(A*dt)
-        phi1 = Diagonal(Float64.((big.(_expA)-I)/A.A))
+        phi1 = Diagonal(Float64.((big.(_expA)-I)/(A.A .* A.α.coeff)))
         expA = Diagonal(_expA)
 
         # Fix zero eigenvalues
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
The IIF methods need quite a bit more work so that's fine. There's a ton of optimizations that can be done on them, and I'd be willing to describe it if you're interested.

src/caches/linear_nonlinear_caches.jl
```diff
@@ -161,7 +162,12 @@ end
 
 function alg_cache(alg::ETDRK4,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
   A = f.f1
-  E,E2,a,b,c,Q = get_etdrk4_oop_operators(dt,A.A)
+  if isa(A, DiffEqArrayOperator)
+    L = A.A .* A.α.coeff # has specail handling is A.A is Diagonal
```
ChrisRackauckas: special

ChrisRackauckas: We should start thinking about matrix types in general.

PR #271, opened 2018-03-08
[WIP] ETD2 Integrator

MSeeker1340:
I have finished the out-of-place portion of the `ETD2` integrator and could use some suggestions & review.

As a reference, the ETD2 time stepping is:

ETD2

One of the major problems for existing exp methods is that only the nonlinear part of the right hand side is FSALed, which means we either have to spend extra time computing `k` in each time or use incorrect derivative values (which the methods currently do).

Fortunately the integrator interface supports custom fsal types, so I defined a `ETD2Fsal` (which is essentially a named tuple) to store both the linear and nonlinear part. The interpolating `k` value at each interval is computed by summing the two parts, which should be relatively fast.

I have also decided to store the previous nonlinear part (Nn in the formula) in the FSAL, which is different from the approach @sipah00 used in AB methods. They both work because FSAL acts somewhat like an additional cache, but my rationale for not storing the previous time steps in the cache is:

- The traditional role of FSAL in RK methods is to carry over the derivative calculations to the next time step, which can be extended naturally to also include the previous time steps for multistep methods.

- If we store the previous time steps in the cache, then the cache would need to be mutable (which is the case for `AB3ConstantCache`). This kinda violates the "constant" assumption of the constant cache.

Finally, unlike `LawsonEuler` and `NorsettEuler` but like `ETDRK4`, the out-of-place style also caches the operator exponentials. In fact, now that I've written `ETD2` I find it weird that the first order methods choose to not cache the operators (they are time-independent anyways). Like `ETDRK4`, I included special handling when the linear operator is scalar or Diagonal, but I could probably further improve the accuracy for small `dt` by switching to `expm1` using.

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -152,7 +152,7 @@ module OrdinaryDiffEq
 
   export GenericIIF1, GenericIIF2
 
-  export LawsonEuler, NorsettEuler, ETDRK4
+  export LawsonEuler, NorsettEuler, ETD1, ETDRK4
 
   export SymplecticEuler, VelocityVerlet, VerletLeapfrog, PseudoVerletLeapfrog,
          McAte2, Ruth3, McAte3, CandyRoz4, McAte4, McAte42, McAte5,
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -17,7 +17,7 @@ module OrdinaryDiffEq
 
   using DiffEqOperators: normbound, DiffEqArrayOperator
 
-  import RecursiveArrayTools: chain
+  import RecursiveArrayTools: chain, recursivecopy!
 
   using Parameters, GenericSVD, ForwardDiff, RecursiveArrayTools,
         NLsolve, Juno, Roots, DataStructures, DiffEqDiffTools
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -152,7 +152,7 @@ module OrdinaryDiffEq
 
   export GenericIIF1, GenericIIF2
 
-  export LawsonEuler, NorsettEuler, ETD1, ETDRK4
+  export LawsonEuler, NorsettEuler, ETD1, ETD2, ETDRK4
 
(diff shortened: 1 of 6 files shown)


MSeeker1340:
Sorry for the decay, got through a lot of midterms this week :/

The in-place version is finished, and next I can take a look at numerical stability issues for the operators and maybe apply what I've learnt writing `ETD2` to other exponential operators to bring them up to standard (at least the incorrect interpolation derivatives should be fixed).

Two questions regarding the development of in-place style integrators:

- What does `u_cache` and `du_cache` do exactly? The information on http://devdocs.juliadiffeq.org/latest/contributing/adding_algorithms.html is a bit vague. They are currently just placeholders copied from `NorsettEuler`.

- How is `u` and `uprev` used in the in-place cache and how are they different from `integrator.u` and `integrator.uprev`? They don't appear inside `perform_step!`, but every algorithm seems to include them.

ChrisRackauckas:
>What does u_cache and du_cache do exactly? The information on http://devdocs.juliadiffeq.org/latest/contributing/adding_algorithms.html is a bit vague. They are currently just placeholders copied from NorsettEuler.

That's for `resizing` the integrator during events. That's quite a usage edge case so I'm not worry about that for now. But events and interpolation are two things that should get testing later, after it all works (interpolation tests can be done via turning on dense tests and checking the L2 error, see the partitioned methods tests).

>How is u and uprev used in the in-place cache and how are they different from integrator.u and integrator.uprev? They don't appear inside perform_step!, but every algorithm seems to include them.

I think this had to do with the interpolations, but I have always hoped to be able to refactor in a way that I could get rid of it. Maybe that's already complete. I would need to track it down.

ChrisRackauckas:
Still says WIP, but tests pass and what it covers looks complete. The Krylov version could be added in an additional PR, so I think this is fine to merge.

MSeeker1340:
I was hoping to finish the numerical stability part before leaving for Mexico so I left the WIP there. I will do another PR for them and maybe improvements to the first-order methods (e.g. use caching across the board) after I get back on Thursday.

ChrisRackauckas:
Sounds good. We just won't update the docs yet. I prefer shorter PRs anyways.

PR #274, opened 2018-03-19
Refactoring of exponential integrators

MSeeker1340:
I have updated the exponential integrators using what I've learnt writing `ETD2`. The major changes are:

1. Rewrote both 1st-order algorithms to also use caching for out-of-place style, which makes more sense for constant linear operator and is consistent with what `ETDRK4` has been doing.

2. Use a common FSAL type `ExpRKFsal` for all exponential Runge-Kutta methods to handle the fsaling of the linear/nonlinear part and calculation of the derivatives used in interpolation.

3. Use the `BigFloat` approach employed by `ETDRK4` for phi-related functions to improve numerical stability for very small timestep. (The update to `NorsettEuler` solves the issue of Julia complaining about deprecation when it is used on a scalar equation).

I have also improved the test script by removing the requirement for _SpecialMatrices_. This package hasn't been updated for a while and gives warning when imported. Since only the 2x2 Strang matrix is used, I have opted to write it out explicitly and drop the requirement for the package. This, combined with the update to `NorsettEuler`, should resolve all warning for the linear-nonlinear convergence tests.

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -12,7 +12,7 @@ isfsal(alg::Vern8) = false
 isfsal(alg::Vern9) = false
 
 fsal_typeof(alg::OrdinaryDiffEqAlgorithm,rate_prototype) = typeof(rate_prototype)
-fsal_typeof(alg::Union{LawsonEuler,NorsettEuler},rate_prototype) = ExpRKFsal{typeof(rate_prototype)}
+fsal_typeof(alg::Union{LawsonEuler,NorsettEuler,ETDRK4},rate_prototype) = ExpRKFsal{typeof(rate_prototype)}
 fsal_typeof(alg::ETD2,rate_prototype) = ETD2Fsal{typeof(rate_prototype)}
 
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
For the FSAL type, have you added an event handling test to see if the resets are able to be computed? That's the part that seems fishy.

MSeeker1340:
Haven't looked at event handling yet. Let me do some digging and see if it works with custom FSAL types.

ChrisRackauckas:
It would probably be good to start checking the `isdtchamgable` behavior when there's a tstop that's not eps of a dt multiple. One way to test this is to set it to not a diviser of the tspan. What should happen is it should interpolate back for the non dtchangable algorithms, but that was added before there was any algorithms to test it with...

ChrisRackauckas:
Other problems were moved to https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/276 since they are somewhat orthogonal to the current PR.

PR #275, opened 2018-03-20
WIP: Stiffness detection for `Tsit5` & `DP5`

YingboMa:
This PR is an implementation of the automatic stiffness detection theory that is mentioned in the Hairer II book.

image

### DIFF
--- a/src/perform_step/low_order_rk_perform_step.jl
+++ b/src/perform_step/low_order_rk_perform_step.jl
@@ -503,7 +503,7 @@ end
   integrator.fsallast = f(u, p, t+dt); k7 = integrator.fsallast
   g7 = u
   # Hairer II, page 22
-  ϱ = norm(k7 - k6, Inf)/norm(g7 - g6, Inf)
+  ϱ = integrator.opts.internalnorm(k7 - k6)/integrator.opts.internalnorm(g7 - g6)
   # 3.3 is the approximate distance of ∂S to the origin
   stiffness = ϱ*dt/3.3
   if integrator.opts.adaptive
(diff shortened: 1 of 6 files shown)


YingboMa:
@ChrisRackauckas I am not sure about what is a good way of passing the result of the stiffness detection to the switching algorithm.

src/perform_step/low_order_rk_perform_step.jl
```diff
@@ -617,10 +629,16 @@ end
...
   k5 = f(uprev+dt*(a51*k1+a52*k2+a53*k3+a54*k4), p, t+c4*dt)
-  k6 = f(uprev+dt*(a61*k1+a62*k2+a63*k3+a64*k4+a65*k5), p, t+dt)
+  g6 = uprev+dt*(a61*k1+a62*k2+a63*k3+a64*k4+a65*k5)
+  k6 = f(g6, p, t+dt)
   update = a71*k1+a73*k3+a74*k4+a75*k5+a76*k6
   u = uprev+dt*update
   integrator.fsallast = f(u, p, t+dt); k7 = integrator.fsallast
+  g7 = u
+  # Hairer II, page 22
+  ϱ = norm(k7 - k6, Inf)/norm(g7 - g6, Inf)
```
ChrisRackauckas: Reason for the Inf norm? Also, is there a way to make this non-allocating (that's safe given the way the aliases in the cache are)

ChrisRackauckas:
>I am not sure about what is a good way of passing the result of the stiffness detection to the switching algorithm.

We'll need an `eigen_est` field in the integrator. Then we'll want to add a `stiffness_check` boolean to the common interface to turn off this calculation (maybe, we'll see if it's even costly at all).

ChrisRackauckas:
Is Van Der Pol what Hairer uses in his book figures?

YingboMa:
> Is Van Der Pol what Hairer uses in his book figures?

Yes

ChrisRackauckas:
I'm going to look into the test failure and find out what that is...

ChrisRackauckas:
`rate_prototype` was uninitialized, and the units were incorrect anyways. I made it do a much simpler calculation which should work. 

I think the next steps are

1) if statement the stiffness calculation so it's only done with the algorithm is a CompositeAlgorithm
2) in the mutable cache version, the calculation should be made non-allocating. `k6-k7` for example is allocating. You can be smart with using `utilde` as a cache I think, but be careful with the way that the aliasing works.

Then I think we can merge.

PR #282, opened 2018-03-28
Add the max norm estimation for the largest eigval on `Rodas5`

YingboMa:
I am not sure about whether `if has_invW(f) && !(typeof(integrator.alg) <: CompositeAlgorithm)` is a good idea (probably bad), since it will introduce a matrix factorization for every `!repeat_step` step, even if the user provides an inverse W matrix for the system. However, one needs to have the Jacobian to do the largest eigenvalue estimation. @ChrisRackauckas

### DIFF
--- a/src/perform_step/low_order_rk_perform_step.jl
+++ b/src/perform_step/low_order_rk_perform_step.jl
@@ -734,11 +734,12 @@ end
     g6 = tmp
     g7 = u
     # Hairer II, page 22
-    @. utilde = k7 - k6
-    ϱu = integrator.opts.internalnorm(utilde)
-    @. utilde = g7 - g6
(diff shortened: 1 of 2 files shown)


YingboMa:
Maybe one is able to perform the stiffness detection on the inverse W matrix, however, I have to read more about the Rosenbrock methods family to be certain.

ChrisRackauckas:
the eigenvalues of W inverse are `1/(1-lambda)` the eigenvalues of `J`. Maybe we can tease out a relationship?

Anyways, even if we can't, there's no reason to make it factorize the Jacobian which is the most expensive step. It can at least calculate the Jacobian and take its norm, but then use `invW` to avoid the factorization. For reference, `W = M-gamma*J` where `M` is the mass matrix, usually `I`.

PR #293, opened 2018-03-31
Refactor in-place Rosenbrock methods

YingboMa:


src/derivative_utils.jl
```diff
@@ -0,0 +1,84 @@
...
+
+  end
+end
+
+function calc_jacobian!(integrator, cache, repeat_step)
+  @inbounds begin
+    is_2332 = typeof(cache) <: Rosenbrock23Cache || typeof(cache) <: Rosenbrock32Cache
+    @unpack t,dt,uprev,u,f,p = integrator
+    @unpack du1,du2,fsalfirst,dT,J,W,uf,tf,linsolve_tmp,linsolve_tmp_vec,jac_config,vectmp = cache
+    d1 = cache.tab.d1
```
ChrisRackauckas: Should refactor the SDIRK methods at the same time. I don't think this trick of relying on the same names here will actually carry over well, and we'll want this to be for all stiff solvers as much as possible. I think there may be some things that need to get passed around.

src/derivative_utils.jl
```diff
@@ -0,0 +1,84 @@
...
+        if has_jac(f)
+          f(Val{:jac}, J, uprev, p, t)
+        else
+          uf.t = t
+          jacobian!(J, uf, uprev, du1, integrator, jac_config)
+        end
+        if typeof(integrator.alg) <: CompositeAlgorithm
+          integrator.eigen_est = norm(J, Inf)
+        end
+        if !is_2332
```
ChrisRackauckas: this should probably be an optional arg on the function.

src/derivative_utils.jl
```diff
@@ -0,0 +1,83 @@
...
+    end
+
+    f( fsalfirst,  uprev, p, t)
+
+    @. linsolve_tmp = fsalfirst + dtd1*dT
+
+  end
+end
+
+function calc_jacobian!(integrator, cache, repeat_step, is_2332=false)
```
ChrisRackauckas: Call it `W_transform = false`

ChrisRackauckas:
Thanks!

PR #295, opened 2018-04-03
Fix `alg_autodiff` error for `CompositeAlgorithm`

YingboMa:
The PR is going to fix the error

```julia
julia> alg = AutoRodas5(Tsit5(); maxstiffstep=10, maxnonstiffstep=10, tol=12//10);

julia> sol = solve(prob_ode_brusselator_2d, alg, saveat=[1.5, 11.5]);
ERROR: This algorithm does not have an autodifferentiation option defined.
...
```

### DIFF
--- a/src/perform_step/composite_perform_step.jl
+++ b/src/perform_step/composite_perform_step.jl
@@ -51,6 +51,7 @@ function choose_algorithm!(integrator,cache::CompositeCache)
     reset_alg_dependent_opts!(integrator,integrator.alg.algs[cache.current],integrator.alg.algs[new_current])
     transfer_cache!(integrator,integrator.cache.caches[cache.current],integrator.cache.caches[new_current])
     cache.current = new_current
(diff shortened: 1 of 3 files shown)


src/perform_step/composite_perform_step.jl
```diff
@@ -51,6 +51,7 @@ function choose_algorithm!(integrator,cache::CompositeCache)
     reset_alg_dependent_opts!(integrator,integrator.alg.algs[cache.current],integrator.alg.algs[new_current])
     transfer_cache!(integrator,integrator.cache.caches[cache.current],integrator.cache.caches[new_current])
     cache.current = new_current
+    integrator.alg.current_alg = new_current
```
ChrisRackauckas: move this to right after where it changes just to make sure there's no bug.

ChrisRackauckas:
Merge when tests pass

PR #299, opened 2018-04-07
Add stiffness detection for Verner's methods

YingboMa:


src/alg_utils.jl
```diff
@@ -2,7 +2,7 @@ isautodifferentiable(alg::OrdinaryDiffEqAlgorithm) = true
 
 isfsal(alg::OrdinaryDiffEqAlgorithm) = true
 isfsal{MType,VType,fsal}(tab::ExplicitRKTableau{MType,VType,fsal}) = fsal
-isfsal(alg::CompositeAlgorithm) = true # Every algorithm is assumed FSAL. Good assumption?
```
ChrisRackauckas: I guess it wasn't 👍

PR #301, opened 2018-04-07
Fix the error in `Rodas5` when using `BigFloat`

YingboMa:
This PR is intended to fix the error

```julia
const linear_bigα = parse(BigFloat,"1.01")
f = (du,u,p,t) -> begin
  for i in 1:length(u)
    du[i] = linear_bigα*u[i]
  end
end
(::typeof(f))(::Type{Val{:analytic}},u0,p,t) = u0*exp(linear_bigα*t)
prob_ode_bigfloat2Dlinear = ODEProblem(f,map(BigFloat,rand(4,2)).*ones(4,2)/2,(0.0,1.0))

julia> sol =solve(prob_ode_bigfloat2Dlinear,AutoVern9(Rodas5());dt=1//2^(4))
ERROR: MethodError: no method matching OrdinaryDiffEq.Rodas5ConstantCache(::BigFloat, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::BigFloat, ::BigFloat, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64)
Closest candidates are:
  OrdinaryDiffEq.Rodas5ConstantCache(::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T, ::T2, ::T, ::T, ::T, ::T, ::T, ::T2, ::T2, ::T2, ::T2) where {T, T2}
```


YingboMa:
The type of `gamma` is not consistent across the Rosenbrock family. E.g.
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/tableaus/rosenbrock_tableaus.jl#L130 and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/tableaus/rosenbrock_tableaus.jl#L333

@ChrisRackauckas

ChrisRackauckas:
Open an issue for that. They won't work with units right now because of the linear solves anyways, but we should track it. Both `c1` and the tableau coefficient are `gamma`, so we need to double it...

PR #303, opened 2018-04-08
Implement the `ABDF2` method

YingboMa:
Here is an implementation of the method that is mentioned in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/5#issuecomment-372030378.

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -70,6 +70,7 @@ alg_extrapolates(alg::Hairer4) = true
 alg_extrapolates(alg::Hairer42) = true
 alg_extrapolates(alg::IRKN4) = true
 alg_extrapolates(alg::IRKN3) = true
+alg_extrapolates(alg::ABDF2) = true
 
 alg_order(alg::OrdinaryDiffEqAlgorithm) = error("Order is not defined for this algorithm")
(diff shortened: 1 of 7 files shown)


src/alg_utils.jl
```diff
@@ -70,6 +70,7 @@ alg_extrapolates(alg::Hairer4) = true
 alg_extrapolates(alg::Hairer42) = true
 alg_extrapolates(alg::IRKN4) = true
 alg_extrapolates(alg::IRKN3) = true
+alg_extrapolates(alg::ABDF2) = true
```
ChrisRackauckas: This is good to add, but I don't think it actually extrapolates yet? The extrapolating initial guess is something that should be added in the future though.

YingboMa:
There are the convergence plots.

Out-of-place:
```julia
julia> sim = test_convergence(dts,prob,ABDF2(extrapolant = :linear));

julia> sim.𝒪est[:l∞]
1.9879754678888957

julia> sim.𝒪est[:l2]
1.9901519044978664

julia> sim.𝒪est[:final]
1.9879754678888957
```

prob1

In-place:
```julia
julia> sim = test_convergence(dts,prob,ABDF2(extrapolant = :linear));

julia> sim.𝒪est[:l∞]
1.987975467904488

julia> sim.𝒪est[:l2]
1.9901519044563416

julia> sim.𝒪est[:final]
1.9879754678564714
```

prob2

PR #309, opened 2018-04-11
Add event handling compatibility for `AB/ABM` methods

YingboMa:


src/integrators/integrator_utils.jl
```diff
@@ -338,7 +338,9 @@ function handle_callbacks!(integrator)
   if integrator.u_modified
     handle_callback_modifiers!(integrator)
   end
-  integrator.u_modified = false
+  if !(typeof(integrator.alg) <: MultistepAlgorithms)
+    integrator.u_modified = false
```
ChrisRackauckas: This one may be unnecessary because of later on anyways. I would try deleting it.

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/53c8950acea4aaf46880fbe5329828d3ba010c97/src/integrators/integrator_utils.jl#L341-L343

src/integrators/integrator_utils.jl
```diff
@@ -442,7 +444,9 @@ function reset_fsal!(integrator)
   end
   # Do not set false here so it can be checked in the algorithm
   # integrator.reeval_fsal = false
-  integrator.u_modified = false
+  if !(typeof(integrator.alg) <: MultistepAlgorithms)
+    integrator.u_modified = false
```
ChrisRackauckas: I don't like this. The multistep algorithms shouldn't be treated separately. Instead, the system should be updated to account for this.

ChrisRackauckas:
I am thinking that `integrator.u_modified=false` here

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/53c8950acea4aaf46880fbe5329828d3ba010c97/src/integrators/integrator_utils.jl#L261

might be all that's needed. Basically, `loopheader` needs to know where `u` was modified to reset FSAL. I originally had it set it to false right after so that way next step it would only do it if another modification happened. However, if we just set it to false at the top of the `loopfooter`, it should be the same thing except for the fact that it will be `true` during `perform_step`. If a rejection occurs it will be `false`, which is what we want as well. It sounds like it would work, but needs tests.

PR #314, opened 2018-04-12
Refactor `calc_W!` to always return a boolean

YingboMa:
Refactoring the `calc_W!` function has been discussed in a Slack conversation with @ChrisRackauckas. Thus, I changed the function `calc_W` such that it will always return a boolean type.


ChrisRackauckas:
Looks like I broke tests last night. I'll get on that.

PR #316, opened 2018-04-13
Use DiffEqBase.check_error!

tkf:
https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/276

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -26,8 +26,6 @@ module OrdinaryDiffEq
 
   import Base: linspace
 
-  import Base: start, next, done, eltype
-
   import ForwardDiff.Dual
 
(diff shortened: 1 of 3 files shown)


devmotion:
Is `@ode_exit_conditions` only used inside of `solve!` in OrdinaryDiffEq and DelayDiffEq? Since it consists only of one line anymore we could remove it completely, I guess.

ChrisRackauckas:
Yeah, we just need to make sure we handle the returns. It's a macro because if it's a function then it won't fully return out of `solve!`.

ChrisRackauckas:
Should restart after https://github.com/JuliaLang/METADATA.jl/pull/14250

ChrisRackauckas:
Test failure is some Rosenbrock thing where Travis doesn't finish them sometimes which I don't understand...

Issue #317, opened 2018-04-13
Properties of CompositeAlgorithm


devmotion:
I'm wondering whether the properties of `CompositeAlgorithm` defined in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/alg_utils.jl should be more consistent and/or have to be extended. More precisely, I'm interested in why some methods such as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/alg_utils.jl#L5 refer to the properties of the current algorithm, whereas others such as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/alg_utils.jl#L197 refer to the first algorithm. Shouldn't properties of a `CompositeAlgorithm` always refer to the properties of the current algorithm? Additionally, many methods are not defined for `CompositeAlgorithm` specifically, e.g. `fsal_typeof`, `is_implicit`, and many more will only return defaults for `OrdinaryDiffEqAlgorithm`. Moreover, it's a bit surprising that in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/alg_utils.jl#L34 a method is defined for `ODECompositeAlgorithm` whereas otherwise all definitions only apply to `CompositeAlgorithm`.

ChrisRackauckas:
>Shouldn't properties of a CompositeAlgorithm always refer to the properties of the current algorithm? 

Yes. This should get updated. 

>Additionally, many methods are not defined for CompositeAlgorithm specifically, e.g. fsal_typeof, is_implicit, and many more will only return defaults for OrdinaryDiffEqAlgorithm

Those are needed. 

>Moreover, it's a bit surprising that in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/alg_utils.jl#L34 a method is defined for ODECompositeAlgorithm whereas otherwise all definitions only apply to CompositeAlgorithm.

Yeah, I wasn't sure at first whether there would be multiple composite algortihms. I think at this point that everything will be `CompositeAlgorithm` with a choice function, so these should probably all be on the concrete type.

ChrisRackauckas:
Actually, I remember why I did this. `CompositeAlgorithm` isn't thread-safe if you use `current_alg` inside of it. All `current_alg` choices should be part of the `cache` instead. That makes the traits much harder though, and should be picked on an all basis?

PR #322, opened 2018-04-14
Make the properties of `CompositeAlgorithm` consistent

YingboMa:
This PR addresses the issue #317 .

CC: @ChrisRackauckas @devmotion

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -14,12 +14,19 @@ isfsal(alg::Vern9) = false
 fsal_typeof(alg::OrdinaryDiffEqAlgorithm,rate_prototype) = typeof(rate_prototype)
 fsal_typeof(alg::Union{LawsonEuler,NorsettEuler,ETDRK4},rate_prototype) = ExpRKFsal{typeof(rate_prototype)}
 fsal_typeof(alg::ETD2,rate_prototype) = ETD2Fsal{typeof(rate_prototype)}
+function fsal_typeof(alg::CompositeAlgorithm,rate_prototype)
+  fsal = unique(fsal_typeof.(alg.algs))
+  @assert length(fsal) == 1 "`fsal_typeof` must be consistent"
+  return fsal[1]
+end
 
 isimplicit(alg::OrdinaryDiffEqAlgorithm) = false
 isimplicit(alg::OrdinaryDiffEqAdaptiveImplicitAlgorithm) = true
 isimplicit(alg::OrdinaryDiffEqImplicitAlgorithm) = true
+isimplicit(alg::CompositeAlgorithm) = any(isimplicit.(alg.algs))
 
 isdtchangeable(alg::OrdinaryDiffEqAlgorithm) = true
+isdtchangeable(alg::CompositeAlgorithm) = all(isdtchangeable.(alg.algs))
 isdtchangeable(alg::GenericIIF1) = false
 isdtchangeable(alg::GenericIIF2) = false
 isdtchangeable(alg::LawsonEuler) = false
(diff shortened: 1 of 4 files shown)


src/alg_utils.jl
```diff
@@ -31,7 +31,7 @@ ismultistep(alg::ETD2) = true
 
 isadaptive(alg::OrdinaryDiffEqAlgorithm) = false
 isadaptive(alg::OrdinaryDiffEqAdaptiveAlgorithm) = true
-isadaptive(alg::OrdinaryDiffEqCompositeAlgorithm) = isadaptive(alg.algs[1])
+isadaptive(alg::OrdinaryDiffEqCompositeAlgorithm) = isadaptive(alg.algs[alg.current_alg])
```
ChrisRackauckas: This is used to choose the default for adaptivity on/off. I think it should be `all(isadaptive(...))`?

ChrisRackauckas:
I agree with @devmotion though that we should be forwarding a lot more of the traits. For some of them, like `fsal_typeof`, we may have to error if they don't all agree.

src/alg_utils.jl
```diff
@@ -14,29 +14,39 @@ isfsal(alg::Vern9) = false
...
+function fsal_typeof(alg::CompositeAlgorithm,rate_prototype)
+  fsal = unique(fsal_typeof.(alg.algs))
+  @assert length(fsal) == 1 "`fsal_typeof` must be consistent"
+  return fsal[1]
+end
 
 isimplicit(alg::OrdinaryDiffEqAlgorithm) = false
 isimplicit(alg::OrdinaryDiffEqAdaptiveImplicitAlgorithm) = true
 isimplicit(alg::OrdinaryDiffEqImplicitAlgorithm) = true
+isimplicit(alg::CompositeAlgorithm) = all(isimplicit.(alg.algs))
```
ChrisRackauckas: I'd go with `any` here. This isn't actually used internally at all though.

src/alg_utils.jl
```diff
@@ -73,7 +84,9 @@ alg_extrapolates(alg::IRKN3) = true
 alg_extrapolates(alg::ABDF2) = true
 
 alg_order(alg::OrdinaryDiffEqAlgorithm) = error("Order is not defined for this algorithm")
+alg_order(alg::CompositeAlgorithm) = (alg_order.(alg.algs)...)
 alg_adaptive_order(alg::OrdinaryDiffEqAdaptiveAlgorithm) = error("Algorithm is adaptive with no order")
+alg_adaptive_order(alg::CompositeAlgorithm) = (alg_adaptive_order.(alg.algs)...)
```
ChrisRackauckas: This might break adaptivity?

src/alg_utils.jl
```diff
@@ -73,7 +84,9 @@ alg_extrapolates(alg::IRKN3) = true
 alg_extrapolates(alg::ABDF2) = true
 
 alg_order(alg::OrdinaryDiffEqAlgorithm) = error("Order is not defined for this algorithm")
+alg_order(alg::CompositeAlgorithm) = (alg_order.(alg.algs)...)
 alg_adaptive_order(alg::OrdinaryDiffEqAdaptiveAlgorithm) = error("Algorithm is adaptive with no order")
+alg_adaptive_order(alg::CompositeAlgorithm) = alg_adaptive_order(alg.algs[alg.current_alg])
```
ChrisRackauckas: this'll not be thread-safe. `current_alg` shouldn't be used, in fact it should probably just be removed here. Instead, the internals of the adaptive algorithm should have a `get_current_adaptive_order(alg,cache)` that returns the current order (for most, it's just `alg_adaptive_order`, and `CompositeAlgorithm` gets a dispatch that it uses the current from the cache, not the alg).

ChrisRackauckas:
MWTP

PR #323, opened 2018-04-15
Make stiff methods other than Rosenbrock compatible with switching algorithms

YingboMa:
Currently, composite algorithms are not compatible with a lot of implicit methods, since lines like https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/kencarp_kvaerno_perform_step.jl#L53. And this PR aims to fix them.


ChrisRackauckas:
Use `integrator.cache.current` instead of the current from the algorithm. In fact, the algorithm should be made a struct and that should be removed.

PR #324, opened 2018-04-16
Make ComposteAlgorithm a struct

ChrisRackauckas:


ChrisRackauckas:
So it's still not thread-safe?

YingboMa:
I am going to give it a try tomorrow.

src/solve.jl
```diff
@@ -35,8 +35,8 @@ function init{algType<:OrdinaryDiffEqAlgorithm,recompile_flag}(
   qsteady_max = qsteady_min_default(alg),
   qoldinit=1//10^4, fullnormalize=true,
   failfactor = 2,
-  beta2=beta2_default(alg),
-  beta1=beta1_default(alg,beta2),
+  beta2=nothing,
```
ChrisRackauckas: does this even work?

src/solve.jl
```diff
@@ -247,10 +227,34 @@ function init{algType<:OrdinaryDiffEqAlgorithm,recompile_flag}(
 
   if typeof(alg) <: OrdinaryDiffEqCompositeAlgorithm
```
ChrisRackauckas: oh got it

ChrisRackauckas:
LGTM if the tests end up passing.

ChrisRackauckas:
A similar change probably needs to be done on stochastic.

Issue #326, opened 2018-04-16
Roadmap for implementing JVODE


YingboMa:
JVODE is an aptly named solver for **Julia** **V**ariable-coefficient **O**rdinary **D**ifferential **E**quations solver, which basically has the same functionalities as CVODE from SUNDIALS. It is going to provide Adams and BDF methods in Nordsieck forms. These two families of methods are able to share Nordsieck vector. It makes the implementation of stiffness detection and automatic switching almost trivial. Here are some documents that can be useful.
https://dl.acm.org/citation.cfm?id=355903
https://dl.acm.org/citation.cfm?id=355636

And here is a checklist for the implementation:
  - [ ] Variable-order variable-stepsize Adams methods in Nordsieck forms.
    - Eigenvalue estimation
  - [ ] Variable-order variable-stepsize BDF methods in Nordsieck forms.

PR #327, opened 2018-04-17
WIP: Add a fifth order Adams method in Nordsieck form

YingboMa:
This is very experimental. It is my first attempt at implementing #326. I am using `MVector`s for storing the coefficients that are used in Nordsieck form. This might be a bad idea, and a better solution would be using a generated function or a macro to generate completely unrolled code from caches. I am going to submit another PR to optimize the coefficient routine if necessary.


src/nordsieck_utils.jl
```diff
@@ -0,0 +1,48 @@
...
+
+# `l` is the coefficients of the polynomial `Λ` that satisfies conditions
+# Λ(0) = 1, Λ(-1) = 0, and Λ̇(-ξᵢ) = 0, where ξᵢ = (tₙ-tₙ₋₁)/dt.
+# It is described in the paper "A Polyalgorithm for the Numerical Solution
+# of Ordinary Differential Equations" by G. D. Byrne and A. C. Hindmarsh in
+# the page 86.
+# https://dl.acm.org/citation.cfm?id=355636
+function calc_coeff!(cache)
+  @inbounds begin
+    @unpack m, l, tau = cache
```
sipah00: unpack `tq` also as used in line `46`

YingboMa: No, `tq` is changed in-place. Also, it is still WIP, so it is not ready for review.

src/perform_step/adams_nordsieck_perform_step.jl
```diff
@@ -0,0 +1,123 @@
...
+  z[1] = uprev
+  # handle callbacks, rewind back to order one.
+  if integrator.u_modified
+    cache.step = 1
+  end
+  # Nordsieck form needs to build the history vector, so one has to start from order one.
+  if cache.step <= 4
+    if cache.step == 1
+      # Grow the Nordsieck vector to length 2+1
+      perform_step!(integrator, EulerConstantCache(), repeat_step)
```
ChrisRackauckas: Just do Tsit5 here?

src/caches/adams_nordsieck_caches.jl
```diff
@@ -0,0 +1,81 @@
...
+  # `m` is a tmp vector that is used for calculating `l`
+  m::MVector{6,lType}
+  # `tq` is used for the error estimation for the current order
+  tq::lType
+  # `tau` stores `dt`s
+  tau::MVector{6, dtType}
+  # `Δ` is the difference between the predictor `uₙ₀` and `uₙ`
+  Δ::uType
+  # `Tsit5` for the first step
+  vern9tab::vern9Type
```
ChrisRackauckas: ? comment

src/nordsieck_utils.jl
```diff
@@ -0,0 +1,141 @@
...
+  end
+end
+
+# Apply the Pascal linear operator
+function perform_predict!(cache::T, undo) where T
+  @inbounds begin
+    isconst = T <: OrdinaryDiffEqConstantCache
+    isconst || (cache = cache.const_cache)
+    @unpack z,step = cache
+    ± = undo ? (-) : (+)
```
ChrisRackauckas: won't this cause a bad type instability in the operator? I assume this could slow it down a lot, so it may be better to add a function barrier or just put multiplication in there.

test/ode/ode_convergence_tests.jl
```diff
@@ -95,6 +95,15 @@ for i = 1:2
   @test abs(sim.𝒪est[:l2]-2) < testTol
   @test abs(sim.𝒪est[:l∞]-2) < testTol
 
+  #####################################
+  # Nordsieck
+  #####################################
+
+  sim = test_convergence(dts1[2:end],prob,AN5())
+  @test abs(sim.𝒪est[:final]-5) < 0.4
```
ChrisRackauckas: stricter

src/nordsieck_utils.jl
```diff
@@ -0,0 +1,141 @@
...
+    return int
+  end
+end
+
+# `l` is the coefficients of the polynomial `Λ` that satisfies conditions
+# Λ(0) = 1, Λ(-1) = 0, and Λ̇(-ξᵢ) = 0, where ξᵢ = (tₙ-tₙ₋₁)/dt.
+# It is described in the paper "A Polyalgorithm for the Numerical Solution
+# of Ordinary Differential Equations" by G. D. Byrne and A. C. Hindmarsh in
+# the page 86.
+# https://dl.acm.org/citation.cfm?id=355636
```
ChrisRackauckas: Should comment about the double for loop Newton polynomial part? That's not in the paper?

PR #341, opened 2018-04-28
Remove undefined exports

tkf:
Writing a patch before I forget https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/105#issuecomment-385138763

It needs new DiffEqBase to be released.

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -37,7 +37,8 @@ module OrdinaryDiffEq
                      resize_non_user_cache!,deleteat_non_user_cache!,addat_non_user_cache!,
                      terminate!,get_du, get_dt,get_proposed_dt,set_proposed_dt!,
                      u_modified!,savevalues!,add_tstop!,add_saveat!,set_reltol!,
-                     set_abstol!, postamble!, last_step_failed
+                     set_abstol!, postamble!, last_step_failed,
+                     isautodifferentiable
 
   using DiffEqBase: check_error!
 
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -135,9 +135,7 @@ module OrdinaryDiffEq
   export ode_addsteps!, ode_interpolant,
         terminate!, savevalues!, copyat_or_push!, isfsal
 
-  export constructDP5, constructVern6, constructDP8,
-         constructDormandPrince, constructFeagin10,
-         constructFeagin12, constructFeagin14
+  export constructDormandPrince
 
   # Reexport the Alg Types
 
--- a/test/export_tests.jl
+++ b/test/export_tests.jl
@@ -1,0 +1,5 @@
+using DiffEqBase
+using OrdinaryDiffEq
+using Base.Test
+
+@test DiffEqBase.undefined_exports(OrdinaryDiffEq) == []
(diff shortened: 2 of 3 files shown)


src/OrdinaryDiffEq.jl
```diff
@@ -134,9 +135,7 @@ module OrdinaryDiffEq
   export ode_addsteps!, ode_interpolant,
         terminate!, savevalues!, copyat_or_push!, isfsal
 
-  export constructDP5, constructVern6, constructDP8,
-         constructDormandPrince, constructFeagin10,
-         constructFeagin12, constructFeagin14
```
tkf: FYI there is `Feagin10ConstantCache`. Maybe you want to export `Feagin10ConstantCache`?

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/47e9711877b3a6767364623fd0af8d404b68488c/src/tableaus/feagin_tableaus.jl#L269-L273

ChrisRackauckas: Nope, those tableaus shouldn't be exported. This should just be removed. The tableaus that should be exported are now in DiffEqDevTools

src/OrdinaryDiffEq.jl
```diff
@@ -37,7 +37,8 @@ module OrdinaryDiffEq
                      resize_non_user_cache!,deleteat_non_user_cache!,addat_non_user_cache!,
                      terminate!,get_du, get_dt,get_proposed_dt,set_proposed_dt!,
                      u_modified!,savevalues!,add_tstop!,add_saveat!,set_reltol!,
-                     set_abstol!, postamble!, last_step_failed
+                     set_abstol!, postamble!, last_step_failed,
+                     isautodifferentiable
```
ChrisRackauckas: I don't think this needs to be exported?

src/OrdinaryDiffEq.jl
tkf: I see.

src/OrdinaryDiffEq.jl
tkf: I think it's handled by `@reexport using DiffEqBase`.

```
julia> using OrdinaryDiffEq

help?> isautodifferentiable
search: isautodifferentiable

  No documentation found.

  DiffEqBase.isautodifferentiable is a Function.

  # 2 methods for generic function "isautodifferentiable":
  isautodifferentiable(alg::OrdinaryDiffEq.OrdinaryDiffEqAlgorithm) in OrdinaryDiffEq at /home/takafumi/.julia/v0.6/OrdinaryDiffEq/src/alg_utils.jl:1
  isautodifferentiable(alg::DiffEqBase.DEAlgorithm) in DiffEqBase at /home/takafumi/.julia/v0.6/DiffEqBase/src/alg_traits.jl:1
```

tkf: Oops. I read your sentence wrongly. But `isautodifferentiable` is jsut imported from `DiffEqBase` so I suppose it's fine?

If you don't want to export `isautodifferentiable` from OrdinaryDiffEq, I guess we need to avoid `@reexport` in OrdinaryDiffEq.jl or remove `export isautodifferentiable` from DiffEqBase.jl.

tkf: Yeah, exporting less from DifferentialEquations packages sounds like a good idea.  I personally like explicit import (`using DiffEqBase: stuff, I, need`) to make code clearer.  Just an idea, but one way to get rid of `export` gradually is to go this explicit import style. Once it's done for all packages, `export` would become irrelevant so you can make it minimal.

ChrisRackauckas: Yeah, it's the upgrading the packages that's a lot of work. I still want to export the public API, but all of those abstract types isn't necessary to export.

PR #344, opened 2018-04-30
3rd order adams explicit variable step size method

sipah00:


src/caches/adams_bashforth_moulton_caches.jl
```diff
@@ -292,6 +292,29 @@ mutable struct VSA3ConstantCache{rateType ,uArrayType, TabType, bool} <: Ordinar
   tab::TabType
 end
 
+mutable struct VSA3Cache{uType,rateType,bool,TabType,uArrayType,bs3Type,uEltypeNoUnits,coefType} <: OrdinaryDiffEqMutableCache
```
sipah00: @ChrisRackauckas

src/caches/adams_bashforth_moulton_caches.jl
```diff
@@ -277,3 +277,80 @@ function alg_cache(alg::ABM54,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
...
+  ϕstar_nm1::uArrayType
+  grid_points::uArrayType
+  k::Int
+  idx::Int
+  order::Int
+  success::bool
+  tab::TabType
+end
+
+mutable struct VSA3Cache{uType,rateType,bool,TabType,uArrayType,bs3Type,uEltypeNoUnits,coefType} <: OrdinaryDiffEqMutableCache
```
ChrisRackauckas: this needs another type parameter

src/caches/adams_bashforth_moulton_caches.jl
```diff
@@ -277,3 +277,80 @@ function alg_cache(alg::ABM54,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
...
+  k3::rateType
+  k4::rateType
+  ϕstar_nm1::coefType
+  grid_points::uArrayType
+  k::Int
+  idx::Int
+  order::Int
+  atmp::uEltypeNoUnits
+  tmp::uType
+  utilde::uArrayType
```
ChrisRackauckas: the value which has the wrong type, the array of array, needs that type parameter. Was it this one?

src/caches/adams_bashforth_moulton_caches.jl
```diff
@@ -292,6 +292,29 @@ mutable struct VSA3ConstantCache{rateType ,uArrayType, TabType, bool} <: Ordinar
...
 
+mutable struct VSA3Cache{uType,rateType,bool,TabType,uArrayType,bs3Type,uEltypeNoUnits,coefType} <: OrdinaryDiffEqMutableCache
+  u::uType
+  uprev::uType
+  fsalfirst::rateType
+  bs3cache::bs3Type
+  k2::rateType
+  k3::rateType
+  k4::rateType
+  ϕstar_nm1::coefType
```
sipah00: @ChrisRackauckas , this is coef array

src/caches/adams_bashforth_moulton_caches.jl
sipah00: here `!Matched::uArrayType`

ChrisRackauckas: yeah, is it supposed to be that type? I don't think utilde is ever supposed to be anything other than a uType if it's following the standard conventions?

sipah00: I followed standard conventions.

sipah00: I just checked it one an example (manually) and it worked fine.

ChrisRackauckas: `grid_points::uArrayType` is that correct? Do `grid_points` and `utilde` have the same type?

sipah00:
```

Comment (sipah00) [reply src/caches/adams_bashforth_moulton_caches.jl]:
```

sipah00: yeah `grid_points` is an array of times (mesh points)

sipah00: but is it causing problem in test?

ChrisRackauckas: yes, that's why it's not matching. `grid_points` is setting `uArrayType` to be `Array{Float64}`, while `uType` is `Array{Float64}, so `uArrayType` should be `Array{Array{Float64}}` and so it doesn't match.

sipah00: ohh, got it, thanks

src/adams_utils.jl
```diff
@@ -0,0 +1,56 @@
+function ϕ_and_ϕstar!(cache, dy, next_point, last_idx)
+	@unpack grid_points, ϕstar_nm1, k = cache
+	ϕ_n = []
```
ChrisRackauckas: Allocation, and this is not strictly typed. This will dramatically slow the code down. You should strictly type this and cache the array.

src/adams_utils.jl
```diff
@@ -0,0 +1,56 @@
...
+	ϕstar_n = []
+	if typeof(dy) <: Array
+		ϕ_n = Array{Array{Float64}}(1,k)
```
ChrisRackauckas: Another allocation, also fixed to Float64

src/adams_utils.jl
```diff
@@ -0,0 +1,56 @@
...
+    	else
+    		ϕ_np1[i] = ϕ_np1[i-1] - ϕstar_n[i-1]
+    	end
+    end
+    return ϕ_np1
+end
+
+function g_coefs!(cache, dt, next_point, last_idx)
+	@unpack grid_points, k = cache
+	c = zeros(Float64, k, k)
```
ChrisRackauckas: another allocation

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -544,3 +544,142 @@ end
...
+
+  if cnt == 1 || cnt == 2
+    perform_step!(integrator, bs3cache)
+    @unpack k4 = bs3cache
+    integrator.fsallast .= k4
+    cache.k = min(k+1, order)
+    if cnt == 1
+        cache.k3 = copy(k1)
+    else
+        cache.k2 = copy(k1)
```
ChrisRackauckas: Another allocation

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -544,3 +544,142 @@ end
...
+  if cnt <= 3
+    idx = cnt
+    grid_points[idx] = t
+   elseif success
+    grid_points[1] = grid_points[2]
+    grid_points[2] = grid_points[3]
+    grid_points[3] = t
+  end
+
+  cache.grid_points = copy(grid_points)
```
ChrisRackauckas: another allocation

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -544,3 +544,142 @@ end
...
+    if integrator.opts.adaptive
+      @. utilde = uprev + (dt/12)*(23*k1 - 16*k2 + 5*k3) - u
+      calculate_residuals!(atmp, utilde, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
+      integrator.EEst = integrator.opts.internalnorm(atmp)
+      if integrator.EEst >= 1
+        cache.success = false
+      end
+    end
+    cache.k = min(k+1, order)
+    cache.k3 = copy(k2)
```
ChrisRackauckas: more allocations

ChrisRackauckas:
It looks like it works, so I'll merge it after a name change, but it's far far far from done.

Name change: Let's call it VCAB3 for Variable Coefficient Adams-Bashforth 3. Variable Coefficient because that's one of the two main ways to do adaptivity of a multistep method, the other being the Nordesiek way that @YingboMa is doing. So do the rename, then we can merge, and then we can optimize this a little bit because it is allocating a ton.

sipah00:
Okay, I will do required changes.

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -574,14 +574,18 @@ end
     grid_points[3] = t
   end
 
-  cache.grid_points = copy(grid_points)
+  cache.grid_points .= grid_points
```
ChrisRackauckas: why is there two sets of grid points?

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -544,3 +544,150 @@ end
...
+  if cnt <= 3
+    idx = cnt
+    grid_points[idx] = t
+   elseif success
+    grid_points[1] = grid_points[2]
+    grid_points[2] = grid_points[3]
+    grid_points[3] = t
+  end
+
+  cache.grid_points .= grid_points
```
ChrisRackauckas: This should be unnecessary since it's the same array? `@unpack k2,k3,grid_points,ϕstar_nm1,k,order,idx,success,tab = cache`

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -544,3 +544,150 @@ end
...
+    for i = 0:k-1
+        u += dt * g[(i)+1] * ϕstar_n[(i)+1]
+    end
+
+    if integrator.opts.adaptive
+      utilde = uprev + (dt/12)*(23*k1 - 16*k2 + 5*k3) - u
+      atmp = calculate_residuals(utilde, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
+      integrator.EEst = integrator.opts.internalnorm(atmp)
+      if integrator.EEst >= 1
+        cache.success = false
```
ChrisRackauckas: @YingboMa how are you doing the check on this? We should make sure it's uniform between algorithms.

src/perform_step/adams_bashforth_moulton_perform_step.jl
sipah00: no need of the second one, I will remove one.

ChrisRackauckas: Neither of them should be necessary though. They are the same array.

src/adams_utils.jl
```diff
@@ -0,0 +1,42 @@
+function ϕ_and_ϕstar!(cache, dy, next_point, last_idx)
```
gabrielgellner: Should these not have 2 space indentations to be in the style of the rest of the package?

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -544,3 +544,150 @@ end
...
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack k2,k3,grid_points,ϕstar_nm1,k,order,idx,success,tab = cache
+  
+  k1 = integrator.fsalfirst
+  cnt = integrator.iter
+
+  if cnt <= 3
+    idx = cnt
+    grid_points[idx] = t
+   elseif success
```
gabrielgellner: looks like extra space at start of elseif.

src/perform_step/adams_bashforth_moulton_perform_step.jl
YingboMa: No, I didn't do this check. Maybe instead of having a boolean type for `cache.success`, we use an enum type for something like `cache.step_retcode` or `integrator.step_retcode`? Then, we can work our way to better logging.

src/adams_utils.jl
ChrisRackauckas: they should. Good catch

src/perform_step/adams_bashforth_moulton_perform_step.jl
YingboMa: This is what I did for `ABDF2`
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/bdf_perform_step.jl#L141-L143

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -544,3 +544,150 @@ end
...
+  integrator.k[1] = integrator.fsalfirst
+  integrator.k[2] = integrator.fsallast
+end
+
+@muladd function perform_step!(integrator,cache::VSA3ConstantCache,repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack k2,k3,grid_points,ϕstar_nm1,k,order,idx,success,tab = cache
+  
+  k1 = integrator.fsalfirst
+  cnt = integrator.iter
```
YingboMa: Instead of using the counter from `integrator`, you may want to use a counter in the cache to handle callbacks. See https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/adams_bashforth_moulton_perform_step.jl#L510-L515

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -544,3 +544,150 @@ end
...
+
+  if cnt <= 3
+    idx = cnt
+    grid_points[idx] = t
```
YingboMa: I think that using a grid of `t`s is not a good idea here, since you are going to compute `grid_points[last_idx] - grid_points[last_idx-i]` and [catastrophic cancellation](https://en.wikipedia.org/wiki/Loss_of_significance) will occur under floating point arithmetic. Store `dt`s instead. By the way, I used the variable `tau` to store past `dt`s. https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/caches/adams_nordsieck_caches.jl#L11

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -544,3 +544,150 @@ end
...
+  ϕ_n, ϕstar_n = ϕ_and_ϕstar!(cache, k1, next_point, last_idx)
+
+  if cnt < 4
+    cache.ϕstar_nm1 = copy(ϕstar_n)
+  else
+    cache.ϕstar_nm1 .= ϕstar_n
+  end
+
+  if cnt == 1 || cnt == 2
+    perform_step!(integrator, tab)
```
YingboMa: As an optimization, we can hot start the algorithm with only one step of RK and use Hermite interpolation to fill the `k` and `k2`.

YingboMa:
This is a good start, though much optimization can go into this algorithm. Also, the fields `order` and `k` seem superfluous. If you have an internal counter in the cache, those two won't be needed.

sipah00:
@ChrisRackauckas @gabrielgellner @YingboMa thanks for the review, I will do suggested changes.

src/adams_utils.jl
```diff
@@ -0,0 +1,42 @@
...
+  @unpack grid_points, ϕstar_nm1, ϕ_n, ϕstar_n,β,k = cache
+  if k < 3
+    if typeof(dy) <: Array
+      ϕ_n = Array{Array{eltype(dy)}}(1,k)
```
ChrisRackauckas: allocations

src/adams_utils.jl
```diff
@@ -0,0 +1,42 @@
...
+      ϕ_n = zeros(typeof(dy),k)
+      ϕstar_n = zeros(typeof(dy),k)
+    end
+    β = zeros(eltype(grid_points),k)
+  end
+  for i = 0:(k)-1
+    if i == 0
+      β[(i)+1] = 1
+      ϕ_n[(i)+1] = copy(dy)
+      ϕstar_n[(i)+1] = copy(dy)
```
ChrisRackauckas: allocations

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -544,3 +544,123 @@ end
...
+    grid_points[cnt] = t
+  else
+    grid_points[1] = grid_points[2]
+    grid_points[2] = grid_points[3]
+    grid_points[3] = t
+  end
+  next_point = t+dt
+  ϕ_n, ϕstar_n = ϕ_and_ϕstar!(cache, k1, next_point, k)
+  if cnt < 4
+    cache.ϕstar_nm1 = copy(ϕstar_n)
```
ChrisRackauckas: allocations

ChrisRackauckas:
I found three spots with allocations, but I don't want this PR to just get held up optimizing. Let's merge and I'll open an issue for further discussion of performance, and it can be addressed in a future PR.

Issue #346, opened 2018-05-10
Allocations in VCAB3


ChrisRackauckas:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/344#pullrequestreview-118962830

We should finish caching all of the little pieces.

PR #347, opened 2018-05-13
WIP : allocations in VCAB3

sipah00:
#346

### DIFF
--- a/src/adams_utils.jl
+++ b/src/adams_utils.jl
@@ -1,15 +1,5 @@
 function ϕ_and_ϕstar!(cache, dy, next_point, last_idx)
   @unpack grid_points, ϕstar_nm1, ϕ_n, ϕstar_n,β,k = cache
-  if k < 3
-    if typeof(dy) <: Array
-      ϕ_n = Array{Array{eltype(dy)}}(1,k)
-      ϕstar_n = Array{Array{eltype(dy)}}(1,k)
-    else
-      ϕ_n = zeros(typeof(dy),k)
-      ϕstar_n = zeros(typeof(dy),k)
-    end
-    β = zeros(eltype(grid_points),k)
-  end
   for i = 0:(k)-1
     if i == 0
       β[(i)+1] = 1
--- a/src/adams_utils.jl
+++ b/src/adams_utils.jl
@@ -3,8 +3,15 @@ function ϕ_and_ϕstar!(cache, dy, next_point, last_idx)
   for i = 0:(k)-1
     if i == 0
       β[(i)+1] = 1
-      ϕ_n[(i)+1] = copy(dy)
-      ϕstar_n[(i)+1] = copy(dy)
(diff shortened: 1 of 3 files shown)


src/adams_utils.jl
```diff
@@ -3,8 +3,15 @@ function ϕ_and_ϕstar!(cache, dy, next_point, last_idx)
   for i = 0:(k)-1
     if i == 0
       β[(i)+1] = 1
-      ϕ_n[(i)+1] = copy(dy)
-      ϕstar_n[(i)+1] = copy(dy)
+      if typeof(dy) <: Array
+        # ϕ_n[(i)+1] = copy(dy)
+        ϕ_n[(i)+1] .= dy
+        # ϕstar_n[(i)+1] .= dy
```
sipah00: @ChrisRackauckas Why broadcasting is not working (convergence tests are failing) here. I could not figure out what is the problem.

americast: :thinking:

ChrisRackauckas: so it passes when it's a copy?

ChrisRackauckas: Are all of the `ϕstar_n` vectors initialized in a way that they alias?

sipah00: yeah

src/caches/adams_bashforth_moulton_caches.jl
```diff
@@ -349,12 +349,17 @@ function alg_cache(alg::VCAB3,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
   k2 = zeros(rate_prototype)
   k3 = zeros(rate_prototype)
   k4 = zeros(rate_prototype)
-  ϕstar_nm1 = Array{typeof(rate_prototype)}(1,3)
   grid_points = zeros(typeof(t),3)
   c = zeros(typeof(t),3,3)
   g = zeros(typeof(t),3)
   ϕ_n = Array{typeof(rate_prototype)}(1,3)
+  ϕstar_nm1 = Array{typeof(rate_prototype)}(1,3)
   ϕstar_n = Array{typeof(rate_prototype)}(1,3)
```
sipah00: @ChrisRackauckas initialization

src/caches/adams_bashforth_moulton_caches.jl
```diff
@@ -349,12 +349,17 @@ function alg_cache(alg::VCAB3,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
...
+  for i in 1:3
+    ϕ_n[i] = zeros(rate_prototype)
+    ϕstar_nm1[i] = zeros(rate_prototype)
+    ϕstar_n[i] = zeros(rate_prototype)
```
ChrisRackauckas: doesn't look like these should alias. Try printing out the `ϕstar_n` values and see if every vector is the same.

PR #348, opened 2018-05-14
WIP: Optimize and clean up `VCAB3`

YingboMa:
TODO:

- [x] Remove unnecessary allocations (by @sipah00)
- [x] Use `dts` instead of `grid_points` to avoid [catastrophic cancellation](https://en.wikipedia.org/wiki/Loss_of_significance)
- [ ] Use an internal counter to handle callbacks

### DIFF
--- a/src/adams_utils.jl
+++ b/src/adams_utils.jl
@@ -1,24 +1,27 @@
 function ϕ_and_ϕstar!(cache, dy, next_point, last_idx)
   @unpack grid_points, ϕstar_nm1, ϕ_n, ϕstar_n,β,k = cache
-  for i = 0:(k)-1
+  for i = 0:k-1
     if i == 0
-      β[(i)+1] = 1
-      if typeof(dy) <: Array
-        # ϕ_n[(i)+1] = copy(dy)
-        ϕ_n[(i)+1] .= dy
-        # ϕstar_n[(i)+1] .= dy
-        ϕstar_n[(i)+1] = copy(dy)
+      β[1] = 1
+      if typeof(cache) <: OrdinaryDiffEqMutableCache
+        ϕ_n[1] .= dy
+        ϕstar_n[1] .= dy
       else
-        ϕ_n[(i)+1] = dy
-        ϕstar_n[(i)+1] = dy
+        ϕ_n[1] = dy
+        ϕstar_n[1] = dy
       end
     else
-      β[(i)+1] = β[i] * (next_point - grid_points[last_idx-i+1])/(grid_points[last_idx] - grid_points[last_idx-i])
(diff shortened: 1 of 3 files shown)


src/adams_utils.jl
```diff
@@ -1,42 +1,49 @@
-function ϕ_and_ϕstar!(cache, dy, next_point, last_idx)
-  @unpack grid_points, ϕstar_nm1, ϕ_n, ϕstar_n,β,k = cache
-  if k < 3
-    if typeof(dy) <: Array
-      ϕ_n = Array{Array{eltype(dy)}}(1,k)
-      ϕstar_n = Array{Array{eltype(dy)}}(1,k)
+function ϕ_and_ϕstar!(cache, du)
+  begin
```
ChrisRackauckas: forgot to put the `@inbounds`

src/adams_utils.jl
```diff
@@ -1,42 +1,49 @@
-function ϕ_and_ϕstar!(cache, dy, next_point, last_idx)
-  @unpack grid_points, ϕstar_nm1, ϕ_n, ϕstar_n,β,k = cache
-  if k < 3
-    if typeof(dy) <: Array
-      ϕ_n = Array{Array{eltype(dy)}}(1,k)
-      ϕstar_n = Array{Array{eltype(dy)}}(1,k)
+function ϕ_and_ϕstar!(cache, du)
```
ChrisRackauckas: can you put a comment for the reference of this function and its implementation?

src/adams_utils.jl
```diff
@@ -1,42 +1,49 @@
...
     end
-    g[(i)] = c[(i),1]
-  end
-  return g
+    return ϕ_n, ϕstar_n
+  end # inbounds
+end
+
+function g_coefs!(cache)
+  begin
```
ChrisRackauckas: `@inbounds`?

src/adams_utils.jl
```diff
@@ -1,42 +1,49 @@
...
       end
     end
-    g[(i)] = c[(i),1]
-  end
-  return g
+    return ϕ_n, ϕstar_n
+  end # inbounds
+end
+
+function g_coefs!(cache)
```
ChrisRackauckas: reference and implementation comment

YingboMa:
Oh, sorry. I just finished the inbounds macro and the reference parts, but I haven't push yet.

sipah00:
cool, then I will do handling callbacks part.

ChrisRackauckas:
Don't forget the unit tests:

> Verify the correctness of coefficients generation for variable step sizes (since the only test that we did is convergence rate on a fixed step size.)

I really like this idea. You can test against the fixed stepsize methods since you know what it should come out to there.

PR #349, opened 2018-05-15
handling callbacks and unit tests for the variable coefficient calculator `VCAB3`

sipah00:
- [x] Use an internal counter to handle callbacks
- [x] Unit tests for the variable coefficient calculator


src/caches/adams_bashforth_moulton_caches.jl
```diff
@@ -322,17 +324,22 @@ du_cache(c::VCAB3Cache) = ()
...
   dts = zeros(typeof(dt),3)
   c = zeros(typeof(t), 3, 3)
   g = zeros(typeof(t), 3)
-  ϕ_n = zeros(typeof(rate_prototype),3)
-  ϕstar_n = zeros(typeof(rate_prototype),3)
+  ϕ_n = Vector{typeof(rate_prototype)}(3)
+  ϕstar_nm1 = Vector{typeof(rate_prototype)}(3)
+  ϕstar_n = Vector{typeof(rate_prototype)}(3)
+  for i in 1:3
+    ϕ_n[i] = rate_prototype
```
ChrisRackauckas: This is going to alias the same array to each of these spots. Does that not cause an issue further down?

test/runtests.jl
```diff
@@ -37,6 +37,7 @@ if group == "All" || group == "Interface"
 end
 
 if group == "All" || group == "Integrators"
+    @time @testset "adams variable coefficients Tests" begin include("ode/adams_tests.jl") end
```
ChrisRackauckas: Match the capital scheme of the others.

src/caches/adams_bashforth_moulton_caches.jl
ChrisRackauckas: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/e50f73d42ea8f5c52e6651170daa4cccc21d0699/src/adams_utils.jl#L17-L24

This will have problems if ϕ_n is an array of all the same array since that loop will essentially not mean anything. How are tests passing with this? And why not use `ϕ_n[i] = zeros(rate_prototype)`?

sipah00: for `out of place`,type of `rate_prototype` would not be `arrayType`

sipah00: but all `perform_step!` for `out-place` have written without considering it as `arrayType`, I think

ChrisRackauckas: no, and for most methods there's a test for it.

sipah00: could you elaborate, any example method I can look?

ChrisRackauckas: I mean, if you take a look at any of them:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/fixed_timestep_perform_step.jl#L235-L265

you'll notice that these will work with `u` an array. It won't be the fastest way because of memory allocations, but it will work.

sipah00: oh, got it but then we have also take care of pointers to arrays(in case of passing to other function etc)

sipah00: if `k1,k2,...` are `arrayType`, they should pass by value not ref in other function, how should I  take care of it?

ChrisRackauckas: in which function call will that be an issue?

sipah00: checking `arrayType` for every `vars` and then passes it by ref or value in `perform_step!`??

ChrisRackauckas: it shouldn't matter. Point to the code where it would matter.

sipah00: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/adams_bashforth_moulton_perform_step.jl#L481

ChrisRackauckas: I see, yes the AB codes will need a few changes to make this work well. But add the `copy` here as a first change in that direction, and make note of this in an issue. That means just edit the capitalization in test description and this is good to merge.

src/caches/adams_bashforth_moulton_caches.jl
```diff
@@ -291,6 +291,7 @@ mutable struct VCAB3ConstantCache{rateType,TabType,tArrayType,rArrayType,cArrayT
   k::Int
```
YingboMa: The field `k` seems superfluous when you have `step`.

test/runtests.jl
```diff
@@ -37,6 +37,7 @@ if group == "All" || group == "Interface"
 end
 
 if group == "All" || group == "Integrators"
+    @time @testset "Adams Variable Coefficients Tests" begin include("ode/adams_tests.jl") end
```
YingboMa: I think that this line belongs to the `"AlgConvergence"` group.

sipah00: in `adams_test.jl` there are only tests for `coefs`, not for `convergence`

ChrisRackauckas: It definitely doesn't make sense in integrators though since it has nothing to do with the integrators. It is a unit test for convergence.

src/caches/adams_bashforth_moulton_caches.jl
sipah00: max value of `k` is 3, but `step` can go above

ChrisRackauckas: so `k = min(3,step)`?

sipah00: yeah

ChrisRackauckas:
Looks like AppVeyor is just a time out. @YingboMa found two minor things. Fix those up and we can merge when the tests re-pass.

src/caches/adams_bashforth_moulton_caches.jl
sipah00: yeah also you are right we can write it as `k = min(3,step)`

YingboMa: Maybe just change https://github.com/sipah00/OrdinaryDiffEq.jl/blob/d3e962903da86b3df80381237cebb5d0dd45afb1/src/perform_step/adams_bashforth_moulton_perform_step.jl#L582 to `cache.k = min(step, order)`?

sipah00:
forgot to delete `k` field and pass as an arg in coef functions, I will do it in WIP: adaptive part `VCAB3`" PR

ChrisRackauckas:
tests failed

sipah00:
I think, tests failed due to other reason.
`travis-ci log`
```
===========================[ ERROR: OrdinaryDiffEq ]============================
failed process: Process(`/home/travis/julia/bin/julia -Cx86-64 -J/home/travis/julia/lib/julia/sys.so --compile=yes --depwarn=yes --check-bounds=yes --code-coverage=user --color=yes --compilecache=yes /home/travis/.julia/v0.6/OrdinaryDiffEq/test/runtests.jl`, ProcessSignaled(9)) [0]
```

ChrisRackauckas:
Add a `gc()` call to help the gc know to free memory after new tests. Travis seems to be overloading their servers.

sipah00:
I will add `gc()` call in WIP: adaptive part VCAB3" PR

PR #353, opened 2018-05-20
Adaptive part `VCAB3`

sipah00:


src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -576,26 +576,21 @@ end
...
     g = g_coefs!(cache)
     u = uprev
     for i = 1:k
         u += dt * g[i] * ϕstar_n[i]
     end
+    du_np1 = f(u,p,t+dt)
+    ϕ_np1 = ϕ_np1!(cache,du_np1)
     if integrator.opts.adaptive
-      utilde = uprev + (dt/12)*(23*k1 - 16*k2 + 5*k3) - u
+      utilde = (g[end] - g[end-1]) * ϕ_np1[end]
```
sipah00: @YingboMa method for `error est` given in hairer 1 is for `predictor-corrector` method, 
but `vcab3` is only explicit method and it also says we can use fixed step size methods (`AB` methods) with current step size for `error est` but we dont know the value at those `prev points`, calculating them would be costly. so I used method given for `predictor-corrector` method, with one less order (`k+1`) because `vcab` is explicit method.

ChrisRackauckas: And you're stepping with the method of higher order? You should check whether this local extrapolation is stable. I think there's a discussion about that in Hairer I?

YingboMa: I think `Hairer I` doesn't really have much information regarding this. In the section `III.7`, he only mentions about VSVO methods and all the error estimations are focused on predictor-corrector methods. Maybe we should implement the predictor-corrector methods after all. @sipah00 could you add a PECE loop here? So that we can use a proper error estimation.

sipah00: sure

ChrisRackauckas: Do the Sundials docs mention local extrapolation at all?

YingboMa: In Sundials CVODE, its predictor is merely Taylor expansion by the Nordsieck vector (without any function evaluation).

ChrisRackauckas: https://www.hindawi.com/journals/mpe/2010/765620/#B11

sipah00:
ping @ChrisRackauckas @YingboMa

src/perform_step/adams_bashforth_moulton_perform_step.jl
ChrisRackauckas: I'm not sure we should do PECE here since the only benefit of pure Adams-Bashforth is to assume high stability isn't necessary and reduce the number of function calls. In that sense, computing the lower order AB coefficients would be more beneficial since it would reduce the number of `f` calls for just another coefficient calculation. Since the calculation of the coefficients are calculated O(1), this is asymtopically cheap. It would be interesting to have this as a contract to PECE to see if there's any large non-stiff problems where this is beneficial.

ChrisRackauckas:
MWTP

PR #355, opened 2018-05-21
Exponential utilities - phi functions

MSeeker1340:
Includes what I've been working for the past week (scalar/dense matrix phi functions using the Sidje formula as well as a simply one-step Krylov `phimv`).

I have written the crucial parts to be non-allocating whenever possible. @ChrisRackauckas can you help take a look at it, especially the code regarding the caches? This is my first time writing performance-intensive code so not sure if I got everything right.

BTW, since `phimv` is currently being imported from Expokit, I named `phimv` and `phimv!` to `_phimv` and `_phimv!` to avoid name clash. The aim is, of course, to drop Expokit dependency and just use our more generic, arbitrary order `phimv` function.


src/exponential_utils.jl
```diff
@@ -0,0 +1,216 @@
...
+    cache = zeros(T, k+1, k+1)
+  else
+    fill!(cache, zero(T))
+  end
+  cache[1,1] = z
+  for i = 1:k
+    cache[i,i+1] = one(T)
+  end
+  P = Base.LinAlg.expm!(cache)
+  return P[1,:]
```
ChrisRackauckas: since `expm!` allocating anyways, we can just use an `@view` here.

src/exponential_utils.jl
```diff
@@ -0,0 +1,216 @@
...
+  cache[1:m, 1:m] = A
+  cache[1:m, m+1] = v
+  for i = m+1:m+k-1
+    cache[i, i+1] = one(T)
+  end
+  P = Base.LinAlg.expm!(cache)
+  # Extract results
+  @views A_mul_B!(w[:, 1], P[1:m, 1:m], v)
+  for i = 1:k
+    w[:, i+1] = P[1:m, m+i]
```
ChrisRackauckas: The `P` slice allocates here. `@view` will a little bit as well. Probably easiest to just double loop this one.

src/exponential_utils.jl
```diff
@@ -0,0 +1,216 @@
...
+  
+```math
+\\varphi_0(z) = \\exp(z),\\quad \\varphi_k(z+1) = \\frac{\\varphi_k(z) - 1}{z} 
+```
+
+Calls `phimv_dense` on each of the basis vectors to obtain the answer.
+"""
+function phim(A::AbstractMatrix{T}, k::Integer; cache=nothing) where {T <: Number}
+  m = size(A, 1)
+  out = [Matrix{T}(m, m) for i = 1:k+1]
```
ChrisRackauckas: should probably split this out to have a `phim!` function

src/exponential_utils.jl
```diff
@@ -0,0 +1,216 @@
...
+  E = eye(A)
```
ChrisRackauckas: is this necessary or can you use `I` instead?

src/exponential_utils.jl
```diff
@@ -0,0 +1,216 @@
...
+  w = Matrix{T}(m, k+1)
+  if cache == nothing
+    cache = Matrix{T}(m+k, m+k)
+  end
+  for i = 1:m
+    e = E[:, i]
```
ChrisRackauckas: this part can allocate quite a bit. There should be a much better way of writing this (creating a lazy type for it?)

src/exponential_utils.jl
```diff
@@ -0,0 +1,216 @@
...
+    phimv_dense!(w, A, e, k; cache=cache) # w = [phi_0(A)*e phi_1(A)*e ... phi_k(A)*e]
+    for j = 1:k+1
+      out[j][:, i] = w[:, j]
```
ChrisRackauckas: This is allocating.

src/exponential_utils.jl
```diff
@@ -0,0 +1,216 @@
...
+      Base.axpy!(-alpha, V[:, i], cache)
+    end
+    beta = norm(cache)
+    H[j+1, j] = beta
+    V[:, j+1] = cache / beta
+  end
+  # Last iteration (j = m)
+  A_mul_B!(cache, A, V[:, m-1])
+  for i = 1:m
+    alpha = dot(V[:, i], cache)
```
ChrisRackauckas: This is allocating

src/exponential_utils.jl
```diff
@@ -0,0 +1,216 @@
...
+    H[i, m] = alpha
+    Base.axpy!(-alpha, V[:, i], cache) # is this still needed?
```
ChrisRackauckas: This is allocating

ChrisRackauckas: Why the comment `# is this still needed?`?

src/exponential_utils.jl
```diff
@@ -0,0 +1,216 @@
...
+  V = Matrix{eltype(b)}(length(b), m)
+  H = Matrix{eltype(b)}(m, m)
+  arnoldi!(V, H, A, b, m; cache=cache)
+end
+"""
+    arnoldi!(V,H,A,b,m) -> V,H
+
+Non-allocating version of `arnoldi`.
+"""
+function arnoldi!(V::Matrix{T}, H::Matrix{T}, A, b::AbstractVector{T},
```
ChrisRackauckas: Is all of this allocating slices for a reason?

src/exponential_utils.jl
MSeeker1340: I changed the implementation of `arnoldi` a bit to not return the extended matrices. Which means `v_m+1` is not needed. And since the previous `v`s are in theory orthogonal this step is not needed if the dot products are all that is needed.
Of course arnoldi does not yield perfectly orthogonal vectors, so it makes sense to play it safe.

MSeeker1340:
Changed all allocating array slicing to views/@inbounds loops.

I'm a little reserved about `phim` though. While `phimv` (and at its base, `phimv_dense!` and `arnoldi!`) are repeatedly used by the integrators so we should aim for maximal performance, `phim` is mainly used by small systems to precompute the expRK coefficients just once, so its performance doesn't really matter (also considering the dimension involved is usually small), so would it be better to use simpler notation which makes the code easier to read?

ChrisRackauckas:
Maybe I've been doing this too long but IMO in-place code just looks more proper package code while code that's allocating a bunch looks like a script. It doesn't matter all too much (these allocations would definitely not change timings too much), but to make it all match I would just prefer to see it use views and inplace operations.

MSeeker1340:
From slack:

Been thinking about Expokit's approach of decomposing `exp(tA)v` into smaller time intervals. While the idea is appealing, I think we should probably stick to plain, one-stage Krylov.
Two reasons: first, while `exp` has this nice semigroup property, the same isn't true for `phi`s. While it is possible to also express `phi_1(tA)v` as a series of steps (Expokit paper (17)), I don't think this is easily generalizable to higher order ones.

Second one has to do with the structure of expRK algorithms. In many of the algorithms we need to evaluate `phi_k(tL)v` for different `t` but the same `L` and `v`. For the one-step Krylov method, we can simply do one Arnoldi iteration and use the same results for different `t`. If we were to use a method similar to Expokit, however, this is not enough because the intermediate stages are not known and depends on `t`.

The advantage of Expokit's method is that we can use a large `t`... but since finer time steps are used under the hood this is most likely an illusion. Since we are doing ODE integration anyways, why not just use a smaller time step to begin with? The whole accuracy control side of Expokit can probably be delegated to the integrator side.

Nevertheless, the error estimation part of Expokit is imo still useful, especially in adaptive expRK that we eventually aim to build.

MSeeker1340:
I changed the caching versions of all the existing exponential integrators to use `phim` and removed functions that calculate the caching operators using mixed precision.

This is more of a sanity check for `phim`, so the computation isn't exactly optimal. On the other hand we need to think about the interface of expRK and related methods in detail in the future, so I wouldn't worry about optimization now.

ChrisRackauckas:
Out of curiousity, what's the timing difference between the old and new dense constructions?

ChrisRackauckas:
Thanks! Looks great. Just curious about the timing difference still. In reality it doesn't make much of a difference since it's the startup stage, so it's really the numerical stability that matters. But it would be good to have documented.

MSeeker1340:
Comparison between the old (mixed precision) and new (Sidje `phim`) approach:

```julia
using OrdinaryDiffEq
using DiffEqOperators
Pkg.status("OrdinaryDiffEq")
```

     - OrdinaryDiffEq                3.12.0+            exponential_utils
    


```julia
srand(0)
n = 20
f1 = DiffEqArrayOperator(randn(n,n))
f2 = (u,p,t) -> u
u0 = randn(n)
dt = 0.01; tspan = (0.0, 1.0)
prob = SplitODEProblem(f1, f2, u0, tspan);
```


```julia
for method in [NorsettEuler, ETD2, ETDRK4]
    println(method)
    @time init(prob, NorsettEuler(); dt=dt)
end
```

    OrdinaryDiffEq.NorsettEuler
      0.005885 seconds (976 allocations: 2.033 MiB)
    OrdinaryDiffEq.ETD2
      0.003104 seconds (976 allocations: 2.033 MiB)
    OrdinaryDiffEq.ETDRK4
      0.005806 seconds (976 allocations: 2.033 MiB)
    


```julia
using OrdinaryDiffEq
using DiffEqOperators
Pkg.status("OrdinaryDiffEq")
```

     - OrdinaryDiffEq                3.12.0+            master
    


```julia
srand(0)
n = 20
f1 = DiffEqArrayOperator(randn(n,n))
f2 = (u,p,t) -> u
u0 = randn(n)
dt = 0.01; tspan = (0.0, 1.0)
prob = SplitODEProblem(f1, f2, u0, tspan);
```


```julia
for method in [NorsettEuler, ETD2, ETDRK4]
    println(method)
    @time init(prob, NorsettEuler(); dt=dt)
end
```

    OrdinaryDiffEq.NorsettEuler
      0.006184 seconds (46.44 k allocations: 2.043 MiB)
    OrdinaryDiffEq.ETD2
      0.007714 seconds (46.44 k allocations: 2.043 MiB)
    OrdinaryDiffEq.ETDRK4
      0.007179 seconds (46.44 k allocations: 2.043 MiB)

MSeeker1340:
Archiving last week's numerical stability analysis for future reference:

The straightforward method of calculating the phi functions is either compute them directly or use the recurrence formula. Either way it is unstable for very small arguments. Using mixed precision calculations (`BigFloat`) can help but also become unstable for higher order phis.

phi_scalar

phi_matrix

Since the existing methods use at most phi_3, the mixed precision approach works but as can be seen above as soon as we go to phi_4 it also breaks.

The alternative method, using a formula from Sidje (Sidje, R. B. (1998). Expokit: a software 
package for computing matrix exponentials. ACM Transactions on Mathematical 
Software (TOMS), 24(1), 130-156. Theorem 1), is unconditionally stable as long as `expm` is stable (that is, no very large positive real eigenvalue for the matrix). This is confirmed by numerical experiments:

phi_stability_alt

stability_phim_sidje

Finally, a notebook for the error analysis of `expm`, which should more or less indicate how `phim` using the Sidje formula would perform. The error grows when the spectral radius of the matrix, which could be problematic in some cases.

[Error analysis of `expm`.pdf](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/files/2028945/Error.analysis.of.expm.pdf)

PR #358, opened 2018-05-23
5th order Adams variable step size explicit method

sipah00:


src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -772,3 +772,138 @@ end
...
+    rk4cache.fsalfirst .= k1
+    perform_step!(integrator, rk4cache)
+    integrator.fsallast .= rk4cache.k
+  else
+    g_coefs!(cache,k)
+    @. u = uprev
+    for i = 1:k
+      @. u += dt * g[i] * ϕstar_n[i]
+    end
+    f(k4,u,p,t+dt)
```
ChrisRackauckas: Extra function evaluation?

ChrisRackauckas: Nevermind. I see that's the FSAL variable, so it's just setting the future past step.

PR #360, opened 2018-05-23
Only run Interface and Integrator groups on AppVeyor

### LINKED ISSUE #359: Fix AppVeyor overload
These tests are overloading AppVeyor. It should probably make use of environments as well and split it up, or if it can't it should just run the interface tests (Interface and Integrator) since it's clear the convergence tests don't really need to be repeated.

YingboMa:
Fix #359

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -15,6 +15,8 @@ else
     group = "All"
 end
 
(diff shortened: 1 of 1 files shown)


YingboMa:
The test error is fixed on master.

PR #361, opened 2018-05-23
Rewriting `phimv` and `arnoldi`

MSeeker1340:
From slack:

The major goal is to have `phimv` able to utilize intermediate results in evaluating `phi_k(tA)b` for different `t`.
My first idea is to have `phimv` accept a list of `t` so it can run `arnoldi` once and compute results for all `t` in the list. But then it occurred to me that we may not know the list a priori. For example in adaptive schemes we may want to first do a time step of `h`, and then reduce `h` if error is too big. But we wounldn't know this beforehand .
What I want to try now is to have the low level `phimv!` to be separate from `arnoldi!`. So instead of `phimv!(w, t, A, b; ...)`, we will do `phimv!(w, t, V, H; ...)` where `V` and `H` are precomputed using `arnoldi!`. Then maybe maintain the old interface, which becomes a wrapper.
In this way we can delegate control of the Krylov methods to the integrator itself instead of making `phimv!` overly complicated.

The update to this is that I think it's best to encapsulate `V` and `H` into a dynamic container type `KrylovSubspace`. Not only does this make the interface simpler, it also answers the question of how to modify the code to handle variable dimension outputs from `arnoldi`.

As a reference, Arnoldi iteration code usually comes with an error criteria and terminates the iterations as soon as it is met (it is referred to as a "happy-breakdown"). For example, in expRK algorithms this error criteria can come from `reltol` of the integrator. This means the output `V` and `H`'s dimension is not known prior to `arnoldi`. By returning an encapsulated `Ks`, however, we can make it so `Ks[:V]` and `Ks[:H]` always have the correct dimensions.

`Ks[:V]` and `Ks[:H]` are just views into the fields `Ks.V` and `Ks.H`. The same `Ks` can be used in multiple `arnoldi!` calls as long as the subspace dimension is smaller than `maxiter`, and do `resize!` if the required dimension indeed exceeds `maxiter`.

### DIFF
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -200,38 +200,47 @@ function Base.show(io::IO, Ks::KrylovSubspace)
 end
 
 """
-    arnoldi(A,b,m) -> V,H
+    arnoldi(A,b,m) -> Ks
 
 Performs `m` anoldi iterations to obtain the Krylov subspace K_m(A,b).
 
-The output is the n x m unitary basis vectors `V` and the m x m upper 
-Heisenberg matrix `H`. They are related by the recurrence formula
+The n x m unitary basis vectors `Ks[:V]` and the m x m upper Heisenberg 
+matrix `Ks[:H]` are related by the recurrence formula
 
 ```
 v_1=b,\\quad Av_j = \\sum_{i=1}^{j+1}h_{ij}v_i\\quad(j = 1,2,\\ldots,m)
 ```
+
+Refer to `KrylovSubspace` for more information regarding the output.
 """
 function arnoldi(A, b, m; cache=nothing)
-  V = Matrix{eltype(b)}(length(b), m)
-  H = Matrix{eltype(b)}(m, m)
-  arnoldi!(V, H, A, b, m; cache=cache)
+  Ks = KrylovSubspace{eltype(b)}(length(b), m)
+  arnoldi!(Ks, A, b, m; cache=cache)
 end
 """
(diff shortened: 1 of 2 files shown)


MSeeker1340:
Currently `arnoldi!` still uses fixed iterations, so the next logical step would be to implement "happy-breakdown". I think our old friend Expokit would be very helpful in this.

BTW the last update to master somehow removed the test script for the phi functions, I re-added them in the commit.

ChrisRackauckas:
>So instead of phimv!(w, t, A, b; ...), we will do phimv!(w, t, V, H; ...) where V and H are precomputed using arnoldi!. Then maybe maintain the old interface, which becomes a wrapper.
In this way we can delegate control of the Krylov methods to the integrator itself instead of making phimv! overly complicated.

I like this.

src/exponential_utils.jl
```diff
@@ -140,40 +140,107 @@ end
...
-  @assert size(V,1) == size(A,1) == size(A,2) == n "Dimension mismatch"
-  @assert size(V,2) == size(H,1) == size(H,2) == m "Dimension mismatch"
+function arnoldi!(Ks::KrylovSubspace{B, T}, A, b::AbstractVector{T}, 
+  m::Integer; cache=nothing) where {B, T <: Number}
+  # Set dimension of the Krylov subspace
+  if m > maxiter(Ks)
+    resize!(Ks, m)
+  end
+  Ks.m = m
+  V, H = Ks[:V], Ks[:H]
```
ChrisRackauckas: just use the fields?

PR #365, opened 2018-05-24
Happy-breakdown detection and `t` parameter for `phimv`

MSeeker1340:
With the addition of happy-breakdown detection and the time parameter, `phimv` now has all functionality supported by Expokit. Ours is actually more powerful in that the code for Arnoldi iterations and function evaluation is separated, so we can reuse the arnoldi results for different `t` easily.

One thing I got wrong though in #361 is how happy-breakdown should be interpreted. It doesn't really have anything to do with the integrator's error control, but rather is an indication of how singular `A` is. In particular, if the minimal polynomial of `A` only has degree $d < n$, then happy-breakdown will occur at the $d$th iteration. In other words, whether happy-breakdown occurs is intrinsic to the problem/discretization method, and doesn't concern the integrator.

What this leaves us is the number of arnoldi iterations `m`, which does have a direct impact on the integrator's accuracy. I think there is some estimation formula available, but I need to consult the literatures on this.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -14,4 +14,3 @@ Compat 0.18.0
 Reexport
 MuladdMacro 0.0.2
 StaticArrays
-Expokit
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -22,8 +22,6 @@ module OrdinaryDiffEq
   using Parameters, GenericSVD, ForwardDiff, RecursiveArrayTools,
         NLsolve, Juno, Roots, DataStructures, DiffEqDiffTools
 
-  using Expokit: expmv, expmv!, phimv, phimv!
-
   import Base: linspace
 
   import ForwardDiff.Dual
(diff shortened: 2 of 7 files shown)


MSeeker1340:
Due to how the Sidje formula is used `phim` and `phimv` does not support the case when `k=0`. The former can be simply substituted with `expm` and I have added an `expmv` to the codebase.

MSeeker1340:
Finished changing `LawsonEuler` and `NorsettEuler` to use our `expmv` & `phimv` and dropped Expokit dependency. The next step would be to also implement Krylov solvers for `ETD2` and `ETDRK4`. 

I should note that this change makes the `perform_step!` function even more difficult to understand than before, especially the in-place one... Definitely need to do an extensive documentation on the internals of expRK methods after we get a few of the higher order algorithms running.

ChrisRackauckas:
This all makes sense. However, I think we should check how our `expmv` and `phimv` differ from ExpoKit's timing and stability wise. Make sure we aren't regressing.

ChrisRackauckas:
That view and performance talk can be in a separate issue to be examined further. Since we most likely want to go this direction anyways (and maybe just fix any performance issue we run into), I'm merging and opening an issue.

PR #369, opened 2018-05-28
Implement JVODE_Adams algorithm

YingboMa:


src/integrators/integrator_utils.jl
```diff
@@ -170,6 +170,13 @@ function step_reject_controller!(integrator,alg)
 end
 
 const StandardControllerAlgs = Union{GenericImplicitEuler,GenericTrapezoid}
+const NordArgs = Union{AN5, JVODE}
+
+function stepsize_controller!(integrator, alg::NordArgs)
+  η = max(stepsize_η!(integrator.cache), 0.1)
+  integrator.qold = integrator.dt/η
```
ChrisRackauckas: you need the other part limiting `qtmp` between `qmax` and `qmin`.

src/nordsieck_utils.jl
```diff
@@ -221,19 +221,36 @@ function stepsize_η!(integrator, cache::T) where T
   isvarorder = isvode && cache.n_wait == 0
   order = get_current_adaptive_order(integrator.alg, integrator.cache)
   L = order+1
-  cache.η = inv( (BIAS2*integrator.EEst)^inv(L) + ADDON )
+  η_next = cache.η = inv( (BIAS2*integrator.EEst)^inv(L) + ADDON )
   if isvarorder
-    cache.η = max(stepsize_η₋₁!(integrator, cache, order), stepsize_η₊₁!(integrator, cache, order), cache.η)
+    η_next = max(stepsize_η₋₁!(integrator, cache, order), stepsize_η₊₁!(integrator, cache, order), cache.η)
+  end
+  # There is no `gamma` in SUNDIALS
```
ChrisRackauckas: yes there is. There is always a safety factor? Their docs mention it.

YingboMa:
@ChrisRackauckas Ready for a code review.

ChrisRackauckas:
I'm in DC but will try to review ASAP. BDF next?

YingboMa:
I plan to benchmark JVODE and optimize it tomorrow. I am not sure about the correctness of the stepsize controller. However, I still want to merge this PR, because it is getting very big. I will implement JVODE BDF as soon as I fully finish Adams.

src/nordsieck_utils.jl
```diff
@@ -186,3 +222,161 @@ function nordsieck_rewind!(cache)
...
+    end
+  end
+  ( integrator.iter == 1 || integrator.u_modified ) && return ( cache.η = min(1e5, cache.η) )
+  cache.η = min(integrator.opts.qmax, max(integrator.opts.qmin, cache.η))
+  return cache.η
+end
+
+function stepsize_η!(integrator, cache::T, order) where T
+  L = order+1
+  cache.η = inv( (BIAS2*integrator.EEst)^inv(L) + ADDON )
```
ChrisRackauckas: I don't get these biases. Are they from Sundials?

YingboMa: Yes, they are in SUNDIALS. https://github.com/LLNL/sundials/blob/master/src/cvode/cvode.c#L3036

src/solve.jl
```diff
@@ -32,7 +32,7 @@ function init{algType<:OrdinaryDiffEqAlgorithm,recompile_flag}(
   reltol=nothing,
   qmax=qmax_default(alg),qmin=qmin_default(alg),
   qsteady_min = qsteady_min_default(alg),
-  qsteady_max = qsteady_min_default(alg),
```
ChrisRackauckas: oh shit, no wonder I never noticed a difference when using it... I wonder if this line effects other methods / benchmarks.

src/nordsieck_utils.jl
ChrisRackauckas: I see. It might just be fine-tuning of the order selection algorithm.

ChrisRackauckas:
Set those biases and ADDON as something that can be changed in the `JVODE` algorithm and then it LGTM. I just don't like hard-coded constants that can't be changed.

PR #370, opened 2018-05-29
Lanczos algorithm for symmetric/Hermitian matrices

MSeeker1340:
For symmetric/Hermitian `A`, we know that Arnoldi iteration will produce an `H` that is also real symmetric. The Lanczos algorithm improves upon Arnoldi by not calculating the zero elements of the upper part of `H`, thus saving time.

```julia
using OrdinaryDiffEq: KrylovSubspace, arnoldi!
```

```julia
n = 1000
m = 50
A = full(Hermitian(randn(n, n)))
Aperm = A + 1e-10 * randn(n, n) # no longer Hermitian
@show ishermitian(A)
@show ishermitian(Aperm);
```

    ishermitian(A) = true
    ishermitian(Aperm) = false
    

```julia
b = randn(n)
cache = similar(b)
Ks = KrylovSubspace{Float64}(n, m);
```

```julia
@time arnoldi!(Ks, A, b; m=m, cache=cache); # uses lanczos!
```

      0.043768 seconds (108 allocations: 13.016 KiB)
    

```julia
@time arnoldi!(Ks, Aperm, b; m=m, cache=cache);
```

      0.046262 seconds (2.61 k allocations: 130.094 KiB)
    
Note that the difference in time is not big: Arnoldi/Lanczos iteration is, after all, bounded in time by `A_mul_B!`. What Lanczos saves is just a bunch of dot products and `axpy!` updates.

Nevertheless, it's still valuable to have `lanczos!` around because it guarantees the results `Ks[:H]` is real symmetric, which can speed up later calculations. For example, `expm` is able to use diagonalization for Hermitian matrices:

```julia
@time expm(A);
```

      0.428920 seconds (36 allocations: 53.774 MiB, 2.35% gc time)
    

```julia
@time expm(Aperm);
```

      0.805784 seconds (103 allocations: 343.342 MiB, 14.75% gc time)

### DIFF
--- a/src/caches/linear_nonlinear_caches.jl
+++ b/src/caches/linear_nonlinear_caches.jl
@@ -174,7 +174,7 @@ function alg_cache(alg::NorsettEuler,u,rate_prototype,uEltypeNoUnits,uBottomElty
     else
       _A = full(A)
     end
-    exphA, phihA = phim(dt*_A, 1)
+    exphA, phihA = phi(dt*_A, 1)
   end
   NorsettEulerCache(u,uprev,similar(u),zeros(rate_prototype),exphA,phihA,Ks,KsCache)
 end
@@ -198,7 +198,7 @@ function alg_cache(alg::NorsettEuler,u,rate_prototype,uEltypeNoUnits,uBottomElty
     else
       _A = full(A)
     end
-    exphA, phihA = phim(dt*_A, 1)
+    exphA, phihA = phi(dt*_A, 1)
   end
   NorsettEulerConstantCache(exphA, phihA)
 end
@@ -233,7 +233,7 @@ function alg_cache(alg::ETD2,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnit
   else
     _A = full(A)
   end
-  Phi = phim(dt*_A, 2)
+  Phi = phi(dt*_A, 2)
   ETD2ConstantCache(Phi[1], Phi[2], Phi[2] + Phi[3], -Phi[3])
 end
 
@@ -256,7 +256,7 @@ function alg_cache(alg::ETD2,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnit
   else
     _A = full(A)
   end
-  Phi = phim(dt*_A, 2)
+  Phi = phi(dt*_A, 2)
   ETD2Cache(u,uprev,zero(u),zero(rate_prototype),zero(rate_prototype),Phi[1],Phi[2],Phi[2]+Phi[3],-Phi[3])
 end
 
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
>I can modify KrylovSubspace to have Ks[:H] be a Tridiagonal in this case. However I don't see how this would be useful in expRK algorithms.

That would change one matrix multiply to O(n) instead of O(n^3) via the dense multiply and also save some memory (which is arguably more important for PDEs). I would do this if it isn't difficult.

MSeeker1340:
Initially I just changed all `Ks[:H]` calls to `@view(Ks[1:Ks.m, 1:Ks.m])`, but then I realized an access method would still be necessary.

The reason is that in order to optimize `lanczos!`, I need to modify `KrylovSubspace` to specifically handle the case of Hermitian inputs. The simplest solution, for example, is to just let `Ks.H` be a `SymTridiagonal` in this case. However here `@view(Ks.H[1:m, 1:m])` is no longer `SymTridiagonal`. So we indeed need elaborate access methods.

MSeeker1340:
Performance comparison:

```julia
using OrdinaryDiffEq: KrylovSubspace, getH, arnoldi!, expv!
```


```julia
n = 1000
m = 50
srand(0)
A = full(Hermitian(randn(n, n)))
Aperm = A + 1e-10 * randn(n, n) # No longer Hermitian
b = randn(n)
t = 1e-3
Ks = KrylovSubspace{Float64}(n, m)
Ksperm = KrylovSubspace{Float64}(n, m)
w = similar(b)
wperm = similar(b)
arnoldi_cache = similar(b)
expv_cache = zeros(m, m);
```

Arnoldi/Lanczos iteration step:


```julia
@time arnoldi!(Ks, A, b; m=m, cache=arnoldi_cache);
```

      0.060559 seconds (110 allocations: 13.141 KiB)
    


```julia
@time arnoldi!(Ksperm, Aperm, b; m=m, cache=arnoldi_cache);
```

      0.067272 seconds (2.61 k allocations: 130.219 KiB)
    


```julia
ishermitian(getH(Ks))
```




    true




```julia
ishermitian(getH(Ksperm))
```




    false



`expv!` step


```julia
@time expv!(w, t, Ks, cache=expv_cache);
```

      0.000516 seconds (78 allocations: 38.313 KiB)
    


```julia
@time expv!(wperm, t, Ksperm, cache=expv_cache);
```

      0.012546 seconds (55 allocations: 355.344 KiB)
    


```julia
w ≈ wperm
```




    true


Again, in the Arnoldi/Lanczos step the difference is small because Lanczos only saves some O(n) operations while the bottleneck lies in `A_mul_B!` of the linear operator. (Come to think of it, the difference *will* be significant for sparse linear operators whose `A_mul_B!` is also of order O(n) right?). For `expv!`, because the symtridiagonal matrix produced by Lanczos has very efficiently eigendecomposition, we have very substantial gains here.

PR #371, opened 2018-05-30
WIP : variable order variable step size Adams method

sipah00:


src/adams_utils.jl
```diff
@@ -64,5 +64,10 @@ function g_coefs!(cache, k)
       end # q
       g[i] = c[i,1]
     end # i
-  end # inbounds
+  # end # inbounds
+end
+
+function γstar(k)
+  γ_star = [1,-1/2,-1/12,-1/24,-19/720,-3/160,-863/60480,-275/24192,-33953/3628800]
```
ChrisRackauckas: This is expensive. Just store the global array as a constant and access it.

src/alg_utils.jl
```diff
@@ -220,6 +220,8 @@ alg_order(alg::VCABM3) = 3
 alg_order(alg::VCABM4) = 4
 alg_order(alg::VCABM5) = 5
 
+alg_order(alg::VCABM) = 7
```
ChrisRackauckas: why 7?

src/caches/adams_bashforth_moulton_caches.jl
```diff
@@ -801,3 +801,38 @@ function alg_cache(alg::VCABM5,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUn
...
+  ϕstar_n = Vector{typeof(rate_prototype)}(9)
+  ϕ_np1 = Vector{typeof(rate_prototype)}(10)
+  for i in 1:9
+    ϕ_n[i] = copy(rate_prototype)
+    ϕstar_nm1[i] = copy(rate_prototype)
+    ϕstar_n[i] = copy(rate_prototype)
+  end
+  β = zeros(typeof(t),9)
+  order = 1
+  max_order = 7
```
ChrisRackauckas: why 7? It usually goes to 12.

sipah00: yeah,I will change it to 12 once I complete it.

src/adams_utils.jl
sipah00: okay

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -1290,3 +1290,69 @@ end
...
+  cache.ϕstar_nm1 .= ϕstar_n
+  g_coefs!(cache,k+1)
+  u = uprev
+  for i = 1:(k-1)
+      u += dt * g[i] * ϕstar_n[i]
+  end
+  du_np1 = f(u,p,t+dt)
+  ϕ_np1!(cache, du_np1, k+1)
+  u += dt * g[k] * ϕ_np1[k]
+  du_np1 = f(u, p, t+dt)
```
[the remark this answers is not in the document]
sipah00: yeah, good point, I will do it.

src/adams_utils.jl
gabrielgellner: Do we care about the precision of these constants? This will just always be a `Float64` array this way, but could we ever need to worry about having to change the precision on this (and do the converting Rational tricks instead)?

src/caches/adams_bashforth_moulton_caches.jl
```diff
@@ -801,3 +801,38 @@ function alg_cache(alg::VCABM5,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUn
...
+  ϕ_np1::rArrayType
+  ϕstar_n::rArrayType
+  β::tArrayType
+  order::Int
+  max_order::Int
+  step::Int
+end
+
+function alg_cache(alg::VCABM,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
+  dts = zeros(typeof(dt),9)
```
gabrielgellner: Shold put a space after comma to be consistent with following lines.

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -1290,3 +1290,69 @@ end
...
+  if integrator.opts.adaptive
+    utilde = dt * γstar(k) * ϕ_np1[k+1]
+    atmp = calculate_residuals(utilde, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
```
gabrielgellner: Add space after last comma to be consistent. Also it later similar calls in this function.

src/alg_utils.jl
```diff
@@ -220,7 +222,7 @@ alg_order(alg::VCABM3) = 3
 alg_order(alg::VCABM4) = 4
 alg_order(alg::VCABM5) = 5
 
-alg_order(alg::VCABM) = 7
+alg_order(alg::VCABM) = 1  #dummy value
```
ChrisRackauckas: just make it nothing so that way if it is used it'll error. We want to make sure it isn't used

src/integrators/integrator_utils.jl
```diff
@@ -161,6 +161,22 @@ function stepsize_controller!(integrator,alg)
   end
   q
 end
+
+function stepsize_controller!(integrator,alg::OrdinaryDiffEqAdamsVarOrderVarStepAlgorithm)
```
ChrisRackauckas: This isn't necessary. It's exactly the same as the other one which would be called if this doesn't exist. Is there a reason you added it?

sipah00: `beta1` and `beta2` are calculated once in starting after that order is changing so `beta1` and `beta2` should be updated.

src/alg_utils.jl
sipah00: making it nothing gives error as this function will be called in starting for calculation of `beta1` and `beta2`

src/integrators/integrator_utils.jl
ChrisRackauckas: I'm not sure that's actually stable though. The PI gains are generally assumed constant. Instead, just add this algorithm to https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L172 for now so that way it doesn't have the PI part. We can play with the extended controllers later.

sipah00: means we should use `std` stepsize controller for now?

ChrisRackauckas: yes for now.

ChrisRackauckas: eliminate what's not necessary to get it working, and then tag on new stuff over time to improve it. Starting with the whole setup, especially when that algorithm together has never been tested or analyze could cause debugging headaches.

sipah00: Okay

sipah00:
@ChrisRackauckas @YingboMa 
`ddeabm vs VCABM`
```julia
using ODEInterfaceDiffEq
using OrdinaryDiffEq
using DiffEqProblemLibrary
using Plots
prob = prob_ode_linear
sol1 = solve(prob, ddeabm())
sol2 = solve(prob, VCABM())
plot(sol1)
plot!(sol2)
```
vcabm

ChrisRackauckas:
That graph doesn't show much. It just shows it's converging to the correct result and will match any method. Instead, show the work-precision diagram. For example:

http://nbviewer.jupyter.org/github/JuliaDiffEq/DiffEqBenchmarks.jl/blob/master/NonStiffODE/Linear%20Work-Precision%20Diagrams.ipynb

sipah00:
Okay

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -1326,3 +1326,135 @@ end
...
+    end
+    integrator.fsallast = f(u, p, t+dt)
+    if step < 4
+      cache.order = min(order+1,3)
+    else
+      utildem2 = dt * γstar[(k-2)+1] * ϕ_np1[k-1]
+      utildem1 = dt * γstar[(k-1)+1] * ϕ_np1[k]
+      ϕ_and_ϕstar!(cache,k1,k+1)
+      ϕ_np1!(cache, integrator.fsallast, k+2)
+      utildep1 = dt * γstar[(k+1)+1] * ϕ_np1[k+2]
```
YingboMa: @sipah00 Could you describe the math behind the error estimation here? It seems like that you are assuming constant `dt` for the next step here? I thought that the order variation needs to calculate the next `dt` for `q-1`, `q` and `q+1` orders, and chooses the order that can maximize the stepsize?

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -1326,3 +1326,135 @@ end
...
+      ϕ_np1!(cache, k4, k+2)
+      @. utildep1 = dt * γstar[(k+1)+1] * ϕ_np1[k+2]
+      calculate_residuals!(atmpm2,utildem2, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
+      calculate_residuals!(atmpm1,utildem1, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
+      calculate_residuals!(atmpp1,utildep1, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
+      errm2 = integrator.opts.internalnorm(atmpm2)
+      errm1 = integrator.opts.internalnorm(atmpm1)
+      errp1 = integrator.opts.internalnorm(atmpp1)
+      if max(errm2,errm1) <= integrator.EEst
+        cache.order = min(order-1,max_order)
```
YingboMa: You don't have to use `min` in here.

src/adams_utils.jl
```diff
@@ -66,3 +66,6 @@ function g_coefs!(cache, k)
     end # i
   end # inbounds
 end
+
+# Coefficients for the implicit Adams methods
+global const γstar = [1,-1/2,-1/12,-1/24,-19/720,-3/160,-863/60480,-275/24192,-33953/3628800]
```
YingboMa: Could add this to a tableau with rational types, e.g., `-19//720`? You don't need to do it in this PR though.

YingboMa:
@sipah00 Could you solve this problem
```julia
julia> using DiffEqBase, DiffEqProblemLibrary

julia> prob = remake(prob_ode_linear, tspan=(0.,10.))
DiffEqBase.ODEProblem with uType Float64 and tType Float64. In-place: false
timespan: (0.0, 10.0)
u0: 0.5
```
and print out the order of `VCABM` with time?

src/perform_step/adams_bashforth_moulton_perform_step.jl
sipah00: yeah

src/adams_utils.jl
sipah00: Okay

ChrisRackauckas:
```
(order, t) = (3, 2.081752986084823)
(order, t) = (3, 2.081752986084823)
(order, t) = (3, 2.081752986084823)
(order, t) = (3, 2.081752986084823)
(order, t) = (3, 2.081752986084823)
(order, t) = (3, 2.081752986084823)
(order, t) = (4, 2.2485837090641185)
```

5 rejections and then an increase? That's very wrong. I would start by finding out how this can happen.

sipah00:
Now, the order goes to 12, but I noticed many rejections.
here is the gist of `(order ,t) and errm2, errm1, errp1`
https://gist.github.com/sipah00/5f0b6c6605c52b3ed05e2acf5fe4d462

```julia
using OrdinaryDiffEq, DiffEqDevTools, Plots, ODEInterfaceDiffEq
gr()
# 2D Linear ODE
f = (du,u,p,t) -> begin
  for i in 1:length(u)
    du[i] = 1.01*u[i]
  end
end
function (::typeof(f))(::Type{Val{:analytic}},u₀,p,t)
  u₀*exp(1.01*t)
end
tspan = (0.0,10.0)
prob = ODEProblem(f,rand(100,100),tspan)

abstols = 1./10.^(3:13)
reltols = 1./10.^(0:10);

setups = [Dict(:alg=>VCABM())
          Dict(:alg=>VCABM3())
          Dict(:alg=>ddeabm())]
names = ["VCABM";"VCABM3";"ddeabm"]
wp = WorkPrecisionSet(prob,abstols,reltols,setups;names=names,save_everystep=false)
plot(wp)
```
plot1

ChrisRackauckas:
You still have the same error estimator kicking out extremely high values. Have you been able to quantify why?

sipah00:
No, I could not figure it out, still trying

ChrisRackauckas:
Look for which value in the summation is large.

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -1326,3 +1326,138 @@ end
...
+      u += dt * g[i] * ϕstar_n[i]
+  end
+  du_np1 = f(u,p,t+dt)
+  ϕ_np1!(cache, du_np1, k+1)
+  u += dt * g[k] * ϕ_np1[k]
+  if integrator.opts.adaptive
+    utilde = dt * (g[k+1]-g[k]) * ϕ_np1[k+1]
+    atmp = calculate_residuals(utilde, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
+    integrator.EEst = integrator.opts.internalnorm(atmp)
+    if integrator.EEst >= one(integrator.EEst)
```
ChrisRackauckas: The check for passing is `<=`, so this should be `>`. Any other method where this is done should be changed as well. @YingboMa informing you too in case you did a `>=` anywhere.

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -1326,3 +1326,138 @@ end
...
+      atmpm2 = calculate_residuals(utildem2, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
+      atmpm1 = calculate_residuals(utildem1, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
+      atmpp1 = calculate_residuals(utildep1, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
+      errm2 = integrator.opts.internalnorm(atmpm2)
+      errm1 = integrator.opts.internalnorm(atmpm1)
+      errp1 = integrator.opts.internalnorm(atmpp1)
+      if max(errm2,errm1) <= integrator.EEst
+        cache.order = order - 1
+      elseif errp1 < integrator.EEst
+        cache.order = min(order+1,max_order)
```
ChrisRackauckas: Hairer notes that when the order is changed the stepsize is kept constant. This is done because the higher order method has a smaller stability region, so changing both can cause it to go unstable. To hack this in, just do `integrator.EEst = one(integrator.EEst)` and add a comment saying that this is for keeping the stepsize constant in the next step.

ChrisRackauckas:
I added two small comments. Fix those up and we can merge.

src/perform_step/adams_bashforth_moulton_perform_step.jl
ChrisRackauckas: @sipah00 can you address @gabrielgellner 's comment?

PR #372, opened 2018-05-30
Additional functionalities for exp utility: error estimate and IOP

MSeeker1340:
The aim of this PR is to address the second point of https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/243#issuecomment-393296491. 

Some numerical results for the corrector:

```julia
n = 100
K = 4
A = diagm(-2 * ones(n)) + diagm(ones(n - 1), -1) + diagm(ones(n - 1), 1)
b = randn(n);
```

```julia
function error_comparison(K, t, m)
    # Compute exact result using dense phi
    w_exact = zeros(n, K)
    P = phi(t*A, K)
    for i = 1:K
        w_exact[:, i] = P[i] * b
    end
    # Krylov with and without correction
    w = phiv(t, A, b, K; m=m, correct=false)[:, 1:K]
    w_correct = phiv(t, A, b, K; m=m, correct=true)[:, 1:K]
    err = [norm(w[:,i] - w_exact[:,i]) / norm(w_exact[:,i]) for i in 1:K]
    err_correct = [norm(w_correct[:,i] - w_exact[:,i]) / norm(w_exact[:,i]) for i in 1:K]
    for i = 1:K
        println("p = $(i-1), before: $(err[i]), after: $(err_correct[i])")
    end
end;
```

```julia
error_comparison(4, 5.0, 20)
```

    p = 0, before: 1.851580144556005e-8, after: 1.1866137192062755e-8
    p = 1, before: 1.1604243216282569e-9, after: 6.930957387640787e-10
    p = 2, before: 1.4084870014088947e-10, after: 7.894351673381101e-11
    p = 3, before: 2.417853000686249e-11, after: 1.2786553243680304e-11
    

```julia
error_comparison(4, 5.0, 15)
```

    p = 0, before: 1.670064384606024e-5, after: 1.3529002061306636e-5
    p = 1, before: 1.523366736059646e-6, after: 1.1210370888360611e-6
    p = 2, before: 2.577756909429866e-7, after: 1.7459774282018524e-7
    p = 3, before: 5.971387392772451e-8, after: 3.7583355289957666e-8
    

```julia
error_comparison(4, 5.0, 10)
```

    p = 0, before: 0.0039036186519269406, after: 0.004777657784517128
    p = 1, before: 0.0005983342152897502, after: 0.0006318794323758182
    p = 2, before: 0.00015660771392520995, after: 0.0001468087636175307
    p = 3, before: 5.3046898782640144e-5, after: 4.495612453983268e-5
    

```julia
error_comparison(4, 10.0, 20)
```

    p = 0, before: 2.7920782891613483e-5, after: 3.5001350862247234e-5
    p = 1, before: 2.3777873965132265e-6, after: 2.740946017107914e-6
    p = 2, before: 3.8466149667992783e-7, after: 4.123593087931721e-7
    p = 3, before: 8.52677289464572e-8, after: 8.57052414404906e-8
    

```julia
error_comparison(4, 10.0, 15)
```

    p = 0, before: 0.0016615112563970835, after: 0.0025777546011967675
    p = 1, before: 0.00021408895792534243, after: 0.00029448496471261217
    p = 2, before: 4.9115234177339934e-5, after: 6.131688135215691e-5
    p = 3, before: 1.477590119497644e-5, after: 1.6994996067541433e-5
    

```julia
error_comparison(4, 10.0, 10)
```

    p = 0, before: 0.041799641994719756, after: 0.08964522220170576
    p = 1, before: 0.008921966901360257, after: 0.016377920187105897
    p = 2, before: 0.0030763723943405454, after: 0.005008332379204382
    p = 3, before: 0.0013105118114586309, after: 0.0019318170942597493
    
So the corrector works best when the error is already small but actually amplifies error when it is big. IMO this is not very practical in real use... on the other hand, using the correction term purely as an error estimate should work well.

### DIFF
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -151,9 +151,9 @@ be smaller than `maxiter`, the maximum allowed arnoldi iterations.
     getV(Ks) -> V
     getH(Ks) -> H
 
-Access methods for orthonormal basis `V` and the Gram-Schmidt coefficients `H`. 
-Both methods return a view into the storage arrays and has the correct 
-dimensions as indicated by `Ks.m`.
+Access methods for the (extended) orthonormal basis `V` and the (extended) 
+Gram-Schmidt coefficients `H`. Both methods return a view into the storage 
+arrays and has the correct dimensions as indicated by `Ks.m`.
 
     resize!(Ks, maxiter) -> Ks
 
@@ -168,14 +168,15 @@ mutable struct KrylovSubspace{B, T}
   V::Matrix{T}  # orthonormal bases
   H::Matrix{T}  # Gram-Schmidt coefficients
   KrylovSubspace{T}(n::Integer, maxiter::Integer=30) where {T} = new{real(T), T}(
-    maxiter, maxiter, zero(real(T)), Matrix{T}(n, maxiter), zeros(T, maxiter, maxiter))
+    maxiter, maxiter, zero(real(T)), Matrix{T}(n, maxiter + 1), 
+    zeros(T, maxiter + 1, maxiter))
 end
 # TODO: switch to overload `getproperty` in v0.7
-getH(Ks::KrylovSubspace) = @view(Ks.H[1:Ks.m, 1:Ks.m])
-getV(Ks::KrylovSubspace) = @view(Ks.V[:, 1:Ks.m])
+getH(Ks::KrylovSubspace) = @view(Ks.H[1:Ks.m + 1, 1:Ks.m])
+getV(Ks::KrylovSubspace) = @view(Ks.V[:, 1:Ks.m + 1])
 function Base.resize!(Ks::KrylovSubspace{B,T}, maxiter::Integer) where {B,T}
-  V = Matrix{T}(size(Ks.V, 1), maxiter)
-  H = zeros(T, maxiter, maxiter)
+  V = Matrix{T}(size(Ks.V, 1), maxiter + 1)
+  H = zeros(T, maxiter + 1, maxiter)
   Ks.V = V; Ks.H = H
   Ks.m = Ks.maxiter = maxiter
   return Ks
@@ -194,8 +195,8 @@ end
 
 Performs `m` anoldi iterations to obtain the Krylov subspace K_m(A,b).
 
-The n x m unitary basis vectors `getV(Ks)` and the m x m upper Heisenberg 
-matrix `getH(Ks)` are related by the recurrence formula
+The n x (m + 1) unitary basis vectors `getV(Ks)` and the (m + 1) x m upper 
+Heisenberg matrix `getH(Ks)` are related by the recurrence formula
 
 ```
 v_1=b,\\quad Av_j = \\sum_{i=1}^{j+1}h_{ij}v_i\\quad(j = 1,2,\\ldots,m)
@@ -248,16 +249,16 @@ function arnoldi!(Ks::KrylovSubspace{B, T}, A, b::AbstractVector{T}; tol=1e-7,
       Base.axpy!(-alpha, @view(V[:, i]), cache)
     end
     beta = norm(cache)
-    if beta < vtol || j == m
-      # happy-breakdown or maximum iteration is reached
-      Ks.m = j
-      return Ks
-    end
     H[j+1, j] = beta
     @inbounds for i = 1:n
       V[i, j+1] = cache[i] / beta
     end
+    if beta < vtol # happy-breakdown
+      Ks.m = j
+      break
+    end
   end
+  return Ks
 end
 """
     lanczos!(Ks,A,b[;tol,m,norm,cache]) -> Ks
@@ -295,16 +296,19 @@ function lanczos!(Ks::KrylovSubspace{B, T}, A, b::AbstractVector{T}; tol=1e-7,
       Base.axpy!(-H[j-1, j], @view(V[:, j-1]), cache)
     end
     beta = norm(cache)
-    if beta < vtol || j == m
-      # happy-breakdown or maximum iteration is reached
-      Ks.m = j
-      return Ks
+    H[j+1, j] = beta
+    if j < m
+      H[j, j+1] = beta
     end
-    H[j+1, j] = H[j, j+1] = beta
     @inbounds for i = 1:n
       V[i, j+1] = cache[i] / beta
     end
+    if beta < vtol # happy-breakdown
+      Ks.m = j
+      break
+    end
   end
+  return Ks
 end
 
 """
@@ -337,7 +341,7 @@ function expv!(w::Vector{T}, t::Number, Ks::KrylovSubspace{B, T};
     # Here we only need a portion.
     cache = @view(cache[1:m, 1:m])
   end
-  @. cache = t * H
+  scale!(t, copy!(cache, @view(H[1:m, :])))
   if ishermitian(cache)
     # Optimize the case for symtridiagonal H
     F = eigfact!(SymTridiagonal(cache)) # Note: eigfact! -> eigen! in v0.7
@@ -346,7 +350,7 @@ function expv!(w::Vector{T}, t::Number, Ks::KrylovSubspace{B, T};
     expH = exp!(cache)
     expHe = @view(expH[:, 1])
   end
-  scale!(beta, A_mul_B!(w, V, expHe)) # exp(A) ≈ norm(b) * V * exp(H)e
+  scale!(beta, A_mul_B!(w, @view(V[:, 1:m]), expHe)) # exp(A) ≈ norm(b) * V * exp(H)e
 end
 
 """
@@ -394,8 +398,8 @@ function phiv!(w::Matrix{T}, t::Number, Ks::KrylovSubspace{B, T}, k::Integer;
     C1 = @view(C1[1:m + k, 1:m + k])
     C2 = @view(C2[1:m, 1:k + 1])
   end
-  @. Hcopy = t * H
+  scale!(t, copy!(Hcopy, @view(H[1:m, :])))
   fill!(e, zero(T)); e[1] = one(T) # e is the [1,0,...,0] basis vector
   phiv_dense!(C2, Hcopy, e, k; cache=C1) # C2 = [ϕ0(H)e ϕ1(H)e ... ϕk(H)e]
-  scale!(beta, A_mul_B!(w, V, C2)) # f(A) ≈ norm(b) * V * f(H)e
+  scale!(beta, A_mul_B!(w, @view(V[:, 1:m]), C2)) # f(A) ≈ norm(b) * V * f(H)e
 end
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -354,7 +354,7 @@ function expv!(w::Vector{T}, t::Number, Ks::KrylovSubspace{B, T};
 end
 
 """
-    phiv(t,A,b,k; kwargs) -> [phi_0(tA)b phi_1(tA)b ... phi_k(tA)b]
+    phiv(t,A,b,k;correct,kwargs) -> [phi_0(tA)b phi_1(tA)b ... phi_k(tA)b]
 
 Compute the matrix-phi-vector products using Krylov. `k` >= 1.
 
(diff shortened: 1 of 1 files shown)


MSeeker1340:
Not very sure about the interface for the error estimate at the moment. Do we just add an output argument `err` to `phiv!`? Seems to over complicate the interface a bit, also we don't always want the error estimate. How can I add a non-allocating output argument that is optional, without having to maintain two separate versions of the function?

MSeeker1340:
So I've probably overcomplicated myself yesterday about error estimates... In both phipm and KIOPS the only necessary error estimate to have is for the highest order phi (in the case of KIOPS it's the exp). So there's no need to return a separate error estimate for all the outputs of `phiv(!)`, at least for now.

Error estimates in action:

```julia
using OrdinaryDiffEq: phi, phiv
```


```julia
n = 100
m = 15
K = 4
A = diagm(-2 * ones(n)) + diagm(ones(n - 1), -1) + diagm(ones(n - 1), 1)
b = randn(n)
t = 5.0;
```


```julia
P = phi(t*A, K)
w_exact = P[K] * b
w, errest = phiv(t, A, b, K; m=m, errest=true)
err = norm(w[:,K] - w_exact)
println("Error: $err, error estimate: $errest")
```

    Error: 4.9003876156387905e-8, error estimate: 7.513596116146876e-8

MSeeker1340:
IOP is easier than I thought to implement so I'll just append the commit here.

Because I'm not using a dedicated data structure for `H` (is there btw? `H` here is an upper Heisenberg matrix with an length-p upper band), all I need to modify is adding a bound to the inner `arnoldi` loop.

Turns out `expv` and `phiv` stays just the same, as well as the code for error estimate and correction. IOP does change how adaptation is performed, however, but this is not the concern of this PR.

As a reference, some papers about using IOP to approximate matrix exponential-vector products:

https://www.researchgate.net/profile/Antti_Koskela/publication/273697107_Approximating_the_Matrix_Exponential_of_an_Advection-Diffusion_Operator_Using_the_Incomplete_Orthogonalization_Method/links/550954880cf27e990e0e5045/Approximating-the-Matrix-Exponential-of-an-Advection-Diffusion-Operator-Using-the-Incomplete-Orthogonalization-Method.pdf

https://onlinelibrary.wiley.com/doi/pdf/10.1002/nla.2090 (Algorithm 1 gives a clear definition of IOP).

MSeeker1340:
Sanity checks: 

```julia
using OrdinaryDiffEq: arnoldi, getH
```


```julia
srand(0)
n = 100
m = 15
A = diagm(-2 * ones(n)) + diagm(ones(n - 1), -1) + diagm(ones(n - 1), 1)
Aperm = A + 1e-5 * randn(n, n)
b = randn(n);
```

For symmetric matrices, IOP(2) should be almost the same as Lanczos:


```julia
Ks = arnoldi(A, b; m=m) # Lanczos
Ksperm = arnoldi(Aperm, b; m=m) # full Arnoldi
Ksperm2 = arnoldi(Aperm, b; m=m, iop=2); # IOP(2)
```


```julia
getH(Ks)
```




    16×15 SubArray{Float64,2,Array{Float64,2},Tuple{UnitRange{Int64},UnitRange{Int64}},false}:
     -2.21741   1.2582    0.0       0.0       …   0.0        0.0        0.0    
      1.2582   -1.84706   1.1293    0.0           0.0        0.0        0.0    
      0.0       1.1293   -2.05088   1.07224       0.0        0.0        0.0    
      0.0       0.0       1.07224  -2.03328       0.0        0.0        0.0    
      0.0       0.0       0.0       0.844917      0.0        0.0        0.0    
      0.0       0.0       0.0       0.0       …   0.0        0.0        0.0    
      0.0       0.0       0.0       0.0           0.0        0.0        0.0    
      0.0       0.0       0.0       0.0           0.0        0.0        0.0    
      0.0       0.0       0.0       0.0           0.0        0.0        0.0    
      0.0       0.0       0.0       0.0           0.0        0.0        0.0    
      0.0       0.0       0.0       0.0       …   0.0        0.0        0.0    
      0.0       0.0       0.0       0.0           0.984855   0.0        0.0    
      0.0       0.0       0.0       0.0          -1.87424    0.946788   0.0    
      0.0       0.0       0.0       0.0           0.946788  -1.89749    0.98828
      0.0       0.0       0.0       0.0           0.0        0.98828   -2.16322
      0.0       0.0       0.0       0.0       …   0.0        0.0        1.00082




```julia
getH(Ksperm)
```




    16×15 SubArray{Float64,2,Array{Float64,2},Tuple{UnitRange{Int64},UnitRange{Int64}},false}:
     -2.2174   1.25819  -1.20995e-6   4.74473e-6  …  -1.56411e-5  -1.09724e-5
      1.2582  -1.84709   1.1293      -9.36381e-6     -1.28886e-5   3.87006e-6
      0.0      1.1293   -2.05088      1.07224        -8.84203e-6   1.21384e-5
      0.0      0.0       1.07224     -2.03327         1.99436e-5   9.47316e-6
      0.0      0.0       0.0          0.844914        1.74995e-5   4.67657e-6
      0.0      0.0       0.0          0.0         …   4.02996e-7   1.5533e-5 
      0.0      0.0       0.0          0.0            -1.25157e-5   2.853e-7  
      0.0      0.0       0.0          0.0             2.05337e-5   1.3109e-6 
      0.0      0.0       0.0          0.0            -1.6915e-5    6.61577e-7
      0.0      0.0       0.0          0.0            -4.99333e-6   1.18318e-5
      0.0      0.0       0.0          0.0         …  -2.15769e-6  -7.03456e-6
      0.0      0.0       0.0          0.0             1.90411e-5   1.22223e-6
      0.0      0.0       0.0          0.0             0.946821     7.6984e-6 
      0.0      0.0       0.0          0.0            -1.89743      0.988291  
      0.0      0.0       0.0          0.0             0.988298    -2.16308   
      0.0      0.0       0.0          0.0         …   0.0          1.00084   




```julia
getH(Ksperm2)
```




    16×15 SubArray{Float64,2,Array{Float64,2},Tuple{UnitRange{Int64},UnitRange{Int64}},false}:
     -2.2174   1.25819   0.0       0.0       …   0.0        0.0        0.0     
      1.2582  -1.84709   1.1293    0.0           0.0        0.0        0.0     
      0.0      1.1293   -2.05088   1.07224       0.0        0.0        0.0     
      0.0      0.0       1.07224  -2.03327       0.0        0.0        0.0     
      0.0      0.0       0.0       0.844914      0.0        0.0        0.0     
      0.0      0.0       0.0       0.0       …   0.0        0.0        0.0     
      0.0      0.0       0.0       0.0           0.0        0.0        0.0     
      0.0      0.0       0.0       0.0           0.0        0.0        0.0     
      0.0      0.0       0.0       0.0           0.0        0.0        0.0     
      0.0      0.0       0.0       0.0           0.0        0.0        0.0     
      0.0      0.0       0.0       0.0       …   0.0        0.0        0.0     
      0.0      0.0       0.0       0.0           0.984791   0.0        0.0     
      0.0      0.0       0.0       0.0          -1.87428    0.946821   0.0     
      0.0      0.0       0.0       0.0           0.946813  -1.89743    0.988291
      0.0      0.0       0.0       0.0           0.0        0.988298  -2.16308 
      0.0      0.0       0.0       0.0       …   0.0        0.0        1.00084 


I'll do an detailed performance-cost analysis for IOP in the afternoon with large sparse `A` in mind.

MSeeker1340:
Some test results for IOP:

The test case is an `n = 1000` sparse 1D Laplacian with reflecting boundary condition. The reason for reflecting BC is that the Laplacian is *not* symmetric, so Lanczos does not apply. But since it is "almost symmetric", we should expect good performance for IOP(2).

As a comparison, the Laplacian is permuted to get a dense version. They differ a lot in their `A_mul_B!` cost:

```julia
@time A_mul_B!(w, Asparse, b);
```

      0.000009 seconds (4 allocations: 160 bytes)
    


```julia
@time A_mul_B!(w, Adense, b);
```

      0.001130 seconds (4 allocations: 160 bytes)

This means that `arnoldi` will be dominated by `dot` and `axpy!` in the sparse case and by `A_mul_B!` in the second case. Indeed this is confirmed by the profiler:

* Sparse `A`:

  sparse

* Dense `A`: 

  dense

Because IOP saves the number of dot products and axpys, we expect to see a large increase in performance for the sparse case but not for the dense case: 

```julia
@time profile_arnoldi(Asparse, 0);
```

      0.000901 seconds (2.61 k allocations: 130.250 KiB)
    


```julia
@time profile_arnoldi(Asparse, 2);
```

      0.000390 seconds (257 allocations: 20.000 KiB)
    


```julia
@time profile_arnoldi(Adense, 0);
```

      0.039249 seconds (2.61 k allocations: 138.250 KiB)
    


```julia
@time profile_arnoldi(Adense, 2);
```

      0.040866 seconds (260 allocations: 28.000 KiB)

(Here an IOP value of 0 indicates full Arnoldi).

Finally, the work-precision diagrams for the sparse case (note I didn't use a lot of trials so the results jiggle a bit):

h5
h20
h100

It is safe to say that in this case using IOP is a strictly better choice.

If `A` is not as symmetric, then the results may not be as obvious and we may need to use longer IOP lengths. But since most of the differential operators from PDE fall in this "almost symmetric" category, I believe we'll be fine using IOP(2).

Full test script:

[Incomplete Orthogonalization Procedure.zip](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/files/2064371/Incomplete.Orthogonalization.Procedure.zip)

ChrisRackauckas:
Looks great!

src/exponential_utils.jl
```diff
@@ -168,14 +168,15 @@ mutable struct KrylovSubspace{B, T}
...
   H::Matrix{T}  # Gram-Schmidt coefficients
   KrylovSubspace{T}(n::Integer, maxiter::Integer=30) where {T} = new{real(T), T}(
-    maxiter, maxiter, zero(real(T)), Matrix{T}(n, maxiter), zeros(T, maxiter, maxiter))
+    maxiter, maxiter, zero(real(T)), Matrix{T}(n, maxiter + 1), 
+    zeros(T, maxiter + 1, maxiter))
 end
 # TODO: switch to overload `getproperty` in v0.7
-getH(Ks::KrylovSubspace) = @view(Ks.H[1:Ks.m, 1:Ks.m])
-getV(Ks::KrylovSubspace) = @view(Ks.V[:, 1:Ks.m])
+getH(Ks::KrylovSubspace) = @view(Ks.H[1:Ks.m + 1, 1:Ks.m])
```
ChrisRackauckas: why is everything +1?

src/exponential_utils.jl
```diff
@@ -228,6 +238,9 @@ function arnoldi!(Ks::KrylovSubspace{B, T}, A, b::AbstractVector{T}; tol=1e-7,
   end
   V, H = getV(Ks), getH(Ks)
   vtol = tol * norm(A, Inf)
+  if iop == 0
+    iop = m
```
ChrisRackauckas: in the PR you only show for 2? What does higher look like?

src/exponential_utils.jl
MSeeker1340: This is to get the "extended" `H` and `V` matrix from Arnoldi. Originally I just left out the last terms to make the outputs of `getH` and `getV` square for easier coding.

Since error estimate & correction is a need now the extra terms are required, so I added them back and also changed the "happy-breakdown" termination logic of the Arnoldi loop so that they are always calculated.

src/exponential_utils.jl
MSeeker1340: For the output of `iop=p`, the Heisenberg matrix `H` will have a length-p upper band (i.e. only p non-zero diagonals for the upper part). I checked this before running the performance tests and it's working as intended.

Performance & accuracy wise, haven't really tested yet but the idea is by increasing `p` you get better accuracy and worse performance. For the almost symmetric operators that I showed in the tests, since IOP(2) is already good enough there's no need to check higher `p`s. For general operators I would expect a more visible performance-accuracy tradeoff for adjusting `p`.

MSeeker1340:
One thing I forgot to mention is that: from the profiling results it is clear that `arnoldi` is computation bounded both for the dense and sparse case, so I don't think there's a need to use a dedicated banded matrix type for `H` as I originally thought.

At best it saves some space when the `KrylovSubspace` is originally constructed. But this is done only once for an integrator so the difference would be negligible.

OK to merge this PR now?

PR #374, opened 2018-06-04
WIP: Implement the ROCK2 method

tyfff:
I am new to Julia, programming, and open-source. If there is any mistake, please tell me. @YingboMa is currently mentoring me.


src/tableaus/rkc_tableaus.jl
```diff
@@ -0,0 +1,1592 @@
+struct ROCK2ConstantCache{T,T2} <: OrdinaryDiffEqConstantCache
+  ms::SVector{46, Int}
+  fp1::SVector{46, T}
+  fp2::SVector{46, T}
+  recfi::SVector{4476, T2}
```
ChrisRackauckas: This should be a Vector. You typically don't want to use `SVector` for large arrays. All of these should probably be arrays, but some testing would need to be done on the `46`.

src/tableaus/rkc_tableaus.jl
```diff
@@ -0,0 +1,1592 @@
...
+end
+
+Base.@pure function ROCK2ConstantCache{T,T2}(::Type{T},::Type{T2})
```
ChrisRackauckas: What code generated these coefficients?

ChrisRackauckas:
Looks like a great start! Could you link what you're getting this from? It looks like there's a lot of hardcoded constants, so if this comes from another code we need to be careful about the license. It should be an MIT-compatible license. If it's generated from some formula in the paper, the generated code should be included (commented) and cited.

tyfff:
The hardcoded constants are from http://www.unige.ch/~hairer/software.html `rock.tar`. The `recfi` constants are the recurrence relations' parameters that are described in [this paper](https://pdfs.semanticscholar.org/20d2/7edfabdcc9d2b6bfdc3106d192f2f4624d14.pdf) from #2  equations (3.4). In A. Abdulle's implementation, there is a routine to compute the spectral radius of the Jacobian matrix and uses that information to select a suitable stage number, as different stage numbers have different stability region. Stepping with different stage numbers have different coefficients for the recurrence relations. They have the parameters for `s=3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,24,26,28,30,32,35,38,41,45,49,53,58,63,68,74,80,87,95,104,114,125,137,150,165,182,200`, thus there are many constants. I am not sure how did they generate the constants though. I will look into it tomorrow. Thanks for the code review!

ChrisRackauckas:
Cool, that page is all MIT licensed: http://www.unige.ch/~hairer/prog/licence.txt so we're good.

src/perform_step/rkc_perform_step.jl
```diff
@@ -0,0 +1,68 @@
...
+  gprev = uprev + temp1 * fsalfirst
+  ms[cache.mdeg] < 2 && u = gprev
+  # for the second to the ms[cache.mdeg] th stages
+  for i in 2:ms[cache.mdeg]
+    temp1 = h * recf[cache.recind + 2 * (i - 2) + 1]
+    temp3 = -recf[cache.recind + 2 * (i - 2) + 2]
+    temp2 = 1 - temp3
+    ci1 = temp1 + temp2 * ci2 + temp3 * ci3
+    u = temp1 * u + temp2 * gprev + temp3 * gprev2
+    i < ms[cache.mdeg] && (gprev2 = gprev, gprev = u)
```
YingboMa: I think `(gprev2 = gprev, gprev = u)` should be `(gprev2 = gprev; gprev = u)`.

src/perform_step/rkc_perform_step.jl
```diff
@@ -0,0 +1,68 @@
...
+  temp1 = h * fp1[cache.mdeg]
+  temp2 = h * fp2[cache.mdeg]
+  gprev = u + temp1 * gprev2
+  ci1 += temp1
+  # error estimate
+  if ntol = 0 # what is ntol?
+    temp3 = temp2 * ( u - gprev2)
+    u = gprev + temp1 * u + temp3
+    ci1 = max(abs(u), abs(uprev)) * rto
+    calculate_residuals!(atmp, utilde, uprev, u, integrator.opts.abstol, integrator.opts.reltol,integrator.opts.internalnorm)
```
YingboMa: Use `calculate_residuals` for the mutable version.

src/perform_step/rkc_perform_step.jl
```diff
@@ -0,0 +1,68 @@
...
+  # two-stage finishing procedure.
+  temp1 = h * fp1[cache.mdeg]
+  temp2 = h * fp2[cache.mdeg]
+  gprev = u + temp1 * gprev2
+  ci1 += temp1
+  # error estimate
+  if ntol = 0 # what is ntol?
+    temp3 = temp2 * ( u - gprev2)
+    u = gprev + temp1 * u + temp3
+    ci1 = max(abs(u), abs(uprev)) * rto
```
YingboMa: This doesn't look right. `u` and `uprev` are vectors in here.

src/perform_step/rkc_perform_step.jl
```diff
@@ -0,0 +1,68 @@
...
+    ci3 = ci2
+    ci2 = ci1
+  end # end if
+  # two-stage finishing procedure.
+  temp1 = h * fp1[cache.mdeg]
+  temp2 = h * fp2[cache.mdeg]
+  gprev = u + temp1 * gprev2
+  ci1 += temp1
+  # error estimate
+  if ntol = 0 # what is ntol?
```
YingboMa: Check https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/low_order_rk_perform_step.jl#L22-L26

src/perform_step/rkc_perform_step.jl
```diff
@@ -0,0 +1,68 @@
...
+  temp1 = h * recf[cache.recind]
+  ci1 = t + temp1
+  ci2 = t + temp1
+  ci3 = t
+  gprev2 = uprev
+  gprev = uprev + temp1 * fsalfirst
+  ms[cache.mdeg] < 2 && u = gprev
+  # for the second to the ms[cache.mdeg] th stages
+  for i in 2:ms[cache.mdeg]
+    temp1 = h * recf[cache.recind + 2 * (i - 2) + 1]
```
YingboMa: `h` -> `dt`

src/perform_step/rkc_perform_step.jl
YingboMa: Also, the conditioning in here seems superfluous.

src/perform_step/rkc_perform_step.jl
```diff
@@ -0,0 +1,68 @@
...
+  err = 0
+  # recurrence
+  # for the first stage
+  temp1 = h * recf[cache.recind]
+  ci1 = t + temp1
+  ci2 = t + temp1
+  ci3 = t
+  gprev2 = uprev
+  gprev = uprev + temp1 * fsalfirst
+  ms[cache.mdeg] < 2 && u = gprev
```
YingboMa: `u = gprev` -> `( u = gprev )`.

src/perform_step/rkc_perform_step.jl
```diff
@@ -16,51 +16,44 @@ end
   # The number of stage.
   mdeg = Int(floor(sqrt((1.5 + dt * integrator.eigen_est)/0.811) + 1))
   if mdeg >= 200
-    h = 0.8 * (200 ^ 2 * 0.811 - 1.5)/integrator.eigen_est
+    dt = 0.8 * (200 ^ 2 * 0.811 - 1.5)/integrator.eigen_est
```
ChrisRackauckas: this won't actually update the `dt`. A more sophisticated approach will be required here. Is the `dt` only ever decreased by this?

ChrisRackauckas:
How's this doing?

ChrisRackauckas:
Looks like there's been a lot of progress and a lot of it looks great. However, I don't understand the implementation of your core inner loop there. What paper did you work from? Looking at http://www.dumkaland.org/publications/content.pdf , I don't understand what the mu, nu, and kappa terms are, or how your loop corresponds to (26) and (27).

tyfff:
I used this [paper](https://pdfs.semanticscholar.org/20d2/7edfabdcc9d2b6bfdc3106d192f2f4624d14.pdf) as a reference. The `recf ` array stores the recurrence parameters (mu, nu and kapa in (3.4)). [Here](https://github.com/tyfff/OrdinaryDiffEq.jl/blob/ebff1738d6dfe08d7c6277d80abda1cde842ce3e/src/perform_step/rkc_perform_step.jl#L94-L96) is how I get these parameters. We don't need to store `nu` because of the relation `-nu-kapa=1`. The indexing for recurrence parameters are computed in [here](https://github.com/tyfff/OrdinaryDiffEq.jl/blob/ebff1738d6dfe08d7c6277d80abda1cde842ce3e/src/rkc_utils.jl#L67).

ChrisRackauckas:
So the `recf = [mu1,nu1,mu2,nu2,...]`? 

Then is `ms` the number of stages in the `n`th order method, so to get to the `n`th order method you offset by `recind = sum(2ms[1:n])`?

I think I am getting it but this can be cleaned up a bit so that way it's clearer at first glance what's going on. First off, why not split this up into `mu` and `nu` arrays? Then instead of `recind`, you can just have `mus[j]` be a static array of all `mu`s for the `j`th order method, and same for `nu`. So then `mus[j][i]` would be the `i`th `mu` for the `j`th order method. Note that there's no `tuple_len` recursion maximum anymore (https://github.com/JuliaLang/julia/pull/27398) and so making this a big `Vector{StaticVector}` (or tuple) should be fine performance-wise on v0.7. I think this would make the algorithm much much easier to read than the Fortran original, and would probably be slightly faster since you'd have more cache efficiency.

Other than that, I'm not entirely sure how you're choosing the maximum stages. You have the eigenvalue computation here https://github.com/tyfff/OrdinaryDiffEq.jl/blob/ebff1738d6dfe08d7c6277d80abda1cde842ce3e/src/rkc_utils.jl#L4 but I don't see it actually used? Could you describe the stage number choice?

ChrisRackauckas:
That test failure is fixed on master

ChrisRackauckas:
This is a pretty straight port of the Fortran code and it passes tests so I decided to merge. I still want to do a reorganization so that way it's much easier to follow, but this will be a separate issue and PR. Thanks! This is a very strong first contribution!

PR #375, opened 2018-06-04
Krylov Phiv internal time stepping

MSeeker1340:
The internal time stepping method of phipm aims to calculate linear combinations of phi-vector products of the form

$$ u = \varphi_0(tA)b_0 + t\varphi_1(tA)b_1 + \cdots + t^p\varphi_p(tA)b_p $$

The interface for the `phiv_timestep` function is far from its finalized version and some parts of the code can still be optimized, but the logic for the time stepping algorithm is complete.

Currently it still uses fixed time step `tau` (except for the last one) and Krylov subspace dimension `m`. But since we already have the error estimate of `phiv` from last week adding adaptation would just be a few lines at the end of each iteration.

Quick sanity checks:

```julia
using OrdinaryDiffEq: phi, phiv_timestep
```

```julia
srand(0)
n = 100
m = 20
p = 4
tol = 1e-7
t = 5.0
A = diagm(-2*ones(n)) + diagm(ones(n-1), -1) + diagm(ones(n-1), 1)
B = randn(n, p+1);
```

```julia
Anorm = norm(A, Inf)
b0norm = norm(B[:,1], Inf)
tau = 10/Anorm * (tol * ((m+1)/e)^(m+1) * sqrt(2*pi*(m+1)) / (4*Anorm*b0norm))^(1/m)
println("Default time step = $tau")
```

    Default time step = 8.832121898706621
    

```julia
Phi = phi(t * A,p)
u_exact = sum(t^i * Phi[i+1] * B[:,i+1] for i = 0:p)
u1 = phiv_timestep(t, A, B; m=m) # effectively only a single step
u2 = phiv_timestep(t, A, B; m=m, tau=1.0); # force multiple steps
```

```julia
err1 = norm(u1 - u_exact) / norm(u_exact)
```

    2.676157436850627e-11

```julia
err2 = norm(u2 - u_exact) / norm(u_exact)
```

    6.637202188498197e-16

### DIFF
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -436,3 +436,68 @@ function phiv!(w::Matrix{T}, t::Number, Ks::KrylovSubspace{B, T}, k::Integer;
     return w
   end
 end
+
+###########################################
+# Krylov phiv with internal time-stepping
+"""
+    phiv_timestep(t,A,B[;tau,m,tol,norm]) -> u
+
+Evaluates the linear combination of phi-vector products using time stepping
+
+```math
+u = \\varphi_0(tA)b_0 + t\\varphi_1(tA)b_1 + \\cdots + t^p\\varphi_p(tA)b_p 
+```
+
+The time stepping formula of Niesen & Wright is used [^1]. If the time step 
+`tau` is not specified, it is chosen according to (17) of Neisen & Wright.
+
+[^1]: Niesen, J., & Wright, W. (2009). A Krylov subspace algorithm for 
+evaluating the φ-functions in exponential integrators. arXiv preprint 
+arXiv:0907.4631.
+"""
+function phiv_timestep(t::Number, A, B::Matrix{T}; tau=nothing, 
+  m=min(30, size(A, 1)), tol=1e-7, norm=Base.norm) where {T <: Number}
+  # Choose initial timestep
+  if tau == nothing
+    Anorm = norm(A, Inf)
+    b0norm = norm(@view(B[:, 1]), Inf)
+    tau = 10/Anorm * (tol * ((m+1)/e)^(m+1) * sqrt(2*pi*(m+1)) / 
+      (4*Anorm*b0norm))^(1/m)
+  end
+  # Initialization
+  n = size(A, 1)
+  p = size(B, 2) - 1
+  W = Matrix{T}(n, p+1)
+  u = Vector{T}(n)
+  copy!(u, @view(B[:, 1])) # u(0) = b0
+
+  tk = 0.0 # current time
+  while tk < t # time stepping loopx
+    if tk + tau > t # last step
+      tau = t - tk
+    end
+    # Compute w0...wp using the recurrence relation (16)
+    copy!(@view(W[:, 1]), u) # w0 = u(tk)
+    @views @inbounds for j = 1:p
+      A_mul_B!(W[:, j+1], A, W[:, j])
+      coeff = 1.0
+      for l = 0:p-j
+        Base.axpy!(coeff, B[:, j+l+1], W[:, j+1])
+        coeff *= (tk / (l + 1))
+      end
+    end
+    # Compute ϕp(tau*A)wp using Krylov
+    # TODO: make this part non-allocating
+    u = tau^p * phiv(tau, A, @view(W[:, end]), p; m=m, tol=tol, norm=norm)[:, end]
+    # Update u using (15)
+    coeff = 1.0
+    @views @inbounds for j = 0:p-1
+      Base.axpy!(coeff, W[:, j+1], u)
+      coeff *= (tau / (j + 1))
+    end
+
+    tk += tau
+  end
+
+  return u
+end
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -478,22 +478,20 @@ function phiv_timestep(t::Number, A, B::Matrix{T}; tau=nothing,
     end
     # Compute w0...wp using the recurrence relation (16)
     copy!(@view(W[:, 1]), u) # w0 = u(tk)
+    coeffs = [1.0; cumprod(tk ./ (1:p - 1))] # cl = tk^l/l!
     @views @inbounds for j = 1:p
       A_mul_B!(W[:, j+1], A, W[:, j])
-      coeff = 1.0
       for l = 0:p-j
-        Base.axpy!(coeff, B[:, j+l+1], W[:, j+1])
-        coeff *= (tk / (l + 1))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
>Actually phipm goes one step further to construct a sparse matrix from the coefficients such that (15) and (16) can be represented as one mat-mat multiplication. While this "vectorized" approach might be a clever trick in Matlab, I think sticking with for loops and axpy! would be enough here.

Indeed, building a sparse matrix unnecessarily would be pretty bad.

MSeeker1340:
Did some analysis on `phiv_timestep` focusing on the performance-accuracy tradeoff of using a smaller time step `tau` vs a larger one.

The problem setup is as follows: `A` is again the sparse 1D Laplacian with width `n=1000`. The target is to calculate

$$ u = \exp(tA)b_0 + t\sum_{i=1}^4 \varphi_i(tA)b_i $$

for randomly constructed normalized $b_i$. This mimics the last step of a typical EPIRK method.

For each value of `t` I varied the Krylov subspace size `m` and the number of steps taken (shown as annotations in the work-precision diagrams) and compared `phiv_timestep!`'s cost & error. The results:

t5
t10
t20
t50
t100

Increasing `m` and the number of steps increases performance cost and reduces error, as expected. What's important to note is that increasing the number of steps seem to have a larger impact on performance. This might be because the parts in `phiv_timestep!` responsible for time stepping is not as optimized as `arnoldi!` and `phiv!`. I intend to investigate this further by doing some profiling on `phiv_timestep!`.

Another intriguing feature for very large `t` is that initially increasing the number of steps actually increases the error. Why is this the case I'm not entirely sure. Anyways it's best to avoid using such a large `t` in the integrators.

ChrisRackauckas:
It looks like the cost of larger Krylov subspaces is almost nothing. What happens if you keep pushing it larger? 100? How high does it have to go before it's more expensive than another step, or until it stops decreasing the error? A larger Krylov subspace is also more parallel of a calculation (since the matrix computations are threaded), so on more optimized HPC-nodes (like 16+ core Intels) it probably has even less of an effect. It seems that the Krylov subspace size should be chosen as large as possible without bounding the memory?

MSeeker1340:
Turns out the previous results are erroneous... Much like the issue in #372 where `norm` ends up taking the majority of time of `arnoldi!`, here it is a similar story but with `ishermitian`.

`ishermitian` is called in `arnoldi!` to determine if the matrix is symmetric/Hermitian and switch to `lanczos` when it is. In the tests above I wrote `A` as a `SymTridiagonal`. Well... a `SymTridiagonal` is technically only symmetric but not necessarily Hermitian, so `ishermitian` just starts to check all elements of `A` as if it is a dense matrix, which takes a lot of time.

The issue if glossed over in #372 because then I chose a non-symmetric `A`, and `ishermitian` is able to terminate quickly for the first few elements. Here for the actually Hermitian `A` it ends up checking *all* elements, resulting in most of the time visible on the work-precision diagrams being just `ishermitian`.

My temporary solution is to switch to using the sparse CSC format, which does have an optimized version of `ishermitian`. I'll post the corrected results in a moment (which are also more in line with those of #372). In the meantime, a simple script to demonstrate my point:



```julia
n = 1000
dd = -2 * ones(n)
dl = ones(n - 1)
_A = SymTridiagonal(dd, dl)
A = sparse(_A); # faster ishermitian
```


```julia
@which ishermitian(_A)
```




ishermitian(A::<b>AbstractArray{T,2} where T</b>) at <a href="https://github.com/JuliaLang/julia/tree/d386e40c17d43b79fc89d3e579fc04547241787c/base/linalg/generic.jl#L933" target="_blank">linalg\generic.jl:933</a>




```julia
@time ishermitian(_A);
```

      0.021926 seconds (4 allocations: 160 bytes)
    


```julia
@which ishermitian(A)
```




ishermitian(A::<b>SparseMatrixCSC</b>) at <a href="https://github.com/JuliaLang/julia/tree/d386e40c17d43b79fc89d3e579fc04547241787c/base/sparse/sparsematrix.jl#L3133" target="_blank">sparse\sparsematrix.jl:3133</a>




```julia
@time ishermitian(A);
```

      0.000028 seconds (5 allocations: 8.156 KiB)
    
Currently I think I'll just keep `ishermitian` as-is inside `arnoldi!`, and have the user be responsible for providing a linear operator type that is guaranteed to have a fast `ishermitian` method. Since we'll be mostly using DiffEqOperators for the integrators this should not be a problem. Nevertheless we must document this requirement in the future.

MSeeker1340:
The corrected work-precision diagrams:

t5
t10
t20
t50
t100

Note that the effect of a larger Krylov subspace on the performance is greater for larger number of time steps, which is to be expected because more iterations will amplify the differences.

As a reference, some profiling results (`m = 20`, `t = 5.0` and 5 steps):

* `arnoldi!`: 0.000378 seconds (59 allocations: 26.469 KiB)

  arnoldi

  If you compare this with the graph in #372, you can see that `A_mul_B!` takes up more proportions. First this may have something to do with `A_mul_B` with `SparseMatrixCSC` being slower than `Tridiagonal`, secondly this is also likely the result of Lanczos being able to save half the inner products/vector additions using symmetricity. 

* `phiv!`: 0.000325 seconds (69 allocations: 182.422 KiB)

  phiv

  Here line 84 is just an `expm!` call. It takes up the majority of time in `phiv`.

* `phiv_timestep`: 0.004737 seconds (1.21 k allocations: 976.734 KiB)

  timestep_5

  Here most of the time is spent on `arnoldi!` and `phiv!`. The front and end parts are extra calculations related to the `W` matrix and the update of `u`.

ChrisRackauckas:
Nice results. Yes, it's looking pretty optimal implementation-wise unless we can speed up that expm!. Though algorithm-wise there's always more tweaking of course. 

It still seems like you should try a higher `m` since with just a few time steps it seems the highest m is still doing the best, so I would like to see if that effect saturates or stays true.

MSeeker1340:
Yep your intuition is correct. Increasing `m` further does result in saturation:

t20
t50
t100

I changed the perspective to fix the number of time steps and vary `m` from 10 to 100. It seem to be the case that the time stepping method will always hit a threshold, which depends on how many internal steps are taken.

The saturation thresholds are fairly low though. For the practical usage case, say `tol ~ 1e-7`, `m = 20 ~ 30` with a few time steps is enough.

MSeeker1340:
@ChrisRackauckas Can we merge this now? The discussions have become a bit long and I think it's better to dedicate a new PR to the adaptation stuff.

ChrisRackauckas:
well the tolerance should be set to match the abstol/reltol of the integrator? But that gives a good default value for m.

Yes, I wasn't merging since you had it as WIP so I thought you were waiting to change something.

PR #376, opened 2018-06-06
Adaptation for `phiv_timestep(!)`

MSeeker1340:
The adaptation framework of phipm (Algorithm 3 & 4 in the paper) is unfortunately a bit messy. Took me a while to sort through the threads and incorporate the framework into `phiv_timestep(!)`. In particular, I changed the do loop of Algorithm 3 to the more manageable form in `phiv_timestep`, which also makes Algorithm 4 (the helper funtion `_phiv_timestep_adapt`) a bit simpler to interpret.

The framework is not yet complete though: the cost comparison part of Algorithm 4 only concerns full Arnoldi and Lanczos, so I need to come up with a similar estimation for IOP(p).

I also entertained myself with the idea of rewriting `phiv_timestep` into an iterator object much like `integrator` does in OrdinaryDiffEq. For the moment I think I'll just keep things as-is, but if there's future need to extend or modify the phipm algorithm, then it may be worthwhile to do so.

### DIFF
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -483,7 +483,7 @@ function phiv_timestep!(u::Vector{T}, t::Float64, A, B::Matrix{T}; tau::Float64=
   @assert n == size(A, 1) == size(A, 2) == size(B, 1) "Dimension mismatch"
   if caches == nothing
     W = Matrix{T}(n, p+1)         # stores the w vectors
-    P = similar(W)                # stores output from phiv!
+    P = Matrix{T}(n, p+2)         # stores output from phiv!
     Ks = KrylovSubspace{T}(n, m)  # stores output from arnoldi!
     phiv_caches = nothing         # caches used by phiv!
   else
@@ -493,7 +493,7 @@ function phiv_timestep!(u::Vector{T}, t::Float64, A, B::Matrix{T}; tau::Float64=
   copy!(u, @view(B[:, 1])) # u(0) = b0
 
   tk = 0.0 # current time
-  while tk < t # time stepping loopx
+  while tk < t # time stepping loop
     if tk + tau > t # last step
       tau = t - tk
     end
@@ -508,8 +508,8 @@ function phiv_timestep!(u::Vector{T}, t::Float64, A, B::Matrix{T}; tau::Float64=
     end
     # Compute ϕp(tau*A)wp using Krylov
     arnoldi!(Ks, A, @view(W[:, end]); tol=tol, m=m, norm=norm, iop=iop, cache=u)
-    phiv!(P, tau, Ks, p; caches=phiv_caches, correct=correct)
-    scale!(tau^p, copy!(u, @view(P[:, end])))
+    _, epsilon = phiv!(P, tau, Ks, p + 1; caches=phiv_caches, correct=correct, errest=true)
+    scale!(tau^p, copy!(u, @view(P[:, end - 1])))
     # Update u using (15)
     coeffs = [1.0; cumprod(tau ./ (1:p - 1))] # cl = tau^l/l!
     @views @inbounds for j = 0:p-1
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -569,10 +569,12 @@ function _phiv_timestep_adapt(m, tau, epsilon, m_old, tau_old, epsilon_old, q, k
     q = log(tau/tau_old) / log(epsilon/epsilon_old) - 1
   end # else keep q the same
   tau_new = tau * (gamma / omega)^(1/(q + 1))
+  tau_new = min(max(tau_new, tau/5), 2*tau, maxtau)
   if m_old < m
     kappa = (epsilon/epsilon_old)^(1/(m_old - m))
   end # else keep kappa the same
   m_new = m + ceil(Int, log(omega / gamma) / log(kappa))
+  m_new = min(max(m_new, div(3*m, 4), 1), Int(ceil(4*m / 3)))
   verbose && println("  - Proposed new m: $m_new, new tau: $tau_new")
   # Compare costs of using new m vs new tau (23)
   cost_tau = _phiv_timestep_estimate_flops(m, tau_new, n, p, NA, iop, Hnorm, maxtau)
@@ -580,9 +582,7 @@ function _phiv_timestep_adapt(m, tau, epsilon, m_old, tau_old, epsilon_old, q, k
   verbose && println("  - Cost to use new m: $cost_m flops, new tau: $cost_tau flops")
   if cost_tau < cost_m
     m_new = m
-    tau_new = min(max(tau_new, tau/5), 2*tau, maxtau)
   else
-    m_new = min(max(m_new, div(3*m, 4), 1), Int(ceil(4*m / 3)))
     tau_new = tau
   end
   return m_new, tau_new, q, kappa
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -468,7 +468,8 @@ Non-allocating version of `phiv_timestep`.
 """
 function phiv_timestep!(u::Vector{T}, t::Float64, A, B::Matrix{T}; tau::Float64=0.0, 
   m::Int=min(10, size(A, 1)), tol::Real=1e-7, norm=Base.norm, iop::Int=0, 
-  correct::Bool=false, caches=nothing, adapt=false, delta=1.2, gamma=0.8) where {T <: Number}
+  correct::Bool=false, caches=nothing, adaptive=false, delta::Real=1.2, 
+  gamma::Real=0.8, NA::Int=countnz(A)) where {T <: Number}
   # Choose initial timestep
   if iszero(tau)
     Anorm = norm(A, Inf)
@@ -487,7 +488,7 @@ function phiv_timestep!(u::Vector{T}, t::Float64, A, B::Matrix{T}; tau::Float64=
     phiv_caches = nothing         # caches used by phiv!
   else
     W, P, Ks, phiv_caches = caches
-    @assert size(W) == size(P) == (n, p+1) "Dimension mismatch"
+    @assert size(W) == (n, p+1) && size(P) == (n, p+2) "Dimension mismatch"
   end
   copy!(u, @view(B[:, 1])) # u(0) = b0
   abstol = tol * norm(A, Inf)
@@ -509,13 +510,14 @@ function phiv_timestep!(u::Vector{T}, t::Float64, A, B::Matrix{T}; tau::Float64=
     # Part 2: compute ϕp(tau*A)wp using Krylov, possibly with adaptation
     arnoldi!(Ks, A, @view(W[:, end]); tol=tol, m=m, norm=norm, iop=iop, cache=u)
     _, epsilon = phiv!(P, tau, Ks, p + 1; caches=phiv_caches, correct=correct, errest=true)
-    if adapt
+    if adaptive
       omega = (t / tau) * (epsilon / abstol)
       epsilon_old = epsilon; m_old = m; tau_old = tau
-      q = m/4; kappa = 2.0
+      q = m/4; kappa = 2.0; maxtau = t - tk
       while omega > delta # inner loop of Algorithm 3
         m_new, tau_new, q, kappa = _phiv_timestep_adapt(
-          m, tau, epsilon, m_old, tau_old, epsilon_old, q, kappa, gamma, omega)
+          m, tau, epsilon, m_old, tau_old, epsilon_old, q, kappa, 
+          gamma, omega, maxtau, n, p, NA, iop, norm(getH(Ks), 1))
         m, m_old = m_new, m
         tau, tau_old = tau_new, tau
         # Compute ϕp(tau*A)wp using the new parameters
@@ -537,7 +539,9 @@ function phiv_timestep!(u::Vector{T}, t::Float64, A, B::Matrix{T}; tau::Float64=
 
   return u
 end
-function _phiv_timestep_adapt(m, tau, epsilon, m_old, tau_old, epsilon_old, q, kappa, gamma, omega)
+# Helper functions for phiv_timestep!
+function _phiv_timestep_adapt(m, tau, epsilon, m_old, tau_old, epsilon_old, q, kappa, 
+  gamma, omega, maxtau, n, p, NA, iop, Hnorm)
   # Compute new m and tau (Algorithm 4)
   if tau_old > tau
     q = log(tau/tau_old) / log(epsilon/epsilon_old) - 1
@@ -547,15 +551,34 @@ function _phiv_timestep_adapt(m, tau, epsilon, m_old, tau_old, epsilon_old, q, k
     kappa = (epsilon/epsilon_old)^(1/(m_old - m))
   end # else keep kappa the same
   m_new = m + log(omega / gamma) / log(kappa)
-  # Compare costs of using new m vs new tau
-  cost_tau = 0 # TODO
-  cost_m = 0 # TODO
+  # Compare costs of using new m vs new tau (23)
+  cost_tau = _phiv_timestep_estimate_flops(m, tau_new, n, p, NA, iop, Hnorm)
+  cost_m = _phiv_timestep_estimate_flops(m_new, tau, n, p, NA, iop, Hnorm)
   if cost_tau < cost_m
     m_new = m
-    tau_new = min(max(tau_new, tau/5), 2*tau, t-tk)
+    tau_new = min(max(tau_new, tau/5), 2*tau, maxtau)
   else
     m_new = min(max(m_new, div(3*m, 4), 1), Int(ceil(4*m / 3)))
     tau_new = tau
   end
   return m_new, tau_new, q, kappa
+end
+function _phiv_timestep_estimate_flops(m, tau, n, p, NA, iop, Hnorm)
+  # Estimate flops for the update of W and u
+  flops_W = 2 * (p - 1) * (NA + n)
+  flops_u = (2 * p + 1) * n
+  # Estimate flops for arnoldi!
+  if iop == 0
+    iop = m
+  end
+  flops_matvec = 2 * m * NA
+  flops_vecvec = 0
+  for i = 1:m
+    flops_vecvec += 3 * min(i, iop)
+  end
+  # Estimate flops for phiv! (7)
+  MH = 44/3 + 2 * ceil(max(0.0, log2(Hnorm / 5.37)))
+  flops_phiv = round(Int, MH * (m + p)^3)
+
+  return flops_W + flops_u + flops_matvec + flops_vecvec + flops_phiv
 end
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -440,7 +440,7 @@ end
 ###########################################
 # Krylov phiv with internal time-stepping
 """
-    phiv_timestep(t,A,B[;tau,m,tol,norm,iop,correct]) -> u
+    phiv_timestep(t,A,B[;adaptive,tol,kwargs...]) -> u
 
 Evaluates the linear combination of phi-vector products using time stepping
 
@@ -449,7 +449,12 @@ u = \\varphi_0(tA)b_0 + t\\varphi_1(tA)b_1 + \\cdots + t^p\\varphi_p(tA)b_p
 ```
 
 The time stepping formula of Niesen & Wright is used [^1]. If the time step 
-`tau` is not specified, it is chosen according to (17) of Neisen & Wright. 
+`tau` is not specified, it is chosen according to (17) of Neisen & Wright. If 
+`adaptive==true`, the time step and Krylov subsapce size adaptation scheme of 
+Niesen & Wright is used, the relative tolerance of which can be set using the 
+keyword parameter `tol`. The delta and gamma parameter of the adaptation 
+scheme can also be adjusted.
+
 For the other keyword arguments, consult `arnoldi` and `phiv`, which are used 
 internally.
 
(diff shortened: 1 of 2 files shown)


src/exponential_utils.jl
```diff
@@ -457,19 +457,18 @@ internally.
...
 """
     phiv_timestep!(u,t,A,B[;kwargs]) -> u
 
 Non-allocating version of `phiv_timestep`.
 """
 function phiv_timestep!(u::Vector{T}, t::Float64, A, B::Matrix{T}; tau::Float64=0.0, 
-  m::Int=min(30, size(A, 1)), tol::Real=1e-7, norm=Base.norm, iop::Int=0, 
-  correct::Bool=false, caches=nothing) where {T <: Number}
+  m::Int=min(10, size(A, 1)), tol::Real=1e-7, norm=Base.norm, iop::Int=0, 
+  correct::Bool=false, caches=nothing, adapt=false, delta=1.2, gamma=0.8) where {T <: Number}
```
ChrisRackauckas: for consistency, use `adaptive`

src/exponential_utils.jl
```diff
@@ -521,3 +537,25 @@ function phiv_timestep!(u::Vector{T}, t::Float64, A, B::Matrix{T}; tau::Float64=
...
+  if tau_old > tau
+    q = log(tau/tau_old) / log(epsilon/epsilon_old) - 1
+  end # else keep q the same
+  tau_new = tau * (gamma / omega)^(1/(q + 1))
+  if m_old < m
+    kappa = (epsilon/epsilon_old)^(1/(m_old - m))
+  end # else keep kappa the same
+  m_new = m + log(omega / gamma) / log(kappa)
+  # Compare costs of using new m vs new tau
+  cost_tau = 0 # TODO
```
ChrisRackauckas: integer?

MSeeker1340: The estimation part here uses theoretical results instead of measuring actual flops. ((23) in the phipm paper).

ChrisRackauckas: and the estimation is with integer values? I was just noting that this might be a place which would cause a further type-instability.

MSeeker1340: Yes. Basically it uses the dimension `n` and `m` and count the number of `A_mul_B!`/`dot`/`axpy!` operations.

ChrisRackauckas: I see

src/exponential_utils.jl
```diff
@@ -468,7 +468,8 @@ Non-allocating version of `phiv_timestep`.
 """
 function phiv_timestep!(u::Vector{T}, t::Float64, A, B::Matrix{T}; tau::Float64=0.0,
```
ChrisRackauckas: `t` isn't necessarily Float64

MSeeker1340:
I was a bit reluctant to finish the cost estimation part, but here it is. I'm not sure whether it is justified to use all these parameters needed just for the sake of choosing whether to adapt `m` or `tau`. But I guess keeping ourselves in line with phipm is the way to go, at least for now. (Also the new KIOPS method basically just copies what phipm does here).

The main difference of my implementation is that I used a crude error estimation for general IOP(p) methods. I haven't factored in Lanczos though --- currently you need to explicitly pass in `iop=2` to get the correct estimate. Will do this in a future update.

One caveat to note is that in order for the cost estimation to work, the number of non-zero elements in `A` (used to estimate the cost of `A_mul_B!`) is needed, which I default to calling `countnz` on `A`). We should take note to also implement this method for the differential operators.

Some simple tests as follows. The premise is the same as the last PR and the previous fixed `m` and `tau` method is changed to use adaptation:

```julia
t = 100.0
tol = 1e-7
u_exact = compute_exact(t)
time = elapsed_timestep_adaptive!(u, t, tol)
err = norm(u - u_exact) / norm(u_exact)
println("time = $time sec, error = $err")
```

    time = 0.042806829 sec, error = 1.3961362635126976e-14

```julia
t = 200.0
tol = 1e-4
u_exact = compute_exact(t)
time = elapsed_timestep_adaptive!(u, t, tol)
err = norm(u - u_exact) / norm(u_exact)
println("time = $time sec, error = $err")
```

    time = 0.011780323 sec, error = 2.8052600778853575e-7

src/exponential_utils.jl
MSeeker1340: By this do you mean it can be other `Real` type or in general complex? I'm not sure if everything will work out OK for complex time though.

ChrisRackauckas: any real

MSeeker1340:
The next step is to look at how `phiv_timestep(!)` with adaptation works. This will not be as straightforward as the fixed `m` and `tau` case, I'm afraid (at least from looking at the previous results on Expokit).

First I shall look at the algorithm as a black box and compare its work-precision diagram with those of the fixed parameter ones. Then I plan to add some print statements inside the adaptation code to get a feel of how the adaptation goes. Finally I can try tweaking the numerous parameters to see how things change (there are *a lot* of magic numbers in the adaptation code...).

Some thoughts on the cost estimation part: if all this trouble is just for determining whether to adapt `m` or `tau`, why not just do both? Or throw a dice and randomly choose one? Another issue is that the adaptation is solely contained within `phiv_timestep`. Can we keep note of the final values of `m` and `tau` and apply them to the next `phiv_timestep` call?

ChrisRackauckas:
> if all this trouble is just for determining whether to adapt m or tau, why not just do both?

Yeah, I don't see why not.

>Can we keep note of the final values of m and tau and apply them to the next phiv_timestep call?

We should make the adaptivity be biased to start from the previous `m` and `tau` for sure.

MSeeker1340:
I want to change the priority and first look at how adaptation goes for a typical `phiv_timestep` call. Some results:

```julia
# High tolerance, no adaptation needed
test(t=50.0, m0=10, tol=1e-4)
```

    Absolute tolerance: 0.0004
    Initial time step unspecified, chosen to be 6.236654789330844
    tk = 0.0, m = 10, tau = 6.236654789330844, error estimate = 1.5574320187111136e-6
    tk = 6.236654789330844, m = 10, tau = 6.236654789330844, error estimate = 1.2696249346678121e-8
    tk = 12.473309578661688, m = 10, tau = 6.236654789330844, error estimate = 4.5194948603321363e-10
    tk = 18.709964367992534, m = 10, tau = 6.236654789330844, error estimate = 4.446341040617938e-11
    tk = 24.946619157323376, m = 10, tau = 6.236654789330844, error estimate = 8.219343868185315e-12
    tk = 31.183273946654218, m = 10, tau = 6.236654789330844, error estimate = 2.1203510051672653e-12
    tk = 37.41992873598506, m = 10, tau = 6.236654789330844, error estimate = 6.69242639537559e-13
    tk = 43.6565835253159, m = 10, tau = 6.236654789330844, error estimate = 2.392124678282721e-13
    tk = 49.893238314646744, m = 10, tau = 0.10676168535325559, error estimate = 5.031273053831423e-29
    ------------------------------
    Absolute error: 3.5973797195618864e-6
    


```julia
# Modest tolerance
test(t=50.0, m0=10, tol=1e-7)
```

    Absolute tolerance: 4.0e-7
    Initial time step unspecified, chosen to be 3.1257317609530046
    tk = 0.0, m = 10, tau = 3.1257317609530046, error estimate = 5.181478597533694e-8
      - Proposed new m: 12, new tau: 2.3815499648380407
      - Cost to use new m: 2641344 flops, new tau: 2798250 flops
      * m = 12, tau = 3.1257317609530046, error estimate = 1.5376256614083216e-9
    tk = 3.1257317609530046, m = 12, tau = 3.1257317609530046, error estimate = 6.833992392665746e-12
    tk = 6.251463521906009, m = 12, tau = 3.1257317609530046, error estimate = 4.841887179015841e-14
    tk = 9.377195282859013, m = 12, tau = 3.1257317609530046, error estimate = 1.0027867476955e-15
    tk = 12.502927043812019, m = 12, tau = 3.1257317609530046, error estimate = 6.180849494289199e-17
    tk = 15.628658804765024, m = 12, tau = 3.1257317609530046, error estimate = 7.758694814192771e-18
    tk = 18.75439056571803, m = 12, tau = 3.1257317609530046, error estimate = 1.2092975942138848e-18
    tk = 21.880122326671035, m = 12, tau = 3.1257317609530046, error estimate = 2.854538160992958e-19
    tk = 25.00585408762404, m = 12, tau = 3.1257317609530046, error estimate = 2.128005772184814e-19
    tk = 28.131585848577046, m = 12, tau = 3.1257317609530046, error estimate = 1.2151354507385144e-19
    tk = 31.257317609530052, m = 12, tau = 3.1257317609530046, error estimate = 5.1935929369796405e-20
    tk = 34.38304937048306, m = 12, tau = 3.1257317609530046, error estimate = 4.001013206213977e-20
    tk = 37.50878113143606, m = 12, tau = 3.1257317609530046, error estimate = 3.9488443521832065e-20
    tk = 40.63451289238906, m = 12, tau = 3.1257317609530046, error estimate = 2.9977845784991696e-20
    tk = 43.76024465334206, m = 12, tau = 3.1257317609530046, error estimate = 1.9151329161101424e-20
    tk = 46.885976414295065, m = 12, tau = 3.1140235857049348, error estimate = 1.1822922834688073e-20
    ------------------------------
    Absolute error: 4.2189824179376445e-10
    


```julia
# Manually set a large tau0
test(t=50.0, m0=10, tau0=10.0 ,tol=1e-7)
```

    Absolute tolerance: 4.0e-7
    tk = 0.0, m = 10, tau = 10.0, error estimate = 6.99495802381299e-6
      - Proposed new m: 17, new tau: 2.6154196898707704
      - Cost to use new m: 1354235 flops, new tau: 2665000 flops
      * m = 14, tau = 10.0, error estimate = 1.5039165133059126e-7
      - Proposed new m: 15, new tau: 7.834070921033023
      - Cost to use new m: 1118070 flops, new tau: 1417843 flops
      * m = 15, tau = 10.0, error estimate = 5.7072379434439144e-8
    tk = 10.0, m = 15, tau = 10.0, error estimate = 1.8119160425942712e-10
    tk = 20.0, m = 15, tau = 10.0, error estimate = 3.954663255810419e-12
    tk = 30.0, m = 15, tau = 10.0, error estimate = 2.300484017861772e-13
    tk = 40.0, m = 15, tau = 10.0, error estimate = 2.2833729849089877e-14
    ------------------------------
    Absolute error: 5.464914920374639e-6
    


```julia
# Use a large m0
test(t=50.0, m0=30, tol=1e-7)
```

    Absolute tolerance: 4.0e-7
    Initial time step unspecified, chosen to be 20.29910245734885
    tk = 0.0, m = 30, tau = 20.29910245734885, error estimate = 1.433124758152144e-11
    tk = 20.29910245734885, m = 30, tau = 20.29910245734885, error estimate = 2.167116955160281e-15
    tk = 40.5982049146977, m = 30, tau = 9.401795085302297, error estimate = 1.6179918794182862e-23
    ------------------------------
    Absolute error: 4.500346691208029e-7
    


```julia
# Very low tolerance
test(t=50.0, m0=10, tol=1e-12)
```

    Absolute tolerance: 4.0e-12
    Initial time step unspecified, chosen to be 0.9884431719340454
    tk = 0.0, m = 10, tau = 0.9884431719340454, error estimate = 1.0779728674291699e-11
      - Proposed new m: 18, new tau: 0.22771261537165308
      - Cost to use new m: 15156792 flops, new tau: 29315000 flops
      * m = 14, tau = 0.9884431719340454, error estimate = 1.0824608125983988e-16
    tk = 0.9884431719340454, m = 14, tau = 0.9884431719340454, error estimate = 1.6124505098434668e-17
    tk = 1.9768863438680908, m = 14, tau = 0.9884431719340454, error estimate = 2.349155589453339e-18
    tk = 2.965329515802136, m = 14, tau = 0.9884431719340454, error estimate = 3.4810977222225085e-19
    tk = 3.9537726877361816, m = 14, tau = 0.9884431719340454, error estimate = 5.165983843921728e-20
    tk = 4.9422158596702275, m = 14, tau = 0.9884431719340454, error estimate = 7.61752065456263e-21
    tk = 5.930659031604273, m = 14, tau = 0.9884431719340454, error estimate = 1.1462839954338764e-21
    tk = 6.919102203538319, m = 14, tau = 0.9884431719340454, error estimate = 1.7763003128771663e-22
    tk = 7.907545375472365, m = 14, tau = 0.9884431719340454, error estimate = 2.6113601646337976e-23
    tk = 8.89598854740641, m = 14, tau = 0.9884431719340454, error estimate = 3.9181927047295455e-24
    tk = 9.884431719340457, m = 14, tau = 0.9884431719340454, error estimate = 1.1459777224757313e-24
    tk = 10.872874891274503, m = 14, tau = 0.9884431719340454, error estimate = 3.370666782417171e-25
    tk = 11.861318063208548, m = 14, tau = 0.9884431719340454, error estimate = 2.0592244260285646e-25
    tk = 12.849761235142594, m = 14, tau = 0.9884431719340454, error estimate = 8.841970376908064e-26
    tk = 13.83820440707664, m = 14, tau = 0.9884431719340454, error estimate = 4.591275635555741e-26
    tk = 14.826647579010686, m = 14, tau = 0.9884431719340454, error estimate = 3.681857842325866e-26
    tk = 15.815090750944732, m = 14, tau = 0.9884431719340454, error estimate = 2.0139549810790861e-26
    tk = 16.803533922878778, m = 14, tau = 0.9884431719340454, error estimate = 1.2103498820705236e-26
    tk = 17.79197709481282, m = 14, tau = 0.9884431719340454, error estimate = 9.306041733899617e-27
    tk = 18.780420266746866, m = 14, tau = 0.9884431719340454, error estimate = 8.561207227940509e-27
    tk = 19.76886343868091, m = 14, tau = 0.9884431719340454, error estimate = 6.4404434949255726e-27
    tk = 20.757306610614954, m = 14, tau = 0.9884431719340454, error estimate = 4.05819148062748e-27
    tk = 21.745749782548998, m = 14, tau = 0.9884431719340454, error estimate = 2.6930969538832687e-27
    tk = 22.734192954483042, m = 14, tau = 0.9884431719340454, error estimate = 2.3698105648064328e-27
    tk = 23.722636126417086, m = 14, tau = 0.9884431719340454, error estimate = 2.2126348917577392e-27
    tk = 24.71107929835113, m = 14, tau = 0.9884431719340454, error estimate = 2.1041022656127387e-27
    tk = 25.699522470285174, m = 14, tau = 0.9884431719340454, error estimate = 1.8878962834906535e-27
    tk = 26.68796564221922, m = 14, tau = 0.9884431719340454, error estimate = 1.5678530300792524e-27
    tk = 27.676408814153262, m = 14, tau = 0.9884431719340454, error estimate = 1.1958897212910044e-27
    tk = 28.664851986087307, m = 14, tau = 0.9884431719340454, error estimate = 9.194609806152343e-28
    tk = 29.65329515802135, m = 14, tau = 0.9884431719340454, error estimate = 7.523175684368832e-28
    tk = 30.641738329955395, m = 14, tau = 0.9884431719340454, error estimate = 6.419850895906014e-28
    tk = 31.63018150188944, m = 14, tau = 0.9884431719340454, error estimate = 6.081745682484846e-28
    tk = 32.61862467382348, m = 14, tau = 0.9884431719340454, error estimate = 5.674626473823003e-28
    tk = 33.60706784575753, m = 14, tau = 0.9884431719340454, error estimate = 5.734198367926971e-28
    tk = 34.59551101769157, m = 14, tau = 0.9884431719340454, error estimate = 6.154401081342939e-28
    tk = 35.583954189625615, m = 14, tau = 0.9884431719340454, error estimate = 4.848000644918654e-28
    tk = 36.57239736155966, m = 14, tau = 0.9884431719340454, error estimate = 5.019530038462463e-28
    tk = 37.5608405334937, m = 14, tau = 0.9884431719340454, error estimate = 4.273523737560822e-28
    tk = 38.54928370542775, m = 14, tau = 0.9884431719340454, error estimate = 3.6381422015269836e-28
    tk = 39.53772687736179, m = 14, tau = 0.9884431719340454, error estimate = 3.0280948929678154e-28
    tk = 40.526170049295835, m = 14, tau = 0.9884431719340454, error estimate = 2.6169463930197427e-28
    tk = 41.51461322122988, m = 14, tau = 0.9884431719340454, error estimate = 2.352065023214744e-28
    tk = 42.503056393163924, m = 14, tau = 0.9884431719340454, error estimate = 2.1856445635327808e-28
    tk = 43.49149956509797, m = 14, tau = 0.9884431719340454, error estimate = 1.9260857896307723e-28
    tk = 44.47994273703201, m = 14, tau = 0.9884431719340454, error estimate = 1.9746560520961e-28
    tk = 45.468385908966056, m = 14, tau = 0.9884431719340454, error estimate = 1.894562941591772e-28
    tk = 46.4568290809001, m = 14, tau = 0.9884431719340454, error estimate = 2.0328003475467205e-28
    tk = 47.445272252834144, m = 14, tau = 0.9884431719340454, error estimate = 1.9882086048049227e-28
    tk = 48.43371542476819, m = 14, tau = 0.9884431719340454, error estimate = 2.091502645366201e-28
    tk = 49.42215859670223, m = 14, tau = 0.5778414032977679, error estimate = 1.6434672214240218e-31
    ------------------------------
    Absolute error: 1.0525739472936103e-13

In general, I found that adaptation is only performed for one or two iterations in the first step. Because the adaptive scheme only tightens `tau` and increases `m`, once you hit the convergence criteria you're stuck with last step's `tau` and `m`, which is sometimes wasteful because as can be seen above a lot of effort is spent on later steps which are very accurate, while the error is dominated by the first step.

In other cases, the actual error may exceed the tolerance level (probably because the error estimate of `phiv!` is not very accurate; could also be because the "exact" result we compare with is in fact not 100% correct due to errors in `expm`).

I think it is fair to say that the scheme is still far from perfect, and improvements can be made. For example, if you look at Algorithm 4 of phipm it is very puzzling why the flops estimation is done *before* the new `m` and `tau` values are constrained (the min/max stuff). If the aim is to judge which is more efficient to adapt, wouldn't it make more sense to estimate the flops of the *actual* new parameters?

Let me finish with the performance side of the tests tomorrow, after which I plan to begin work on the integrators themselves. I don't think it's worthwhile to spend too much time on improving the adaptation scheme at this stage, and we can always revisit this later.

ChrisRackauckas:
>If the aim is to judge which is more efficient to adapt, wouldn't it make more sense to estimate the flops of the actual new parameters?

Yes, this is more in line with how things like stiffness detection is done.

MSeeker1340:
Profiling result:

capture

The small stack to the right of `phiv!` is `arnoldi!` and `phiv!` in the adaptation loop. Which also goes to show that adaptation is only performed scarcely.

The major difference from the non-adaptive profile is the leftmost and rightmost red stack: they corresponds to the lines

```julia
coeffs = [1.0; cumprod(tk ./ (1:p - 1))] # cl = tk^l/l!
```

and

```julia
coeffs = [1.0; cumprod(tau ./ (1:p - 1))] # cl = tau^l/l!
```

Not sure what triggers the change. Previously the compiler is able to do this very efficiently without all the broadcast calls. Should be easy to fix though, I'll just switch back to using regular loops instead of having everything in one go using `cumprod`.

ChrisRackauckas:
```
cumprod(tau ./ (1:p - 1))
```

This allocates `tau ./ (1:p - 1)` and then does the product.

ChrisRackauckas:
Wait, can you explain that profile again? Where is this code and how much is it effecting timing? Is the typing on `tk` and `tau` the same?

MSeeker1340:
They are the same. But I think you touched upon the source of the problem. Previously `t` and `tau` are both `Float64`, so `cumprod` probably has an optimized version for them. But I have changed them to the more general `Real` which might have triggered the difference.

Anyway, let's just use the safer loops instead. And it can be argued that the loops are actually easier to understand ;)

New profile:

capture

Again the small stack to the right of `phiv!` is the adaptation step `arnoldi!` and `phiv!`. The blue section to the left is the line for calculating the `W` matrix in between time steps, and consists mainly of `A_mul_B!`.

ChrisRackauckas:
>Previously t and tau are both Float64, so cumprod probably has an optimized version for them. But I have changed them to the more general Real which might have triggered the difference.

What do you mean you changed them to `Real`? That's an abstract type so `t` and `tau` can never be `Real`.

MSeeker1340:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/376#pullrequestreview-126864311

Changed here:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/376/commits/d384530e10ae8fa0f8819023c94883bcd5049205

ChrisRackauckas:
That doesn't have an effect on typing, just on dispatching behavior. See http://www.stochasticlifestyle.com/type-dispatch-design-post-object-oriented-programming-julia/

MSeeker1340:
Did some experiments on the performance-error tradeoff of the adaptive schemes and their comparison with the fixed `m` and `tau` method. The viewpoint is from the side of the integrator/user. Here the integration time step `h` (not to be confused with `phiv_timestep`'s internal time step) is the parameter that the integrator/user chooses. Since they have no knowledge of `phiv_timestep`'s internal, they either choose a priori `m` and `tau` for the fixed version (say from experience) or use the adaptive and let `phiv_timestep` automates the choice (the default being `m0=10` and `tau0` using the estimation formula from phipm).

The setup is, as usual, the `n=1000` 1D Laplacian. Results: performance vs error

time
error

Performance side, the fixed methods' time does not change much with respect to `h`, as it should be. The adaptive methods have to take more steps in order to satisfy the tolerance criteria, thus they take up more time.

Error side, we see the fixed methods doing worse as `h` increases and the adaptive schemes' error decreases. The latter is a bit surprising, considering that the adaptive scheme has fixed tolerance criterion and we should expect the error to remain on a similar level. This suggest that we may be too conservative in some aspect and it's possible to modify the adaptation scheme for better performance. For example, currently the scheme always goes in the direction of more work, less error. I believe it is possible to also include the other direction, and loosen up `m` and `tau` if the error estimate is way smaller than the tolerance.

Having said that, I think the `phiv_timestep` subroutine is now in a workable state for both the fixed and adaptive version. We should switch priority to the integrators now, and if time permits, go back to this to look at ways to improve the adaptation scheme.

ChrisRackauckas:
>This suggest that we may be too conservative in some aspect and it's possible to modify the adaptation scheme for better performance.

Agreed. Before we move on we might want to see if any simple tweaks can fix this.

PR #378, opened 2018-06-11
Time-stepping Krylov: special case for `expv_timestep`

MSeeker1340:
As suggested by @ChrisRackauckas, added a special case for `phiv_timestep` which calculates just the first term $exp(tA)*b$.

We can already do this using `phiv_timestep(!)` alone, but that requires using a n-by-1 matrix `B`, which is a bit unintuitive. `expv_timestep(!)` is basically a wrapper for `phiv_timestep(!)` that allows a vector `b` as input. (Internally it is reshaped to the n-by-1 matrix form).

I was initially a bit worried about doing extra work for the special case. After more carefully examining the code for updating the `W` matrix and `u`, I found that all the loops are no-ops for `p == 0`, so we're not losing performance here.

### DIFF
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -440,6 +440,47 @@ end
 ###########################################
 # Krylov phiv with internal time-stepping
 """
+    exp_timestep(t,A,b[;adaptive,tol,kwargs...]) -> u
+
+Evaluates the matrix exponentiation-vector product using time stepping
+
+```math
+u = \\exp(tA)b 
+```
+
+The time stepping formula of Niesen & Wright is used [^1]. If the time step 
+`tau` is not specified, it is chosen according to (17) of Neisen & Wright. If 
+`adaptive==true`, the time step and Krylov subsapce size adaptation scheme of 
+Niesen & Wright is used, the relative tolerance of which can be set using the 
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
>I was initially a bit worried about doing extra work for the special case. After more carefully examining the code for updating the W matrix and u, I found that all the loops are no-ops for p == 0, so we're not losing performance here.

Great! Yes, most things should be no-ops. The only thing to be careful about is the interior `expm!` call. The Sidje algorithm in the case of just `expv` has the inner part simplify to just using `exp`, right? I assume that right now you're doing `expm!` on a 1x1 in `expv_timestep`? If that's the case, that's something to specialize on.

MSeeker1340:
> The Sidje algorithm in the case of just expv has the inner part simplify to just using exp, right?

I see the misunderstanding here. The Sidje algorithm computes `expm!` on a `(m + p) x (m + p)` matrix where `m` is the Krylov subspace size and `p` is the highest order phi. Even for `p=0` (that is, `expv_timestep`) we still need to compute `expm!` on the original `m x m` Heisenberg matrix `H`. So there's nothing to optimize/specialize here.

ChrisRackauckas:
oh perfect. Then yeah, this is good.

PR #382, opened 2018-06-12
fixed dts issue in step rejection case and optimized exp of phi vec

sipah00:
current benchmark (linear ODE)
linef


ChrisRackauckas:
Wow, that looks much better! That's what I was expecting it to look like. Cool. Once the renaming is done and tests pass this can be merged. Then we should take a look at some profiles to see if anything can be improved.

PR #384, opened 2018-06-13
Crank-Nicolson Admas Bashforth (ABCN2)

sipah00:


src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -1536,3 +1536,59 @@ end
...
+  f1 = integrator.f.f1
+  f2 = integrator.f.f2
+  if cnt == 1
+    halfdt = dt/2
+    k = f1(uprev + halfdt*integrator.fsalfirst, p, t+halfdt) + f2(uprev + halfdt*integrator.fsalfirst, p, t+halfdt)
+    u = uprev + dt*k  # Midpoint
+    cache.k2 = f2(uprev, p ,t)
+  else
+    # Explicit part
+    k1 = f2(uprev, p, t)
```
ChrisRackauckas: using the `zprev = dt*f1(uprev, p, t)` and the fsallast you can get `k1` without another `f2` evaluation. You can assume that `f2` is the expensive part since it's the non-stiff part that isn't being iterated.

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -1536,3 +1536,59 @@ end
...
+
+function perform_step!(integrator,cache::ABCN2ConstantCache,repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack k2,uf,κ,tol = cache
+  cnt = integrator.iter
+  f1 = integrator.f.f1
+  f2 = integrator.f.f2
+  if cnt == 1
+    halfdt = dt/2
+    k = f1(uprev + halfdt*integrator.fsalfirst, p, t+halfdt) + f2(uprev + halfdt*integrator.fsalfirst, p, t+halfdt)
```
ChrisRackauckas: No reason to do midpoint here for the extra function call. Euler is fine, and there's other theoretical justifications for using Euler here.

src/algorithms.jl
```diff
@@ -167,6 +167,27 @@ struct VCABM5 <: OrdinaryDiffEqAdaptiveAlgorithm end
 
 struct VCABM <: OrdinaryDiffEqAdamsVarOrderVarStepAlgorithm end
 
+# IMEX Multistep methods
+
+struct ABCN2{CS,AD,F,FDT,K,T,T2} <: OrdinaryDiffEqAlgorithm
```
ChrisRackauckas: This is a Newton algorithm. OrdinaryDiffEqImplicitAlgorithm

src/perform_step/adams_bashforth_moulton_perform_step.jl
sipah00: I thought it would be good to use 2nd order method for starting value as `ABCN2` is a 2nd order method.

ChrisRackauckas: Euler is 1st order globally, but in a single step it's second order. Generally that's the result, so you can always use a starter of one order less (that's why RK4 for 5th order Adams is fine). If you assume that the previous derivative and the current derivative are the same, AB2 = forward Euler, so it's a very natural starting assumption, and you can show this is stable.

src/algorithms.jl
sipah00: or `OrdinaryDiffEqImplicitExplicitAlgorithm` ?

ChrisRackauckas: no reason to categorize it that deep. At that point it wouldn't be as useful. This is to catch the implicit restart features.

src/perform_step/adams_bashforth_moulton_perform_step.jl
sipah00: Okay

src/perform_step/adams_bashforth_moulton_perform_step.jl
sipah00: that should be `fsalfirst`? `k1 = dt*fsalfirst - zprev`

ChrisRackauckas: yes. And be careful with the `dt` by changing the later calculation and you're good.

sipah00: okay

PR #386, opened 2018-06-15
Crank-Nicolson Leapfrog (CNLF)

sipah00:


sipah00:
Getting `convergence test` fail when `f1` is not zero, also `sol` is far away from actual sol. It only works when `f1` is zero.

sipah00:
@ChrisRackauckas could you look once, where am I doing mistake.

ChrisRackauckas:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/386/files#diff-1cc7ad80e3fb1d3afc1680b3a07c2ca3R1679

that part is incorrect. You need to skip over `k2` for one round. So you need two caches in order to have it wait one iteration before being reused.

sipah00:
why? we need value of `f1` at `t-dt`.

ChrisRackauckas:
Oh I see what you're doing. Clever 😄 . Yeah, I'm not sure of the issue. But I do see something else. Slack chat.

sipah00:
@ChrisRackauckas In `CNLF2`, we can not use leapfrog method for first step because we need value at the previous point

ChrisRackauckas:
Yeah, so do implicit Euler for the implicit part in the first step. Then, when the first step is at its end, update the historical value to be `f1(uprev,p,t)`, so then it's `t-dt` in the next step.

sipah00:
and for explicit part, should I use `explicit euler` ?

ChrisRackauckas:
Yes, for the explicit part do explicit Euler. Details for a simplified implementation of it:

make the `t-dt` term 0
and double the `t+dt` term
that makes the `g` part equivalent to implicit Euler and the implicit code is all the same in this step except for the `gamma` coefficient.

Then the explicit term will need a bit more handling to switch from explicit Euler to the Leapfrog, but this will just require a quick loop.

sipah00:
cnlf1
`blue` -> `Tsit5`
`Orange` -> `ABCN2`
`green` -> `CNLF2`
this is what I was talking about, `sol` is far away from actual sol (green (CNLF2)).
```julia
f1 = (u,p,t) -> 2u
f2 = (u,p,t) -> 2u
prob = SplitODEProblem(f1,f2,1.0,(0.0,1.0))
sol = solve(prob,CNLF2(),dt=1//256)
```

ChrisRackauckas:
The tests do the two parts separate. Which part is converging incorrectly? The explicit or implicit part?

sipah00:
implicit

sipah00:
okay

sipah00:
no improvement

src/perform_step/adams_bashforth_moulton_perform_step.jl
```diff
@@ -1608,32 +1607,86 @@ function perform_step!(integrator, cache::ABCN2Cache, repeat_step=false)
...
+  if cnt == 1
+    tmp = uprev + dt * (integrator.fsalfirst - du₁)
+  else
+    tmp = uprev2 + 2 * dt * (integrator.fsalfirst - du₁)
   end
-  integrator.f(k,u,p,t+dt)
+  # Implicit part
+  # precalculations
+  if cnt == 1
+    γ = 1
```
ChrisRackauckas: this isn't type-stable. Make it `1//1`. There's probably a similar change that should be done in other places.

ChrisRackauckas:
Then I would start printing out values. You know that the `k1` and `k2` are derivative calculations, so for the linear ODE they should be `alpha * u`. Check to make sure they all line up with the correct values. Check if `W = I - gamma*dt*J` is correct.

sipah00:
yeah, I will start checking every value with theoretical for a test example.

sipah00:
```julia
f1 = (u,p,t) -> 2u
f2 = (u,p,t) -> 2u
prob = SplitODEProblem(f1,f2,1.0,(0.0,1.0))
sol = solve(prob,CNLF2(),dt=0.2)
```
I solved this problem using pen and paper, and compared with values given by code, and both are same.

sipah00:
@YingboMa this method is not performing good, see above plot

ChrisRackauckas:
Can you show the plot with f1=0, and the plot with f2=0?

sipah00:
slack chat

sipah00:
ohh.. I was just going to push mutable cache method.

sipah00:
will open new PR.

ChrisRackauckas:
oh shit, I missed that that one was missing. Just do it in a new PR

PR #388, opened 2018-06-15
WIP : optimization of VCABM

sipah00:
- [x] use circular pointers
- [ ] optimize coefficients calculations


sipah00:
current benchmark (linear_ode)
linear

ChrisRackauckas:
You can do the swapping in the other methods too: https://github.com/sipah00/OrdinaryDiffEq.jl/blob/ef9c8b006773bc04e0a59528c7f1961c526657d1/src/perform_step/adams_bashforth_moulton_perform_step.jl#L669

sipah00:
Okay

PR #389, opened 2018-06-18
WIP: optimization of VCABM 2

YingboMa:
This PR is built upon the work by @sipah00. The optimization that I did includes adding `@inbounds` in the coefficient calculation, fully unrolling `VCABM5` and scaling the `g` coefficients by a factor of `dt`.

### DIFF
--- a/src/adams_utils.jl
+++ b/src/adams_utils.jl
@@ -76,17 +76,17 @@ function ϕ_np1!(cache, du_np1, k)
   @inbounds begin
     @unpack ϕ_np1, ϕstar_n = cache
     for i = 1:k
-      if i == 1
-        if typeof(cache) <: OrdinaryDiffEqMutableCache
-          ϕ_np1[i] .= du_np1
-        else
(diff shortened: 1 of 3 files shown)


YingboMa:
The benchmark problem is from http://nbviewer.jupyter.org/github/JuliaDiffEq/DiffEqBenchmarks.jl/blob/master/NonStiffODE/Pleiades%20Work-Precision%20Diagrams.ipynb

Master:
```
julia> @btime solve($prob,$(VCABM()),abstol=$(1e-12),reltol=$(1e-10), save_everystep=$false);
  6.349 ms (3458 allocations: 185.53 KiB)
```
This branch:
```
julia> @btime solve($prob,$(VCABM()),abstol=$(1e-12),reltol=$(1e-10), save_everystep=$false);
  6.014 ms (3436 allocations: 184.50 KiB)
```

plot

ChrisRackauckas:
Those tests need to be divided by dt as well.

PR #390, opened 2018-06-19
added order condition in const cache method (VCABM)

sipah00:


ChrisRackauckas:
why?

sipah00:
in order changing condtion we need `err` at current order `k` , `k-1` , `k-2` and `k+1`.
This is the reason @YingboMa got bound error and he fixed it in the mutable cache method.

ChrisRackauckas:
I see.

Issue #394, opened 2018-06-19
Full list of exponential integrators to implement


MSeeker1340:
Note: many integrators for linear systems also use exponential-like matrix functions and Krylov approximations (Magnus expansion, adiabatic integrators, symplectic exponential methods, etc). Since they require special structure for the intended problems I have opted to not include them in the list below.

# Classical ExpRK methods

The classical exponential Runge-Kutta type integrators use the semilinear formulation and can be summarized using the Butcher table (see the Hochbruck review [1]). As such they can be easily constructed for small semilinear systems by precomputing the operators $a_{ij}$ and $b_i$. For systems that do not have a semilinear structure, the exponential Rosenbrock formulation casts each integration step to a semilinear problem using the system's Jacobian.

For large systems, Krylov approximation should be used, which typically use fixed subspace dimension `m` and does not use internal time-stepping or adaptation. An optimized Krylov implementation should go in the column direction (i.e. use a single Krylov subspace for all terms in a column) and an $s$-stage ExpRK scheme should require at most $s$ Arnoldi iterations.

## ETDRK schemes

Oldest of the ExpRK methods, originally constructed to mimic the classical Runge-Kutta methods.

- [x] Norsett-Euler/ETDRK1
- [x] Exponential Trapezoid/ETDRK2
- [x] ETDRK3
- [x] ETDRK4

## Stiffly accurate methods of Hochbruck

- [ ] Third order method families (5.8), (5.9) and (5.12) in [2].
- [ ] Hochbruck-Ostermann (5.19) in [2], also in [3].

## Other methods

- [ ] Exponential Midpoint ((2.40) of [1] with $c_2 = 1/2$)
- [ ] Strehmel and Weiner's third order method in [2].
- [ ] ExpRK methods included in [3]:
  - [ ] Krogstad
  - [ ] Strehmel and Weiner's fourth order method.
  - [ ] Friedli
  - [ ] ETD5RKF
- [ ] Affine Lie group schemes in [3]:
  - [ ] RKMK2e
  - [ ] ETD2CF3
  - [ ] Cfree4
  - [ ] RKMK4t

## Adaptive exponential Rosenbrock methods

Not to be confused with adaptive Krylov methods in EPIRK, these are ExpRK methods that use a lower-order error estimator to perform step size adaptation, much like classical adaptive RK methods.

- [ ] exprb32 in [1].
- [ ] exprb43 in [1].

# Exponential multistep methods

[...]

PR #395, opened 2018-06-20
Hochbruck-Ostermann Method

MSeeker1340:
The aim of this PR is to add a few more ExpRK integrators from #394 as well as optimize/organize the existing implementations.

The Hochbruck-Ostermann method is a 5-stage, 4th order ExpRK method (the name is coined by EXPINT). This is so far the only fourth order method proposed that is stiffly accurate, and as such it's an important one to include.

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -160,7 +160,7 @@ module OrdinaryDiffEq
 
   export GenericIIF1, GenericIIF2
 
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
That LGTM. I like the flow that you have going there: the algorithm is very easy to read and I agree that it's quite straightforward now to add other tableaus. Is there a reason why this is still labelled WIP?

MSeeker1340:
I originally intended to also put the optimization work here. Should be better to keep it separate though.

PR #397, opened 2018-06-22
Clean up Nordsieck methods

YingboMa:


YingboMa:
Currently, the stepsize controller is more robust than it used to be. However, there are still many places that need improvement. SUNDIALS is more precise and it takes fewer steps than `JVODE`, even for a trivial linear ODE. Also, I have tried the Pleides problem on `JVODE`. It sometimes diverges from the correct solution.

plot


Should I open another PR to make JVODE more robust, or should I do it in this PR? @ChrisRackauckas

ChrisRackauckas:
Different PR. This is big and I want to review it a bit

src/caches/nordsieck_caches.jl
```diff
@@ -8,53 +9,58 @@ mutable struct AN5ConstantCache{zType,lType,dtType,uType,tsit5Type} <: OrdinaryD
...
 
-function AN5ConstantCache(u, uprev, rate_prototype, uBottomEltypeNoUnits, tTypeNoUnits, dt)
+function alg_cache(alg::AN5,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
   N = 5
   z = [zero(rate_prototype) for i in 1:N+1]
   Δ = u
   l = zeros(tTypeNoUnits,N+1); m = zeros(l)
-  c_LTE = zero(tTypeNoUnits)
-  tau = zeros(typeof(dt), 6)
+  constant = zero(tTypeNoUnits)
```
ChrisRackauckas: That's the least informative variable name.

ChrisRackauckas:
Just that one change for now.

PR #399, opened 2018-06-22
1st order Numerical Differentiation Method (QNDF1)

sipah00:


src/perform_step/bdf_perform_step.jl
```diff
@@ -143,3 +143,73 @@ end
...
+  κ = -0.1850
+  γₖ = 1//1
+  k = 1
+  ρ = dt/dtₙ₋₁
+  D[1,1] = uprev - uprev2 # backward diff
+  temp_D .= D
+  if ρ != 1
+    U!(k,cache)
+    R!(k,ρ,cache)
+    cache.D .= D * (R * U)
```
ChrisRackauckas: you don't need to do matrix multiplications. This will allocate and isn't a generally good way to write it down. Just write the update out.

sipah00: for the method of order more than one, we have to do it by mat mul?, I tried to write in way that it would be suitable for higher order method implementation.

ChrisRackauckas: No it is just a small loop. Shampine wrote it as a matmul because MATLAB, but there should be a much simpler way to loop it

sipah00: Okay, I will try.

sipah00: for this method (order- one), `D`,`R` and `U` are only one element mat, so here we can use variable instead of `1*1` mat.

ChrisRackauckas: Yeah that's fine

sipah00:
It is `BDF` with `kappa = 0`.

ChrisRackauckas:
You should set a value for `qsteady_max` which is greater than 1, like `2//1` so that way it's a quasi-constant step-size. We can optimize the default later.

sipah00:
Okay

src/caches/bdf_caches.jl
```diff
@@ -199,17 +200,18 @@ function alg_cache(alg::QNDF1,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
   fsalfirstprev = zeros(rate_prototype)
   k = zeros(rate_prototype)
 
-  D = Vector{typeof(rate_prototype)}(1)
-  temp_D = Vector{typeof(rate_prototype)}(1)
-  R = Vector{typeof(rate_prototype)}(1)
-  U = Vector{typeof(rate_prototype)}(1)
-  D2 = Vector{typeof(rate_prototype)}(1)
+  D = Vector{typeof(u)}(1)
```
ChrisRackauckas: Shouldn't these be matrices? Well for QNDF1 they should just be `similar(u)` though.

ChrisRackauckas:
Don't forget to alias `QBDF` to `QNDF` with `kappa=0`, and make `kappa` a parameter in `QNDF`.

src/caches/bdf_caches.jl
sipah00: I think it should be vector, we need `del^1(p), del^2(p) ...`, and we can store it in vector of `typeof(u)`

ChrisRackauckas: For 1st order isn't there only one though?

sipah00: yeah

sipah00: only `del^1(p)`

ChrisRackauckas: So in this case you don't need the vector. I guess it's fine anyways if you want to make the code easily generalize to the others.

test/ode/ode_convergence_tests.jl
```diff
@@ -13,128 +13,133 @@ testTol = 0.2
 
 for i = 1:2
   prob = probArr[i]
-  sim = test_convergence(dts,prob,Euler())
```
ChrisRackauckas: This looks like a mistake.

sipah00: yeah I forgot to uncomment. just pushed changes

ChrisRackauckas:
Is this still WIP? Looks done?

PR #401, opened 2018-06-23
Optimizing the classical ExpRK integrators

MSeeker1340:
The first update adds a version of `phiv` that uses a pre-constructed Krylov subspace, which should make the out-of-place versions of the ExpRK methods a bit faster due to fewer `arnoldi` calls. The difference between the out-of-place and in-place versions now should only be the extra allocation of caches, but I need to check this using profiling.

### DIFF
--- a/src/perform_step/exponential_rk_perform_step.jl
+++ b/src/perform_step/exponential_rk_perform_step.jl
@@ -107,12 +107,7 @@ end
 function perform_step!(integrator, cache::NorsettEulerCache, repeat_step=false)
   @unpack t,dt,uprev,u,f,p = integrator
   @unpack rtmp,Jcache,Ks,KsCache = cache
-  if isa(f, SplitFunction)
-    A = f.f1
-  else
-    f.jac(Jcache, uprev, p, t)
-    A = Jcache
-  end
+  A = isa(f, SplitFunction) ? f.f1 : f.jac(uprev, p, t) # get linear operator
   alg = typeof(integrator.alg) <: CompositeAlgorithm ? integrator.alg.algs[integrator.cache.current] : integrator.alg
 
   if alg.krylov
--- a/src/perform_step/exponential_rk_perform_step.jl
+++ b/src/perform_step/exponential_rk_perform_step.jl
@@ -212,18 +212,19 @@ function perform_step!(integrator, cache::ETDRK3ConstantCache, repeat_step=false
   Au = A * uprev
   F1 = integrator.fsalfirst
   if alg.krylov
+    kwargs = [(:m, min(alg.m, size(A,1))), (:norm, integrator.opts.internalnorm), (:iop, alg.iop)]
     # Krylov on F1 (first column)
-    # TODO: reuse Krylov subspace for w1_half
-    w1_half = phiv(dt/2, A, F1, 1; m=min(alg.m, size(A,1)), norm=integrator.opts.internalnorm, iop=alg.iop)
-    w1 = phiv(dt, A, F1, 3; m=min(alg.m, size(A,1)), norm=integrator.opts.internalnorm, iop=alg.iop)
+    Ks = arnoldi(A, F1; kwargs...)
+    w1_half = phiv(dt/2, Ks, 1)
+    w1 = phiv(dt, Ks, 3)
     U2 = uprev + dt/2 * w1_half[:, 2]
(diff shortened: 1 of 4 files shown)


src/perform_step/exponential_rk_perform_step.jl
```diff
@@ -258,19 +259,20 @@ function perform_step!(integrator, cache::ETDRK3Cache, repeat_step=false)
   halfdt = dt/2
   if alg.krylov
     w1_half, w1, w2, w3, phiv_caches = KsCache
+    kwargs = [(:m, min(alg.m, size(A,1))), (:norm, integrator.opts.internalnorm), (:iop, alg.iop), (:cache, tmp)]
```
ChrisRackauckas: this will need to be a named tuple on v0.7 but that's fine.

MSeeker1340: I'll add a comment for this.

MSeeker1340:
Some profiling results, all using `HochOst4`:

The setup is the `n` dimensional system

$$ u' = \Delta u - 0.1u $$

where $\Delta u$ is discretized using the central differencing method as the linear part and $-0.1u$ is taken as the nonlinear part. The time step is chosen to be `dt = 0.1` (shouldn't influence performance in any way).

Profiling results for `step!(integrator)` is shown below. I used different values of `n` for the caching and Krylov versions, as it really does not make sense to compare those two together --- we will almost always use the caching version for small systems and the Krylov version for large systems.

# Caching (n = 50)

For the not-in-place version: 0.000089 seconds (64 allocations: 30.203 KiB)

caching_nip

For the in-place version: 0.000060 seconds (9 allocations: 1.859 KiB)

caching_ip

# Krylov (n = 1000, m = 20)

For the not-in-place version: 0.004848 seconds (940 allocations: 3.288 MiB)

krylov_nip

For the in-place version: 0.004825 seconds (1.28 k allocations: 1.204 MiB)

krylov_ip

# Conclusion

It's a bit hard to interpret the profiling results, but it's safe to say that most of the time is indeed spent on computation (gemv in the caching case and arnoldi/phiv in the Krylov case). Well except for the red section of the in-place Krylov version though: it is likely the case that this causes way more allocations than expected. The line this section correspond to is exponential_rk_perform_step.jl:538

```julia
    @muladd @. @views tmp = uprev + dt * (
      0.5w1_half[:,2] - 0.75w1_half[:,3] + 0.5w1_half[:,4] + w1[:,4] - 0.25w1[:,3] + 
      0.5w2_half[:,3] - w2[:,4] + 0.25w2[:,3] - 0.5w2_half[:,4] + 
      0.5w3_half[:,3] - w2[:,4] + 0.25w3[:,3] - 0.5w3_half[:,4] + 
      w4[:,4] - 0.25w4[:,3] - 0.25w4_half[:,3] + 0.5w4_half[:,4]) # tmp is U5
```

Maybe it's because the calculation is too long and Julia needs to allocate temporaries to store intermediate results? 

BTW, using Coverage.jl could be more helpful than using Julia's internal statistical profiling tools to analyze memory allocations. I should try using it.

ChrisRackauckas:
Yes that's the 12+ broadcast issue. It is fixed on Julia v0.7 so don't work around it.

ChrisRackauckas:
>Again why is this allocating?

This kind of profiling can misattribute lines due to how the profiler samples. So it could be a few lines before/after instead.

ChrisRackauckas:
>40640 V[:, 1] = b / Ks.beta. This can be fixed by @. the whole line right?

b and beta scalar? Then yes `@.` will fix it.

MSeeker1340:
That's indeed the case:

```
0   @. V[:, 1] = b / Ks.beta
```

I think optimization-wise the profiling and malloc analysis indicate that we're in a pretty good shape now. Next I'm going to see if I can tidy up the code a bit and move some of the repetitive sections into functions.

MSeeker1340:
I think that's about it. While I can group together some commonly occuring operations in `perform_step!`, I think writing them out makes it easier to understand while wrapping in a function with lots of arguments can be too obscure.

ChrisRackauckas:
Yes, I prefer simple to read over concise as well. Ltgm

MSeeker1340:
@ChrisRackauckas OK to merge this now?

ChrisRackauckas:
Yes sorry, I meant to merge it when I commented.

PR #402, opened 2018-06-23
WIP : IMEX Euler method

sipah00:


ChrisRackauckas:
Since in the test problem the implicit and explicit parts are the same, the IMEX algorithm was equivalent to the Trapezoid method and so it got super-convergence of order 2.

PR #405, opened 2018-06-26
Implementing the Anas5 method

EllarBooher:
Hello, this is my first time doing this sort of thing and I was able to update my fork and get things to this point so I'm excited.
Yingbo Ma said he wanted to make comments and I want to get this thing off the ground so I am making this pull request.
Please, tell me how to improve this as I am unsure about many things and certain I have made many mistakes. I have not yet added tests because I'm still uncertain of that, but i have added everything else.
Thank you.


src/tableaus/low_order_rk_tableaus.jl
```diff
@@ -1386,3 +1386,100 @@ function DP5_dense_bs(T)
...
+    a51 = T(-11277/8000)
+    a52 = T(171/80)
+    a53 = T(-459/2000)
+    a54 = T(3213/8000)
+    a61 = T(-519/100)
+    a62 = T(5)
+    a63 = T(189/100)
+    a64 = T(-459/200)
+    a65 = T(1)
+    c1 = T2(0)
```
YingboMa: It is a convention that tableaux in `OrdinaryDiffEq.jl` do not have zero entries, and `c1` is actually `c2`. See https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/tableaus/low_order_rk_tableaus.jl#L186 and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/tableaus/high_order_rk_tableaus.jl#L78-L79

src/tableaus/low_order_rk_tableaus.jl
```diff
@@ -1386,3 +1386,100 @@ function DP5_dense_bs(T)
...
+    b2 = T(0.0)
+    b3 = T(0.4632352941176471)
+    b4 = T(0.1607142857142857)
+    b5 = T(0.3112356053532524)
+    b6 = T(-.04166666666666667)
+    Anas5ConstantCache(a21,a31,a32,a41,a42,a43,a51,a52,a53,a54,a61,a62,a63,a64,a65,c1,c2,c3,c4,c5,c6,b1,b2,b3,b4,b5,b6)
+end
+
+Base.@pure function Anas5ConstantCache{T,T2}(::Type{T},::Type{T2})
+    a21 = T(1/10)
```
YingboMa: You can use `1//10` in here to use a [rational number type](https://docs.julialang.org/en/stable/manual/complex-and-rational-numbers/#Rational-Numbers-1) which is a representation that has no error. `1/10` will return a floating point which is not completely accurate in the underlying representation. i.e.
```julia
julia> @printf("%.100lf", 1/10)
0.1000000000000000055511151231257827021181583404541015625000000000000000000000000000000000000000000000
```

EllarBooher:
I have added convergence tests, updated the tableaus to remove the 0s, update the rational numbers in the tableaus, and implemented w as a required parameter of the algorithm.

ChrisRackauckas:
Is this still WIP? Looks good to me.

src/algorithms.jl
```diff
@@ -37,6 +37,9 @@ struct Heun <: OrdinaryDiffEqAdaptiveAlgorithm end
 struct Ralston <: OrdinaryDiffEqAdaptiveAlgorithm end
 struct Midpoint <: OrdinaryDiffEqAdaptiveAlgorithm end
 struct RK4 <: OrdinaryDiffEqAdaptiveAlgorithm end
+struct Anas5{T} <: OrdinaryDiffEqAlgorithm
+ w::T		
+end
```
YingboMa: `w` is not a keyword argument in here. You need to write an outer constructor.
```julia
Anas5(; w=1) = Anas5(w)
```

ChrisRackauckas: Why would 1 be a good default? I don't think it should have a default so the current setup is good.

EllarBooher:
I was able to fix mistakes to the point that it compiles without error, but it does not pass the test for reasons I don't fully understand. I am going to double check the constants.

EllarBooher:
It seems a65, as specified by the author, tends to infinity as dt gets smaller.

ChrisRackauckas:
Where are we at with this?

EllarBooher:
I caught another mistake but it still fails the convergence test, and I don't know why.

ChrisRackauckas:
what order is it getting?

EllarBooher:
Right around 1, sometimes it dips below and sometimes it's slightly above.

ChrisRackauckas:
Check some of the relations to make sure that your coefficients are correct. For example, each row of the a's should equal the corresponding c.

EllarBooher:
I was able to greatly increase the order. It now is 4th order and sometimes gets even higher. The author states that the method tends to 5th order as v = dt*w becomes smaller, is this satisfactory or does it need more work? As it is this fails the convergence test for fifth order, but not catastrophically.

ChrisRackauckas:
You need to wipe those DS store things.

ChrisRackauckas:
Good job! I merged this manually to remove the DS_Store files.

EllarBooher:
Thank you. In the future I will be careful which files I commit.

PR #406, opened 2018-06-26
Exp4 using EPIRK

MSeeker1340:
Mostly reviewing literature today. To fully utilize the EPIRK framework I need to modify `phiv_timestep` to be able to output intermediate results instead of just a single one, which I plan on tackling this week.

Nevertheless, the basic framework of an EPIRK integrator should be pretty much similar to the classical ones and I can even reuse some of the existing infrastructure. The only thing missing from `Exp4` now is code responsible for the actual propagation of the integrator using `phiv_timestep(!)`.

### DIFF
--- a/src/perform_step/exponential_rk_perform_step.jl
+++ b/src/perform_step/exponential_rk_perform_step.jl
@@ -582,9 +582,34 @@ function perform_step!(integrator, cache::Exp4ConstantCache, repeat_step=false)
   @unpack t,dt,uprev,f,p = integrator
   A = f.jac(uprev, p, t)
   alg = typeof(integrator.alg) <: CompositeAlgorithm ? integrator.alg.algs[integrator.cache.current] : integrator.alg
-  f0 = integrator.fsalfirst # f(u0) is fsaled
+  f0 = integrator.fsalfirst # f(uprev) is fsaled
+  ts = [dt/3, 2dt/3, dt]
+  kwargs = [(:tol, integrator.opts.reltol), (:iop, alg.iop), (:norm, integrator.opts.internalnorm), (:adaptive, true)]
 
-  # TODO
+  # Krylov for f(uprev)
+  B1 = [zeros(f0) f0]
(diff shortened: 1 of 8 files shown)


MSeeker1340:
The modification to `phiv_timestep(!)` is complete and I can return my attention to `Exp4` now. Basically what I did is allow `phiv_timestep` to return the value of `u` at multiple time snapshots. As suggested by Tokman, I did not modify the internal timestep to match the snapshot, but instead compute them a posteriori using the already constructed Krylov subspace (only an actual `phiv!` and some vector additions are needed, so this is very efficient).

The reason for doing this is that, as I pointed out before, the EPIRK methods (with Exp4 as an example) are specially structured to use one `phiv_timestep` call to compute multiple entities. By allowing to do so I can have major performance boost for the integrator.

Nevertheless, with the modification `phiv_timestep` is beginning to look overly complicated. There definitely exists opportunities to improve its structure and I should go back to this in the future, along with improvements to the adaptation scheme that I mentioned in the previous weeks.

src/perform_step/exponential_rk_perform_step.jl
```diff
@@ -576,6 +576,95 @@ function perform_step!(integrator, cache::HochOst4Cache, repeat_step=false)
...
+
+  # Krylov for f(uprev)
+  B1 = [zeros(f0) f0]
+  K1 = phiv_timestep(ts, A, B1; kwargs...) # tϕ(tA)f0
+  @inbounds for i = 1:3
+    K1[:,i] ./= ts[i]
+  end
+  w4 = K1 * [-7/300, 97/150, -37/300]
+  u4 = uprev + dt * w4
+  d4 = f(u4, p, t+dt) - f0 - dt * (A*w4) # TODO: what should be the time?
```
ChrisRackauckas: If you sum together the row of a_ij coefficients you should get the c_i coefficients for the time. However, there's no guarantee that this is higher order though since there would be extra order conditions which are dropped in the derivation for autonomous systems. Thus it's probably best to just leave it like this, and document this as first-order for non-autonomous systems.

Issue #366, opened 2018-05-25
expmv and phimv performance


ChrisRackauckas:
I think there's still some things to investigate.

1. Does this view hurt performance moreso than allocating? https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/365#discussion_r190770827
2. How's the new expmv and phimv vs the ones from ExpoKit.jl. What's the performance difference and numerical stability difference? I know ExpoKit.jl does some extra stepping, so is it more numerically stable but with an extra cost? Curious to have this mapped out.

@MSeeker1340

MSeeker1340:
So the preliminary accuracy/performance analysis comparison with Expokit is here:
https://gist.github.com/MSeeker1340/5f5f9f1333824ee7aa7052b98a0e6dc2
The results are... a bit alarming, at least for big `h`. It seems that if the size of Krylov subspace is small then our non-subdividing implementation performs rather poorly.
Of course, when you look at the performance analysis, you can see that Expokit spends much more time in the case of large `h` because of the subdividing. But the gap in accuracy is just too big to justify using the non-subdividing version. It turns out Expokit's approach does make sense after all.
Now, say that we switch back to using Expokit's subdivision approach. Then the problem would be how to formulate a time-stepping method analogous to what Expokit does in the case of `phimv` to higher order phi functions. Guess it's time to start doing some algebra then.

MSeeker1340:
I figured it might be best to use a real-life example to test the accuracy of `expmv`, as well as compare it with explicit RK methods.

So I recorded the two versions of `expmv` as well as RK4's time stepping accuracy when applied to the 1D diffusion equation with central differencing discretization. The system's dimension is `n = 500` and `h` is the Courant number `dt/dx^2`.

![m500](https://user-images.githubusercontent.com/30965945/40592803-4c686f8a-61f1-11e8-83b9-94398c450115.png)
![m30](https://user-images.githubusercontent.com/30965945/40592805-4f22f286-61f1-11e8-966f-f1cf27da2f7d.png)
![m20](https://user-images.githubusercontent.com/30965945/40592809-50cdfda6-61f1-11e8-950e-5531489e15ce.png)
![m10](https://user-images.githubusercontent.com/30965945/40592810-522fb996-61f1-11e8-87d8-dfcc8965c54c.png)
[...]

PR #408, opened 2018-07-02
Packaged cache type for `expv` and `phiv`

MSeeker1340:
One problem I found while writing `Exp4` is the resizing issue of caches used by `phiv(!)`. Since we're using adaptation for the EPIRK methods the Krylov space size `m` changes, but this is not reflected in the caches. In fact, previously we can only set a maximum size and hope that the adaptation does not produce an `m` that is bigger than the limit.

We need to dynamically resize the caches as adaptation proceeds. However one complication is that Julia does not support `resize!` on multidimensional arrays. Therefore the only choice left is to implement our own dynamic cache object.

To demonstrate how `PhivCache` is use:

```julia
using OrdinaryDiffEq: phiv, PhivCache
```

```julia
n = 20; m = 5; p = 3
srand(0)
A = randn(n, n)
t = 1e-2
b = randn(n);
```

```julia
cache = PhivCache{Float64}(m, p)
length(cache.mem)
```

    114

```julia
w = phiv(t, A, b, p; m=m, cache=cache)
```

    20×4 Array{Float64,2}:
      1.84056     1.81877      0.905763     0.301319  
      ...
     -0.054289   -0.00951268   0.00271059   0.00214844

```julia
# Use a bigger m with the same cache
w = phiv(t, A, b, p; m=2*m, cache=cache)
```

    20×4 Array{Float64,2}:
      1.84056     1.81877      0.905763     0.301319  
      ...
     -0.054289   -0.00951268   0.00271059   0.00214844

```julia
# cache is resized under the hood
length(cache.mem)
```

    638

Note that we're not actually calling `reshape` as Julia does not allow resizing vectors whose data is reference by other objects (e.g. through `reshape`). Instead if resizing is needed, a completely new memory chunk is allocated. While this may cause extra time for garbage collection, this should be a rare operation.

Incidentally, because we're using linear memory for `PhivCache` and returning reshaped arrays of the memory chunk instead of views into a larger array, this also solves the contiguity issue of #366.

### DIFF
--- a/src/caches/linear_nonlinear_caches.jl
+++ b/src/caches/linear_nonlinear_caches.jl
@@ -175,10 +175,11 @@ function alg_cache_expRK(alg::OrdinaryDiffEqExponentialAlgorithm, u, f, dt, plis
   if alg.krylov
     ops = nothing # no caching
     # Build up caches used by Krylov phiv
-    Ks = KrylovSubspace{T}(n, min(alg.m, n))
-    phiv_caches = construct_phiv_caches(Ks, maximum(plist))
+    m = min(alg.m, n)
+    Ks = KrylovSubspace{T}(n, m)
+    phiv_cache = PhivCache{T}(m, maximum(plist))
     ws = [Matrix{T}(n, plist[i] + 1) for i = 1:length(plist)]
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
I really like this. I think it makes it cleaner.

PR #410, opened 2018-07-03
Fourth order EPIRK methods

MSeeker1340:
The main reference I'm using for the EPIRK integrators is [this preprint](https://arxiv.org/pdf/1604.00583) by  Rainwater and Tokman. The two fourth order EPIRK integrators are EPIRK4s3A and EPIRK4s3B (the s3 refers to them having effectively three stages), which I plan to add to OrdinaryDiffEq.

With Exp4 in place the implementation is not very difficult. I made a small improvement to allow `phiv_timestep` to accept sub arrays (from `@view`) for both input and output, so that different calls to `phiv_timestep!` can share the same cache.

It should be noted that EPIRK4s3A has three "flavors": horizontal, vertical and mixed. I choose to use the mixed scheme as it requires one less call of `phiv_timestep` and should have better performance (the paper's numerical results also agree with this).

### DIFF
--- a/src/perform_step/exponential_rk_perform_step.jl
+++ b/src/perform_step/exponential_rk_perform_step.jl
@@ -759,6 +759,46 @@ function perform_step!(integrator, cache::EPIRK4s3BConstantCache, repeat_step=fa
   integrator.u = u
 end
 
+function perform_step!(integrator, cache::EPIRK4s3BCache, repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack tmp,rtmp,rtmp2,K,A,B,KsCache = cache
+  f.jac(A, uprev, p, t)
+  alg = typeof(integrator.alg) <: CompositeAlgorithm ? integrator.alg.algs[integrator.cache.current] : integrator.alg
+  f0 = integrator.fsalfirst # f(u0) is fsaled
+  kwargs = [(:tol, integrator.opts.reltol), (:iop, alg.iop), (:norm, integrator.opts.internalnorm), 
+    (:adaptive, true), (:caches, KsCache)]
+
+  # Compute U2 and U3 vertically
+  fill!(@view(B[:, 2]), zero(eltype(B)))
+  B[:, 3] .= f0
+  phiv_timestep!(K, [dt/2, 3dt/4], A, @view(B[:, 1:3]); kwargs...)
+  K[:, 1] .*= 8 / (3*dt)
+  K[:, 2] .*= 16 / (9*dt)
+  ## U2 and R2
+  @. tmp = uprev + @view(K[:, 1]) # tmp is now U2
+  f(rtmp, tmp, p, t + dt/2); A_mul_B!(rtmp2, A, @view(K[:, 1]))
+  @. rtmp = rtmp - f0 - rtmp2 # rtmp is now R2
+  B[:, 4] .= (54/dt^2) * rtmp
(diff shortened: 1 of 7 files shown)


test/linear_nonlinear_krylov_tests.jl
```diff
@@ -49,8 +49,8 @@ end
     prob = ODEProblem(f, u0, (0.0, 1.0))
     prob_ip = ODEProblem{true}(f_ip, u0, (0.0, 1.0))
 
-    dt = 0.1; tol=1e-5
-    Algs = [Exp4]
+    dt = 0.05; tol=1e-5
```
ChrisRackauckas: was there an issue here?

MSeeker1340: If you recall our numerical experiments on `phiv_timestep!`, the provided tolerance is more of a general guideline than a hard limit. This is especially true for small time steps as there's not much room of improvement for the Krylov part and most of the error comes directly from the truncation error of the schemes. 

Using the previous `dt` would fail the test (by a small margin), so I turned it down a bit.

Now that you mentioned it, comparing the results to Tsit5 is not a good test and should probably be rewritten using `test_convergence`. It's best if we can find a nonlinear system with known solution which we can test on.

ChrisRackauckas: Oh I just meant the `dt` change.

We don't need analytical solutions for convergence testing, though we should probably switch to using a better analytical solution when https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/issues/8 is there. @Vaibhavdixit02 did you ever add the analytical solution? https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/pull/14

src/perform_step/exponential_rk_perform_step.jl
```diff
@@ -665,6 +666,69 @@ function perform_step!(integrator, cache::Exp4Cache, repeat_step=false)
   # integrator.k is automatically set due to aliasing
 end
 
+function perform_step!(integrator, cache::EPIRK4s3AConstantCache, repeat_step=false)
+  @unpack t,dt,uprev,f,p = integrator
+  A = f.jac(uprev, p, t)
```
ChrisRackauckas: The EPIRKs should probably use the standard route for grabbing Jacobians, or everything should be updated at the same time.

ChrisRackauckas:
Is there any specialization that can be done on semilinear problems that would be helpful?

src/perform_step/exponential_rk_perform_step.jl
ChrisRackauckas: It's fine, keep it. https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/113 makes DiffEqFunction-based ODEProblems the default, so this'll be the new style.

MSeeker1340:
> Is there any specialization that can be done on semilinear problems that would be helpful?

Definitely. Both the Jacobian and the remainder term can be calculated more conveniently using the knowledge of the linear term. I think we can do this along with the addition of sparse Jacobian support.

I should note that in the arxiv paper the author stated that the remainder term can be taken as the nonlinear term for a semilinear problem. *However*, I believe this is an error on the author's part because EPIRK methods all use Taylor expansion as their basis and this is different from the semilinear partition. The original paper by Tokman, for example, stated clearly that the remainder term should take this particular form.

MSeeker1340:
This should conclude the fourth order EPIRK methods. I shall leave the optimization part as a whole after the fifth order methods are implemented.

PR #413, opened 2018-07-04
0.7 compatibility

ChrisRackauckas:


ChrisRackauckas:
@keno is there a reason why Femtocleaner won't work for this repo? Is it down again?

PR #414, opened 2018-07-04
WIP : QNDF2 and some changes in QNDF1

sipah00:


src/algorithms.jl
```diff
@@ -252,6 +252,28 @@ Base.@pure QNDF1(;chunk_size=0,autodiff=true,diff_type=Val{:central},
...
+  extrapolant::Symbol
+  min_newton_iter::Int
+  max_newton_iter::Int
+  new_jac_conv_bound::T2
+  kappa::κType
+end
+Base.@pure QNDF2(;chunk_size=0,autodiff=true,diff_type=Val{:central},
+                      linsolve=DEFAULT_LINSOLVE,κ=nothing,tol=nothing,
+                      extrapolant=:linear,min_newton_iter=1,
+                      max_newton_iter=7,new_jac_conv_bound = 1e-3,kappa = -1//9,
```
ChrisRackauckas: Shouldn't this be a tuple of two kappas?

src/caches/bdf_caches.jl
```diff
@@ -240,6 +230,57 @@ function alg_cache(alg::QNDF1,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
...
+
+# QNDF2
+
+mutable struct QNDF2ConstantCache{F,uToltype,coefType,coefType1,uType,dtType} <: OrdinaryDiffEqConstantCache
+  uf::F
+  ηold::uToltype
+  κ::uToltype
+  tol::uToltype
+  newton_iters::Int
+  eulercache::ImplicitEulerConstantCache
```
ChrisRackauckas: it's almost the same, but why not start with QNDF1?

src/perform_step/bdf_perform_step.jl
```diff
@@ -279,3 +275,86 @@ function perform_step!(integrator,cache::QNDF1Cache,repeat_step=false)
...
+  k = 2
+  ρ = dt/dtₙ₋₁
+
+  # backward diff
+  D[1] = uprev - uprev2
+  D[2] = D[1] - (uprev2 - uprev3)
+
+  if ρ != 1
+    R!(k,ρ,cache)
+    cache.D .= D * (R * U)
```
ChrisRackauckas: we'll want to optimize this in the future, but for now keep it.

src/algorithms.jl
sipah00: why two values? as given in the paper, for 2nd order `qndf` `kappa` is `-1/9`.

src/caches/bdf_caches.jl
sipah00: `qndf1` also requires one starting value.

sipah00:
I have doubt, in `qndf2` step size should be changed after two consecutive const step, I tried to do that, but it is failing in case where taking const step forcefully results to `EEst > 1`.
Atually scheme given for changing `D` is valid when prev `D` is constructed using values at `t = t_{n}−jh` for `j = 0, 1, 2`.

ChrisRackauckas:
Oh, I see where the confusion is. It's my bad for not catching this in the last review.

>qndf1 also requires one starting value.

No it doesn't. It's indexed with one value from the future. 

https://en.wikipedia.org/wiki/Backward_differentiation_formula#Specific_formulas

BDF1 = Implicit Euler. NDF1 is implicit Euler with an extra constant for more accuracy. Notice the first difference starts from {n+1}, not {n}, which is why it doesn't need a historical value. You shouldn't need to use Euler to start that. And then BDF1 is typically used to start BDF2, etc. which is how the adaptive order algorithm hotstarts itself. This confusion is likely why the NDF2 algorithm is acting weird. And then

>why two values? as given in the paper, for 2nd order qndf kappa is -1/9.

Two so that way you can set the kappa in the BDF1 starter.

sipah00:
As given in the paper, `ndf` methods have extra term `kappa*gamma_k*(y_{n+1}-y_{n+1}^0))`  (Eq. 4) and due to this term `qndf1` becomes
ss

ChrisRackauckas:
Oh, I didn't realize that the predictor wasn't Adams! I see, that's interesting. That means I would do BDF1 as the starter then, i.e. start it with one step that's `kappa=0`.

sipah00:
> start it with one step that's kappa=0

Okay

> I have doubt, in qndf2 step size should be changed after two consecutive const step, I tried to do that, but it is failing in case where taking const step forcefully results to EEst > 1.
Atually scheme given for changing D is valid when prev D is constructed using values at t = t_{n}−jh for j = 0, 1, 2.

?

ChrisRackauckas:
I don't know what you mean by that second question.

sipah00:
step changing scheme -> when step size changes we have to calculate new backward diff `D` by equation `new_D = old_D * R * U`, where `R` depends upon `rho=dt/dt_{n-1}` and `old_D` should be backward diff with step size `dt_{n-1}`, so, for 2nd order method, when we are going to change step size then we have to calculate `new_D` for which we need `old_D` which requires values of function at `t = t_{n} - j*h` where `j = 0, 1, 2`  and `h=dt_{n-1}` (old_step_size), **so we can change step size for next iteration after two consecutive const step**. Now suppose we have taken initial two step with const step size using `implicitEuler` now  in third iteration, if step size has been changed we can calculate `new_D` because we have values of function at `t=t_n, t_n-h, t_n-2*h` (h=prev step size) for creating `old_D`, 
Now we have to keep this step size const for next iteration suppose and if we do it forcefully for next iteration and if it results `EEst>1` (step failed) then what to do in this case?

sipah00:
Now, It is clear to you?

ChrisRackauckas:
Oh I see what you mean. This is why BDFs have to chain. To start it, use BDF1 (just like with QNDF1). So just the same algorithm, but with 1 backward difference and `kappa = 0` to start. Then if you reject, change the stepsize and do a BDF1 step again. This is almost the adaptive order algorithm, so it probably only makes sense to do this for QNDF2 and then do a generic adaptive order QNDF after this

ChrisRackauckas:
The extra cache is overkill. Why not just do the BDF 1? It can reuse almost all of the stepping code

sipah00:
yeah, I will do it in next commits, I just wanted to get the review for lines 311 to 323. 
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/414/files#diff-a47f650c5d66e06a667d8a85cec5c42eR311

ChrisRackauckas:
On rejection, just bring cnt down to one

sipah00:
It would break assignment flow of `uprevs` and `dts`, to correct it we have to add    
extra `if-else` conditions.

sipah00:
matlab paper describes truncation error as `(1/(k+1))*(∇^(k+1)yₙ₊₁)` (formula no. 3) for bdf methods and for ndf it is `(κ*γₖ + inv(k+1)) * (∇^(k+1)yₙ₊₁)` (formula no. 5). 
so we can get error for any order (k), only problem is that we can apply it after required number of const steps.

src/perform_step/bdf_perform_step.jl
```diff
@@ -262,20 +256,129 @@ function perform_step!(integrator,cache::QNDF1Cache,repeat_step=false)
...
+
+  # initial guess
+  z = dt*integrator.fsalfirst
+
+  z, η, iter, fail_convergence = diffeq_nlsolve!(integrator, cache, W, z, tmp, γ, 1, Val{:newton})
+  fail_convergence && return
+  u = tmp + γ*z
+
+  if integrator.opts.adaptive && integrator.success_iter > 0 && flag
+    # bdf1_EEst!(integrator, u, uprev, cache)
```
sipah00: this error est function is the same as in `ImplicitEuler`.

ChrisRackauckas: You probably shouldn't use the `ImplicitEuler` one here. For BDF1, i.e. the first step, it should probably be the `(1/(k+1))*(∇^(k+1)yₙ₊₁)` instead of the SPICE estimator of `ImplicitEuler`.

ChrisRackauckas: oh hmm, that needs a second difference?

src/perform_step/bdf_perform_step.jl
```diff
@@ -262,20 +256,129 @@ function perform_step!(integrator,cache::QNDF1Cache,repeat_step=false)
...
+    D[1] = uprev - uprev2
```
sipah00: from line 343 to 349, I have implemented error est of `bdf1` from Matlab paper (formula no. 3) i.e `EEst = 1//2 * (u - 2*uprev + uprev2)` but when using this error est instead of error est of `ImplicitEuler` results to following error.

```
ERROR: Something went wrong. Integrator stepped past tstops but the algorithm was dtchangeable. Please report this error.
Stacktrace:
 [1] handle_tstop!(::OrdinaryDiffEq.ODEIntegrator{OrdinaryDiffEq.QNDF2{0,true,DiffEqBase.LinSolveFactorize{Base.LinAlg.#lufact!},DataType,Void,Void,Float64,Rational{Int64},:Predictive},Float64,Float64,Void,Float64,Float64,Float64,Array{Float64,1},DiffEqBase.ODESolution{Float64,1,Array{Float64,1},Void,Void,Array{Float64,1},Array{Array{Float64,1},1},DiffEqBase.ODEProblem{Float64,Float64,false,Void,#f,Void,Void,UniformScaling{Int64},DiffEqBase.StandardODEProblem},OrdinaryDiffEq.QNDF2{0,true,DiffEqBase.LinSolveFactorize{Base.LinAlg.#lufact!},DataType,Void,Void,Float64,Rational{Int64},:Predictive},OrdinaryDiffEq.InterpolationData{#f,Array{Float64,1},Array{Float64,1},Array{Array{Float64,1},1},OrdinaryDiffEq.QNDF2ConstantCache{DiffEqDiffTools.UDerivativeWrapper{#f,Float64,Void},Float64,Array{Float64,2},Array{Float64,2},Float64,Float64}}},#f,Void,OrdinaryDiffEq.QNDF2ConstantCache{DiffEqDiffTools.UDerivativeWrapper{#f,Float64,Void},Float64,Array{Float64,2},Array{Float64,2},Float64,Float64},OrdinaryDiffEq.DEOptions{Float64,Float64,Float64,Float64,DiffEqBase.#ODE_DEFAULT_NORM,DiffEqBase.CallbackSet{Tuple{},Tuple{}},DiffEqBase.#ODE_DEFAULT_ISOUTOFDOMAIN,DiffEqBase.#ODE_DEFAULT_PROG_MESSAGE,DiffEqBase.#ODE_DEFAULT_UNSTABLE_CHECK,DataStructures.BinaryHeap{Float64,DataStructures.LessThan},DataStructures.BinaryHeap{Float64,DataStructures.LessThan},Void,Void,Int64,Array{Float64,1},Array{Float64,1},Array{Float64,1}},Float64}) at /home/shubham00/.julia/v0.6/OrdinaryDiffEq.jl/src/integrators/integrator_utils.jl:417

```

ChrisRackauckas: And the EEst values you were getting were sensible? You don't modify `dt` anywhere? This error wouldn't mean that this approach is incorrect or anything, it means there's some egregious `dt` modification going on.

src/perform_step/bdf_perform_step.jl
```diff
@@ -262,20 +256,129 @@ function perform_step!(integrator,cache::QNDF1Cache,repeat_step=false)
...
+      integrator.dt = dtₙ₋₁
+    elseif pass2
+      integrator.dt = pdt
+    end
+  end
+
+  dt = integrator.dt
+
+  if cnt == 1 || cnt == 2 || flag
+    κ = zero(integrator.alg.kappa)
```
sipah00: these changes turns `qndf2` to `bdf1` (line 308 to 310).

ChrisRackauckas: Slick!

sipah00:
[adp.pdf](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/files/2180030/adp.pdf)
this paper describes adaptive order and step conditions of `ode15s` program. please have a look.

ChrisRackauckas:
Yes, this looks perfect. Follow this.

src/perform_step/bdf_perform_step.jl
```diff
@@ -262,20 +256,129 @@ function perform_step!(integrator,cache::QNDF1Cache,repeat_step=false)
...
+
+function perform_step!(integrator,cache::QNDF2ConstantCache,repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack pass1,pass2,pdt,uprev2,uprev3,dtₙ₋₁,dtₙ₋₂,D,D2,R,U = cache
+  cnt = integrator.iter
+  pdt = dt
+  k = 2
+  flag = dtₙ₋₁ != dtₙ₋₂
+
+  if flag && cnt != 1
```
sipah00: here, I have tried to changes `dt` forcefully, this is what I have done -> when prev two step size are not same that means we can't use `qndf2`  so first I tried `bdf1` with last step size so that if this gets success, we can apply `qndf2` in next iteration but if it gets failed try with actual step size that was in last iteration.

ChrisRackauckas: oh yes, that would cause this error. Instead, follow the description described in the paper you found and we should be good. Just calculate the `est`, `est_k-1`, etc.

sipah00: We have to apply `bdf1` in case we can't apply `qndf`, so should I apply `bdf1` until we get the required number of previous constant steps because trick of stepping with the last step in hope that it gets success and would use `qndf` in next iteration, could cause problem (as you mentioned)?
also second paper says -> 
The ode15s is not allowed to change either the order or the step size until a minimum of
(k + 2) consecutive steps are given using the same order formula and the same step size

ChrisRackauckas: The paper is showing that ode15s isn't doing as quasi-constant as I thought. Instead, think of it as a normal BDF and implement it that way. We can tag the quasi-constant part on later, which is just that if `<qsteady_max` then keep `dt` constant, and if `dt` and the order are constant then re-use the Jacobian. In all other cases it's fine to change `dt` around.

ChrisRackauckas:
You'll need to merge master into this or start fresh (i.e. branch and copy over) for v0.7.

sipah00:
@ChrisRackauckas are backwards tests passing now?

ChrisRackauckas:
they should be. Don't worry about the full tests though, just worry about your tests.

sipah00:
my tests (convergence) are passing :grinning:

ChrisRackauckas:
I'll take a look, test, and merge if it seems already. Thanks for notifying.

src/perform_step/bdf_perform_step.jl
```diff
@@ -262,19 +256,200 @@ function perform_step!(integrator,cache::QNDF1Cache,repeat_step=false)
...
+  integrator.k[1] = integrator.fsalfirst
+  integrator.k[2] = integrator.fsallast
+end
+
+function perform_step!(integrator,cache::QNDF2ConstantCache,repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack uprev2,uprev3,dtₙ₋₁,dtₙ₋₂,D,D2,R,U = cache
+  cnt = integrator.iter
+  k = 2
+  if cnt == 1 || cnt == 2
```
ChrisRackauckas: why if cnt == 2?

src/perform_step/bdf_perform_step.jl
```diff
@@ -262,19 +256,200 @@ function perform_step!(integrator,cache::QNDF1Cache,repeat_step=false)
...
+  elseif dtₙ₋₁ != dtₙ₋₂
+    κ = integrator.alg.kappa
+    γ₁ = 1//1
+    γ₂ = 1//1 + 1//2
+    ρ₁ = dt/dtₙ₋₁
+    ρ₂ = dt/dtₙ₋₂
+    D[1] = uprev - uprev2
+    D[1] = D[1] * ρ₁
+    D[2] = D[1] - ((uprev2 - uprev3) * ρ₂)
+  else
```
ChrisRackauckas: this is probably an over-optimization but it's fine.

src/perform_step/bdf_perform_step.jl
sipah00: oh, I forgot to remove.

ChrisRackauckas:
Tests pass for me locally. Just needs to address the last review point.

src/perform_step/bdf_perform_step.jl
sipah00: oh, It is Okay after two-step we can apply qndf2, remember it has an extra term `kappa*gamma_k*(y_{n+1}-y_{n+1}^0))`.

PR #417, opened 2018-07-05
Fifth order EPIRK methods

MSeeker1340:
Again, the integrators EPIRK5s3 and EXPRB53s3 comes from [this paper](https://arxiv.org/pdf/1604.00583). Unfortunately current EPIRK5s3 produces large test error and seems to have only first order convergence. This is most likely a result of a typo/bug and not because of the underlying framework (since the other fifth order method EXPRB53s3 works without problem).

### DIFF
--- a/src/caches/linear_nonlinear_caches.jl
+++ b/src/caches/linear_nonlinear_caches.jl
@@ -513,6 +513,10 @@ function alg_cache(alg::EPIRK5P1,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNo
   EPIRK5P1Cache(u,uprev,tmp,rtmp,rtmp2,K,A,B,KsCache)
 end
 
+struct EPIRK5P2ConstantCache <: ExpRKConstantCache end
+alg_cache(alg::EPIRK5P2,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,
+  uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}}) = EPIRK5P2ConstantCache()
+
 ####################################
 # Multistep exponential method caches
 
--- a/src/caches/linear_nonlinear_caches.jl
+++ b/src/caches/linear_nonlinear_caches.jl
@@ -483,6 +483,10 @@ function alg_cache(alg::EXPRB53s3,u,rate_prototype,uEltypeNoUnits,uBottomEltypeN
   EXPRB53s3Cache(u,uprev,tmp,rtmp,rtmp2,K,A,B,KsCache)
 end
 
+struct EPIRK5P1ConstantCache <: ExpRKConstantCache end
+alg_cache(alg::EPIRK5P1,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,
+  uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}}) = EPIRK5P1ConstantCache()
+
 ####################################
 # Multistep exponential method caches
 
--- a/src/caches/linear_nonlinear_caches.jl
+++ b/src/caches/linear_nonlinear_caches.jl
@@ -428,6 +428,31 @@ struct EPIRK5s3ConstantCache <: ExpRKConstantCache end
 alg_cache(alg::EPIRK5s3,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,
   uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}}) = EPIRK5s3ConstantCache()
 
+struct EPIRK5s3Cache{uType,rateType,matType,KsType} <: ExpRKCache
+  u::uType
+  uprev::uType
+  tmp::uType
+  k::uType
+  rtmp::rateType
+  rtmp2::rateType
+  A::matType
+  B::matType
+  KsCache::KsType
+end
+function alg_cache(alg::EPIRK5s3,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,
+  tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{true}})
+  tmp, k = (similar(u) for i = 1:2)                   # uType caches
+  rtmp, rtmp2 = (zeros(rate_prototype) for i = 1:2)   # rateType caches
+  # Allocate matrices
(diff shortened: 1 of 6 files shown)


src/perform_step/exponential_rk_perform_step.jl
```diff
@@ -799,6 +799,167 @@ function perform_step!(integrator, cache::EPIRK4s3BCache, repeat_step=false)
...
+  B = [zeros(f0) f0]
+  K = phiv_timestep([dt/2, 9dt/10], A, B; kwargs...)
+  U2 = uprev + K[:, 1]
+  R2 = f(U2, p, t + dt/2)  - f0 - A*K[:, 1] # remainder of U2
+  U3 = uprev + K[:, 2] # partially
+
+  # Compute the second group for U3
+  B = [zeros(R2) zeros(R2) zeros(R2) R2]
+  K = phiv_timestep([dt/2, 9dt/10], A, B; kwargs...)
+  U3 .+= 216/(25*dt^2) .* K[:, 1] + 8/dt^2 .* K[:, 2]
```
ChrisRackauckas: For my own future reference, 216 = 8*27, the 8 = 1/2^3 is the factor from the Krylov subspace algorithm because this calculates phi_i / t^i

ChrisRackauckas: Note that this line isn't fusing because of the scalar operations BTW. Just `@.` it. But optimizations can come later.

MSeeker1340:
I wrote a classical ExpRK implementation of EPIRK5s3 (using single step `phiv` instead of `phiv_timestep`) here: https://github.com/MSeeker1340/OrdinaryDiffEq.jl/commit/cb14d529f8f18039d34331941c8fe453ede9a9a9. For the same test problem in the test script, the two versions produce about the same error (~10^-4) whereas other ExpRK methods, be it classical or EPIRK, gives about 10^-7 error.

I have high suspicion that there is a mistake in the original formula. Since the coefficients are originally derived by symbolically solving a set of order conditions using Mathematica by the authors, I can try to substitute back the coefficients to check consistency.

Another angle I can tackle is to implement more EPIRK methods, like the original EPIRK5-P1/P2 integrators by Tokman. If all these don't have problems, then the likelihood that EPIRK5s3 is wrong will certainly increase.

ChrisRackauckas:
Please do EPIRK5-P1. That one seems to be the most efficient when it doesn't have order loss (i.e. not too stiff) so it would be a pretty essential algorithm to have. EPIRK5s3 is really beat out in any case by EXPRB53s3 anyways, so if it doesn't exist then it doesn't have much of a practical difference (though I do like the completionism to be able to verify benchmarks)

src/caches/linear_nonlinear_caches.jl
```diff
@@ -424,6 +424,65 @@ function alg_cache(alg::EPIRK4s3B,u,rate_prototype,uEltypeNoUnits,uBottomEltypeN
...
+  rtmp::rateType
+  rtmp2::rateType
+  A::matType
+  B::matType
+  KsCache::KsType
+end
+function alg_cache(alg::EPIRK5s3,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,
+  tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{true}})
+  tmp, k = (similar(u) for i = 1:2)                   # uType caches
+  rtmp, rtmp2 = (zeros(rate_prototype) for i = 1:2)   # rateType caches
```
YingboMa: Use `zero(rate_prototype)` please. This will trigger a deprecation warning on Julia 0.7.

src/caches/linear_nonlinear_caches.jl
```diff
@@ -424,6 +424,65 @@ function alg_cache(alg::EPIRK4s3B,u,rate_prototype,uEltypeNoUnits,uBottomEltypeN
...
+  # Allocate matrices
+  n = length(u); T = eltype(u)
+  A = Matrix{T}(n, n) # TODO: sparse Jacobian support
+  B = zeros(T, n, 5)
```
YingboMa: Same here, `fill(zero(T), n, 5)`.

src/perform_step/exponential_rk_perform_step.jl
```diff
@@ -799,6 +799,167 @@ function perform_step!(integrator, cache::EPIRK4s3BCache, repeat_step=false)
...
 
+function perform_step!(integrator, cache::EPIRK5s3ConstantCache, repeat_step=false)
+  @unpack t,dt,uprev,f,p = integrator
+  A = f.jac(uprev, p, t)
+  alg = typeof(integrator.alg) <: CompositeAlgorithm ? integrator.alg.algs[integrator.cache.current] : integrator.alg
+  f0 = integrator.fsalfirst # f(uprev) is fsaled
+  kwargs = [(:tol, integrator.opts.reltol), (:iop, alg.iop), (:norm, integrator.opts.internalnorm), (:adaptive, true)]
+
+  # Compute U2 horizontally
+  B = zeros(eltype(f0), length(f0), 4)
```
YingboMa: `fill(zero(eltype(f0)), length(f0), 4)`

src/perform_step/exponential_rk_perform_step.jl
```diff
@@ -799,6 +799,167 @@ function perform_step!(integrator, cache::EPIRK4s3BCache, repeat_step=false)
...
+
+function perform_step!(integrator, cache::EXPRB53s3ConstantCache, repeat_step=false)
+  @unpack t,dt,uprev,f,p = integrator
+  A = f.jac(uprev, p, t)
+  alg = typeof(integrator.alg) <: CompositeAlgorithm ? integrator.alg.algs[integrator.cache.current] : integrator.alg
+  f0 = integrator.fsalfirst # f(uprev) is fsaled
+  kwargs = [(:tol, integrator.opts.reltol), (:iop, alg.iop), (:norm, integrator.opts.internalnorm), (:adaptive, true)]
+
+  # Compute the first group for U2 and U3
+  B = [zeros(f0) f0]
```
YingboMa: `zero(f0)`

MSeeker1340:
Note on the choice coefficients for EPIRK5-P1/P2: the $p$ coefficients aren't really free as they are cancelled out in the final form of the formula. The only free coefficient is $g_{22}$ in both cases. I made the simplest choice of $g_{22} = g_{32}$, so that no intermediate output is needed for the second column. I'm not sure if this is the optimal choice, and the paper doesn't have a discussion about this either.

PR #424, opened 2018-07-09
v0.7 upgrade for exponential integrators

MSeeker1340:
The aim is to get tests related to the exponential integrators (utility_tests.jl, linear_nonlinear_convergence_tests.jl and linear_nonlinear_krylov_tests.jl) to pass and then fix all depwarns. 

For the error fixes, I have temporarily disabled the choice of custom opnorm for the integrators (i.e. no longer uses `integrators.opts.internalnorm` but instead just `LinearAlgebra.opnorm`). In the future this can be added back by expanding `internalnorm` to a tuple of `(norm, opnorm)`, as suggested by @YingboMa.

The convergence tests currently still doesn't pass, because it still uses the old trait-based method to specify the analytic solution. I probably need to follow up on the update progress on DiffEqDevTools to get this to work properly.

### DIFF
--- a/src/exponential_utils.jl
+++ b/src/exponential_utils.jl
@@ -276,7 +276,7 @@ function arnoldi!(Ks::KrylovSubspace{B, T}, A, b::AbstractVector{T}; tol::Real=1
   return Ks
 end
 """
-    lanczos!(Ks,A,b[;tol,m,norm,cache]) -> Ks
+    lanczos!(Ks,A,b[;tol,m,opnorm,cache]) -> Ks
 
 A variation of `arnoldi!` that uses the Lanczos algorithm for Hermitian matrices.
 """
@@ -597,7 +597,7 @@ function phiv_timestep!(U::AbstractMatrix{T}, ts::Vector{tType}, A, B::AbstractM
   verbose && println("Absolute tolerance: $abstol")
   if iszero(tau)
     Anorm = opnorm(A, Inf)
-    b0norm = norm(@view(B[:, 1]), Inf)
+    b0norm = maximum(abs.(@view(B[:, 1])))
     tau = 10/Anorm * (abstol * ((m+1)/e)^(m+1) * sqrt(2*pi*(m+1)) /
       (4*Anorm*b0norm))^(1/m)
     verbose && println("Initial time step unspecified, chosen to be $tau")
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
>The convergence tests currently still doesn't pass, because it still uses the old trait-based method to specify the analytic solution. I probably need to follow up on the update progress on DiffEqDevTools to get this to work properly.

We need to add analytical solutions to `SplitFunction`.

src/exponential_utils.jl
```diff
@@ -597,8 +597,8 @@ function phiv_timestep!(U::AbstractMatrix{T}, ts::Vector{tType}, A, B::AbstractM
   verbose && println("Absolute tolerance: $abstol")
   if iszero(tau)
     Anorm = opnorm(A, Inf)
-    b0norm = norm(@view(B[:, 1]), Inf)
-    tau = 10/Anorm * (abstol * ((m+1)/e)^(m+1) * sqrt(2*pi*(m+1)) /
+    b0norm = maximum(abs.(@view(B[:, 1])))
```
YingboMa: This will allocate a new vector.

YingboMa:
It looks great! There is just one very minor issue that I have commented.

src/exponential_utils.jl
MSeeker1340: Ah shoot, I was still under the impression that the `norm` here is the user-specified one, and going back to use `ODE_DEFAULT_NORM` doesn't work because of the `Inf`.

ChrisRackauckas:
Put a PR in to change the norm interface if necessary. Now is good because if it's breaking then we can do it now without extra deprications due to the 0.7 update.

MSeeker1340:
> Put a PR in to change the norm interface if necessary. Now is good because if it's breaking then we can do it now without extra deprications due to the 0.7 update.

@YingboMa suggested changing `internalnorm` to a tuple of `(norm=..., opnorm=...)`. We can still keep the default vecnorm (`ODE_DEFAULT_NORM`) as is and for the moment use `LinearAlgebra.opnorm` as the default for `opnorm`. In the future we can probably add in a fast normest for `opnorm`.

@ChrisRackauckas if this looks good to you, I shall go ahead and implement the change.

ChrisRackauckas:
Should we instead just add `internalopnorm`?

MSeeker1340:
> Should we instead just add internalopnorm?

Yep this is the simpler solution.

By the way I gave up on my original plan to turn `getV`, `getH`, etc into properties. Turns out the custom property support of v0.7 is not that versatile (you can't shadow existing fieldnames and need to write all properties in one `getproperty` function). Also as Yingbo pointed out we don't really need fancy interfaces for low level utilities.

This should conclude the upgrade of ExpRK methods to v0.7. I'll do a final error/depwarn check on ExpRK after the SplitFunction issue is resolved, and then @YingboMa can merge this.

PR #429, opened 2018-07-12
Change save_start to take saveat into consideration

Vaibhavdixit02:


ChrisRackauckas:
Need to check this test to see if it's the desired behavior: https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/403297358#L957-L1037

ChrisRackauckas:
I realized that this was a bit more intricate than I had at first thought and the best definition would be implementing the one in my head, so I went ahead and did that. We need to get all of the solver packages uniform to this though.

ChrisRackauckas:
Don't forget the PR to DiffEqBayes to remove the special handling.

PR #435, opened 2018-07-16
Remove `prob.f<:Tuple` checks

YingboMa:


src/solve.jl
```diff
@@ -58,7 +58,7 @@ function DiffEqBase.__init(
   allow_extrapolation = alg_extrapolates(alg),
   initialize_integrator=true,kwargs...) where {algType<:OrdinaryDiffEqAlgorithm,recompile_flag}
 
-  if typeof(prob.f)<:Tuple
+  if typeof(prob.f)<:DynamicalODEFunction || typeof(prob.f)<:SplitFunction
```
ChrisRackauckas: `if typeof(prob.f)<:DynamicalODEFunction` is all that's needed. Split functions can only have one `mm`.

PR #439, opened 2018-07-18
Add `JVODE_BDF` algorithm

YingboMa:


YingboMa:
The tests passed locally, but I am sure that this algorithm is not ready for any serious use. However, I am inclined to merge this nevertheless, and I will make another PR to make `JVODE` more robust.

PR #441, opened 2018-07-19
Refactor ROCK2

YingboMa:


YingboMa:
The test failure is unrelated.

src/rkc_utils.jl
```diff
@@ -54,6 +54,69 @@ function maxeig!(integrator, cache::OrdinaryDiffEqConstantCache)
...
       z = uprev + quot*tmp
+    else
+      # An arbitrary change on `z`
+      cache.z = shuffle(cache.z)
+    end
+  end
+  return false
+end
+
+function maxeig!(integrator, cache::OrdinaryDiffEqMutableCache)
```
ChrisRackauckas: where is this power iteration described?

YingboMa: It is a translation from the original Fortran code.
```fortran
c------------------------------------------------------------
c     Rocktrho compute eigmax, a close upper bound of the
c     spectral radius of the Jacobian matrix using a
c     power method (J.N. Franklin (matrix theory)).
c     The algorithm used is a small change (initial vector
c     and stopping criteria) of that of
c     Sommeijer-Shampine-Verwer, implemented in RKC.
c-------------------------------------------------------------
```
I want to move this function into the `DiffEqEigTool.jl` repo.

Issue #416, opened 2018-07-05
Sparse Jacobians


ChrisRackauckas:
https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/220 defines an interface an the whole of DiffEq. Implementing it into OrdinaryDiffEq.jl well will be an ordeal as well. We will need to change the W calculations to take into account possible structure, sparsity, or laziness:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/derivative_utils.jl#L35

Specifically these lines:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/derivative_utils.jl#L66-L73

We will probably want to handle https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/259 and https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/263 simultaneously by updating this one https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/derivative_utils.jl#L82 to have similar options to the in-place one. 

Then we just need to make https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/caches/kencarp_kvaerno_caches.jl#L73-L74 build the correct cache depending on the `jac_prototype`. That should probably be some dispatching function in `derivative_utils.jl` that handles it that we then apply to every cache similarly. In the caches, we will want to make the default linsolve change depending on the type of J/W here: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/caches/kencarp_kvaerno_caches.jl#L92 . So it should probably be `nothing` in the algorithms and then `if nothing` use a smarter default `else` use the user-override. For sparse it's probably smart to just use the sparse-LU, and for matrix-free we can just default to GMRES (maybe add in a simple preconditioner?) from IterativeSolvers.jl.

Then we will need to incorporate it into https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/nlsolve_utils.jl . We should have a dispatch to avoid the whole nonlinear solving if the function is linear. While `u' = Au` isn't much of an interesting case, the IMEX algorithms applied to semilinear equations `u' = Au + f(t,u)` is one case where this will show up so we might as well handle it. I don't think any more modifications will be needed there.

[...]

PR #443, opened 2018-07-19
WIP: Lazy W operator support

MSeeker1340:
Implements the lazy W operators as suggested in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/416#issuecomment-406323306.

I decided to write `WOperator` in OrdinaryDiffEq.jl so that DiffEqOperators is not required for OrdinaryDiffEq. The fact that the interface for `AbstractDiffEqLinearOperator` in DiffEqBase helps a lot.

(By the way, it occurs to me that a lot of methods in common_defaults.jl for DiffEqOperators should, in fact, be written in DiffEqBase instead. I should do a separate PR for this in the future).

Currently the lazy part of `WOperator` is already in place and what remains is to modify `alg_cache` and `calc_W!` to use it. (Also I need to migrate `mass_matrix` to `*DEFunction` in DiffEqBase). The suggestions here https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/416#issuecomment-406330098 is also worth looking into.

I also intend to add more tests for the methods in `derivative_utils.jl` to `utility_tests.jl`, starting with `calc_J!` and `calc_W!`.

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -101,6 +101,67 @@ function calc_W!(integrator, cache::OrdinaryDiffEqMutableCache, dtgamma, repeat_
   end
 end
 
+"""
+    WOperator(mass_matrix,gamma,J[;cache])
+
+A linear operator that represents the W matrix of an ODEProblem, defined as
+
+```math
+W = MM - \\gamma J
+```
+
+where `MM` is the mass matrix (a regular `AbstractMatrix` or a `UniformScaling`),
+`γ` is a real number proportional to the time step, and `J` is the Jacobian
+operator (must be a `AbstractDiffEqLinearOperator`).
+
+`WOperator` supports lazy `*` and `mul!` operations, the latter utilizing an
+internal cache (can be specified in the constructor; default to regular `Vector`).
+It supports all of `AbstractDiffEqLinearOperator`'s interface.
+"""
+mutable struct WOperator{T,
+  MType <: Union{AbstractMatrix,UniformScaling},
+  GType <: Real,
+  JType <: DiffEqBase.AbstractDiffEqLinearOperator{T},
+  CType <: AbstractVector
+  } <: DiffEqBase.AbstractDiffEqLinearOperator{T}
+  mass_matrix::MType
+  gamma::GType
+  J::JType
+  cache::CType
+  function WOperator(mass_matrix, gamma, J; cache=nothing)
+    T = eltype(J)
+    # Convert mass_matrix, if needed
+    if !isa(mass_matrix, Union{AbstractMatrix,UniformScaling})
+      mass_matrix = convert(AbstractMatrix, mass_matrix)
+    end
+    # Construct the cache, default to regular vector
+    if cache == nothing
+      cache = Vector{T}(undef, size(J, 1))
+    end
+    new{T,typeof(mass_matrix),typeof(gamma),typeof(J),typeof(cache)}(mass_matrix,gamma,J,cache)
+  end
+end
+set_gamma!(W::WOperator, gamma) = (W.gamma = gamma; W)
+DiffEqBase.update_coefficients!(W::WOperator,u,p,t) = (update_coefficients!(W.J,u,p,t); W)
+Base.convert(::Type{AbstractMatrix}, W::WOperator) = W.mass_matrix - W.gamma * convert(AbstractMatrix,W.J)
+Base.size(W::WOperator, args...) = size(W.J, args...)
+Base.getindex(W::WOperator, i::Int) = W.mass_matrix[i] - W.gamma * W.J[i]
+Base.getindex(W::WOperator, I::Vararg{Int,N}) where {N} =
+  W.mass_matrix[I...] - W.gamma * W.J[I...]
+Base.:*(W::WOperator, x::Union{AbstractVecOrMat,Number}) = W.mass_matrix*x - W.gamma * (W.J*x)
+function LinearAlgebra.mul!(Y::AbstractVecOrMat, W::WOperator, B::AbstractVecOrMat)
+  # Compute mass_matrix * B
+  if isa(W.mass_matrix, UniformScaling)
+    @. Y = W.mass_matrix.λ * B
+  else
+    mul!(Y, W.mass_matrix, B)
+  end
+  # Compute J * B
+  mul!(W.cache, W.J, B)
+  # Subtract result
+  axpy!(-W.gamma, W.cache, Y)
+end
+
 function calc_W!(integrator, cache::OrdinaryDiffEqConstantCache, dtgamma, repeat_step, W_transform=false)
   @unpack t,uprev,f = integrator
   @unpack uf = cache
--- a/test/utility_tests.jl
+++ b/test/utility_tests.jl
@@ -1,5 +1,5 @@
 using OrdinaryDiffEq: phi, phi, phiv, phiv_timestep, expv, expv_timestep, arnoldi, getH, getV
-using LinearAlgebra, SparseArrays, Random, Test
+using LinearAlgebra, SparseArrays, Random, Test, DiffEqOperators
 
 @testset "Exponential Utilities" begin
   # Scalar phi
@@ -66,4 +66,17 @@ using LinearAlgebra, SparseArrays, Random, Test
   u_exact = Phi[1] * B[:, 1]
   u = expv_timestep(t, A, B[:, 1]; adaptive=true, tol=tol)
   @test norm(u - u_exact) / norm(u_exact) < tol
+end
+
+@testset "Derivative Utilities" begin
+  @testset "WOperator" begin
+    srand(0); y = zeros(2); b = rand(2)
+    mm = I; _J = rand(2,2)
+    W = OrdinaryDiffEq.WOperator(mm, 1.0, DiffEqArrayOperator(_J))
+    OrdinaryDiffEq.set_gamma!(W, 2.0)
+    _W = mm - 2.0 * _J
+    @test convert(AbstractMatrix,W) ≈ _W
+    @test W * b ≈ _W * b
+    mul!(y, W, b); @test y ≈ _W * b
+  end
 end


MSeeker1340:
I will be using ImplicitEuler as my test solver as opposed to KenCarp3 as suggested in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/416#issue-338401389 because apparently KenCarp3 doesn' support problems with custom mass matrices at the moment.

(Speaking of which, I should also remind myself to give a similar warning for the ExpRK methods.)

The out-of-place version turns out to be fairly simple: all we need is add an additional clause in `calc_W!` that checks if a lazy W can be constructed (having a `AbstractDiffEqLinearOperator` `jac_prototype`), and construct one every time it is called. No change is required for `perform_step!` or `diffeq_nlsolve!`.

I'm still using the `mass_matrix` field in the problem for the moment, as I don't want to change too many things at once. A simple test script can be found [here](https://gist.github.com/MSeeker1340/d38af7f47649d8ba30af329a76316432) and the results are:

```
-------------------------
Linear univariate problem
Finite difference error = 0.000756116458477063
Exact Jacobian error = 0.000756116458477063
-------------------------
Linear bivariate problem
Finite difference error = 0.001981690609776791
Exact Jacobian error = 0.001981690609776791
-------------------------
Linear bivariate problem with mass matrix
Finite difference error = 0.2085322284227815
Exact Jacobian error = 0.2085322284227815
```

MSeeker1340:
@ChrisRackauckas I have a question regarding the `W_transform` parameter in `calc_W!`.

If I understand it correctly, a transformed W operator is

$$ W = \frac{1}{\gamma}MM - J $$

right, basically dividing the original W by $\gamma$? When is this form of W used?

I can probably add a `transform` field in `WOperator` to differentiate between the two.

ChrisRackauckas:
>I'm still using the mass_matrix field in the problem for the moment, as I don't want to change too many things at once. A simple test script can be found here and the results are:

This might be nice to keep in OrdinaryDiffEqExtendedTests.jl

>right, basically dividing the original W by $\gamma$? When is this form of W used?

Yes, some methods are written in one way and others in the other way. The transformed version is simply better, but I think Rosenbrock23 hasn't upgraded to it yet.

ChrisRackauckas:
>I decided to write WOperator in OrdinaryDiffEq.jl so that DiffEqOperators is not required for OrdinaryDiffEq. The fact that the interface for AbstractDiffEqLinearOperator in DiffEqBase helps a lot.

Note we will want to upgrade StochasticDiffEq.jl to get similar implicit handling soon after. We want to keep this part of them in sync.

ChrisRackauckas:
I guess it doesn't need a cache anyone could just build a cache into the linsolve.

MSeeker1340:
> I guess it doesn't need a cache anyone could just build a cache into the linsolve?

I think it's more natural to keep `WOperator` self-contained, e.g. for krylov linsolve we need the `mul!` method.

ChrisRackauckas:
>I took an attempt at the second route in #416 (comment) (i.e. migrate the mass_matrix field from *DEProblem to *DEFunction). While this works, I need to change too many things at once both in DiffEqBase and OrdinaryDiffEq. Considering the amount of work left for v0.7 compatibility, I think this might not be a good idea for the time being.

The upgrade is pretty much done. Be bold, make the change at least to ODEs. The change then needs to be propagated to SDEs and DDEs though.

MSeeker1340:
Newest version intended for https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/126.

src/derivative_utils.jl
```diff
@@ -78,7 +78,7 @@ internal cache (can be specified in the constructor; default to regular `Vector`
 It supports all of `AbstractDiffEqLinearOperator`'s interface.
 """
 mutable struct WOperator{T,
-  MType,
+  MType <: Union{UniformScaling,AbstractMatrix},
```
ChrisRackauckas: diffeqoperators are <: AbstractMatrix?

MSeeker1340:
After several trials, I think I've found a good way to optimize the factorization of `WOperator`. As a reference, the requirement for efficient factorization can be found here https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/416#issuecomment-406330098. In particular:

>  "W has the possibility of having a dense cache associated with it. If it's not allocated, for my linsolve I'll need it so allocate it". This should only allocate the first iteration. Then the next iterations we use it: "If W updates then build the dense/sparse W matrix in memory (into the already allocated cache) and factorize that".

Actually, this requirement is already met by the default linear solver `DEFAULT_LINSOLVE` https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/linear_nonlinear.jl. Both the constructor and the solver method are written in a lazy way so that a factorization call to W is only made when its content changes (as specified by the `update_matrix` flag).

However, there's still a hidden issue: the factorization call `p.factorization(A)` expands to `p.factorization(convert(AbstractMatrix, A))` for `A::WOperator`. Since the conversion is just `mm - gamma*J`, it is allocating and thus so is the factorization.

The solution I came up with is to have a lazy cache in `WOperator` (`_nonlazy_form`) which serves as a buffer for `convert`. The cache is allocated not in the constructor, but when `convert` is called for the first time. Subsequent `convert` calls are non-allocating thanks to the cache. This, combined with the lazy `DEFAULT_LINSOLVE`, should satisfy the requirement (It's not entirely non-allocating though, as `lu!(A)` is allocating; however there's not much we can do about this at the moment).

In the same spirit, I have also modified the function cache used by `mul!` to also be lazy (allocated as needed). This means that algorithms that only use factorization for linsolve will never allocate the cache, whereas algorithms such as gmres that linsolves using `mul!` only will never allocate the factorization cache (`_nonlazy_form`).

src/derivative_utils.jl
```diff
@@ -147,6 +157,10 @@ function Base.:\(W::WOperator, x::Union{AbstractVecOrMat,Number})
   end
 end
 function LinearAlgebra.mul!(Y::AbstractVecOrMat, W::WOperator, B::AbstractVecOrMat)
+  if W._func_cache == nothing
+    # Allocate cache only if needed
+    W._func_cache = Vector{eltype(W)}(undef, size(Y, 1))
```
ChrisRackauckas: I think algorithm caches should be able to supply this during construction when possible. That's an optimization we can add later.

src/derivative_utils.jl
```diff
@@ -167,17 +181,27 @@ function LinearAlgebra.mul!(Y::AbstractVecOrMat, W::WOperator, B::AbstractVecOrM
...
   end
 end
 
+"""
+    lazy_W(f) -> t/f
+
+Predicate for determining what kind of function supports the use of lazy W.
+Also used in cache construction.
+"""
+lazy_W(f) = false # default
```
ChrisRackauckas: Can't it always support lazy W? It's non-lazy that not always possible.

ChrisRackauckas:
>The solution I came up with is to have a lazy cache in WOperator (_nonlazy_form) which serves as a buffer for convert. The cache is allocated not in the constructor, but when convert is called for the first time. Subsequent convert calls are non-allocating thanks to the cache. This, combined with the lazy DEFAULT_LINSOLVE, should satisfy the requirement (It's not entirely non-allocating though, as lu!(A) is allocating; however there's not much we can do about this at the moment).

Good idea. And yeah, there will be some allocations but since the `lu!` dominates the time I don't think they are the biggest deal. We should actually benchmark whether this cache is actually worthwhile. 

>In the same spirit, I have also modified the function cache used by mul! to also be lazy (allocated as needed). This means that algorithms that only use factorization for linsolve will never allocate the cache, whereas algorithms such as gmres that linsolves using mul! only will never allocate the factorization cache (_nonlazy_form).

Smart. This vector cache likely will have a big effect. Making it work with `resize!` though might be a little difficult? I can handle that later.

src/derivative_utils.jl
MSeeker1340: To use lazy W the `jac_prototype` needs to be an diffeq operator.

Of course I can modify the constructor so that regular `jac` & `jac_prototype` are automatically wrapped in a `DiffEqArrayOperator`, so it's mostly a design choice. If you think this is preferrable then I can make the change (I would need to add DiffEqOperators to the requirement list though; this was actually my reason for going against it, but now that I think about this you need to import DiffEqOperators anyway for the operator interface so this argument doesn't make sense).

There's also the problem of dual mass matrix of `DynamicalFunction`, which I do not know how to handle just yet. For `SplitFunction` I need to probably add a `jac_prototype` field, while for other `*DEFunction` I think it's better to handle them in their respective packages.

ChrisRackauckas: >There's also the problem of dual mass matrix of DynamicalFunction, which I do not know how to handle just yet.

We don't have any implicit methods for partitioned ODEs, so we can just avoid it for now. I think that for anything like a second order ODE or Hamiltonian problem you'll have that the second mass matrix function has to be I (?), so it may be a non-issue. We'll get to that when we get there.

>For SplitFunction I need to probably add a jac_prototype field, while for other *DEFunction I think it's better to handle them in their respective packages.

Yeah, though remember that StochasticDiffEq.jl will have the same structure as here so we should make sure this same setup gets ported there soon after this merges.

MSeeker1340:
> Making it work with resize! though might be a little difficult?

If `resize!` does not work (I think Julia forbids resizing vectors whose memory are referenced elsewhere, so this might be a problem?), there's always the option of reallocating the cache as a whole since `WOperator` is mutable.

MSeeker1340:
Note that since the part in `calc_W!` that explicitly uses concrete W is removed, tests for implicit methods fail at the moment and I need to update them the same way as `ImplicitEuler`. The tests regarding `WOperator` and `calc_W!` in utility_tests.jl should pass without problem though.

src/derivative_utils.jl
```diff
@@ -54,10 +54,161 @@ function calc_J!(integrator, cache::OrdinaryDiffEqMutableCache, is_compos)
...
+  _concrete_form         # non-lazy form (matrix/number) of the operator
+  WOperator(mass_matrix, gamma, J; transform=false) = new{eltype(J),typeof(mass_matrix),
+    typeof(gamma),typeof(J)}(mass_matrix,gamma,J,transform,nothing,nothing)
+end
+function WOperator(f::DiffEqBase.AbstractODEFunction, gamma; transform=false)
+  @assert DiffEqBase.has_jac(f) "f needs to have an associated jacobian"
+  if isa(f, Union{SplitFunction, DynamicalODEFunction})
+    error("WOperator does not support $(typeof(f)) yet")
+  end
+  # Convert mass matrix, if needed
```
ChrisRackauckas: That's fine for now. I am hoping to support time-dependent mass matrices but we can just add an issue that it's a feature to implement in the future.

src/derivative_utils.jl
```diff
@@ -54,10 +54,161 @@ function calc_J!(integrator, cache::OrdinaryDiffEqMutableCache, is_compos)
...
+    W._concrete_form = W.mass_matrix / W.gamma - convert(Number,W.J)
+  else
+    W._concrete_form = W.mass_matrix - W.gamma * convert(Number,W.J)
+  end
+  W._concrete_form
+end
+Base.size(W::WOperator, args...) = size(W.J, args...)
+function Base.getindex(W::WOperator, i::Int)
+  if W.transform
+    W.mass_matrix[i] / W.gamma - W.J[i]
```
ChrisRackauckas: Since this shows up everywhere and it's constant given by the algorithm, it might make sense to make this a type parameter so the code can simply dispatch off of it, or at least not have large performance cuts because of this conditioning in indexing. I'm not sure this indexing is really used all that much though.

src/derivative_utils.jl
```diff
@@ -102,27 +246,42 @@ function calc_W!(integrator, cache::OrdinaryDiffEqMutableCache, dtgamma, repeat_
...
+      W = WOperator(mass_matrix, dtgamma, J; transform=false)
     else
-      J = ForwardDiff.derivative(uf,uprev)
-      W = 1 - dtgamma*J
+      if isarray
+        J = ForwardDiff.jacobian(uf,uprev)
+      else
+        J = ForwardDiff.derivative(uf,uprev)
+      end
+      W = mass_matrix - dtgamma*J
```
ChrisRackauckas: out of place isn't using WOperator yet?

src/derivative_utils.jl
MSeeker1340: > I'm not sure this indexing is really used all that much though.

This is true both for factorization (it indexes the concrete matrix instead of W) and `mul!` (which is lazy).

src/derivative_utils.jl
MSeeker1340: It's used in the branch where `DiffEqBase.has_jac(f) == true` (https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/443/files/21b1a5b7fae562c8f514cc445002453fa916da6c#diff-35ab888388acf2e7239e17f33e5eefe7R257 and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/443/files/21b1a5b7fae562c8f514cc445002453fa916da6c#diff-35ab888388acf2e7239e17f33e5eefe7R272).

MSeeker1340:
Finally got everything to pass. @ChrisRackauckas If everything so far looks good to you, then we can close this after the CI tests finish. I'll be working on StochasticDiffEq in the meantime.

src/derivative_utils.jl
ChrisRackauckas: Why not all of the time?

ChrisRackauckas: I guess it's not necessary all of the time, and if we don't allow a linsolve option on it then it's pointless for now.

src/derivative_utils.jl
ChrisRackauckas: Sounds good

test/differentiation_traits_tests.jl
```diff
@@ -26,7 +26,7 @@ using OrdinaryDiffEq, Test
     nothing
   end
 
-  Lotka_f = ODEFunction(Lotka,jac=Lotka_jac,tgrad=Lotka_tgrad)
+  Lotka_f = ODEFunction(Lotka,jac=Lotka_jac,jac_prototype=zeros(2,2),tgrad=Lotka_tgrad)
```
ChrisRackauckas: but `nothing` still works?

test/REQUIRE
```diff
@@ -4,3 +4,4 @@ ParameterizedFunctions 3.0.0
 StaticArrays
 DiffEqCallbacks
 DiffEqOperators
+Compat
```
ChrisRackauckas: where is this used?

src/derivative_utils.jl
```diff
@@ -240,6 +240,32 @@ function calc_W!(integrator, cache::OrdinaryDiffEqMutableCache, dtgamma, repeat_
       else
         new_W = false
       end
+    else # concrete W using jacobian from `calc_J!`
```
ChrisRackauckas: This branch is taken when?

test/REQUIRE
MSeeker1340: iterator_test.jl

ChrisRackauckas:
An integration test where an ODE is integrated with the mass matrix being a sparse matrix is required to show that this all captures the new functionality.

test/REQUIRE
ChrisRackauckas: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/commit/6eeabcb4024c054718df761cd2ce03ecf4b02302

ChrisRackauckas:
Fixes https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/259

test/utility_tests.jl
```diff
@@ -106,4 +106,27 @@ end
...
+    A = sparse([-1.0 0.0; 0.0 -0.5])
+    mm = sparse([2.0 0.0; 0.0 1.0])
+    u0 = [1.0, 1.0]; tspan = (0.0,1.0)
+
+    _f = (u,p,t) -> t*(A*u); _f_ip = (du,u,p,t) -> lmul!(t,mul!(du,A,u))
+    fun1 = ODEFunction(_f; mass_matrix=mm)
+    fun2 = ODEFunction(_f; mass_matrix=mm, jac=(u,p,t) -> t*A)
+    fun1_ip = ODEFunction(_f_ip; mass_matrix=mm)
+    fun2_ip = ODEFunction(_f_ip; mass_matrix=mm,
+      jac_prototype=DiffEqArrayOperator(similar(A); update_func=(J,u,p,t) -> (J .= t .* A; J)))
```
ChrisRackauckas: extra tab

ChrisRackauckas:
Looks great. Merge this when the tests pass.

MSeeker1340:
About the integration test: In the temporary test script, I compared solution using 1) lazy W (by providing a jac function/jac prototype) and 2) ForwardDiff (by providing an `ODEFunction` without `jac`) against the analytical solution. I have changed it so that the two solutions are compared against each other, which would be more suitable for checking the lazy W implementation indeed work.

I have also used a time-dependent jacobian so that the update functionality can be checked. As per your suggestion, sparse mass matrix and jacobian are used. One caveat is that I have to manually set the algorithm's linsolve to `lu` instead of the default `lu!`, because the latter does not support sparse arrays yet.

ChrisRackauckas:
>I have also used a time-dependent jacobian so that the update functionality can be checked. As per your suggestion, sparse mass matrix and jacobian are used. One caveat is that I have to manually set the algorithm's linsolve to lu instead of the default lu!, because the latter does not support sparse arrays yet.

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/linear_nonlinear.jl#L14-L17

We can make the default be `nothing` and have that smartly check `f` and use that to determine `lu` vs `lu!`. I think that default can handle most cases well (until we get to DAEs, when we will want to make it smarter and automatically switch to qr, and on BandedMatrices it should default to the banded QR, etc). Separate issue.

PR #451, opened 2018-07-22
Restructure nonlinear solvers

YingboMa:


src/nlsolve/type.jl
```diff
@@ -0,0 +1,36 @@
...
+  ηold::uToltype
+  # The following fields will alias for mutable cache
+  zprev::uType
+  dz::uType
+  tmp::uType
+  b::uType
+  k::rateType
+end
+
+struct Anderson{iip,N} <: AbstractNLsolveSolver
```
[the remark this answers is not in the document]
YingboMa: `N` is the number of vectors that are stored in Anderson acceleration.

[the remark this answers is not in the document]
ChrisRackauckas: How are we benefiting by having it type information?

src/nlsolve/type.jl
```diff
@@ -31,6 +31,6 @@ NLSolverCache(κ, tol, min_iter, max_iter, true,
                 (nothing for i in 1:10)...)
 
 # Default `iip` to `true`, but the whole type will be reinitialized in `alg_cache`
-Anderson{N}(;kwargs...) where {N, iip} = Anderson{N,true}(NLSolverCache(kwargs...))
+Anderson{N}(;kwargs...) where N = Anderson{N,true}(NLSolverCache(kwargs...))
```
ChrisRackauckas: In order to do this `Anderson{N}` it has to be an inner constructor.

src/OrdinaryDiffEq.jl
```diff
@@ -194,4 +196,6 @@ module OrdinaryDiffEq
 
   export AutoSwitch, AutoTsit5, AutoDP5,
          AutoVern6, AutoVern7, AutoVern8, AutoVern9
+
+  export Newton, Anderson, Functional
```
ChrisRackauckas: Those are too common to export. Do something like NLNewton

src/caches/sdirk_caches.jl
```diff
@@ -27,60 +24,84 @@ du_cache(c::ImplicitEulerCache)   = (c.k,c.fsalfirst)
...
+  ImplicitEulerCache(u,uprev,uprev2,du1,fsalfirst,k,z,dz,b,tmp,atmp,J,W,uf,jac_config,linsolve,nlsolve)
 end
 
-mutable struct ImplicitEulerConstantCache{F,uToltype} <: OrdinaryDiffEqConstantCache
+mutable struct ImplicitEulerConstantCache{F} <: OrdinaryDiffEqConstantCache
   uf::F
-  ηold::uToltype
-  κ::uToltype
-  tol::uToltype
-  newton_iters::Int
```
ChrisRackauckas: It should  be strictly typed if we are dispatching on it often. Parameterize it

ChrisRackauckas:
Lgtm

ChrisRackauckas:
Needs to include https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/450

ChrisRackauckas:
One thing to add to the nonlinear solvers is that, if the nonlinear part is actually an `AbstractDiffEqOperator`, then it's linear and the `linsolve` can be used instead. This is important in IMEX algorithms where you might have the stiff part be the linear PDE discretization, and thus instead of doing all of the extra steps you just do a linear solve and get "immediate convergence".

ChrisRackauckas:
This looks good to me. I think that you should just modify it for a linear fast path as mentioned above and then apply it to each implicit algorithm and then we should merge. Another PR can come back and implement Anderson acceleration and tweak some more, but this structure looks very good.

ChrisRackauckas:
I think we merge now and add the linear fast path + Anderson in future PRs.

PR #460, opened 2018-07-30
Lazy W bugfix

MSeeker1340:
Discovered this typo while working on StochasticDiffEq. This isn't technically a bug so it wasn't caught in the test suite, but this would result in the jacobian being unnecessarily wrapped in `DiffEqArrayOperator` if it is already a diffeq operator.


ChrisRackauckas:
👍

PR #461, opened 2018-07-30
Bugfix for WOperator utility tests

MSeeker1340:
Found these when modifying the test script for StochasticDiffEq.

1. In the calc_W! test set, fixed `W` to actually point to the operator in the mutable cache.

2. In the integration tests, the algorithms in the list are now actually used instead of just `ImplicitEuler`. As it turns out most of these algorithms do not actually support custom mass matrices, and are thus removed.


ChrisRackauckas:
>In the integration tests, the algorithms in the list are now actually used instead of just ImplicitEuler. As it turns out most of these algorithms do not actually support custom mass matrices, and are thus removed.

Yup it's just http://docs.juliadiffeq.org/latest/solvers/dae_solve.html#OrdinaryDiffEq.jl-1 Getting more methods working with mass matrices isn't difficult but hasn't been a high priority, but we will revisit that later.

PR #462, opened 2018-07-31
SBDF2, SBDF3 and SBDF4.

sipah00:


src/perform_step/sbdf_perform_step.jl
```diff
@@ -0,0 +1,102 @@
...
+  if cnt == 1
+    tmp = uprev + dt*du₂
+  else
+    tmp = (4*uprev - uprev2)/3 + (dt/3)*(4*du₂ - 2*k₂)
+  end
+  # Implicit part
+  # precalculations
+  γ = 1//1
+  if cnt != 1
+   γ = 1//3
```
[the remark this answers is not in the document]
sipah00: Yeah, fixed it.

ChrisRackauckas:
Is this still WIP?

sipah00:
I will push `SBDF3` in next one hour.

ChrisRackauckas:
ahh okay, that's why. I was looking at it and was thinking SBDF2 looked good to me 👍

src/perform_step/sbdf_perform_step.jl
```diff
@@ -100,3 +100,108 @@ function perform_step!(integrator, cache::SBDF2Cache, repeat_step=false)
...
+  γ = 1//1
+  if cnt > 2
+   γ = 6//11
+  end
+  γdt = γ*dt
+  W = calc_W!(integrator, cache, γdt, repeat_step)
+
+  # initial guess
+  zprev = dt*du₁
+  z = zprev # Constant extrapolation
```
ChrisRackauckas: we'll want to add minimal correction extrapolation as well, which would just be `z=0`. That can be another PR though.

src/perform_step/sbdf_perform_step.jl
```diff
@@ -0,0 +1,207 @@
...
+
+function perform_step!(integrator,cache::SBDF3ConstantCache,repeat_step=false)
+  @unpack t,dt,uprev,u,f,p,alg = integrator
+  @unpack uprev2,uprev3,k₁,k₂ = cache
+  cnt = integrator.iter
+  f1 = integrator.f.f1
+  f2 = integrator.f.f2
+  du₁ = f1(uprev,p,t)
+  du₂ = integrator.fsalfirst - du₁
+  if cnt <= 2
```
ChrisRackauckas: two steps of Euler? It would be better to make the second step be SBDF2.

sipah00: Okay

src/perform_step/sbdf_perform_step.jl
```diff
@@ -0,0 +1,344 @@
...
+  integrator.fsallast = cache.k
+  resize!(integrator.k, integrator.kshortsize)
+  integrator.k[1] = integrator.fsalfirst
+  integrator.k[2] = integrator.fsallast
+  integrator.f(integrator.fsalfirst, integrator.uprev, integrator.p, integrator.t)
+end
+
+function perform_step!(integrator, cache::SBDF3Cache, repeat_step=false)
+  @unpack t,dt,uprev,u,f,p,alg = integrator
+  @unpack tmp,uprev2,uprev3,k,k₁,k₂,du₁,z = cache
```
ChrisRackauckas: missing a du2, need it in the cache

ChrisRackauckas:
Fix the du2 stuff so that way the SBDF3 and 4 can run, but they likely won't converge either so I started marking some tests as `@test_broken`. When 3 and 4 are able to converge (though at order 2), let's merge and then in future PRs we can keep working on these.

ChrisRackauckas:
I also made the split method test asymmetric so that way it can catch that the bug doesn't occur when `f1 = f2`.

src/perform_step/sbdf_perform_step.jl
sipah00: or we can replace `du2` with `(integrator.fsalfirst - du₁)`.

ChrisRackauckas: Even better. Good idea.

ChrisRackauckas:
Do more tests need to be labeled broken?

sipah00:
yeah

ChrisRackauckas:
Merge when tests pass

PR #467, opened 2018-08-03
Refactor interpolants

devmotion:
I tried to come up with some refactorization of the interpolants as discussed in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/214#issuecomment-338249220:
- I renamed the concrete implementations to `_ode_interpolant` and `_ode_interpolant!`
- Now `ode_interpolant` and `ode_interpolant!` just call these methods, depending on the cache type and whether the desired output is `nothing`
- The `@def` macro is used to avoid code duplications in (among others) unpacking of tableaus and computation of coefficients of the polynomials

### DIFF
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -332,7 +332,7 @@ function ode_interpolant(Θ,dt,y₀,y₁,k,cache,idxs,T::Type{Val{TI}}) where TI
 end
 
 function ode_interpolant(Θ,dt,y₀,y₁,k,cache::OrdinaryDiffEqMutableCache,idxs,T::Type{Val{TI}}) where TI
-  if typeof(idxs) <: Number || typeof(y₀) <: Union{Number,SArray}
+  if typeof(idxs) <: Number
     _ode_interpolant(Θ,dt,y₀,y₁,k,cache,idxs,T)
   else
     # determine output type
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -324,9 +324,16 @@ By default, simpledense
   nothing
 end
 
+"""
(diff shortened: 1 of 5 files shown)


src/dense/generic_dense.jl
```diff
@@ -324,9 +324,16 @@ By default, simpledense
...
 
+"""
+ode_interpolant and ode_interpolant! dispatch
+"""
+function ode_interpolant(Θ,dt,y₀,y₁,k,cache,idxs,T::Type{Val{TI}}) where TI
+  _ode_interpolant(Θ,dt,y₀,y₁,k,cache,idxs,T)
+end
+
 function ode_interpolant(Θ,dt,y₀,y₁,k,cache::OrdinaryDiffEqMutableCache,idxs,T::Type{Val{TI}}) where TI
   if typeof(idxs) <: Number || typeof(y₀) <: Number
```
ChrisRackauckas: I think this should have `typeof(y₀) <: Union{Number,SArray}`

ChrisRackauckas:
Smart! I like this a lot, thanks! Merge when tests pass.

src/dense/generic_dense.jl
devmotion: This is just copied, nothing new here. Can it actually happen that this is called with static vectors and a mutable cache?

ChrisRackauckas: Oh I missed the mutable cache part. I don't think it's possible. I'm pretty sure any method would error, so I guess just `if typeof(idxs) <: Number` is all that's necessary.

ChrisRackauckas:
Thanks! This is quite a massive improvement in code quality.

PR #471, opened 2018-08-05
Add derivatives of interpolation for DP5, SSPRK, and OwrenZen

devmotion:
I added derivatives to some of the interpolations (see https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/427), using mostly broadcasts and `@views` (see https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/214#issuecomment-410526011).

### DIFF
--- a/src/dense/interpolants.jl
+++ b/src/dense/interpolants.jl
@@ -9,53 +9,143 @@ end
 """
 Hairer Norsett Wanner Solving Ordinary Differential Euations I - Nonstiff Problems Page 192
 """
+@def dp5pre0 begin
+  b10 = @evalpoly(Θ, 0, 1)
+  b20 = @evalpoly(Θ, 0, 1, -1)
+  b30 = @evalpoly(Θ, 0, 0, 1, -1)
+  b40 = @evalpoly(Θ, 0, 0, 1, -2, 1)
+end
+
 @muladd function _ode_interpolant(Θ,dt,y₀,y₁,k,cache::Union{DP5ConstantCache,DP5Cache,DP5ThreadedCache},idxs::Nothing,T::Type{Val{0}})
-  Θ1 = 1-Θ
-  #@. y₀ + dt*Θ*(k[1]+Θ1*(k[2]+Θ*(k[3]+Θ1*k[4])))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Test failures may just be due to timing?

devmotion:
The (main) issue is the test of the 5th derivative of the OwrenZen5 interpolation. Even on my local computer testing got stuck there. However, I can run this test manually by calling it directly in the REPL, aborting it (the first run is extremely slow), and calling it again (then it succeeds without any noticeable delay in less than a second). So it seems compilation is extremely slow. Do you know why that's the case? Should we just remove the slowest test (testing only up to the 4th derivative still takes quite some time but succeeds on my computer)?

ChrisRackauckas:
It's because of how ForwardDiff works. It's building successively larger tuples, and it's multiplicative because it's nesting duals of the duals.

devmotion:
Can we circumvent that issue somehow?

ChrisRackauckas:
Not with current autodiff given the tuple usage. Maybe mixed auto and numerical differencing could do it sufficiently well.

devmotion:
Could we just test derivatives sequentially, i.e. test the implementation of the first derivative by differentiating the interpolation of the solution, the implementation of the second derivative by differentiating the implementation of the first derivative and so on? In that way we have to compute only first derivatives via ForwardDiff.

ChrisRackauckas:
Yeah that would do it

devmotion:
I added some out of place tests that have been performed only for in place methods so far and a test of Rodas4. Moreover, since the computation of higher-order derivatives is a lot faster now I enabled tests of all derivatives for Vern9, so this fixes https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/423 as well. Interestingly the relative tolerance for the Vern9 tests has to be set to a quite high value (2e-2) compared to the relative tolerances in all other tests - could this indicate that there's a bug in the implementation of the Vern9 interpolation?

ChrisRackauckas:
it could indicate a bug or a numerical instability, or that the steps are just much larger and so this has a negative effect on the higher order derivatives since they will be a lower order approximation

PR #472, opened 2018-08-06
Update ExpRK alongside ExponentialUtilities

MSeeker1340:
Migrates exponential_utils.jl to an independent package ExponentialUtilities (https://github.com/JuliaDiffEq/ExponentialUtilities.jl). Also contains updates for EPIRK and convergence tests.

BTW @ChrisRackauckas how can I publish ExponentialUtilities to Julia's registry and set up the CI?


ChrisRackauckas:
Did you not generate the package via Pkg.generate? I don't see the CI ymls in there.

ChrisRackauckas:
Registration is done via attobot. Just set a release. 

I think it's easiest to just `Pkg.generate` from v0.6 and copy the code over. However you get the ymls over is fine though.

MSeeker1340:
> I think it's easiest to just Pkg.generate from v0.6 and copy the code over.

I'll give this a try, thanks!

MSeeker1340:
Added autodiff support, mimicking what the implicit methods do.

ChrisRackauckas:
The callbacks are fixed up so if you rebase then it should be done.

PR #497, opened 2018-09-06
Interpolation at discontinuity toggle

YingboMa:
Ref: https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/203

### DIFF
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -102,7 +102,7 @@ ode_interpolation(tvals,ts,timeseries,ks)
 Get the value at tvals where the solution is known at the
 times ts (sorted), with values timeseries and derivatives ks
 """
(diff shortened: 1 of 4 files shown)


src/dense/generic_dense.jl
```diff
@@ -102,7 +102,7 @@ ode_interpolation(tvals,ts,timeseries,ks)
 Get the value at tvals where the solution is known at the
 times ts (sorted), with values timeseries and derivatives ks
 """
-function ode_interpolation(tvals,id,idxs,deriv,p)
+function ode_interpolation(tvals,id,idxs,deriv,p,continuity)
```
ChrisRackauckas: we should have a stub method with no continuity choice that expands to one with `:left` so that it's easier to make the change. Otherwise we need a bunch of version bounds.

PR #498, opened 2018-09-07
New DiffEq testing structure

YingboMa:


YingboMa:
Ref: https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/319

PR #499, opened 2018-09-13
InterpolationData fix

YingboMa:


ChrisRackauckas:
Merge and tag a patch release when this passes.

PR #505, opened 2018-09-26
Add a `saveat` test with a callback & fix the test case

YingboMa:


YingboMa:
When the `saveat` time (in this case it is `3.192754284070482`) coincide with the callback saving, we will end up with having three exact same timepoints saved.

```julia
sol = solve(prob,Tsit5(),callback=callback_single,saveat=3.192754284070482)
@test_broken count(x->x==3.192754284070482, sol.t) == 2
```

src/integrators/integrator_utils.jl
```diff
@@ -89,8 +90,9 @@ function savevalues!(integrator::ODEIntegrator,force_save=false,reduce_size=true
         copyat_or_push!(integrator.sol.alg_choice,integrator.saveiter,integrator.cache.current)
       end
     end
+    saved = true
   end
-  if force_save || (integrator.opts.save_everystep && integrator.iter%integrator.opts.timeseries_steps==0)
+  if (force_save && !saved) || (integrator.opts.save_everystep && integrator.iter%integrator.opts.timeseries_steps==0)
```
ChrisRackauckas: if a saveat point is just below the callback point, does this block it from being saved? Or has savevalues already been called for that?

YingboMa: Ah, I didn't think about that case.
```julia
julia> t -= eps(t)
3.1927542840704817

julia> sol = solve(prob,Tsit5(),callback=callback_single,saveat=t); sol.t
7-element Array{Float64,1}:
  0.0
  3.1927542840704817
  3.192754284070482
  6.385508568140963
  9.578262852211415
  9.578262852211445
 15.0
```
Could you give me some pointers about this problem?

src/callbacks.jl
```diff
@@ -203,9 +203,15 @@ function apply_callback!(integrator,callback::ContinuousCallback,cb_time,prev_si
     change_t_via_interpolation!(integrator,integrator.tprev+cb_time)
   end
   saved_in_cb = false
+  oldopt = integrator.opts.save_everystep
 
   @inbounds if callback.save_positions[1]
```
ChrisRackauckas: This still only runs when `callback.save_positions[1] == true`, so the `saveat`s are skipped. I would do a normal `savevalues!` before checking this, and then check to whether you should force one additional save by checking `sol.t` for if it was saved already.

ChrisRackauckas:
AppVeyor tests are failing. Since events touch a lot of floating point handling which can change between platforms, I would take those seriously.

PR #507, opened 2018-09-28
Make `src/callbacks.jl` generic across JuliaDiffEq

YingboMa:
Ref: https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/164


src/integrators/integrator_utils.jl
```diff
@@ -464,8 +464,18 @@ function reset_fsal!(integrator)
   # integrator.reeval_fsal = false
 end
 
-function (integrator::ODEIntegrator)(t,deriv::Type=Val{0};idxs=nothing)
+function (integrator::ODEIntegrator)(t,deriv::Type=Val{0},relative=false;idxs=nothing)
+  relative && (t = integrator.tprev+integrator.dt*t)
```
ChrisRackauckas: if we do this, then it should be added to every integrator type.

src/integrators/integrator_interface.jl
```diff
@@ -12,14 +12,17 @@ function DiffEqBase.change_t_via_interpolation!(integrator::ODEIntegrator,t,modi
...
     integrator.t = t
     integrator.dt = integrator.t - integrator.tprev
-    reeval_internals_due_to_modification!(integrator)
+    DiffEqBase.reeval_internals_due_to_modification!(integrator)
     if T
       solution_endpoint_match_cur_integrator!(integrator)
     end
   end
 end
+DiffEqBase.change_t_via_interpolation!(integrator::ODEIntegrator,
```
ChrisRackauckas: why two functions?

YingboMa: We want to make sure that the first arg of `change_t_via_interpolation!` is specialized, yet, it needs to take both `ODEIntegrator` and `DDEIntegrator`.

ChrisRackauckas: why not a union?

ChrisRackauckas: oooooh I see. Ahh yeah, since we need something for DelayDiffEq.jl to overload. Add a comment.

src/alg_utils.jl
```diff
@@ -1,5 +1,7 @@
 isautodifferentiable(alg::OrdinaryDiffEqAlgorithm) = true
 
+DiffEqBase.isdiscrete(alg::FunctionMap) = true
```
ChrisRackauckas: This should probably be used in place of things like https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/solutions/solution_interface.jl#L61

PR #509, opened 2018-09-30
Add Dual of Float32/Float64 to CompiledFloats

YingboMa:
This helps the performance of doing sensitivity analysis via `ForwarDiff.jl`.

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -44,7 +44,9 @@ module OrdinaryDiffEq
    :($(esc(ex)))
   end
 
-  const CompiledFloats = Union{Float32,Float64}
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
We should definitely add some tests that use dual numbers.

YingboMa:
I don't know how to write a test for this, but I have confirmed it locally by profiling.

BTW, could we annotate all of the tableaux with `Base.@pure`?

ChrisRackauckas:
>BTW, could we annotate all of the tableaux with Base.@pure?

I think we can? There was an issue with BigFloat I think since those all have a mutable part in the background, but for the compiled floats it should be fine. I'd check the compile time though, and see if it has a noticable runtime difference.

>I don't know how to write a test for this, but I have confirmed it locally by profiling.

PR #513, opened 2018-10-01
`f.f1` is always an `ODEFunction`

YingboMa:
Ref: https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/158


ChrisRackauckas:
Why is this helpful?

YingboMa:
This would avoid a lot of the corner cases where the f1 function is an operator. For instance, we need to check if f1 is an operator before every has_jac(f), but with PR, we can avoid this happily.

ChrisRackauckas:
This needs a lower bound then?

YingboMa:
Right. I will add a lower bound in an hour or so, thanks!

YingboMa:
An operator can be handled by using `A = f.f1.f`.

ChrisRackauckas:
So an ODEFunction can hold a DiffEqOperator?

YingboMa:
Yes

PR #514, opened 2018-10-02
`save_on` option in `solve`

MSeeker1340:
Add a `save_on` boolean flag to `DEOptions` that forces the integrator to not save values when stepping, giving the user more control over the saving behavior. The option has a higher precedence over `save_everystep` and `saveat` but has lower precedence than `save_start` and `save_end`.

### DIFF
--- a/src/integrators/integrator_utils.jl
+++ b/src/integrators/integrator_utils.jl
@@ -55,30 +55,54 @@ function modify_dt_for_tstops!(integrator)
 end
 
 function savevalues!(integrator::ODEIntegrator,force_save=false,reduce_size=true)
-  while !isempty(integrator.opts.saveat) && integrator.tdir*top(integrator.opts.saveat) <= integrator.tdir*integrator.t # Perform saveat
-    integrator.saveiter += 1
-    curt = pop!(integrator.opts.saveat)
-    if curt!=integrator.t # If <t, interpolate
-      ode_addsteps!(integrator)
-      Θ = (curt - integrator.tprev)/integrator.dt
-      val = ode_interpolant(Θ,integrator,integrator.opts.save_idxs,Val{0}) # out of place, but no force copy later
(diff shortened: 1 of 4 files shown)


YingboMa:
> Reason: in apply_callback! and apply_discrete_callback!, if save_positions[1] == true, we will first attempt savevalues!.

This is incorrect. The first `savevalues!` is unconditional in `apply_callback!` and `apply_discrete_callback!`.

> which is checked by last(integrator.sol.t) == integrator.t; we should probably change this hack btw

Why is it a hack? Would explicitly return `issaved` in the `savevalues!` be better?

> I think we really should have a SaveOption struct to group together these options.

Agreed. Could you please open an issue for this? Thanks!

MSeeker1340:
> This is incorrect. The first savevalues! is unconditional in apply_callback! and apply_discrete_callback!.

Yes the call to `savevalues!` is unconditional but it might do nothing.

> Would explicitly return issaved in the savevalues! be better?

Yeah this is what I have in mind. The check itself is not very problematic so calling it a hack may be unjustified, but using a boolean return value is definitely a better practice imo.

MSeeker1340:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/515

ChrisRackauckas:
If we want to do this then we should make sure it's universal. We'll need both the other integrator types to similarly apply this. Each at least needs an issue

MSeeker1340:
> If we want to do this then we should make sure it's universal. We'll need both the other integrator types to similarly apply this. Each at least needs an issue

By *this* you mean?

ChrisRackauckas:
>By this you mean?

The `save_on` option should work on Sundials, ODEInterfaceDiffEq, and StochasticDiffEq (it should just naturally work with DelayDiffEq without modification), and we should modify the common interface / integrator interface docs to explain it.

MSeeker1340:
Rebased onto current master.

PR #519, opened 2018-10-10
Bump ExponentialUtilities to v1.2.0

MSeeker1340:


ChrisRackauckas:
MWTP and tag.

ChrisRackauckas:
This should've dropped the `cache =` parts too (the reason for the lower bound)

PR #522, opened 2018-10-12
Document `savevalues!` and make it return booleans

YingboMa:
Per the discussion in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/514#issuecomment-426416101.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -205,11 +205,11 @@ function apply_callback!(integrator,callback::ContinuousCallback,cb_time,prev_si
   saved_in_cb = false
 
   # handle saveat
-  savevalues!(integrator)
+  _, saved_in_saveat = savevalues!(integrator)
 
   @inbounds if callback.save_positions[1]
     # if already saved then skip saving
-    last(integrator.sol.t) == integrator.t || savevalues!(integrator,true)
+    saved_in_saveat || savevalues!(integrator,true)
     saved_in_cb = true
   end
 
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
If we're going to do this then every `saveat` should act the same.

MSeeker1340:
For some strange reason PuMaS errors on this branch:

```
SS=2 and Overlap Tests: Error During Test at /home/mseeker/.julia/dev/PuMaS/test/runtests.jl:12
  LoadError: BoundsError: attempt to access false
    at index [2]
  Stacktrace:
   [1] indexed_iterate(::Bool, ::Int64, ::Nothing) at ./tuple.jl:69
   [2] apply_discrete_callback! at /home/mseeker/.julia/dev/OrdinaryDiffEq/src/callbacks.jl:248 [inlined]
```

The stack trace is really cryptic... I might need to consult a debugger for this.

MSeeker1340:
OK now. So the mismatch of lhs arguments results in the `BoundsError`... I'll keep this in mind ;)

I think Python did a better job at indicating errors in argument unpacking:

```python
>>> a, b, c = 1, 2
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected 3, got 2)
```

ChrisRackauckas:
Docstrings should get added here: https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/integrator_interface.jl

PR #533, opened 2018-10-14
Default `jac_prototype` in ExpRK caches

MSeeker1340:
Fixes an issue where the Jacobian is not default initialized to a dense matrix if `has_jac(f) == true` but `f.jac_prototype == nothing` and added a regression test. Similar fixes have been applied to implicit methods before but ExpRK was left out.

### DIFF
--- a/src/perform_step/exponential_rk_perform_step.jl
+++ b/src/perform_step/exponential_rk_perform_step.jl
@@ -2,9 +2,9 @@ using LinearAlgebra: axpy!
 
 # Helper function to compute the G_nj factors for the classical ExpRK methods
 @inline _compute_nl(f::SplitFunction, u, p, t, A) = f.f2(u, p, t)
-@inline _compute_nl(f::ODEFunction, u, p, t, A) = f(u, p, t) - A * u
+@inline _compute_nl(f, u, p, t, A) = f(u, p, t) - A * u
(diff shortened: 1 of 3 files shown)


MSeeker1340:
The function contructed by `@ode_def` is not an ODEFunction, which makes `_compute_nl` fail. Fixed in newest commit.

src/perform_step/exponential_rk_perform_step.jl
```diff
@@ -2,9 +2,9 @@ using LinearAlgebra: axpy!
 
 # Helper function to compute the G_nj factors for the classical ExpRK methods
 @inline _compute_nl(f::SplitFunction, u, p, t, A) = f.f2(u, p, t)
-@inline _compute_nl(f::ODEFunction, u, p, t, A) = f(u, p, t) - A * u
```
ChrisRackauckas: it should be outputing an `ODEFunction`?

ChrisRackauckas:
It's an `AbstractODEFunction`. But yes, we probably don't need a restriction here

MSeeker1340:
Rebased onto master.

MSeeker1340:
@ChrisRackauckas Test failures seem related to floating point errors and the complex tests should not be affected by this PR. OK to merge now?

Issue #511, opened 2018-09-30
Setup SSPRK method adaptivity with newer error estimate function


ChrisRackauckas:
these things:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/ssprk_perform_step.jl#L474

should be `calculate_residuals` calls:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/low_order_rk_perform_step.jl#L150

PR #534, opened 2018-10-15
Use calculate_residuals for SSPRK and Heun/Ralston

ranocha:
Cf. #511.

### DIFF
--- a/src/perform_step/ssprk_perform_step.jl
+++ b/src/perform_step/ssprk_perform_step.jl
@@ -471,7 +471,9 @@ end
 
   integrator.fsallast = f(u, p, t+dt)
   if integrator.opts.adaptive
-    integrator.EEst = integrator.opts.internalnorm(@. (utilde-u)/(integrator.opts.abstol+max(integrator.opts.internalnorm(uprev),integrator.opts.internalnorm(u))*integrator.opts.reltol))
+    utilde = utilde - u
+    atmp = calculate_residuals(utilde, uprev, u, integrator.opts.abstol, integrator.opts.reltol, integrator.opts.internalnorm)
+    integrator.EEst = integrator.opts.internalnorm(atmp)
   end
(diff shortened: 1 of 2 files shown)


src/perform_step/ssprk_perform_step.jl
```diff
@@ -514,7 +516,8 @@ end
   step_limiter!(u, f, t+dt)
 
   if integrator.opts.adaptive
-    @. atmp = (utilde-u)/(integrator.opts.abstol+max(integrator.opts.internalnorm(uprev),integrator.opts.internalnorm(u))*integrator.opts.reltol)
+    @. utilde = utilde - u
```
YingboMa: Is it possible to transform the coefficients such that we only need to perform one linear combination, but not to subtract two arrays? e.g. https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/low_order_rk_perform_step.jl#L23

ranocha: I don't see how this could be possible without using more memory to save more temporary arrays.

YingboMa: Ah, I see. Thanks for explaining!

ChrisRackauckas:
No, there are ways. For example

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/9a2823d3d88cce0438d6dcadb90eb0d21e2ea914/src/perform_step/ssprk_perform_step.jl#L462-L465

```julia
  u = u + dt_2*k
  if integrator.opts.adaptive
    utilde = (uprev + 2*u) / 3
  end
```

could instead be

```julia
  u = u + dt_2*k
  if integrator.opts.adaptive
    utilde = (uprev - u) / 3
  end
```

and now that's what used to be `utilde-u`. I wouldn't expect these transformations to have a large timing or numerical stability effect though, but it could be worth a further PR.

PR #537, opened 2018-10-16
Solvers for linear ODEs

MSeeker1340:
Things to do:

- [x] `LinearExponential` solver that internally calls `expv`/`expv_timestep`. Best choice if the linear operator is time-independent and a good first-order approximation for time-dependent ones.

- [ ] ~~Update `LinearImplicitEuler`.~~ Deleted since `ImplicitEuler` can already handle linear systems.

- [ ] ~~Determine what to do about `MidpointSplitting`: this is both a split ODE solver and a linear ODE solver; which should take precedence in terms of organization? (Also this links to DiffEqOperators with the new linear composition interface.)~~

- [ ] ~~Magnus integrator for time-dependent operators: @jagot interested in taking a shot at this?~~ We should do this as a separate project.

Related issue: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/90.

For `LinearExponential`, currently this is still fixed-dt but I'd like it to be able to just integrate to a list of `tstops` points. Not sure how to do this though; maybe define the algorithm as "adaptive" and hack at stepsize control?


ChrisRackauckas:
>For LinearExponential, currently this is still fixed-dt but I'd like it to be able to just integrate to a list of tstops points.

That will already work. Just don't check it off as `isdtchangable` false.

jagot:
I am definitely interested in having a good framework available that simplifies building propagators based on the Magnus expansion. However, I have come to believe that there is no hard-and-fast rule for how operators should be split and grouped in the expansion (what goes where), but rather that it has to be decided on a case-to-case basis. _However_, I _do_ think there should be a succinct way of expressing "this operator should be integrated over the time step, to the k:th order", or "this operator needs to be evaluated as a commutator over the time-step" (see e.g. https://arxiv.org/pdf/1801.06913.pdf for the different kinds of operators that arise from a Magnus expansion).

I should also note, with regards to the first bullet point above (unless I misunderstood something), that a linear exponential is as good as the time integral of the operator in the exponent. That is, the order of the propagator is `min(the order of the matrix exponent approximation, the order of the time integral of the matrix in the exponent)`.

jlperla:
cc: @chiyahn  track this one.  We can also contribute in whatever way we can.

ChrisRackauckas:
I think the Magnus integrator can be its own PRs. I don't think there will be a single one. IMO the easiest thing to do is just to hard code a bunch like MagnusRK4 and stuff like that, and then get fancy if we really want to flesh it out.

natschil:
Do any of these methods support mass-matrices?

ChrisRackauckas:
No. Exponential integrators cannot naturally support mass matrices (or DAEs) since they do not solve an implicit equation. Just stiff ODEs.

natschil:
Makes sense. What about the LinearImplicitEuler method?

ChrisRackauckas:
That was can probably just be deleted. It was a test for these linear-only integrators and the operator interface. With @YingboMa 's specialization on linearity in the nonlinear solvers, this algorithm has no use anymore.

natschil:
Ah makes sense.  Btw, what is the status of solving Mu' = A(t)u for sparse (linear) M and A(t) at the moment? I remember trying to do this a while ago, but back then nothing was working well in DifferentialEquations, so I ended up using a hand-built linear implicit euler. Do @YingboMa's changes solve this?

YingboMa:
I think we are really close to solving Mu' = A(t)u efficiently. We just need to handle `update_coefficients!` and mass matrix in Newton and functional iteration.

MSeeker1340:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/537/commits/d08f04cec7a337427da1982c9b215eba74b11133#diff-36711e348dff6f0a056a775b1776623dL84

The check here is preventing me from using `LinearExponential` with no `dt` set and default `tstops` (i.e. only at `tspan[1]` and `tspan[2]`), so I hotfixed it by adding a special case. This is still preventing algorithms like `Euler`, which is dt changeable, from running this way. I think there's definitely a bug here, e.g. look at the following example:

```julia
julia> prob = ODEProblem((u,p,t) -> 0.0, 1.0, (0.0,1.0));

julia> solve(prob, Euler())
ERROR: Fixed timestep methods require a choice of dt or choosing the tstops
...

julia> solve(prob, Euler(); tstops=[1.0])
retcode: Success
t: 2-element Array{Float64,1}:
 0.0
 1.0
u: 2-element Array{Float64,1}:
 1.0
 1.0
```

The first invocation should behave exactly like the second one, but instead it gets characterized as an error.

MSeeker1340:
I'm assuming the Travis failure is still due to floating point errors? Other than that @ChrisRackauckas I think this should be good to merge.

ChrisRackauckas:
>I'm assuming the Travis failure is still due to floating point errors?

Still? That implies master is failing but it's not? It is worthwhile to see why this would get an instability while master does not have one.

ChrisRackauckas:
It was a fluke.

src/solve.jl
```diff
@@ -81,7 +81,7 @@ function DiffEqBase.__init(
 
   t = tspan[1]
 
-  if (((!(typeof(alg) <: OrdinaryDiffEqAdaptiveAlgorithm) && !(typeof(alg) <: OrdinaryDiffEqCompositeAlgorithm)) || !adaptive) && dt == tType(0) && isempty(tstops)) && !(typeof(alg) <: FunctionMap)
+  if (((!(typeof(alg) <: OrdinaryDiffEqAdaptiveAlgorithm) && !(typeof(alg) <: OrdinaryDiffEqCompositeAlgorithm)) || !adaptive) && dt == tType(0) && isempty(tstops)) && !(typeof(alg) <: Union{FunctionMap,LinearExponential})
```
ChrisRackauckas: Why is this needed? `isempty(tstops)` should catch it?

ChrisRackauckas: ooooh, I see. You want it to accept no `dt` so that it just solves with `dt` equal to the timespan? That makes sense.

src/alg_utils.jl
```diff
@@ -194,8 +193,8 @@ alg_order(alg::TsitPap8) = 8
 alg_order(alg::GenericImplicitEuler) = 1
 alg_order(alg::GenericTrapezoid) = 2
 alg_order(alg::ImplicitEuler) = 1
-alg_order(alg::LinearImplicitEuler) = 1
 alg_order(alg::MidpointSplitting) = 2
+alg_order(alg::LinearExponential) = 0
```
ChrisRackauckas: I would put 1 since it's 1st order on A(t) problems.

ChrisRackauckas:
>This is still preventing algorithms like Euler, which is dt changeable, from running this way. I think there's definitely a bug here, e.g. look at the following example:

I don't think it's a bug. If someone runs Euler without a dt, I think it's good to error. Technically you can always make `dt` be the size of the time span, but for most algorithms that doesn't make practical sense and thus for those algorithms you want to tell the user that an appropriate dt is required. The exception is LinearExponential, where if they give a non time-dependent A, this algorithm has zero truncation error, and thus it can make sense to take `dt` as large as possible.

Issue #541, opened 2018-10-25
Restructure Exprb


MSeeker1340:
From slack:

> Xingjian Guo: @chrisrackauckas I double checked the Hochbruck review paper. The coefficients for Exprb32/Exprb43 are correct according to the condition in Table 2.3.
My current suspicion is that unlike the preceding ExpRK methods like ETDRK4, Exprb methods can only be used if the linear operator in each step is strictly the Jacobian of the right hand side (i.e. you cannot arbitrary set out a linear and nonlinear partition and expect the algorithm to work for such a split problem). It is not explicitly said so but the derivations in the Exprb section always uses the Jacobian, so I can only assume this is the case. In fact, the coefficients for Exprb32/Exprb43 do not satisfy the stiff order conditions of Table 2.2, which could explain the reduction in order in our tests.
To prove this I have prepared the following script, modified from our current test set:

https://gist.github.com/MSeeker1340/5a21009bf774a3e3b6049a646efd584d

> Xingjian Guo: So you can see the Exprb algorithms only achieve their theoretical order when applied to a problem in the general nonlinear formulation, unlike `ETDRK4` which can handle both.

I think there isn't anything wrong with Hochbruck's review, technically speaking, but the structure of the paper makes it easy to make the mistake of thinking they can apply to the same problems.
e.g. both methods are given in a Butcher table, but in reality one can use any general linear operator `A` while the other needs the jacobian.
I'll double check on the Exprb code once more to make sure there isn't anything wrong with our implementation. If my conjecture is correct, we need to restructure the codebase to reflect this fact (e.g. it doesn't make sense for Exprb to have support for caching methods now; it should have an interface similar to the EPIRK methods).
Ah and fyi, in the test script I used Krylov for the second suite since you can't use caching if the linear operator changes with time; and since full Krylov (in this case, `m=2`) is used there should be minimal error resulting from Krylov approximation. I've also tested the first suite with `krylov=true` and the results are the same.

> Chris Rackauckas: okay yeah, this makes sense
so we should re-structure it like EPIRK to be first order ODE only
and the defaults should be changed to make it make sense for that.
nice work. I think that with this it'll be good to add this to every stiff ODE benchmark to see how it does against the other adaptive stiff ODE solvers.

MSeeker1340:
Resolved via #543 .

PR #543, opened 2018-10-26
Restructure Exprb

MSeeker1340:
#541 

@ChrisRackauckas Is there any good reference for writing test cases for adaptive algorithms? The current one in `linear_nonlinear_krylov_test.jl` certainly is not good enough.


ChrisRackauckas:
>Is there any good reference for writing test cases for adaptive algorithms? The current one in linear_nonlinear_krylov_test.jl certainly is not good enough.

We usually just do fixed dt convergence and a regression test on the length of the solution on a test problem. The regression test just makes sure that the error estimates are sane and don't change.

MSeeker1340:
Adapted the regression tests from `ode_dense_tests.jl` here. After this is merged, I'll update the docs for adaptive Exprb.

PR #546, opened 2018-10-28
`MArray` compatibility

YingboMa:
There is
```julia
    f₀ = zero.(u0./t)
    u0 isa MArray && (f₀ = MArray(f₀))
```
because broadcasting an `MArray` results an `SArray`.
```julia
julia> ode.u0
2-element MArray{Tuple{2},Float64,1,2}:
 1.0
 2.0

julia> ode.u0 ./ 1
2-element SArray{Tuple{2},Float64,1,2}:
 1.0
 2.0
```

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -28,7 +28,7 @@ module OrdinaryDiffEq
 
   # Required by temporary fix in not in-place methods with 12+ broadcasts
   # `MVector` is used by Nordsieck forms
-  import StaticArrays: SArray, MVector, SVector, @SVector
+  import StaticArrays: SArray, MArray, MVector, SVector, @SVector
 
   # Integrator Interface
   import DiffEqBase: resize!,deleteat!,addat!,full_cache,user_cache,u_cache,du_cache,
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
looks like SArray tests fail here?

YingboMa:
`eltype(m)` can be an `MArray`, too. https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/test/static_array_tests.jl#L4

ChrisRackauckas:
`recursive_bottom_eltype(m)`.

YingboMa:
That doesn't seem to work either...
```julia
julia> u0 = fill(zero(MVector{2,Float64}), 2); t = 0.1;

julia> similar(u0)
2-element Array{MArray{Tuple{2},Float64,1,2},1}:
 #undef
 #undef

julia> similar(u0, eltype(eltype(u0)))
2-element Array{Float64,1}:
 6.9502444917253e-310
 6.9502444933561e-310

julia> similar(u0, typeof(oneunit(eltype(eltype(u0)))/oneunit(t)))
2-element Array{Float64,1}:
 6.9501798101687e-310
 6.9501798097537e-310
```

ChrisRackauckas:
This fix is really sketch and fails tests.

YingboMa:
In favor of https://github.com/JuliaArrays/StaticArrays.jl/pull/536

PR #552, opened 2018-11-02
nothing really matters?

ChrisRackauckas:


YingboMa:
Should we adapt https://github.com/JuliaLang/julia/pull/29727 in JuliaDiffEq?

src/integrators/integrator_utils.jl
```diff
@@ -466,10 +484,11 @@ function reset_fsal!(integrator)
...
   # integrator.reeval_fsal = false
+  nothing
 end
 
 function (integrator::ODEIntegrator)(t,deriv::Type=Val{0};idxs=nothing)
   current_interpolant(t,integrator,idxs,deriv)
 end
 
-(integrator::ODEIntegrator)(val::AbstractArray,t::Union{Number,AbstractArray},deriv::Type=Val{0};idxs=nothing) = current_interpolant!(val,t,integrator,idxs,deriv)
+(integrator::ODEIntegrator)(val::AbstractArray,t::Union{Number,AbstractArray},deriv::Type=Val{0};idxs=nothing) = (current_interpolant!(val,t,integrator,idxs,deriv); nothing)
```
YingboMa: This change might be breaking. Also, in Base, mutating functions generally return something, e.g.
```julia
julia> reverse!(collect(1:3))
3-element Array{Int64,1}:
 3
 2
 1

julia> push!(ones(2), 1)
3-element Array{Float64,1}:
 1.0
 1.0
 1.0
```

ChrisRackauckas:
>Should we adapt JuliaLang/julia#29727 in JuliaDiffEq?

This was discussed at triage. Making sure to return `nothing` will be the recommendation, but not necessarily `return nothing`.

src/integrators/integrator_utils.jl
ChrisRackauckas: something non-deterministically broke a split methods test. I don't think it's this though.

ChrisRackauckas: Sure, we can drop this nothing

ChrisRackauckas:
I was doing this manually, so if you have a better method then that would be helpful

ChrisRackauckas:
FWIW test timings don't improve from this, so maybe perform_step! would be more helpful.

PR #553, opened 2018-11-04
Refactor dt handling

devmotion:
Move dt handling to separate function, as suggested in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/88#discussion_r230598636.


ChrisRackauckas:
Mwtp

PR #558, opened 2018-11-09
WIP: Handle transposed state vector

YingboMa:
```julia
using OrdinaryDiffEq, LinearAlgebra
prob = ODEProblem((du, u, p, t)->du.=u, [1.]', (0, 10.))
solve(prob, Rodas5())
```
results error because of lines like https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/rosenbrock_perform_step.jl#L43 and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/caches/rosenbrock_caches.jl#L88. We need to be careful about the discrepancy between its meaning in linear algebra vs its meaning in a differential equation solver. I think we always want to use the non-adjoint version for `k`s in the Rosenbrock family and `tmp` and `dz` in the Newton-based solvers. What do you think @ChrisRackauckas?

### DIFF
--- a/src/caches/rosenbrock_caches.jl
+++ b/src/caches/rosenbrock_caches.jl
@@ -735,7 +735,7 @@ struct Rosenbrock5ConstantCache{TF,UF,Tab} <: OrdinaryDiffEqConstantCache
   tab::Tab
 end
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Does this even compute the correct results? It seems like it would do the `ldiv!` in the Jacobian part incorrectly unless the Jacobian knows to use the transpose of the matrix? It seems this is better handled on the sensitivity side than each integrator unless we can easily fix that.

PR #562, opened 2018-11-15
EEst bugfix

MSeeker1340:


ChrisRackauckas:
Oops. I messed that up a few other similar spots in that same commit. Mind adding them to this PR?

src/integrators/integrator_utils.jl
```diff
@@ -233,7 +233,7 @@ end
 function stepsize_controller!(integrator,alg::Union{StandardControllerAlgs,
                               OrdinaryDiffEqNewtonAdaptiveAlgorithm{:Standard}})
   # Standard stepsize controller
-  if EEst == zero(EEst)
+  if integrator.EEst == zero(integrator.EEst)
```
devmotion: Out of curiosity: why not just `iszero(integrator.EEst)` since then types could specialize on this and the fallback is `integrator.EEst == zero(integrator.EEst)` anyway (see https://github.com/JuliaLang/julia/blob/36e21d47b5bea100c54e4a4d1be845cbfa83a669/base/number.jl#L40)? I've seen these comparisons multiple times in JuliaDiffEq and I'm really interested in whether there's a reason to use them instead of `iszero` :smiley:

ChrisRackauckas: I didn't know about `iszero` 😃

devmotion:
Test fails because of the change that was introduced in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/commit/5e5cc30bae522e8101c17efad0c034cf8b403e6d#diff-36711e348dff6f0a056a775b1776623dR180

The problem is that the test in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/test/data_array_test.jl#L75 results in
```julia
round(Int, abs((4//1-0//1)/(1//10)), RoundUp)
```
which errors (even though it shouldn't IMO) whereas previously it lead to
```julia
round(Int, float((4//1-0//1)/(1//10)), RoundUp)
```
which succeeds. To me it seems like an upstream problem (why should this line fail?) but replacing the affected line in OrdinaryDiffEq by
```julia
               steps = ceil(Int, internalnorm((tspan[2]-tspan[1])/dt)) 
```
seems to fix the problems (at least
```julia
ceil(Int, abs(4//1-0//1)/(1//10))
```
doesn't fail).

Issue #17, opened 2016-12-31
Reduce memory footprint of RK methods


ChrisRackauckas:
There are tricks that can be done to reduce the memory footprint of the RK methods. Many times a certain `k` vector is unused after a step, and there there's no need to allocate memory for every step. Instead, some of the `k`s can actually point to the same piece of memory (but can be named differently for easy maintenance since there's no cost). 

In the current setup, this is as simple as identifying the vectors for which this can be done and doing something like `k13 = k3`, and getting rid of the line where `k13` is first allocated. Care needs to be taken to note that the folded vectors aren't used in interpolations. Hairer's `dopri5` and `dop853` implementations have some examples of where this can be done (though it's hard to know it's there unless you already know about it!)

ChrisRackauckas:
The commit of DP5 shows how this can be done.

PR #570, opened 2018-12-04
Reduces cache in Vern6 Algorithm

kanav99:
This PR is a part of issue #17 . Reduces the number of `k` variables by 2.

Also, `a1102`, `a1103`, `a1202`, `a1203` are all zero. I tried removing them from the cache but that lead to failing of tests.


devmotion:
If these coefficients are actually zero you should be able to remove them from the cache. I guess you just have to remove any occurrences of them in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/dense/verner_addsteps.jl as well.

kanav99:
Oh I missed that one. I will remove them and run tests if that works out.

kanav99:
Removed all the references of a1102, a1103, a1202, a1203. Test passed successfully.

PR #572, opened 2018-12-06
Reduces cache in Vern9 Algorithm

kanav99:
This PR is part of the issue #17 . Currently this only removes the obvious `k`s which could reuse the previous `k`'s allocated memory.


src/perform_step/verner_rk_perform_step.jl
```diff
@@ -675,43 +675,42 @@ end
...
-  integrator.k[7]=k7; integrator.k[8]=k8;
-  integrator.k[9]=k9; integrator.k[10]=k10;
-  integrator.k[11]=k11; integrator.k[12]=k12;
-  integrator.k[13]=k13; integrator.k[14]=k14;
-  integrator.k[15]=k15; integrator.k[16]=k16
+  integrator.k[1]=k1; integrator.k[2]=k8;
+  integrator.k[3]=k9; integrator.k[4]=k10;
+  integrator.k[5]=k11; integrator.k[6]=k12;
+  integrator.k[7]=k13; integrator.k[8]=k14;
+  integrator.k[9]=k15; integrator.k[10]=k16
```
devmotion: Maybe add the same comment as below here as well?

ChrisRackauckas: Yes, it's probably a good idea.

kanav99: Okay sure I will make the edit.

src/dense/verner_addsteps.jl
```diff
@@ -632,28 +632,30 @@ end
...
-    copyat_or_push!(k,14,f(uprev+dt*(a1401*k[1]                                  +a1406*k[6]+a1407*k[7]+a1408*k[8]+a1409*k[9]+a1410*k[10]+a1411*k[11]+a1412*k[12]+a1413*k[13]),p,t+c13*dt))
-    copyat_or_push!(k,15,f(uprev+dt*(a1501*k[1]                                  +a1506*k[6]+a1507*k[7]+a1508*k[8]+a1509*k[9]+a1510*k[10]+a1511*k[11]+a1512*k[12]+a1513*k[13]+a1514*k[14]),p,t+dt))
-    copyat_or_push!(k,16,f(uprev+dt*(a1601*k[1]                                  +a1606*k[6]+a1607*k[7]+a1608*k[8]+a1609*k[9]+a1610*k[10]+a1611*k[11]+a1612*k[12]+a1613*k[13]),p,t+dt))
+    copyat_or_push!(k,2,f(uprev+dt*(a0801*k[1]                                  +a0806*k[6]+a0807*k[7]),p,t+c7*dt))
+    copyat_or_push!(k,3,f(uprev+dt*(a0901*k[1]                                  +a0906*k[6]+a0907*k[7]+a0908*k[2]),p,t+c8*dt))
+    copyat_or_push!(k,4,f(uprev+dt*(a1001*k[1]                                  +a1006*k[6]+a1007*k[7]+a1008*k[2]+a1009*k[3]),p,t+c9*dt))
+    copyat_or_push!(k,5,f(uprev+dt*(a1101*k[1]                                  +a1106*k[6]+a1107*k[7]+a1108*k[2]+a1109*k[3]+a1110*k[4]),p,t+c10*dt))
+    temp6 = recursivecopy(k[6])
+    temp7 = recursivecopy(k[7])
```
devmotion: Just out of curiosity: what's the advantage of copying `k[6]` and `k[7]` compared to just adding two additional elements to `k` (such that it's `copyat_or_push!(k,8,...)` up to `copyat_or_push!(k,12,..)` in the next lines)?

ChrisRackauckas: Adding it to the `k` vector would increase the size of the saved vector.

ChrisRackauckas:
Thanks!

PR #574, opened 2018-12-08
Fix performance regression

YingboMa:
Fixes https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/571

```julia
julia> @btime solve($probs,$(Tsit5())); # master
  2.981 ms (60568 allocations: 4.10 MiB)

julia> @btime solve($probs,$(Tsit5())); # PR
  487.315 μs (2664 allocations: 473.02 KiB)
```

### DIFF
--- a/src/misc_utils.jl
+++ b/src/misc_utils.jl
@@ -39,6 +39,12 @@ end
 get_chunksize(x) = 0
 get_chunksize(x::NLSOLVEJL_SETUP{CS,AD}) where {CS,AD} = CS
 
+@inline @muladd calculate_residuals(ũ::Number, u₀::Number, u₁::Number,
+                                    α, ρ, internalnorm) = ũ / (α + max(internalnorm(u₀), internalnorm(u₁)) * ρ)
+
+@inline @muladd calculate_residuals(u₀::Number, u₁::Number,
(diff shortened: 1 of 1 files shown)


devmotion:
I guess the implementation of `calculate_residuals(u₀, u₁, α, ρ, internalnorm)` in lines 114 to 116 is affected by the same problem, isn't it?

dkarrasch:
What types of `internalnorm` can be used? I thought one could, for instance, define an `internalnorm` that computes the norm based on only few of the components. That couldn't be called on numbers then?

devmotion:
At the moment you have to define a scalar norm as well but that will be changed in the next DiffEq release. There's a discussion about this issue in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/569.

YingboMa:
> By the way, did you test whether the different implementations are still required?

Yes, it is still needed.

ChrisRackauckas:
Sweet, I find this to be cleaner as well.

dkarrasch:
Can we please get a version tag for this fix? Or are there more things currently in the pipeline?

ChrisRackauckas:
Release made

PR #578, opened 2018-12-22
Fix SymplecticEuler

### LINKED ISSUE #577: SymplecticEuler implementation is wrong
While working on some numerical comparisons, we noticed that the SymplecticEuler seems to be in fact a simple Euler method. You can easily see that it is not symplectic applying it to a damped harmonic oscillator. To give credit where credit is due, the first to spot the issue was @mvermeeren.

Also the Velocity Verlet seems to be initialised slightly off-energy, but that’s not a big issue.

YingboMa:
Fixes #577

Master:
seuler_before

PR:
seuler_after

### DIFF
--- a/src/perform_step/symplectic_perform_step.jl
+++ b/src/perform_step/symplectic_perform_step.jl
@@ -22,7 +22,7 @@ end
   u = uprev + dt*kuprev
   # Now actually compute the step
   # Do it at the end for interpolations!
-  kdu = f.f1(duprev,uprev,p,t)
+  kdu = f.f1(duprev,u,p,t)
   du = duprev + dt*kdu
   ku = f.f2(du,uprev,p,t)
 
@@ -62,7 +62,7 @@ end
   @. u = uprev + dt*kuprev
   # Now actually compute the step
   # Do it at the end for interpolations!
-  f.f1(kdu,duprev,uprev,p,t)
+  f.f1(kdu,duprev,u,p,t)
   @. du = duprev + dt*kdu
   f.f2(ku,du,uprev,p,t)
 end


YingboMa:
```julia
using OrdinaryDiffEq, Plots; gr()

f2(v,u,p,t) = (v-2)u
f1(v,u,p,t) = (1-u)v
prob = DynamicalODEProblem{false}(f1,f2,2.,4.,(0., 20.))
sol = solve(prob, SymplecticEuler(), dt=0.05)
f2(dv,v,u,p,t) = @. dv=(v-2)u
f1(du,v,u,p,t) = @. du=(1-u)v
prob1 = DynamicalODEProblem{true}(f1,f2,[2.],[4.],(0., 20.))
sol1 = solve(prob1, SymplecticEuler(), dt=0.05)
plt1=plot(sol, vars=(2,1), aspect_ratio=1, title="Lotka-Volterra out-of-place")
plt2=plot(sol1, vars=(2,1), aspect_ratio=1, title="Lotka-Volterra in-place")
plot(plt1, plt2, dpi=150)
```

devmotion:
IMO the implementation of `SymplecticEuler` is still not completely correct since the plots show a significant difference between the in-place and out-of-place computations.

Moreover, it seems the Lotka-Volterra model and the damped harmonic oscillator in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/577 do not satisfy the assumptions listed in http://docs.juliadiffeq.org/latest/solvers/dynamical_solve.html (BTW there's a typo in the documentation, f2 should be independent of t and u (and not v)):
> The functions should be specified as f1(dv,v,u,p,t) and f2(du,v,u,p,t) (in the inplace form), where f1 is independent of v (unless specified by the solver), and f2 is independent of t and v.

For the given examples one of the integration steps (of u or v) of the symplectic Euler method is an implicit step (see e.g. https://www.unige.ch/~hairer/poly_geoint/week2.pdf), which is not supported by the explicit implementation of `SymplecticEuler`.

Moreover, I noticed a redundant line https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/symplectic_perform_step.jl#L11 (I still don't fully understand the current implementation, I'm a bit confused by the updating scheme and the pre-computations in `initialize!`).

ChrisRackauckas:
>For the given examples one of the integration steps (of u or v) of the symplectic Euler method is an implicit step (see e.g. https://www.unige.ch/~hairer/poly_geoint/week2.pdf), which is not supported by the explicit implementation of SymplecticEuler.

That page from Hairer has a typo in it. It should be like https://en.wikipedia.org/wiki/Semi-implicit_Euler_method . The semi-implicitness is that it would be implicit if a single ODE, but using the partitions it becomes an explicit method. 

>and it seems fsallast is initialized differently for the in-place and out-of-place method

The out of place method seems wrong. @YingboMa take a look at it?

>However, I still don't fully understand the current implementation, I'm a bit confused by the updating scheme and the pre-computations in initialize!.

I agree it's weird. It's because I wanted to have the first derivatives computed, so half of the computation is done before and then each step updates the second half and then the first half. This is so that we have derivatives on both ends for the interpolation, the same thing we do for the other Hermite methods, but with the partitioning scheme is gets quite a bit more complicated. I think I messed up here.

PR #580, opened 2019-01-01
Implements Anderson Acceleration for functional Iteration

kanav99:
Based on the paper "Anderson Acceleration for Fixed-point Iterations", Walker and Ni (https://users.wpi.edu/~walker/Papers/Walker-Ni,SINUM,V49,1715-1735.pdf), this PR serves to implement for solving non-linear equations.

Got the idea and help from the implementation in JuliaNLSolvers/NLsolve.jl

### DIFF
--- a/src/nlsolve/anderson.jl
+++ b/src/nlsolve/anderson.jl
@@ -13,6 +13,7 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
   zs = zeros(S.n+1)
   gs = zeros(S.n+1)
   residuals = zeros(S.n+1)
+  alphas = zeros(S.n)
   # initial step of NLAnderson iteration
   zs[1] = z
   iter = 1
@@ -22,8 +23,12 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
   gs[1] = z₊
   dz = z₊ - z
   ndz = integrator.opts.internalnorm(dz)
-  xs = circshift(xs, 1)
-  gs = circshift(gs, 1)
+  for t = (S.n+1):-1:2
+    zs[t] = zs[t-1]
+    gs[t] = gs[t-1]
+  end
+  # zs = circshift(zs, 1)
+  # gs = circshift(gs, 1)
   z = z₊
   zs[1] = z
   η = nlcache.ηold
@@ -39,16 +44,21 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
     
     mk = min(S.n, iter-1)
     residuals[1:mk] = (gs[2:mk+1] .- zs[2:mk+1]) .- (gs[1] - zs[1])
-    alphas[1:mk] .= residuals[1:mk] \ (zs[1] - gs[1])
+    alphas[1:mk] .= residuals[1:mk] \ [(zs[1] - gs[1])]
     for i = 1:mk
         z₊ += alphas[i]*(gs[i+1] - gs[1])
     end
-    xs = circshift(xs, 1)
-    gs = circshift(gs, 1)
+    for t = (S.n+1):-1:2
+      zs[t] = zs[t-1]
+      gs[t] = gs[t-1]
+    end
+    # zs = circshift(zs, 1)
+    # gs = circshift(gs, 1)
     zs[1] = z₊
     ndzprev = ndz
     dz = z₊ - z
     ndz = integrator.opts.internalnorm(dz)
+    θ = ndz/ndzprev
     if θ > 1 || ndz*(θ^(max_iter - iter)/(1-θ)) > κtol
       fail_convergence = true
       break
--- a/src/nlsolve/anderson.jl
+++ b/src/nlsolve/anderson.jl
@@ -1,0 +1,60 @@
+function (S::NLAnderson{false,<:NLSolverCache})(integrator)
+  nlcache = S.cache
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack z,tmp,W,κ,tol,c,γ,max_iter,min_iter = nlcache
+  if typeof(integrator.f) <: SplitFunction
+    f = integrator.f.f1
+  else
+    f = integrator.f
+  end
+  # precalculations
+  κtol = κ*tol
+
+  zs = zeros(S.n+1)
+  gs = zeros(S.n+1)
+  residuals = zeros(S.n+1)
+  # initial step of NLAnderson iteration
+  zs[1] = z
+  iter = 1
+  tstep = t + c*dt
+  u = tmp + γ*z
+  z₊ = dt*f(u, p, tstep)
+  gs[1] = z₊
+  dz = z₊ - z
+  ndz = integrator.opts.internalnorm(dz)
+  xs = circshift(xs, 1)
+  gs = circshift(gs, 1)
+  z = z₊
+  zs[1] = z
+  η = nlcache.ηold
+  do_anderson = true
+
+  # anderson acceleration for fixed point iteration
+  fail_convergence = false
+  while (do_anderson || iter < min_iter) && iter < max_iter
+    iter += 1
+    u = tmp + γ*z
+    z₊ = dt*f(u, p, tstep)
+    gs[1] = z₊
+    
+    mk = min(S.n, iter)
+    residuals[1:mk] = (gs[2:mk+1] .- zs[2:mk+1]) .- (gs[1] - zs[1])
+    alphas[1:mk] .= residuals[1:mk] \ (zs[1] - gs[1])
+    for i = 1:mk
+        z₊ += alphas[i]*(gs[i+1] - gs[1])
+    end
+    xs = circshift(xs, 1)
+    gs = circshift(gs, 1)
+    zs[1] = z₊
+    ndzprev = ndz
+    dz = z₊ - z
+    ndz = integrator.opts.internalnorm(dz)
+    if θ > 1 || ndz*(θ^(max_iter - iter)/(1-θ)) > κtol
+      fail_convergence = true
+      break
+    end
+    η = θ/(1-θ)
+    do_functional = (η*ndz > κtol)
+    z = z₊
+  end
+end
--- a/src/nlsolve/anderson.jl
+++ b/src/nlsolve/anderson.jl
@@ -75,3 +75,91 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
 
   return (z, η, iter, false)
 end
+
+function (S::NLAnderson{true})(integrator)
+  nlcache = S.cache
+  @unpack t,dt,uprev,u,f,p = integrator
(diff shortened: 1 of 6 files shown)


src/nlsolve/anderson.jl
```diff
@@ -0,0 +1,67 @@
...
+  if typeof(integrator.f) <: SplitFunction
+    f = integrator.f.f1
+  else
+    f = integrator.f
+  end
+  # precalculations
+  κtol = κ*tol
+
+  zs = zeros(S.n+1)
+  gs = zeros(S.n+1)
```
devmotion: Shouldn't these definitions of `zs` and `gs` cause problems if `f` returns general number types and/or vectors? I realized that the current implementation in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/nlsolve/functional.jl#L25 only considers the scalar case - is there a reason for not considering vector valued derivatives? I guess, the implementation there should be fixed (i.e. generalized) as well.

src/nlsolve/anderson.jl
```diff
@@ -0,0 +1,67 @@
...
+  residuals = zeros(S.n+1)
+  # initial step of NLAnderson iteration
+  zs[1] = z
+  iter = 1
+  tstep = t + c*dt
+  u = tmp + γ*z
```
devmotion: Needs dots, if we want to deal with vector valued derivatives.

src/nlsolve/anderson.jl
```diff
@@ -0,0 +1,67 @@
...
+  z₊ = dt*f(u, p, tstep)
+  gs[1] = z₊
+  dz = z₊ - z
```
devmotion: Same here

src/nlsolve/anderson.jl
devmotion: Basically the generalization should only require adding dots to all possibly vector valued computations.

devmotion:
Nice! As a general note, you can reduce allocations by using views, as done in the implementation in [NLsolve](https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/solvers/anderson.jl#L36).

src/nlsolve/anderson.jl
kanav99: @devmotion The dots are there in the NLFunctional{true} implementation, (see here: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/nlsolve/newton.jl#L87 ). I have currently done the same for the sake of uniformity. Implementation for `iip=true` is to be done in the coming commits, but we for sure can work to generalize it.

kanav99:
Thanks for the suggestion, I am trying to follow NLSolve as much as possible. I was unfamiliar with views, so I will make the change in next update!

src/nlsolve/anderson.jl
YingboMa: All out-of-place solvers don't have broadcasts, so I also didn't add dots in functional iteration.

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/fixed_timestep_perform_step.jl#L181-L182

src/nlsolve/anderson.jl
```diff
@@ -0,0 +1,67 @@
...
+  ndz = integrator.opts.internalnorm(dz)
+  xs = circshift(xs, 1)
+  gs = circshift(gs, 1)
```
YingboMa: Two-argument `circshift` allocates.

YingboMa:
Thank you! Looks like a good start. Could you include the file and test it? E.g. https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/b90e3d9a5eb4fba3a9de66544dd0894489c5965f/test/ode/ode_convergence_tests.jl#L66

ChrisRackauckas:
Worth mentioning: https://github.com/JuliaNLSolvers/NLsolve.jl/issues/128

antoine-levitt:
Just out of curiosity, why is this not calling nlsolve?

YingboMa:
NLsolve has high allocation.

antoine-levitt:
Which parts specifically? Can anything be done about that?

YingboMa:
In-place non-linear solving in OrdinaryDiffEq should have zero allocation because we preallocate all the cache arrays. Maybe NLsolve can a functionality to enable the user to pass in a cache struct.

Note that Newton iteration in OrdinaryDiffEq takes the advantage of the special form of the target function, hence it is unlikely for NLsolve to replace that.

ChrisRackauckas:
NLsolve could in theory be used, in practice it's not there yet. It needs zero allocations, the ability to re-use the cache arrays in other parts of diffeq, compatibility with all of the number and array types we support (units and MultiScaleArrays in particular are the hard ones). It needs a quasi-Newton method, functional iteration, and AA. It needs these special convergence checks in there. NLsolve didn't have a lot of that in the first iterations, and it still doesn't have a lot of it, so it doesn't make sense to only have AA be NLsolve but more limited than the others. When NLsolve gets everything we need (and I am pushing for it to be in that direction), then it could make sense to use it. But for now, the easiest thing to do is a quick AA routine to finish up the trio and call it day.

src/nlsolve/anderson.jl
ChrisRackauckas: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/582 merged so this PR should take it into consideration.

pkofod:
I'm doing an efficient AA implementation myself these days because I need it for some work myself, and Antoine is also interested in a better implementation, so it's interesting to see the discussion here!

> In-place non-linear solving in OrdinaryDiffEq should have zero allocation because we preallocate all the cache arrays. Maybe NLsolve can add a functionality that enables the user to pass in a cache struct.

you can?

> Note that Newton iteration in OrdinaryDiffEq takes the advantage of the special form of the target function, hence it is unlikely for NLsolve to replace that.

How special is the objective? Do the "taking advantage of" lend itself to quasi-Newton methods more generally, or is it strictly for Newton's method?


> . It needs zero allocations, the ability to re-use the cache arrays in other parts of diffeq

This *should* be possible. If not, let's find those allocations!

>  It needs a quasi-Newton method, functional iteration, and AA.

What is functional iteration? Fixed point iteration? Use AA with memory = 1. If some things are handled inefficiently and should be special-cased for m == 1 I'm fine with that.

> compatibility with all of the number and array types we support (units and MultiScaleArrays in particular are the hard ones).

This is potentially a hard one. Happy to help if there are specific things I can test out.


> It needs these special convergence checks in there.

This is something NLsolve does not do well enough, and neither does Optim. I want to extend the callback-interface to handle this.


>  NLsolve didn't have a lot of that in the first iterations, and it still doesn't have a lot of it, so it doesn't make sense to only have AA be NLsolve but more limited than the others. When NLsolve gets everything we need (and I am pushing for it to be in that direction), then it could make sense to use it. But for now, the easiest thing to do is a quick AA routine to finish up the trio and call it day.

I think you're 100% right. I think NLsolve needs quite some love, and I've explained several times in several places that I cannot commit to doing it right now, though I'm following discussions still. I'm going to graduate some time after the summer, and I'm actively pursuing possibilities of funding extended full-time development on NLsolve and it's siblings.

YingboMa:
> you can?

> This should be possible. If not, let's find those allocations!

Even with `AndersonCache`, `anderson_` still allocates. E.g:
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/solvers/anderson.jl#L54
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/solvers/anderson.jl#L84
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/solvers/anderson.jl#L91
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/solvers/anderson.jl#L97-L98
https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/solvers/anderson.jl#L116

> How special is the objective? Do the "taking advantage of" lend itself to quasi-Newton methods more generally, or is it strictly for Newton's method?

Yes, it is quasi-Newton, and the details are described in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/e69b2187a8431cd34fea036137253eb8b295f316/src/nlsolve/newton.jl#L1-L33.

src/nlsolve/anderson.jl
```diff
@@ -0,0 +1,173 @@
...
+  else
+    @. z₊ = dt*k + z
+    mul!(ztmp, mass_matrix, z)
+    @. z₊ -= ztmp
+  end
+  gs[:,1] .= vec(z₊)
+  @. dz = z₊ - z
+  ndz = integrator.opts.internalnorm(dz)
+  for t = (S.n+1):-1:2
+    @. zs[:,t] = zs[:,t-1]
```
ChrisRackauckas: should use views on the rhs here.

src/nlsolve/anderson.jl
```diff
@@ -0,0 +1,173 @@
...
+      @. z₊ = dt*k
+    else
+      @. z₊ = dt*k + z
+      mul!(ztmp, mass_matrix, z)
+      @. z₊ -= ztmp
+    end
+    gs[:,1] .= vec(z₊)
+
+    mk = min(S.n, iter-1)
+    residuals[:,1:mk] .= (gs[:,2:mk+1] .- zs[:,2:mk+1]) .- (gs[:,1] - zs[:,1])
```
ChrisRackauckas: views

src/nlsolve/type.jl
```diff
@@ -26,6 +26,8 @@ end
 struct NLAnderson{iip,T<:NLSolverCache} <: AbstractNLsolveSolver
   cache::T
   n::Int
+  NLAnderson{iip,T}(nlcache::T) where {iip, T<:NLSolverCache} = new(nlcache, 5)
+  NLAnderson{iip,T}(nlcache::T, n) where {iip, T<:NLSolverCache} = new(nlcache, n)
```
ChrisRackauckas: Just one method: `NLAnderson{iip,T}(nlcache::T, n=5) where {iip, T<:NLSolverCache} = new(nlcache, n)`

ChrisRackauckas:
If you use an array of arrays representation, then only `residuals` needs to be a concrete matrix? Then a lot of the swapping could be pointer swapping instead of copying values around. That might clean up the code, make it more versatile, and faster.

kanav99:
I have made the asked changes. It's still incomplete, but I believe its in a stable condition right now. Mass matrix tests pass partially, (one problem set fails while other passes). Convergence tests pass. And most probably the logic is correct (I tested solving same problems on Functional and Anderson and both gave nearly same solution).

I am listing down the current limitations of the code:
1. No broadcast and Mass matrices in out-of place implementation `NLAnderson{false}`
2. Mass matrix test fails
3. Change to array of arrays format required

I believe this PR is in a state that can be merged.

ChrisRackauckas:
Looks great, thanks!

PR #581, opened 2019-01-01
Added low storage SSPRK53

dgan181:
I have added a low storage five stages, third order explicit RK method as describe in [Higueras and T. Roldán (2018)]( http://arxiv.org/abs/1809.04807v1). I would like to know if these additions are acceptable.


ChrisRackauckas:
Looks great. Thanks!

PR #583, opened 2019-01-02
Add broadcasts to Heun, Ralston, Euler, FunctionMap, and Midpoint

devmotion:
Add broadcasts to some of the fixed time step algorithms.


ChrisRackauckas:
Is this not a major performance hit, especially with static arrays?

PR #584, opened 2019-01-02
WIP : Implementation of SSPRK53_2N2

dgan181:
@ranocha I tried to implement the second method in [Higueras and Roldan](https://arxiv.org/abs/1809.04807) but my code fails the non-linear ode test. I can't seem to figure out where I am going wrong. I'd really appreciate some help.


ChrisRackauckas:
What does the plot of the convergence test look like? Just do `plot(sim)`. It should be in the region where the test is loglinear (it automatically log scales). Use that to shift the dts. It will bottom out if at floating point error, but that shouldn't happen with this low of an order. It won't be linear if the dts are too large, which could be happening here.

dgan181:
This is the plot obtained without changing `dts`, i.e. `dts = 1. //2 .^(8:-1:4)`
firt

This is when every value in `dts` is halved, i.e. `dts = 1. //2 .^(9:-1:5)`
second

For the latter choice of `dts`, the difference falls within the tolerance. Should I proceed with making the change?

ChrisRackauckas:
>For the latter choice of dts, the difference falls within the tolerance. Should I proceed with making the change?

Yes, this shows that it wasn't in the asymtopic range before.

dgan181:
@ChrisRackauckas Is what I've done accetable? Should I change something?

ChrisRackauckas:
Looks great. Sorry for disappearing a few days. I'm moving across the US very soon so it's frantic time right now. Just ping me when you need me.

Issue #592, opened 2019-01-09
binary_minheap is deprecated


baggepinnen:
The function `binary_minheap` is deprecated in favor of a constructor. It is used twice in this package, see here https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/search?q=binary_minheap&unscoped_q=binary_minheap

```julia
└ @ OrdinaryDiffEq C:\Users\Helga\.julia\packages\OrdinaryDiffEq\7mKtS\src\solve.jl:415
┌ Warning: `binary_minheap(xs::AbstractVector{T}) where T` is deprecated, use `BinaryMinHeap(xs)` instead.
│   caller = tstop_saveat_disc_handling at solve.jl:389 [inlined]
└ @ Core C:\Users\Helga\.julia\packages\OrdinaryDiffEq\7mKtS\src\solve.jl:389
[...]

PR #593, opened 2019-01-09
Update deprecated functions binary_min/maxheap

baggepinnen:
Solves #592


oxinabox:
Probably want to add bounds of `>=0.15` for DataStructures.jl in the Require/project.toml

Issue #493, opened 2018-08-29
More Runge-Kutta Methods for (HYperbolic) PDEs


ranocha:
Here are some additional articles describing/investigating/developing Runge-Kutta methods, especially for hyperbolic PDEs:
[ ] [Kubatko, Yeager, Ketcheson (2014)](https://link.springer.com/article/10.1007%2Fs10915-013-9796-7)
[ ] [Parsani, Ketcheson, Deconinck (2013)](https://epubs.siam.org/doi/10.1137/120885899)
[ ] [Toulorge, Desmet (2012)](https://www.sciencedirect.com/science/article/pii/S0021999111006796)
[ ] [Kennedy, Carpenter, Lewis (2000)](https://www.sciencedirect.com/science/article/pii/S0168927499001415)
[ ] [Mead, Renaut (1999)](https://www.sciencedirect.com/science/article/pii/S0021999199962605)

These methods do not necessarily have high-priority but may be implemented.

ChrisRackauckas:
Thanks! Since we have good infrustructure for Runge-Kutta methods I think it's best to leave this as an introductory issue (this is a good "first algorithm" for potential GSoC students!)

ChrisRackauckas:
There's this one as well: https://arxiv.org/abs/1806.08693

ranocha:
> There's this one as well: https://arxiv.org/abs/1806.08693

I've added it to the list.

ChrisRackauckas:
And a new one: https://arxiv.org/abs/1809.04807

ranocha:
> And a new one: https://arxiv.org/abs/1809.04807

Thanks.  I've added that as well to the list.

dgan181:
I have tried to implement the first method in the list given above in #581. I'd like to know if it is alright.

ranocha:
Thanks, @dgan181. You implemented the method `SSPRK53_2N1` of Higueras and Roldan as `SSPRK53_2N`, didn't you? It may be good to add also the other scheme `SSPRK53_2N2` (eq. (60) of Higueras and Roldan).

It would be good to add the new citation for `SSPRK53_2N` (or `SSPRK53_2N1` and `SSPRK53_2N2` if both are added) to https://github.com/JuliaDiffEq/juliadiffeq.github.io/blob/master/citing.md.

dgan181:
@ranocha I tried implementing the second method here #584 . Once that is resolved, I will add the new citations.

[...]

PR #596, opened 2019-01-13
WIP: LDDRK 6-stage 4th order

saurabhkgp21:
I tried implementing 6 stage, 4th order Low Dissipation and Dispersion RK(LDDRK) method as described in  [Stanescu, Habashi (1998)](https://www.sciencedirect.com/science/article/pii/S0021999198959861). But my code gives `DtLessThanMin` instead of `Success` as retcode. I am not able to figure out where I'm going wrong. 
As this is my first contribution in this package, I am certain that there will be other mistakes also in my PR. I'd really appreciate some help.
Ref :- #493


src/algorithms.jl
```diff
@@ -43,6 +43,7 @@ Anas5(; w=1) = Anas5(w)
 struct OwrenZen3 <: OrdinaryDiffEqAdaptiveAlgorithm end
 struct OwrenZen4 <: OrdinaryDiffEqAdaptiveAlgorithm end
 struct OwrenZen5 <: OrdinaryDiffEqAdaptiveAlgorithm end
+struct LDDRK64 <: OrdinaryDiffEqAdaptiveAlgorithm end
```
ChrisRackauckas: You accidentally marked it as an adaptive algorithm, which then sets the trait https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/cc3dc9d1f78215c3b125a6fe84b4fc461f7121d3/src/alg_utils.jl#L45 so that way the integrator by default uses adaptive time stepping. Because your method isn't actually adaptive, it thinks every step has infinite error and drops dt to zero, giving your error. To fix this, use `<: OrdinaryDiffEqAlgorithm`

saurabhkgp21: Done. But algorithms is still not passing convergence test. Probably there is some mistake in constants or `perform_step!`. I will again take a look by tonight.

saurabhkgp21:
@ChrisRackauckas I tried debugging this method by cross checking intemediate values of `u` with manually calculated values, and error between two values were always less than `1e-10` till I calculated. Although error at final point is relatively too high in the range of `1e-6`. I also followed same format to implement RK46-NL described by [Berland, Bogey, Bailly (2006)](https://www.sciencedirect.com/science/article/pii/S0045793005000575) and order is approximately equal to 4 after shifting dts. 
I surely misunderstood something and had tried these two methods in similar way. I'd be really happy to learn something new which can be followed to debug something like this.

ChrisRackauckas:
There really isn't a silver bullet to debugging this kind of code, though throwing some math knowledge at it usually works. For example, from the RK coefficient derivations, the rows of the tableau coefficients should add up to the c values. The betas should add up to 1. Etc. Double check the numbers and these relations.

YingboMa:
See https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/599#issuecomment-455361276

src/caches/low_order_rk_caches.jl
```diff
@@ -193,6 +193,66 @@ end
 
 alg_cache(alg::BS3,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}}) = BS3ConstantCache(real(uBottomEltypeNoUnits),real(tTypeNoUnits))
 
+@cache struct LDDRK64Cache{uType,rateType,TabType} <: OrdinaryDiffEqMutableCache
```
ChrisRackauckas: move this to the file for SSPRK methods

src/perform_step/fixed_timestep_perform_step.jl
```diff
@@ -323,6 +323,85 @@ end
   end
 end
 
+function initialize!(integrator,cache::LDDRK64ConstantCache)
```
ChrisRackauckas: move this to the file for SSPRK methods

test/ode/ode_convergence_tests.jl
```diff
@@ -17,6 +17,8 @@ for i = 1:2
   prob = probArr[i]
   sim = test_convergence(dts,prob,Euler())
   @test sim.𝒪est[:final] ≈ 1 atol=testTol
+  sim1 = test_convergence(dts1,prob,LDDRK64())
```
ChrisRackauckas: set it up with the ode_ssprk_tests.jl file

saurabhkgp21:
Should we create separate files for LDD methods similar to SSP methods?

ChrisRackauckas:
LDD methods are SSP. I don't think there's enough to justify splitting it. Maybe.

saurabhkgp21:
Papers which I have read for SSP methods weren't specifying LDD properties, that's why I wasn't sure.
Anyway, If all SSP methods follow LDD properties, there isn't any reason to split.

PR #598, opened 2019-01-16
RK46-NL Implementation

saurabhkgp21:
Implementation of 4th order 6 stage RK method (RK46-NL) as described in [Berland, Bogey, Bailly (2006)](https://www.sciencedirect.com/science/article/pii/S0045793005000575).

Ref :- #493


saurabhkgp21:
I hadn't checked the discussion on the issue yesterday regarding this method. It will be alright to close this PR if @arnav-t wants to open another one.

ChrisRackauckas:
It's fine. Since this is done @arnav-t can pick another one out of the list. No reason to discard working code. 

It looks like you're quite situated with what you're doing here, so if you want to chat now about what other methods to start looking at please ping me in the Slack channel.

PR #599, opened 2019-01-17
WIP: Implemented LDDRK25 from Stanescu, Habashi (1998)

arnav-t:
Implemented second order, five-stage scheme LDDRK25 from [Stanescu, Habashi (1998)].(https://www.sciencedirect.com/science/article/pii/S0021999198959861). 
I haven't added the tests in yet because I'm not sure which file I should put them in.

### DIFF
--- a/src/caches/low_order_rk_caches.jl
+++ b/src/caches/low_order_rk_caches.jl
@@ -534,3 +534,71 @@ function alg_cache(alg::Anas5,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
 end
 
 alg_cache(alg::Anas5,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}}) = Anas5ConstantCache(real(uBottomEltypeNoUnits),real(tTypeNoUnits))
+
+@cache struct LDDRK25Cache{uType,rateType,TabType} <: OrdinaryDiffEqMutableCache
+
+	u::uType
+	uprev::uType
+	k::rateType
+	tmp::uType
+	fsalfirst::rateType
+	tab::TabType
+
+end
+
+struct LDDRK25ConstantCache{T,T2} <: OrdinaryDiffEqConstantCache
+
+	α2::T
+	α3::T
+	α4::T
+	α5::T
+
+	β1::T
+	β2::T
+	β3::T
+	β4::T
+	β5::T
+
+	c2::T2
+	c3::T2
+	c4::T2
+	c5::T2
+
+	function LDDRK25ConstantCache(::Type{T}, ::Type{T2}) where {T,T2}
+
+		α2 = T(-0.6913065)
+		α3 = T(-2.655155)
+		α4 = T(-0.8147688)
+		α5 = T(-0.6686587)
+
+		β1 = T(0.1)
+		β2 = T(0.75)
+		β3 = T(0.7)
+		β4 = T(0.479313)
+		β5 = T(0.310392)
+
+		c2 = T2(0.1)
+		c3 = T2(0.3315201)
+		c4 = T2(0.4577796)
+		c5 = T2(0.8666528)
+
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
tests should go in the tests/convergence_tests.jl file. You'll see the other SSPRK methods, so just put it there. Make sure it converges to the right order and do `plot(sim)` to check the convergence plot and make sure it looks like a nice straight line in log-space.

arnav-t:
@ChrisRackauckas The SSPRK tests seem to be in a different file (`ode_ssprk_tests.jl`).

ChrisRackauckas:
Yes indeed. My bad. Please add it to that one.

ChrisRackauckas:
The SSPRK methods are in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/ssprk_perform_step.jl, so this should be moved over to there.

src/caches/low_order_rk_caches.jl
```diff
@@ -534,3 +534,71 @@ function alg_cache(alg::Anas5,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
 end
 
 alg_cache(alg::Anas5,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}}) = Anas5ConstantCache(real(uBottomEltypeNoUnits),real(tTypeNoUnits))
+
+@cache struct LDDRK25Cache{uType,rateType,TabType} <: OrdinaryDiffEqMutableCache
+
+	u::uType
+	uprev::uType
+	k::rateType
+	ω::uType
```
ChrisRackauckas: keep this as `tmp` for consistency with the other implementations.

ChrisRackauckas:
Just those two comments and then it looks good to go.

arnav-t:
It isn't actually passing the convergence test, though.

ChrisRackauckas:
Oh yes, if the test fails it won't merge. I wrote that assuming it will pass the test. Take a look at `plot(sim)` to see if the convergence behavior is linear and not flat-lining at floating point epsilon. If the slope is just off, look for an error in the coefficients.

YingboMa:
It looks that you are using tab to indent. Could you use 2-space indentation instead?

arnav-t:
@YingboMa Sure.

arnav-t:
@ChrisRackauckas The convergence behavior isn't linear.
Linear ODE:
ode_linear
Linear 2D ODE:
ode_2dlinear
Why would this happen?

YingboMa:
```julia
julia> alg = LDDRK25();

julia> dts1 = 1 .//2 .^(12:-1:2);

julia> sim = test_convergence(dts1, test_problems_only_time[1], alg);

julia> plot(sim)

julia> plot(sim, dpi=200)
```
plot

In this plot, we can see that the convergence rate drops when the error is around 10^(-6). This is due to the fact that only 6 digits of the coefficients are given in the paper. This is more noticeable in the non-linear case.

plot


You can change the range of the `dts`, and mark the non-linear test as broken.
```julia
alg = LDDRK25()
dts1 = 1 .//2 .^(5:-1:2)
for prob in test_problems_only_time
	sim = test_convergence(dts1, prob, alg)
end
for prob in test_problems_linear
	sim = test_convergence(dts1, prob, alg)
end
for prob in test_problems_nonlinear
	sim = test_convergence(dts1, prob, alg)
	@test_broken sim.𝒪est[:l2] ≈ OrdinaryDiffEq.alg_order(alg) atol=testTol
end
```

ranocha:
> > @ChrisRackauckas The SSPRK tests seem to be in a different file (`ode_ssprk_tests.jl`).

> Yes indeed. My bad. Please add it to that one.

I don't want to play the devil's advocate, but it might be confusing if this method is tested in `ode_ssprk_tests.jl` because it is not SSP.

ChrisRackauckas:
True, we need another level to differentiate these methods. Probably a set of low_storage_rk.jl files.

ranocha:
> True, we need another level to differentiate these methods. Probably a set of low_storage_rk.jl files.

That would be good. We should also move the implementation of #596 to that file since the method implemented there is also not SSP (because it has negative Butcher coefficients).

PR #604, opened 2019-01-18
Adams methods: only update dts when accepted

ChrisRackauckas:
Accepted is EEst < 1, so there seems to be an error in the adaptivity handling here.


YingboMa:
The original looks correct to me. Because doing

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/4a2c592d01a575aa76def8d4482f8bb8176ceffc/src/perform_step/adams_bashforth_moulton_perform_step.jl#L1307-L1311

is unconditional, and if there is `integrator.EEst > one(integrator.EEst)`, then we undo the change in `dts`. IMO, the fix should be `integrator.EEst >= one(integrator.EEst)`

ChrisRackauckas:
I see. If it's equal to 1 we accept, so it shouldn't undo either, so it's already correct.

PR #605, opened 2019-01-18
Converts Matrices in anderson.jl to arrays of arrays

kanav99:
Though the `ode_convergence_tests.jl` pass successfully, I wrote a small experiment code which throws instability warning (it ran perfectly on my previous version):
```
using OrdinaryDiffEq
function f(du,u,p,t)
	du[1] = 2*u[1] + 3*u[2]
	du[2] = 4*u[1] + 5*u[2]
end
u0 = [1.0;1.0]
tspan = (0.0,1.0)
prob = ODEProblem(f,u0,tspan)
sol = solve(prob,ImplicitEuler())
print("hello")
sol2 = solve(prob,ImplicitEuler(nlsolve=NLAnderson()))


println(sol(0.5))
println(sol2(0.5))
println(sol(0.6))
println(sol2(0.6))
println(sol(0.7))
println(sol2(0.7))
println(sol(0.8))
println(sol2(0.8))
```
Which throws:
```
hello┌ Warning: Instability detected. Aborting
└ @ DiffEqBase ~/.julia/packages/DiffEqBase/8usQ9/src/integrator_interface.jl:162
[28.2652, 49.038]
ERROR: LoadError: Solution interpolation cannot extrapolate past the final timepoint. Either solve on a longer timespan or use the local extrapolation from the integrator interface.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] macro expansion at /home/kanav/.julia/packages/Parameters/NholY/src/Parameters.jl:743 [inlined]
 [3] ode_interpolation(::Float64, ::Function, ::Nothing, ::Type, ::Nothing, ::Symbol) at /home/kanav/.julia/dev/OrdinaryDiffEq/src/dense/generic_dense.jl:236
 [4] #call#6 at /home/kanav/.julia/dev/OrdinaryDiffEq/src/interp_func.jl:70 [inlined]
 [5] ODESolution at /home/kanav/.julia/packages/DiffEqBase/8usQ9/src/solutions/ode_solutions.jl:14 [inlined] (repeats 2 times)
 [6] top-level scope at none:0
 [7] include at ./boot.jl:317 [inlined]
 [8] include_relative(::Module, ::String) at ./loading.jl:1044
 [9] include(::Module, ::String) at ./sysimg.jl:29
 [10] exec_options(::Base.JLOptions) at ./client.jl:231
 [11] _start() at ./client.jl:425
in expression starting at /home/kanav/Projects/DiffEqExperiments/anderson-inplace.jl:15
```

### DIFF
--- a/src/caches/sdirk_caches.jl
+++ b/src/caches/sdirk_caches.jl
@@ -56,13 +56,8 @@ DiffEqBase.@def iipnlcachefields begin
     jac_config = nothing
     linsolve = nothing
     z₊ = similar(z)
-    if typeof(z) <: AbstractArray
-      zs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
-      gs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
-    else
-      zs = [zero(z) for i in 1:alg.nlsolve.n+1]
-      gs = [zero(z) for i in 1:alg.nlsolve.n+1]
-    end
+    zs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
+    gs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
   end
 
   if κ !== nothing
--- a/src/caches/sdirk_caches.jl
+++ b/src/caches/sdirk_caches.jl
@@ -20,6 +20,8 @@ DiffEqBase.@def iipnlcachefields begin
     jac_config = nothing
     linsolve = alg.linsolve(Val{:init},nf,u)
     z₊ = z
+    zs = []
+    gs = []
   elseif alg.nlsolve isa NLNewton
     if DiffEqBase.has_jac(f) && !DiffEqBase.has_invW(f) && f.jac_prototype !== nothing
       W = WOperator(f, dt, true)
@@ -34,6 +36,8 @@ DiffEqBase.@def iipnlcachefields begin
     jac_config = build_jac_config(alg,nf,uf,du1,uprev,u,tmp,dz)
     linsolve = alg.linsolve(Val{:init},uf,u)
     z₊ = z
+    zs = []
+    gs = []
   elseif typeof(alg.nlsolve) <: NLFunctional
     J = nothing
     W = nothing
@@ -42,6 +46,8 @@ DiffEqBase.@def iipnlcachefields begin
     jac_config = nothing
     linsolve = nothing
     z₊ = similar(z)
+    zs = []
+    gs = []
   elseif typeof(alg.nlsolve) <: NLAnderson
     J = nothing
     W = nothing
@@ -50,6 +56,13 @@ DiffEqBase.@def iipnlcachefields begin
     jac_config = nothing
     linsolve = nothing
     z₊ = similar(z)
+    if typeof(z) <: AbstractArray
+      zs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
+      gs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
+    else
+      zs = [zero(z) for i in 1:alg.nlsolve.n+1]
+      gs = [zero(z) for i in 1:alg.nlsolve.n+1]
+    end
   end
 
   if κ !== nothing
@@ -98,17 +111,29 @@ DiffEqBase.@def oopnlcachefields begin
   else
     tol = uToltype(min(0.03,first(reltol)^(0.5)))
   end
+  if typeof(alg.nlsolve) <: NLAnderson
+    if typeof(z) <: AbstractArray
+      zs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
+      gs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
+    else
+      zs = [zero(z) for i in 1:alg.nlsolve.n+1]
+      gs = [zero(z) for i in 1:alg.nlsolve.n+1]
+    end
+  else
+    zs = []
+    gs = []
+  end
   z₊,dz,tmp,b,k = z,z,z,z,rate_prototype
 end
 
 DiffEqBase.@def iipnlsolve begin
   _nlsolve = typeof(alg.nlsolve).name.wrapper
-  nlcache = NLSolverCache(κ,tol,min_iter,max_iter,10000,new_W,z,W,γ,c,ηold,z₊,dz,tmp,b,k)
+  nlcache = NLSolverCache(κ,tol,min_iter,max_iter,10000,new_W,z,W,γ,c,ηold,z₊,dz,tmp,b,k,zs,gs)
   nlsolve = _nlsolve{true, typeof(nlcache)}(nlcache)
 end
 DiffEqBase.@def oopnlsolve begin
   _nlsolve = typeof(alg.nlsolve).name.wrapper
-  nlcache = NLSolverCache(κ,tol,min_iter,max_iter,10000,new_W,z,W,γ,c,ηold,z₊,dz,tmp,b,k)
+  nlcache = NLSolverCache(κ,tol,min_iter,max_iter,10000,new_W,z,W,γ,c,ηold,z₊,dz,tmp,b,k,zs,gs)
   nlsolve = _nlsolve{false, typeof(nlcache)}(nlcache)
 end
 
--- a/src/caches/sdirk_caches.jl
+++ b/src/caches/sdirk_caches.jl
@@ -20,8 +20,8 @@ DiffEqBase.@def iipnlcachefields begin
     jac_config = nothing
     linsolve = alg.linsolve(Val{:init},nf,u)
     z₊ = z
-    zs = []
-    gs = []
+    zs = nothing
+    gs = nothing
   elseif alg.nlsolve isa NLNewton
     if DiffEqBase.has_jac(f) && !DiffEqBase.has_invW(f) && f.jac_prototype !== nothing
       W = WOperator(f, dt, true)
@@ -36,8 +36,8 @@ DiffEqBase.@def iipnlcachefields begin
     jac_config = build_jac_config(alg,nf,uf,du1,uprev,u,tmp,dz)
     linsolve = alg.linsolve(Val{:init},uf,u)
     z₊ = z
-    zs = []
-    gs = []
+    zs = nothing
+    gs = nothing
   elseif typeof(alg.nlsolve) <: NLFunctional
     J = nothing
     W = nothing
@@ -46,8 +46,8 @@ DiffEqBase.@def iipnlcachefields begin
     jac_config = nothing
     linsolve = nothing
     z₊ = similar(z)
-    zs = []
-    gs = []
+    zs = nothing
+    gs = nothing
   elseif typeof(alg.nlsolve) <: NLAnderson
     J = nothing
     W = nothing
@@ -120,8 +120,8 @@ DiffEqBase.@def oopnlcachefields begin
       gs = [zero(z) for i in 1:alg.nlsolve.n+1]
     end
   else
-    zs = []
-    gs = []
+    zs = nothing
+    gs = nothing
   end
   z₊,dz,tmp,b,k = z,z,z,z,rate_prototype
 end
--- a/src/caches/sdirk_caches.jl
+++ b/src/caches/sdirk_caches.jl
@@ -108,11 +108,11 @@ DiffEqBase.@def oopnlcachefields begin
   end
   if typeof(alg.nlsolve) <: NLAnderson
     if typeof(z) <: AbstractArray
-      zs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
-      gs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
+      zs = [vec(z) for i in 1:alg.nlsolve.n+1]
+      gs = [vec(z) for i in 1:alg.nlsolve.n+1]
     else
-      zs = [zero(z) for i in 1:alg.nlsolve.n+1]
-      gs = [zero(z) for i in 1:alg.nlsolve.n+1]
+      zs = [z for i in 1:alg.nlsolve.n+1]
+      gs = [z for i in 1:alg.nlsolve.n+1]
     end
   else
     zs = nothing
--- a/src/nlsolve/anderson.jl
+++ b/src/nlsolve/anderson.jl
@@ -79,8 +79,8 @@ end
 function (S::NLAnderson{true})(integrator)
   nlcache = S.cache
   @unpack t,dt,uprev,u,f,p = integrator
-  @unpack z,z₊,dz,tmp,κ,tol,k,c,γ,min_iter,max_iter = nlcache
-  ztmp = u
+  @unpack z,z₊,dz,b,tmp,κ,tol,k,c,γ,min_iter,max_iter = nlcache
+  ztmp = b
   mass_matrix = integrator.f.mass_matrix
   if typeof(integrator.f) <: SplitFunction
     f = integrator.f.f1
@@ -141,8 +141,7 @@ function (S::NLAnderson{true})(integrator)
 
     mk = min(S.n, iter-1)
     ztmp = vec(u)
-    ztmp .= gs[1]
-    ztmp .-= zs[1]
+    @. ztmp = gs[1] - zs[1]
     for i in 2:mk+1
       residuals[:,i-1] .= (gs[i] .- zs[i]) .- ztmp
     end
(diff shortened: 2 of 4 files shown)


src/nlsolve/anderson.jl
```diff
@@ -90,12 +90,16 @@ function (S::NLAnderson{true})(integrator)
   # precalculations
   κtol = κ*tol
 
-  zs = zeros(length(vec(z)), S.n+1)
-  gs = zeros(length(vec(z)), S.n+1)
+  zs = fill(Any[], S.n+1)
```
ChrisRackauckas: this should fill with `typeof(u)`. In fact, I would just do the comprehension `zs = [similar(u) for i in 1:S.n+1]`

src/nlsolve/anderson.jl
```diff
@@ -90,12 +90,16 @@ function (S::NLAnderson{true})(integrator)
...
+  gs = fill(Any[], S.n+1)
   residuals = zeros(length(vec(z)), S.n+1)
+  for e = 1:1:(S.n+1)
```
ChrisRackauckas: no need for this if you do a comprehension

src/nlsolve/anderson.jl
```diff
@@ -108,20 +112,19 @@ function (S::NLAnderson{true})(integrator)
...
     @. z₊ -= ztmp
   end
-  gs[:,1] .= vec(z₊)
+  gs[1] .= vec(z₊)
   @. dz = z₊ - z
   ndz = integrator.opts.internalnorm(dz)
   for t = (S.n+1):-1:2
-    @. zs[:,t] = zs[:,t-1]
-    @. gs[:,t] = gs[:,t-1]
+    zs[t] = zs[t-1]
```
ChrisRackauckas: .= otherwise these now alias

src/nlsolve/anderson.jl
```diff
@@ -108,20 +112,19 @@ function (S::NLAnderson{true})(integrator)
...
+    gs[t] = gs[t-1]
```
ChrisRackauckas: .=

src/nlsolve/anderson.jl
YingboMa: If you want to use pointer swaps, check out this https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/602/files#diff-a47f650c5d66e06a667d8a85cec5c42eR865

ChrisRackauckas: The issue is that the top one doesn't become the bottom one, so after enough steps they all are the same array.

kanav99: Oh got it. Thanks!

src/nlsolve/anderson.jl
```diff
@@ -135,22 +138,25 @@ function (S::NLAnderson{true})(integrator)
...
     end
-    gs[:,1] .= vec(z₊)
+    gs[1] .= vec(z₊)
 
     mk = min(S.n, iter-1)
-    residuals[:,1:mk] .= (gs[:,2:mk+1] .- zs[:,2:mk+1]) .- (gs[:,1] - zs[:,1])
-    alphas[1:mk] .= residuals[:,1:mk] \ (zs[:,1] .- gs[:,1])
+    tmp = (gs[1] - zs[1])
+    for i in 2:mk+1
+      residuals[:,i-1] .= (gs[i] .- zs[i]) .- (gs[1] - zs[1])
```
YingboMa: You should use tmp here

src/nlsolve/anderson.jl
```diff
@@ -135,22 +136,27 @@ function (S::NLAnderson{true})(integrator)
       mul!(ztmp, mass_matrix, z)
       @. z₊ -= ztmp
     end
-    gs[:,1] .= vec(z₊)
+    gs[1] .= vec(z₊)
 
     mk = min(S.n, iter-1)
-    residuals[:,1:mk] .= (gs[:,2:mk+1] .- zs[:,2:mk+1]) .- (gs[:,1] - zs[:,1])
-    alphas[1:mk] .= residuals[:,1:mk] \ (zs[:,1] .- gs[:,1])
+    tmp = (gs[1] - zs[1])
```
YingboMa: Parentheses here aren’t needed.

YingboMa: Consider using ztmp here.

src/nlsolve/anderson.jl
```diff
@@ -135,22 +136,27 @@ function (S::NLAnderson{true})(integrator)
...
+    for i in 2:mk+1
+      residuals[:,i-1] .= (gs[i] .- zs[i]) .- (gs[1] - zs[1])
+    end
+    alphas[1:mk] .= residuals[:,1:mk] \ (zs[1] .- gs[1])
```
YingboMa: Same here

kanav99:
Using `tmp` was the cause of the divergence. The code works perfectly fine now. We could decrease the memory requirement now (by reusing some arrays). But the source seems to be in a pretty stable condition now.

kanav99:
The output now was:
```
[28.2652, 49.038]
[28.2122, 48.9447]
[58.8456, 102.826]
[58.731, 102.624]
[122.813, 215.317]
[122.583, 214.913]
[256.638, 450.64]
[256.131, 449.748]
```
These seem to be perfectly fine now.

src/nlsolve/anderson.jl
```diff
@@ -135,22 +136,27 @@ function (S::NLAnderson{true})(integrator)
       mul!(ztmp, mass_matrix, z)
       @. z₊ -= ztmp
     end
-    gs[:,1] .= vec(z₊)
+    gs[1] .= vec(z₊)
 
     mk = min(S.n, iter-1)
-    residuals[:,1:mk] .= (gs[:,2:mk+1] .- zs[:,2:mk+1]) .- (gs[:,1] - zs[:,1])
-    alphas[1:mk] .= residuals[:,1:mk] \ (zs[:,1] .- gs[:,1])
+    ztmp = (gs[1] - zs[1])
```
ChrisRackauckas: get rid of this allocation?

ChrisRackauckas:
>Using tmp was the cause of the divergence. 

Is there a different cache array you can use there to get rid of the allocation? It's probably worthwhile to have just one more for this, or find a clever way to use an existing one.

src/nlsolve/anderson.jl
kanav99: Would replacing `=` with `.=` serve the purpose?

ChrisRackauckas: If `ztmp` already exists.

kanav99:
Currently the ztmp comes from cache (reusing `b`s memory). So should I remove the entry of `b` completely from the cache? Caches size would decrease then.
Also I am not sure, Would there be any allocation if I do`ztmp .= gs[1] - zs[1]` assuming ztmp is already allocated? Or should I do `ztmp .= gs[1]; ztmp .-=zs[1]` ?

ChrisRackauckas:
>Also I am not sure, Would there be any allocation if I doztmp .= gs[1] - zs[1] assuming ztmp is already allocated? Or should I do ztmp .= gs[1]; ztmp .-=zs[1] ?

Broadcasts fuse, so `ztmp .= gs[1] .- zs[1]` will not allocate, but if you forget the `.` on the `-` it will allocate. For simplicity you can do `@. ztmp = gs[1] - zs[1]` which means "dot everything".

kanav99:
Okay so I think I should leave the ztmp associated with b only as I cannot remove that from cache(All nlsolve algorithms use the same cache). Resuing `u` may have adverse effects. But we could reduce the memory footprint once we confirm that `u` is not used anymore. Do you agree?

src/nlsolve/anderson.jl
```diff
@@ -131,26 +132,33 @@ function (S::NLAnderson{true})(integrator)
...
       @. z₊ -= ztmp
     end
-    gs[:,1] .= vec(z₊)
+    gs[1] .= vec(z₊)
 
     mk = min(S.n, iter-1)
-    residuals[:,1:mk] .= (gs[:,2:mk+1] .- zs[:,2:mk+1]) .- (gs[:,1] - zs[:,1])
-    alphas[1:mk] .= residuals[:,1:mk] \ (zs[:,1] .- gs[:,1])
+    ztmp = vec(u)
+    @. ztmp = gs[1] - zs[1]
```
YingboMa: `@. ztmp = zs[1] - gs[1]`

src/nlsolve/anderson.jl
```diff
@@ -131,26 +132,33 @@ function (S::NLAnderson{true})(integrator)
...
+    for i in 2:mk+1
+      residuals[:,i-1] .= (gs[i] .- zs[i]) .- ztmp
```
YingboMa: `residuals[:,i-1] .= (gs[i] .- zs[i]) .+ ztmp`

src/nlsolve/anderson.jl
YingboMa: `alphas[1:mk] .= @view(residuals[:,1:mk]) \ ztmp`

src/nlsolve/anderson.jl
```diff
@@ -90,12 +90,12 @@ function (S::NLAnderson{true})(integrator)
   # precalculations
   κtol = κ*tol
 
-  zs = zeros(length(vec(z)), S.n+1)
-  gs = zeros(length(vec(z)), S.n+1)
+  zs = [zero(vec(z)) for i in 1:S.n+1]
+  gs = [zero(vec(z)) for i in 1:S.n+1]
```
YingboMa: If possible, move `zs` and `gs` to https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/nlsolve/type.jl#L3

kanav99: But moving there would make arrays for other methods as well. That would be I guess a lot of memory.

YingboMa: Just make them default to `nothing`.

kanav99: As they are arrays, I have default them to `[]` instead.

ChrisRackauckas: Parameterize it

kanav99:
I have in my last commit moved the `gs` and `zs` to the cache, along with requested changes. While working on this, I also got the idea on how to add broadcast to NLAnderson{false} too :-D
Requesting reviews @ChrisRackauckas @YingboMa

src/nlsolve/type.jl
```diff
@@ -18,6 +18,8 @@ mutable struct NLSolverCache{rateType,uType,W,uToltype,cType,gType} <: AbstractN
   tmp::uType
   b::uType # can be aliased with `k` if no unit
   k::rateType
+  zs::Array
```
ChrisRackauckas: this won't be inferred. Just parameterize it, and change the `[]`s to `nothing`

src/caches/sdirk_caches.jl
```diff
@@ -50,6 +56,13 @@ DiffEqBase.@def iipnlcachefields begin
     jac_config = nothing
     linsolve = nothing
     z₊ = similar(z)
+    if typeof(z) <: AbstractArray
```
ChrisRackauckas: for iip it has to be <:AbstractArray, so this branch isn't doing anything. Just keep the top

src/nlsolve/type.jl
kanav99: On changing `Array` to `Array{uType}` and `[]` to `nothing`, I get this error:
```
ERROR: LoadError: MethodError: no method matching OrdinaryDiffEq.NLSolverCache(::Float64, ::Float64, ::Int64, ::Int64, ::Int64, ::Bool, ::Float64, ::Float64, ::Int64, ::Int64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Float64, ::Nothing, ::Nothing)
Closest candidates are:
  OrdinaryDiffEq.NLSolverCache(::uToltype, ::uToltype, ::Int64, ::Int64, ::Int64, ::Bool, ::uType, ::W, ::gType, ::cType, ::uToltype, ::uType, ::uType, ::uType, ::uType, ::rateType, !Matched::Array{uType,N} where N, !Matched::Array{uType,N} where N) where {rateType, uType, W, uToltype, cType, gType} at /home/kanav/.julia/dev/OrdinaryDiffEq/src/nlsolve/type.jl:4
```

YingboMa: Use ‘zsType’, and put it in the above curly bracket.

ChrisRackauckas: no, parameterize the field

kanav99:
Sorry I had to bother you. Thank you for pointing it out. I have done all the requested changes

YingboMa:
Remember to do the same ztmp transformation to the out-of-place Anderson,
too.

kanav99:
Yes I get it. I think that I should work on that now, Making out-of-place Anderson support broadcast along with these transformations. May I do this in a seperate PR? Or should I do it over here only?

YingboMa:
Another PR would be good. This PR is getting long.

kanav99:
Okay I believe this PR is in a stable state. I would start working on the next PR.

src/nlsolve/type.jl
```diff
@@ -36,7 +38,7 @@ NLSolverCache(;κ=nothing, tol=nothing, min_iter=1, max_iter=10) =
 NLSolverCache(κ, tol, min_iter, max_iter, 0, true,
               ntuple(i->nothing, 4)...,
               κ === nothing ? κ : zero(κ),
-              ntuple(i->nothing, 5)...)
+              ntuple(i->nothing, 5)..., [], [])
```
YingboMa: ntuple(i->nothing, 7)...)

src/caches/sdirk_caches.jl
```diff
@@ -50,6 +56,8 @@ DiffEqBase.@def iipnlcachefields begin
     jac_config = nothing
     linsolve = nothing
     z₊ = similar(z)
+    zs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
+    gs = [zero(vec(z)) for i in 1:alg.nlsolve.n+1]
```
YingboMa: The OOP version prehapes doesn’t need explicit initialization. [vec(z) for i in 1:alg.nlsolve.n+1]

kanav99: Okay I think I should deal with this in the coming PR.
But I couldn't get why? Wouldn't doing only vec would make all the elements of the array same?

YingboMa: Because you are going to populate them in the OOP AA anyway.

kanav99: This is resolved in the latest commit.

ChrisRackauckas:
Looks like a test fails

src/nlsolve/anderson.jl
```diff
@@ -43,15 +45,17 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
     gs[1] = z₊
     
     mk = min(S.n, iter-1)
-    residuals[1:mk] = (gs[2:mk+1] .- zs[2:mk+1]) .- (gs[1] - zs[1])
+    residuals[1:mk] .= (gs[2:mk+1] .- zs[2:mk+1]) .- (gs[1] - zs[1])
     alphas[1:mk] .= residuals[1:mk] \ [(zs[1] - gs[1])]
```
ChrisRackauckas:
```suggestion
    alphas[1:mk] .= residuals[1:mk] \ (zs[1] - gs[1])
```

src/nlsolve/anderson.jl
```diff
@@ -131,26 +136,33 @@ function (S::NLAnderson{true})(integrator)
...
 
     mk = min(S.n, iter-1)
-    residuals[:,1:mk] .= (gs[:,2:mk+1] .- zs[:,2:mk+1]) .- (gs[:,1] - zs[:,1])
-    alphas[1:mk] .= residuals[:,1:mk] \ (zs[:,1] .- gs[:,1])
+    ztmp = vec(b)
+    @. ztmp = zs[1] - gs[1]
+    for i in 2:mk+1
+      @. residuals[:,i-1] = gs[i] - zs[i] + ztmp
+    end
+    alphas[1:mk] .= @view(residuals[:,1:mk]) \ ztmp
```
ChrisRackauckas: should use `ldiv!` here, and maybe not the default factorization. Is this where the singularity comes from?

YingboMa:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/commit/d821c5aaed92e9f2f35d049221da9e80fc3b8379

kanav99:
This commit fixes the singularity problem. It is strange how most tests passed before even the bug existed. Also some of the tests marked broken now work nicely. All the tests pass now. According to slack discussions, I should optimize this code in a next seperate PR.

src/nlsolve/anderson.jl
kanav99: Currently the out-of place version is limited to scalars only thats why I did so. I will make a seperate PR for introducing broadcast to oop method.

PR #606, opened 2019-01-19
Implemented RKM from Mead, Renaut (1999)  

keshavseksaria:
Implemented RKM from [Mead, Renaut (1999)](https://ac.els-cdn.com/S0021999199962605/1-s2.0-S0021999199962605-main.pdf?_tid=aa0cbff2-e2d7-42ed-894b-0a451ba5c77d&acdnat=1547857405_9deab4571d33961a2729b9ebd3680d98) #493

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -153,7 +153,7 @@ module OrdinaryDiffEq
 
   export FunctionMap, Euler, Heun, Ralston, Midpoint, SSPRK22,
          SSPRK33, SSPRK53, SSPRK53_2N1,SSPRK53_2N2, SSPRK63, SSPRK73, SSPRK83, SSPRK432, SSPRK932,
-         SSPRK54, SSPRK104, RK4, RKM ExplicitRK, OwrenZen3, OwrenZen4, OwrenZen5,
+         SSPRK54, SSPRK104, RK4, RKM, ExplicitRK, OwrenZen3, OwrenZen4, OwrenZen5,
          LDDRK64, LDDRK46, BS3, BS5, CarpenterKennedy2N54,
          ORK256, RK46NL, DP5, DP5Threaded, Tsit5, DP8, Vern6, Vern7, Vern8, TanYam7, TsitPap8,
          Vern9,Feagin10, Feagin12, Feagin14, CompositeAlgorithm, Anas5
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -153,7 +153,7 @@ module OrdinaryDiffEq
 
   export FunctionMap, Euler, Heun, Ralston, Midpoint, SSPRK22,
          SSPRK33, SSPRK53, SSPRK53_2N1,SSPRK53_2N2, SSPRK63, SSPRK73, SSPRK83, SSPRK432, SSPRK932,
-         SSPRK54, SSPRK104, RK4, ExplicitRK, OwrenZen3, OwrenZen4, OwrenZen5,
+         SSPRK54, SSPRK104, RK4, RKM, ExplicitRK, OwrenZen3, OwrenZen4, OwrenZen5,
          LDDRK64, CFRLDDRK64, BS3, BS5, CarpenterKennedy2N54,
          ORK256, RK46NL, DP5, DP5Threaded, Tsit5, DP8, Vern6, Vern7, Vern8, TanYam7, TsitPap8,
          Vern9,Feagin10, Feagin12, Feagin14, CompositeAlgorithm, Anas5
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
It looks like the test fails and you're only getting order 1.45 convergence. Can you plot the convergence simulation to see what the error behavior is like?

keshavseksaria:
`sim = test_convergence(dts, test_problems_only_time[1], alg);
plot(sim)`
screenshot from 2019-01-19 16-39-53
 look's like something is wrong in implementation?

keshavseksaria:
need to implement this 
screenshot from 2019-01-20 00-00-56

so i followed a simple RK4 type algorithm
Unable to figure out where i went wrong.
Kindly help.

src/perform_step/ssprk_perform_step.jl
```diff
@@ -1119,4 +1119,87 @@ end
...
+
+@muladd function perform_step!(integrator,cache::RKMConstantCache,repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack α2,α3,α4,α5,α6,β1,β2,β3,β4,β5,β6,c2,c3,c4,c5,c6 = cache
+
+  # u1
+  k1 =integrator.fsalfirst
+  u   = uprev + β1*dt*k1
+  # u2
+  k2 = f(uprev + α2*k1, p, t+c2*dt)
```
ChrisRackauckas: `α2*dt*k1`

ChrisRackauckas:
You're missing `dt*...` with a lot of the `k`s used in the function evaluation, but only for the out-of-place version. I assume the in-place one converges fine?

ChrisRackauckas:
Are you plotting the in-place or out of place implementation? Do they both have the same behavior?

What does your solutions look like? It seems that they might be constant, which would be an indicator of the issue.

src/perform_step/ssprk_perform_step.jl
```diff
@@ -1119,4 +1119,87 @@ end
...
+  k1 =integrator.fsalfirst
+  u   = uprev + β1*dt*k1
+  # u2
+  k2 = f(uprev + α2*k1*dt, p, t+c2*dt)
+  u  = u + β2*dt*k2
+  # u3
+  k3 = f(uprev + α3*k2*dt, p, t+c3*dt)
+  u   = u + β3*dt*k3
+  # u4
+  k4 = dt*f(uprev + α4*k3*dt, p, t+c4*dt)
```
ChrisRackauckas: don't do `dt*f` if the `dt` is handled later!

ChrisRackauckas:
Your issue is the 3 `dt*f`s in there. When you add the value to other u's, you do `dt*k`, so you have `dt^2`. For implementations we use the `dt*k` and not the `dt*f` they mention in the paper. There's a reason for this. Essentially, `k` should be a rate type, something with units `u/t`, and keeping it in this scaling is important for interpolations.

src/OrdinaryDiffEq.jl
```diff
@@ -153,7 +153,7 @@ module OrdinaryDiffEq
 
   export FunctionMap, Euler, Heun, Ralston, Midpoint, SSPRK22,
          SSPRK33, SSPRK53, SSPRK53_2N1,SSPRK53_2N2, SSPRK63, SSPRK73, SSPRK83, SSPRK432, SSPRK932,
-         SSPRK54, SSPRK104, RK4, ExplicitRK, OwrenZen3, OwrenZen4, OwrenZen5,
+         SSPRK54, SSPRK104, RK4, RKM ExplicitRK, OwrenZen3, OwrenZen4, OwrenZen5,
```
YingboMa: Comma is missing

src/perform_step/ssprk_perform_step.jl
keshavseksaria: I am aware of rateType and uType. Actually i changed the implementation, firstly i was using tmp (uType) then i implemented using k1,k2.. instead. I should have checked carefully.
Sorry for the inconvenience caused.

ChrisRackauckas:
This still fails the convergence tests. What do the plots look like?

keshavseksaria:
`julia> include("ode_ssprk_tests.jl")`
`julia> alg = RKM();`
`julia> sim = test_convergence(dts,test_problems_linear[1], alg);`
`julia>plot(sim)`

screenshot from 2019-01-20 16-05-55


`julia> sim = test_convergence(dts,test_problems_nonlinear[1], alg);`
`julia>plot(sim)`
screenshot from 2019-01-20 16-09-53

also on solving, it is showing the **`dt <= dtmin`**

YingboMa:
Could you create a new PR? Merging went wrong.

keshavseksaria:
Sure

PR #612, opened 2019-01-20
WIP: implement fifth order Radau IIA

YingboMa:
```julia
julia> using OrdinaryDiffEq

julia> function lorenz(u,p,t)
        du = similar(u)
        du[1] = 10.0(u[2]-u[1])
        du
       end
lorenz (generic function with 1 method)

julia> solve(prob, Vern9(), dt=0.01, adaptive=false)[end]
3-element Array{Float64,1}:
 -5.857685382424004
 -5.831082486429051
 23.932132987022584

julia> solve(prob, RadauIIA5(), dt=0.01, adaptive=false)[end]
3-element Array{Float64,1}:
 -5.857676300291801
 -5.831062113159235
 23.932140720000515
```

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -186,7 +186,8 @@ alg_order(alg::OwrenZen3) = 3
 alg_order(alg::OwrenZen4) = 4
 alg_order(alg::OwrenZen5) = 5
 alg_order(alg::LDDRK64) = 4
-alg_order(alg::LDDRK46) = 4
+alg_order(alg::CFRLDDRK64) = 4
+alg_order(alg::NDBLSRK144) = 4
 
 alg_order(alg::DP5) = 5
 alg_order(alg::DP5Threaded) = 5
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -101,7 +101,7 @@ get_current_alg_order(alg::JVODE,cache) = get_current_adaptive_order(alg,cache)
 get_current_alg_order(alg::QNDF,cache) = cache.order
 get_current_adaptive_order(alg::QNDF,cache) = cache.order
 
-alg_adaptive_order(alg::OrdinaryDiffEqAdaptiveAlgorithm) = error("Algorithm is adaptive with no order")
+#alg_adaptive_order(alg::OrdinaryDiffEqAdaptiveAlgorithm) = error("Algorithm is adaptive with no order")
 get_current_adaptive_order(alg::OrdinaryDiffEqAlgorithm,cache) = alg_adaptive_order(alg)
 get_current_adaptive_order(alg::CompositeAlgorithm,cache) = alg_adaptive_order(alg.algs[cache.current])
 
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -52,6 +52,7 @@ qmin_default(alg::DP8) = 1//3
 qmax_default(alg::OrdinaryDiffEqAlgorithm) = 10
 qmax_default(alg::CompositeAlgorithm) = minimum(qmax_default.(alg.algs))
 qmax_default(alg::DP8) = 6
+qmax_default(alg::RadauIIA5) = 8
 
 get_chunksize(alg::OrdinaryDiffEqAlgorithm) = error("This algorithm does not have a chunk size defined.")
 get_chunksize(alg::OrdinaryDiffEqAdaptiveImplicitAlgorithm{CS,AD}) where {CS,AD} = CS
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -199,6 +199,7 @@ alg_order(alg::TanYam7) = 7
 alg_order(alg::TsitPap8) = 8
 alg_order(alg::GenericImplicitEuler) = 1
 alg_order(alg::GenericTrapezoid) = 2
+alg_order(alg::RadauIIA5) = 5
 alg_order(alg::ImplicitEuler) = 1
 alg_order(alg::MidpointSplitting) = 2
 alg_order(alg::LinearExponential) = 1
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -281,6 +281,8 @@ alg_adaptive_order(alg::Feagin14) = 12
 alg_adaptive_order(alg::Rosenbrock23) = 3
 alg_adaptive_order(alg::Rosenbrock32) = 2
 
+alg_adaptive_order(alg::RadauIIA5) = 3
+
 alg_adaptive_order(alg::GenericImplicitEuler) = 0
 alg_adaptive_order(alg::GenericTrapezoid) = 1
 alg_adaptive_order(alg::ImplicitEuler) = 0
--- a/src/perform_step/firk_perform_step.jl
+++ b/src/perform_step/firk_perform_step.jl
@@ -107,16 +107,16 @@ end
     e1dt, e2dt, e3dt = e1/dt, e2/dt, e3/dt
     tmp = @. e1dt*z1 + e2dt*z2 + e3dt*z3
     mass_matrix != I && (tmp = mass_matrix*tmp)
-    err = @. integrator.fsalfirst + tmp
-    alg.smooth_est && (err = LU1 \ err)
-    atmp = calculate_residuals(err, uprev, u, abstol, reltol, internalnorm)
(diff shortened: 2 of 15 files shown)


YingboMa:
The adaptivity of the current implementation is very weird.
```julia
julia> using OrdinaryDiffEq, DiffEqDevTools, Test, LinearAlgebra

julia> using DiffEqProblemLibrary.ODEProblemLibrary: importodeproblems; importodeproblems()

julia> import DiffEqProblemLibrary.ODEProblemLibrary: van

julia> solve(remake(vanstiff, p=1e7), RadauIIA5(), abstol=1e-5, reltol=1e-7) |> length
668

julia> solve(remake(vanstiff, p=1e7), RadauIIA5()) |> length
┌ Warning: Interrupted. Larger maxiters is needed.
└ @ DiffEqBase ~/.julia/packages/DiffEqBase/8usQ9/src/integrator_interface.jl:150
95

julia> solve(remake(vanstiff, p=1e7), Kvaerno5()) |> length
110

julia> solve(remake(vanstiff, p=1e7), Kvaerno5(), abstol=1e-5, reltol=1e-7) |> length
401
```

src/alg_utils.jl
```diff
@@ -101,7 +101,7 @@ get_current_alg_order(alg::JVODE,cache) = get_current_adaptive_order(alg,cache)
 get_current_alg_order(alg::QNDF,cache) = cache.order
 get_current_adaptive_order(alg::QNDF,cache) = cache.order
 
-alg_adaptive_order(alg::OrdinaryDiffEqAdaptiveAlgorithm) = error("Algorithm is adaptive with no order")
+#alg_adaptive_order(alg::OrdinaryDiffEqAdaptiveAlgorithm) = error("Algorithm is adaptive with no order")
```
ChrisRackauckas: this is used in the adaptivity algorithms?

src/perform_step/firk_perform_step.jl
```diff
@@ -0,0 +1,129 @@
...
+  do_newton && return
+
+  u = @. uprev + z3
+
+  if adaptive
+    e1dt, e2dt, e3dt = e1/dt, e2/dt, e3/dt
+    tmp = @. e1dt*z1 + e2dt*z2 + e3dt*z3
+    mass_matrix != I && (tmp = mass_matrix*tmp)
+    err = @. integrator.fsalfirst + tmp
+    alg.smooth_est && (err = LU1 \ err)
```
ChrisRackauckas: I'm not sure you need estimation smoothing on Radau. At least not by default?

src/perform_step/firk_perform_step.jl
```diff
@@ -0,0 +1,129 @@
...
+    atmp = calculate_residuals(err, uprev, u, abstol, reltol, internalnorm)
+    @show integrator.EEst = internalnorm(atmp)^(6/4)
```
ChrisRackauckas: why ^(6/4)?

src/perform_step/firk_perform_step.jl
```diff
@@ -0,0 +1,129 @@
...
+
+    if integrator.iter == 1 || integrator.u_modified || integrator.EEst > oneunit(integrator.EEst)
```
ChrisRackauckas: what's this?

src/perform_step/firk_perform_step.jl
YingboMa: `internalnorm(atmp)^(1/4)` is the LTE, but the method order is `5`, so I did `^(6/4)` to compensate it.

src/perform_step/firk_perform_step.jl
YingboMa: It is to avoid the hump phenomenon.

src/perform_step/firk_perform_step.jl
ChrisRackauckas: the stepsize controller has an exponentiation in it for the order.

src/perform_step/firk_perform_step.jl
YingboMa: image

src/perform_step/firk_perform_step.jl
ChrisRackauckas: What method is the embedded one? Radau3? Isn't that L-stable?

src/perform_step/firk_perform_step.jl
YingboMa: In the stepsize controller, it is going to do `EEst^(1/(5+1))`, so `(EEst^(6/4))^(1/(5+1))` is going to be `EEst^(1/4)`.

src/perform_step/firk_perform_step.jl
```diff
@@ -0,0 +1,129 @@
...
+  cache.ηold = η
+  integrator.force_stepfail = do_newton
+  cache.nl_iters = iter
+  do_newton && return
+
+  u = @. uprev + z3
+
+  if adaptive
+    e1dt, e2dt, e3dt = e1/dt, e2/dt, e3/dt
+    tmp = @. e1dt*z1 + e2dt*z2 + e3dt*z3
```
ChrisRackauckas: is this not the err tilde term?

YingboMa: The err tilde term is `integrator.fsalfirst + tmp`.

src/perform_step/firk_perform_step.jl
YingboMa: There is still place like https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L259 , so I am not sure what to do. Should it be `get_current_adaptive_order`?

src/perform_step/firk_perform_step.jl
```diff
@@ -0,0 +1,129 @@
...
+    mass_matrix != I && (tmp = mass_matrix*tmp)
+    err = @. integrator.fsalfirst + tmp
```
ChrisRackauckas: shouldn't this not be err tilde?

src/perform_step/firk_perform_step.jl
ChrisRackauckas: Yes, that's an error. Please fix in this PR.

src/perform_step/firk_perform_step.jl
YingboMa: No, the embedded cannot be Radau3, because that require different nodes.
image

src/perform_step/firk_perform_step.jl
```diff
@@ -0,0 +1,158 @@
...
+
+  # precalculations
+  c1m1 = c1-1
+  c2m1 = c2-1
+  c1mc2= c1-c2
+  κtol = κ*tol # used in Newton iteration
+  γdt, αdt, βdt = γ/dt, α/dt, β/dt
+  J = calc_J(integrator, cache, is_compos)
+  if u isa Number
+    LU1 = γdt*mass_matrix - J
```
ChrisRackauckas: this term should be mass_matrix/(gamma*dt), but it's not

ChrisRackauckas: nevermind

ChrisRackauckas: wait, yes it is.

ChrisRackauckas: capture

ChrisRackauckas: In all cases you should have `gamma*dt`. The transformation is to divide both sides by `gdt = gamma*dt`. The reason is because `J` is O(n^2) but `I` is `O(n)` (without mass matrices of course), so this decreases the overall calculations. This is exactly the same as the `invW_t` transformation on the Rosenbrock and SDIRK methods.

src/perform_step/firk_perform_step.jl
```diff
@@ -0,0 +1,158 @@
...
+  end
+  cache.ηold = η
+  integrator.force_stepfail = do_newton
+  cache.nl_iters = iter
+  do_newton && return
+
+  u = @. uprev + z3
+
+  if adaptive
+    e1dt, e2dt, e3dt = e1/dt, e2/dt, e3/dt
```
ChrisRackauckas: These should all be divided by γdt

YingboMa: `e1`, `e2`, and `e3` are already divided by `γ0` in the tableau
image

PR #613, opened 2019-01-20
WIP : Adding Linear multistep SSPRK

dgan181:
I tried to implement the LMM SSPRK methods. (https://arxiv.org/abs/1504.04107v2). The fixed step size method doesn't work for either of the two methods ((3,2) or (4,3)). The (4,3) methods is still quite incomplete.


dgan181:
Evaluated: 0.9796304194960388 ≈ 3 (atol=0.25)

sin

Evaluated: 3.938423505065563 ≈ 3 (atol=0.25)

nonlin

ChrisRackauckas:
The first one is a very strong indicator that there is a problem. Most higher order methods, when there is a problem with the implementation, will still converge, just with order one. The second one is superconvergence.

dgan181:
@ChrisRackauckas  The above plots are both for SSPRKMSVS43. I thought there was something fishy with the implementation as well. However, it is very similar to how I implemented SSPRKMSVS32, and that passes all the tests. The 43 methods seems to superconverge for the non-linear tests, it converges at order 3 for the linear ones and at order 1 for `prob_ode_sin` and `prob_ode_sin_inplace`.

ChrisRackauckas:
yeah, from the plot you can see that the final error has super convergence while the other 2 norms are just 3rd order. But the linear one not going 3rd order is troublesome. There must be something in there that's messing up still if that's not getting higher order, though it's very weird that nonlinear tests could pass (and superconverge!) with an error. I'll give you some more time to work with it before taking a deep look. Try dropping the dts on the linear and see if there's a weird slope change that happens?

test/ode/ode_ssprk_tests.jl
```diff
@@ -228,6 +228,34 @@ sol = solve(test_problem_ssp, alg, dt=8/5, adaptive=false)
 sol = solve(test_problem_ssp_inplace, alg, dt=8/5, adaptive=false)
 @test mapreduce(t->all(0 .<= sol(t) .<= 1), (u,v)->u&&v, range(0, stop=8, length=50), init=true)
 
+alg = SSPRKMSVS32()
+for prob in test_problems_only_time
+  sim = test_convergence(dts, prob, alg)
+  # higher order as pure quadrature
+  @test abs(sim.𝒪est[:final]-1-OrdinaryDiffEq.alg_order(alg)) < testTol
```
ChrisRackauckas: why subtract by 1? It looks like this causes the tests to fail

ChrisRackauckas:
Bump the superconverging one to 4. It looks like real 4th order convergence and that is fine.

test/ode/ode_ssprk_tests.jl
dgan181: My mistake. I thought I had pushed that change. I''ll just do that

ChrisRackauckas:
Great!

dgan181:
@ChrisRackauckas This has just been the fixed step method. The variable step method still needs to be added. I thought I was expected to do that as well in this PR. There is a bit of code in the SSPRKMSVS32 with the variable step implementation which doesn't quite work yet. Should I open a PR to remove that bit?

ChrisRackauckas:
Do it in a near PR. I much prefer short and sweet PRs that keep the work moving forward. Long PRs making it harder to revert when there are errors and are harder to reviewer. So I would recommend just having one general idea per PR, getting them working, merge, repeat. You'll make a lot steadier progress like this!

PR #621, opened 2019-01-21
Use recursivecopy for numbers as well

devmotion:
Minor simplification due to https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/pull/48.


devmotion:
By the way, could we just use `recursivecopy` in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L218-L222 and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L224-L228?

PR #627, opened 2019-01-23
WIP: Richardson extrapolation for Euler method

saurabhkgp21:
Implemented perform_step method only for constant cache.

### DIFF
--- a/src/caches/extrapolation_caches.jl
+++ b/src/caches/extrapolation_caches.jl
@@ -1,0 +1,31 @@
+@cache struct RichardsonEulerCache{uType,rateType,tabType} <: OrdinaryDiffEqMutableCache
+  u::uType
+  uprev::uType
+  tmp::uType
+  k::rateType
+  fsalfirst::rateType
+  tab::tabType
+  T::Array{uType,2}
+end
+
+struct RichardsonEulerConstantCache{Int64} <: OrdinaryDiffEqConstantCache
+  m::Int64 #Generalise type of m
+  # Can add different order of n_{i} here
+  function RichardsonEulerConstantCache()
+    m = Int64(2)
+    new{Int64}(m)
+  end
+end
+
+function alg_cache(alg::RichardsonEuler,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{true}})
+  tmp = similar(u)
+  k = zero(rate_prototype)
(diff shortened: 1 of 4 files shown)


src/caches/extrapolation_caches.jl
```diff
@@ -0,0 +1,31 @@
...
+  uprev::uType
+  tmp::uType
+  k::rateType
+  fsalfirst::rateType
+  tab::tabType
+  T::Array{uType,2}
+end
+
+struct RichardsonEulerConstantCache{Int64} <: OrdinaryDiffEqConstantCache
+  m::Int64 #Generalise type of m
```
saurabhkgp21: Is it fine to use Int64 here or should I generalise it?

ChrisRackauckas: Use `Int` here. `Int` matches the machine integer so it stops a lot of errors that an come up with 32-bit machines

src/caches/extrapolation_caches.jl
```diff
@@ -0,0 +1,31 @@
...
+  u::uType
+  uprev::uType
+  tmp::uType
+  k::rateType
+  fsalfirst::rateType
+  tab::tabType
+  T::Array{uType,2}
+end
+
+struct RichardsonEulerConstantCache{Int64} <: OrdinaryDiffEqConstantCache
```
ChrisRackauckas: No need to parameterize by a concrete type. Delete `{Int64}` from here.

src/caches/extrapolation_caches.jl
```diff
@@ -0,0 +1,31 @@
...
+  m::Int64 #Generalise type of m
+  # Can add different order of n_{i} here
+  function RichardsonEulerConstantCache()
+    m = Int64(2)
```
ChrisRackauckas: All `Int`

saurabhkgp21:
How to decide the value of `m`? As of now, I have taken `m=2` for testing this method.

ChrisRackauckas:
From memory, this looks good. My books should arrive today in the mail (just moved, so they got all boxed up and shipped), and when they do, I sift through the algorithm to look for correctness and likely merge. It's much easier to keep the PRs compact and one-purpose, so we merge this and then continue on with the next parts. 

Next PR would be to use the error estimator to adaptively choose how far to go order-wise. Essentially, the `m` in the algorithm should be the maximum order, and the algorithm uses error estimates to choose how far to go. You should also have an option in the algorithm `order_adapt=true` for turning that on/off (useful for testing!). The book should describe how the embedding is used for order selection and stepsize selection. Keep adaptive stepsize for a separate PR.

PR #629, opened 2019-01-25
WIP: Implemented MEBDF for nonlinear ODEs

Phiro333:
I have also tried to implement a nonlinear version of the MEBDF method. It seems to converge correctly, but maybe there are still some optimizations or corrections to be done.


src/algorithms.jl
```diff
@@ -21,6 +21,26 @@ FunctionMap(;scale_by_time=false) = FunctionMap{scale_by_time}()
 
 ###############################################################################
 
+# MEBDF
+
+struct MEBDF{CS,AD,F,F2,FDT,T2} <: OrdinaryDiffEqNewtonAlgorithm{CS,AD}
```
ChrisRackauckas: looks like this was doubled? Delete the first

src/caches/sdirk_caches.jl
```diff
@@ -549,3 +549,50 @@ function alg_cache(alg::Union{Hairer4,Hairer42},u,rate_prototype,uEltypeNoUnits,
   Hairer4Cache(u,uprev,du1,fsalfirst,k,z₁,z₂,z₃,z₄,z₅,dz,b,tmp,atmp,J,
                W,uf,jac_config,linsolve,nlsolve,tab)
 end
+
+@cache mutable struct MEBDFCache{uType,rateType,uNoUnitsType,JType,WType,UF,JC,F,N} <: SDIRKMutableCache
```
ChrisRackauckas: I wouldn't classify it as SDIRK. It should go in the bdf files.

test/ode/ode_convergence_tests.jl
```diff
@@ -123,6 +123,15 @@ for i = 1:2
   sim17 = test_convergence(dts,prob,KenCarp3())
   @test sim17.𝒪est[:final] ≈ 3 atol=testTol
 
+  sim18 = test_convergence(dts,prob,MEBDF(extrapolant = :linear))
```
ChrisRackauckas: move it down a notch to the BDF section

src/caches/sdirk_caches.jl
Phiro333: true, it fits better into this category. I have moved it over and did the other updates aswell.

src/perform_step/bdf_perform_step.jl
```diff
@@ -882,3 +882,114 @@ function perform_step!(integrator,cache::QNDFCache,repeat_step=false)
...
+    @. z = dt*integrator.fsalfirst
+  else # :constant
+    z .= zero(eltype(u))
+  end
+
+### STEP 1
+ nlcache.tmp = uprev
+ z,η,iter,fail_convergence = nlsolve!(integrator)
+ fail_convergence && return
+ @. z₁ = uprev + z
```
ChrisRackauckas: why not just broadcast directly to `nlcache.tmp`?

src/algorithms.jl
```diff
@@ -803,11 +803,27 @@ end
 
 ################################################################################
 
+struct MEBDF{CS,AD,F,F2,FDT,T2} <: OrdinaryDiffEqNewtonAlgorithm{CS,AD}
```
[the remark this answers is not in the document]
Phiro333: I changed the name into MEBDF2

src/perform_step/bdf_perform_step.jl
Phiro333: I will check that later. z1 is needed in line 979 and 982 again.

src/perform_step/bdf_perform_step.jl
```diff
@@ -882,3 +882,114 @@ function perform_step!(integrator,cache::QNDFCache,repeat_step=false)
...
+ nlcache.c = 1
+ nlcache.new_W = false
+ z,η,iter,fail_convergence = nlsolve!(integrator)
+ fail_convergence && return
+ @. u = tmp2 + z
+
+### finalize
+ nlcache.ηold = η
+ nlcache.nl_iters = iter
+ integrator.EEst = 1
```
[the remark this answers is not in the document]
Phiro333: Ok. I would like to add adaptivity, but still have to find a way to do that.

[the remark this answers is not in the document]
ChrisRackauckas: Yeah we can discuss this in the channel a bit more. There's a few things to work out, or maybe a paper on it here.

ChrisRackauckas:
Fantastic work. Thanks! We should definitely make sure to add this to the DAE page.

src/perform_step/bdf_perform_step.jl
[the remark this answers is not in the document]
Phiro333: Sounds good! Thanks.

Phiro333:
That was fast. Thank you!

PR #630, opened 2019-01-25
Adds broadcast to NLAnderson{false} implementation

kanav99:
Makes it compatible with arrays. Mass matrix is also working fine with this (Except the test that were broken with the in-place version too).

### DIFF
--- a/src/nlsolve/anderson.jl
+++ b/src/nlsolve/anderson.jl
@@ -1,7 +1,9 @@
 function (S::NLAnderson{false,<:NLSolverCache})(integrator)
   nlcache = S.cache
   @unpack t,dt,uprev,u,f,p = integrator
-  @unpack z,tmp,W,κ,tol,c,γ,max_iter,min_iter,zs,gs = nlcache
+  @unpack z,tmp,W,κ,b,tol,c,γ,max_iter,min_iter,zs,gs = nlcache
+  mass_matrix = integrator.f.mass_matrix
+  ztmp = b
   if typeof(integrator.f) <: SplitFunction
     f = integrator.f.f1
   else
@@ -12,16 +14,21 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
 
   # zs = zeros(S.n+1)
   # gs = zeros(S.n+1)
-  residuals = zeros(S.n+1)
+  residuals = zeros(length(z), S.n+1)
   alphas = zeros(S.n)
   # initial step of NLAnderson iteration
   zs[1] = z
   iter = 1
   tstep = t + c*dt
-  u = tmp + γ*z
-  z₊ = dt*f(u, p, tstep)
+  u = @. tmp + γ*z
+  if mass_matrix == I
+    z₊ = dt*f(u, p, tstep)
+  else
+    mz = mass_matrix * z
+    z₊ = dt .* f(u, p, tstep) .- mz .+ z
+  end
   gs[1] = z₊
-  dz = z₊ - z
+  dz = z₊ .- z
   ndz = integrator.opts.internalnorm(dz)
   zptr = zs[S.n+1]
   gptr = gs[S.n+1]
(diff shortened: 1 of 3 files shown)


src/nlsolve/anderson.jl
```diff
@@ -1,7 +1,9 @@
 function (S::NLAnderson{false,<:NLSolverCache})(integrator)
   nlcache = S.cache
   @unpack t,dt,uprev,u,f,p = integrator
-  @unpack z,tmp,W,κ,tol,c,γ,max_iter,min_iter,zs,gs = nlcache
+  @unpack z,tmp,W,κ,b,tol,c,γ,max_iter,min_iter,zs,gs = nlcache
+  mass_matrix = integrator.f.mass_matrix
+  ztmp = b
```
YingboMa: This line is not necessary.

src/nlsolve/anderson.jl
```diff
@@ -12,16 +14,21 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
 
   # zs = zeros(S.n+1)
   # gs = zeros(S.n+1)
-  residuals = zeros(S.n+1)
+  residuals = zeros(length(z), S.n+1)
```
YingboMa: `fill(zero(eltype(z)), length(z), S.n+1)`

src/nlsolve/anderson.jl
```diff
@@ -12,16 +14,21 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
...
   alphas = zeros(S.n)
   # initial step of NLAnderson iteration
   zs[1] = z
   iter = 1
   tstep = t + c*dt
-  u = tmp + γ*z
-  z₊ = dt*f(u, p, tstep)
+  u = @. tmp + γ*z
+  if mass_matrix == I
+    z₊ = dt*f(u, p, tstep)
```
YingboMa: missing dot

src/nlsolve/anderson.jl
```diff
@@ -12,16 +14,21 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
 
   # zs = zeros(S.n+1)
   # gs = zeros(S.n+1)
-  residuals = zeros(S.n+1)
+  residuals = zeros(length(z), S.n+1)
   alphas = zeros(S.n)
```
YingboMa: `fill(zero(eltype(z)), S.n)`

ChrisRackauckas:
>Note that we are still mutating residuals

Let's at least get rid of that and add a test with StaticArray

kanav99:
I have currently only done the small changes you stated. I am now seeing what I can do about the array residuals. This time I will try to keep things modular.

ChrisRackauckas:
Re-running tests to see if QNDF fails again.

src/nlsolve/anderson.jl
```diff
@@ -42,15 +48,28 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
...
     gs[1] = z₊
 
     mk = min(S.n, iter-1)
-    residuals[1:mk] .= (gs[2:mk+1] .- zs[2:mk+1]) .- (gs[1] - zs[1])
-    alphas[1:mk] .= residuals[1:mk] \ [(zs[1] - gs[1])]
+    ztmp = zs[1] .- gs[1]
+    for i in 2:mk+1
+      @. residuals[:,i-1] = gs[i] - zs[i] + ztmp
+    end
+    resqr = qr(residuals[:,1:mk], Val(true))
```
YingboMa: Using `qr` won't help. You have `residuals = fill(zero(eltype(z)), length(z), S.n+1)` anyway.

YingboMa: Using `qr` instead of `qr!` here does nothing other than increases memory allocation. However, let's leave it as-is, and write our custom QR factorization in another PR.

PR #632, opened 2019-01-28
Add `vec` and `reshape` in out-of-place stiff solvers

YingboMa:
Fixes https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/416
regex:
```
%s/\(k. = \)W\\\(.*\)/\1reshape(W\\vec(\2), axes(uprev))/g
%s/\(est = .*\)W\\\(.*\)/\1reshape(W\\vec(\2), axes(tmp))/g
```

### DIFF
--- a/src/misc_utils.jl
+++ b/src/misc_utils.jl
@@ -165,3 +165,8 @@ macro cache(expr)
     $(esc(:jac_iter))($(esc(:c))::$name) = tuple($(jac_vars...))
   end
 end
+
+_reshape(v, siz) = reshape(v, siz)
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
Nice. Can you apply the same fix to StochasticDiffEq?

YingboMa:
Sure

ChrisRackauckas:
That's annoying that `vec` and `reshape` won't work. Maybe `_vec` and `_reshape` that passes numbers right through?

PR #634, opened 2019-01-29
Implemented RKC

deeepeshthakur:
@ChrisRackauckas @YingboMa I tried to implement RKC from #2 but I'm getting errors than I tried but couldn't resolve and also I have some doubts about changing some constants. It's basically the Julia translation of the fortran code from the paper [RKC: An explicit solver for parabolic PDEs](https://ir.cwi.nl/pub/4764).

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -275,6 +275,7 @@ alg_order(alg::SBDF) = alg.order
 
 alg_order(alg::ROCK2) = 2
 alg_order(alg::ROCK4) = 4
+alg_order(alg::RKC) = 2
 
 alg_order(alg::MEBDF2) = 2
 
@@ -320,6 +321,7 @@ beta1_default(alg::DP5,beta2) = typeof(beta2)(1//alg_order(alg)) - 3beta2/4
 beta1_default(alg::DP5Threaded,beta2) = typeof(beta2)(1//alg_order(alg)) - 3beta2/4
 
 gamma_default(alg::OrdinaryDiffEqAlgorithm) = 9//10
+gamma_default(alg::RKC) = 8//10
 
 qsteady_min_default(alg::OrdinaryDiffEqAlgorithm) = 1
 qsteady_max_default(alg::OrdinaryDiffEqAlgorithm) = 1
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -294,6 +294,7 @@ alg_adaptive_order(alg::Rosenbrock23) = 3
 alg_adaptive_order(alg::Rosenbrock32) = 2
 
 alg_adaptive_order(alg::RadauIIA5) = 3
+alg_adaptive_order(alg::RKC) = 2
 
 alg_adaptive_order(alg::GenericImplicitEuler) = 0
 alg_adaptive_order(alg::GenericTrapezoid) = 1
--- a/src/caches/rkc_caches.jl
+++ b/src/caches/rkc_caches.jl
@@ -88,3 +88,39 @@ end
 function alg_cache(alg::ROCK4,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
   ROCK4ConstantCache(real(uEltypeNoUnits), real(uEltypeNoUnits),u)
 end
+
+mutable struct RKCConstantCache{zType} <: OrdinaryDiffEqConstantCache
+  #to match the types to call maxeig!
+  zprev::zType
+  function RKCConstantCache(zprev)
+    new{typeof(zprev)}(zprev)
+  end
+end
+@cache struct RKCCache{uType,rateType,uNoUnitsType} <: OrdinaryDiffEqMutableCache
+  u::uType
+  uprev::uType
+  gprev::uType
+  gprev2::uType
+  tmp::uType
+  atmp::uNoUnitsType
+  fsalfirst::rateType
+  k::rateType
+  # k2::rateType
+  constantcache::RKCConstantCache
+end
+
+function alg_cache(alg::RKC,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{true}})
+  constantcache = RKCConstantCache(u) # WIP: not sure about what type to use in here
+  gprev = similar(u)
+  gprev2 = similar(u)
+  tmp = similar(u)
+  atmp = similar(u,uEltypeNoUnits)
+  fsalfirst = zero(rate_prototype)
+  k = zero(rate_prototype)
+  # k2 = zero(rate_prototype)
+  RKCCache(u, uprev, gprev, gprev2, tmp, atmp, fsalfirst, k, constantcache)
+end
+
+function alg_cache(alg::RKC,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
+  RKCConstantCache(u) # WIP: not sure about what type to use in here
+end
--- a/src/caches/rkc_caches.jl
+++ b/src/caches/rkc_caches.jl
@@ -105,22 +105,20 @@ end
   atmp::uNoUnitsType
   fsalfirst::rateType
   k::rateType
-  # k2::rateType
   constantcache::RKCConstantCache
 end
 
 function alg_cache(alg::RKC,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{true}})
-  constantcache = RKCConstantCache(u) # WIP: not sure about what type to use in here
+  constantcache = RKCConstantCache(u)
   gprev = similar(u)
   gprev2 = similar(u)
   tmp = similar(u)
   atmp = similar(u,uEltypeNoUnits)
   fsalfirst = zero(rate_prototype)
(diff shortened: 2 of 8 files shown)


src/alg_utils.jl
```diff
@@ -321,6 +322,7 @@ beta1_default(alg::DP5,beta2) = typeof(beta2)(1//alg_order(alg)) - 3beta2/4
 beta1_default(alg::DP5Threaded,beta2) = typeof(beta2)(1//alg_order(alg)) - 3beta2/4
 
 gamma_default(alg::OrdinaryDiffEqAlgorithm) = 9//10
+gamma_default(alg::RKC) = 8//10
```
deeepeshthakur: @ChrisRackauckas @YingboMa Changed gamma to match the constants of the paper

src/integrators/integrator_utils.jl
```diff
@@ -237,40 +237,45 @@ function stepsize_controller!(integrator,alg::Union{StandardControllerAlgs,
...
 end
 function step_reject_controller!(integrator,alg::Union{StandardControllerAlgs,
-                                 OrdinaryDiffEqNewtonAdaptiveAlgorithm{CS,AD,:Standard}}) where {CS, AD}
+  OrdinaryDiffEqNewtonAdaptiveAlgorithm{CS,AD,:Standard}}) where {CS, AD}
   integrator.dt = integrator.qold
 end
 
-function stepsize_controller!(integrator,
-                              alg::OrdinaryDiffEqNewtonAdaptiveAlgorithm{CS,AD,:Predictive}) where {CS, AD}
+const PredictiveControllerAlgs = Union{RKC}
```
deeepeshthakur: @ChrisRackauckas @YingboMa I had to use the Gustafsson predictive stepsize controller.

src/integrators/integrator_utils.jl
```diff
@@ -237,40 +237,45 @@ function stepsize_controller!(integrator,alg::Union{StandardControllerAlgs,
...
   # Gustafsson predictive stepsize controller
 
   if iszero(integrator.EEst)
     q = inv(integrator.opts.qmax)
   else
     gamma = integrator.opts.gamma
-    if alg isa RadauIIA5
-      @unpack nl_iters = integrator.cache
-      @unpack max_iter = alg
+    if alg isa RKC
```
deeepeshthakur: @ChrisRackauckas @YingboMa Added this condition because earlier the calls were designed for Implicit methods but except choosing fac, rest of the mechanics was exactly same as described in the paper

deeepeshthakur:
@ChrisRackauckas @YingboMa If it helps, the convergence results for test problems are
**Evaluated: 0.0 ≈ 2 (atol=0.1)
Evaluated: 5.225007222098151e-7 ≈ 2 (atol=0.1)**

ChrisRackauckas:
You should change that because it's based on the stability which is different for the two methods

deeepeshthakur:
> You should change that because it's based on the stability which is different for the two methods

Okay, I'll do that. Any suggestion on other issues?
Specially why `integrator.opts.adaptive` is false

ChrisRackauckas:
Adaptiveity is disabled during convergence tests

deeepeshthakur:
> Adaptiveity is disabled during convergence tests

So how can I enable it?

YingboMa:
Convergence tests are done in fixed timestep.

deeepeshthakur:
> Convergence tests are done in fixed timestep.

So, the difference in the evaluated order is not due because adaptivity is off?

YingboMa:
Yes. Check how you apply the coefficients.

ChrisRackauckas:
>So, the difference in the evaluated order is not because adaptivity is off?

Yes, it has nothing to do with time stepping adaptivity. The order being essentially zero is a sign that something is really off. What does the convergence plot look like?

deeepeshthakur:
I'll check the coefficients again and because the order was coming out to be zero, I assumed I did something wrong and never plotted the convergence plot.

deeepeshthakur:
@ChrisRackauckas @YingboMa I think I figured out what was going haywire.
This snippet is from Wikipedia. It shows how the Chebyshev polynomials are selected over different domains of `x`
image

So in the fortran code they have selected the second formula assuming that `x >= 1` 

**Snippet from the paper**
image

**My Code**
```
  w0 = 1.0 + 2.0/(13.0*(mdeg^2.0))
  temp1 = w0^2.0 - 1.0
  temp2 = sqrt(temp1)
  arg   = mdeg*log(w0 + temp2)
```
As you can see `arg` is equivalent to `x` and `arg < 1` in our test cases. So we are using the wrong formula. I thinking of implementing both these formulas 
<a href="https://www.codecogs.com/eqnedit.php?latex=\large&space;\mathbf{T_{s}&space;=&space;\left\{\begin{matrix}&space;\cosh(s\cosh^{-1}(x))\hspace{1cm}&space;1&space;\leq&space;x\\\hspace{1cm}\cos(s\cos^{-1}(x))\hspace{1cm}&space;-1&space;\leq&space;x&space;\leq&space;1&space;\end{matrix}\right.}" target="_blank"></a>
But that would be a little different from the original fortran implementation. Should I?

deeepeshthakur:
@ChrisRackauckas

src/caches/rkc_caches.jl
```diff
@@ -88,3 +88,39 @@ end
 function alg_cache(alg::ROCK4,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
   ROCK4ConstantCache(real(uEltypeNoUnits), real(uEltypeNoUnits),u)
 end
+
+mutable struct RKCConstantCache{zType} <: OrdinaryDiffEqConstantCache
+  #to match the types to call maxeig!
+  zprev::zType
+  function RKCConstantCache(zprev)
```
YingboMa: You don't need to write a constructor explicitly.

src/alg_utils.jl
ChrisRackauckas: yeah that's fine

src/integrators/integrator_utils.jl
ChrisRackauckas: add `OrdinaryDiffEqNewtonAdaptiveAlgorithm{CS,AD,:Predictive}` to the union

src/integrators/integrator_utils.jl
ChrisRackauckas: that works

src/integrators/integrator_utils.jl
```diff
@@ -479,8 +483,8 @@ end
 function reset_fsal!(integrator)
   # Under these condtions, these algorithms are not FSAL anymore
   if typeof(integrator.cache) <: OrdinaryDiffEqMutableCache ||
-     (typeof(integrator.cache) <: CompositeCache &&
-      typeof(integrator.cache.caches[1]) <: OrdinaryDiffEqMutableCache)
```
ChrisRackauckas: keep these kinds of changes in a different PR.

src/rkc_utils.jl
```diff
@@ -126,27 +153,26 @@ function maxeig!(integrator, cache::OrdinaryDiffEqMutableCache)
   end
   return false
 end
-
 """
-    choosedeg!(cache) -> nothing
+choosedeg!(cache) -> nothing
```
YingboMa: Why the indentation change?

src/perform_step/rkc_perform_step.jl
```diff
@@ -296,3 +296,169 @@ end
...
+      z2   = z1
+      z1   = z
+      dz2  = dz1
+      dz1  = dz
+      d2z2 = d2z1
+      d2z1 = d2z
+    end
+  end
+  # error estimate
+  tmp = 0.8*(uprev - u) + 0.4*dt*(fsalfirst + gprev)
```
ChrisRackauckas: this only needs to be calculated if adaptive?

src/perform_step/rkc_perform_step.jl
```diff
@@ -296,3 +296,169 @@ end
...
+  arg   = mdeg*log(w0 + temp2)
+  w1    = (sinh(arg)*temp1) / (cosh(arg)*mdeg*temp2 - w0*sinh(arg))
+  b1    = 1.0/((2.0*w0)^2.0)
+  b2    = b1
+
+  # stage-1
+  gprev2 = copy(uprev)
+  μs     = w1*b1
+  gprev  = uprev + dt*μs*fsalfirst
+  th2  = 0.0
```
ChrisRackauckas: this code is not type safe. These all assume Float64. Use `zero(eltype(u))` and `one(eltype(u))` instead

src/rkc_utils.jl
deeepeshthakur: I auto-indented it. I'll change it back.

ChrisRackauckas: PRs for fixing indentation are good, but muddling the diff so it's hard to see what you did in a PR that's about adding/changing an algorithm is bad.

src/caches/rkc_caches.jl
deeepeshthakur: But every constant cache has one!

src/rkc_utils.jl
deeepeshthakur: Yeah, I'll make sure it never happens again.

src/caches/rkc_caches.jl
YingboMa:
```julia
julia> struct A{T}
           a::T
       end

julia> A(1)
A{Int64}(1)

julia> struct B{T}
           b::T
           B(x) = new{typeof(x)}(x)
       end

julia> B(1)
B{Int64}(1)
```
You don't need to add a trivial constructor. Julia adds it automatically.

ChrisRackauckas:
The constructor issue is the last thing?

deeepeshthakur:
Yes. I made all the other corrections.

PR #636, opened 2019-02-02
Order and stepsize control RichardsonEuler

saurabhkgp21:
This PR is still in its initial stage. There are still bugs in this implementation.
@ChrisRackauckas is this a correct way to go on for Variable Order?

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -32,7 +32,12 @@ ExplicitRK(;tableau=ODE_DEFAULT_TABLEAU) = ExplicitRK(tableau)
 @inline trivial_limiter!(u, f, t) = nothing
 
 struct Euler <: OrdinaryDiffEqAlgorithm end
-struct RichardsonEuler <: OrdinaryDiffEqExtrapolationVarOrderVarStepAlgorithm end
+struct RichardsonEuler <: OrdinaryDiffEqExtrapolationVarOrderVarStepAlgorithm
+  max_order::Int
+  init_order::Int
+end
+RichardsonEuler(;max_order=9,init_order=1) = RichardsonEuler(max_order,init_order)
+
 struct RK46NL <: OrdinaryDiffEqAlgorithm end
 struct Heun <: OrdinaryDiffEqAdaptiveAlgorithm end
 struct Ralston <: OrdinaryDiffEqAdaptiveAlgorithm end
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -37,7 +37,7 @@ struct RichardsonEuler <: OrdinaryDiffEqExtrapolationVarOrderVarStepAlgorithm
   min_order::Int
   init_order::Int
 end
-RichardsonEuler(;max_order=4,min_order=1,init_order=1) = RichardsonEuler(max_order,min_order,init_order)
+RichardsonEuler(;max_order=9,min_order=1,init_order=9) = RichardsonEuler(max_order,min_order,init_order)
 
 struct RK46NL <: OrdinaryDiffEqAlgorithm end
 struct Heun <: OrdinaryDiffEqAdaptiveAlgorithm end
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -34,9 +34,10 @@ ExplicitRK(;tableau=ODE_DEFAULT_TABLEAU) = ExplicitRK(tableau)
 struct Euler <: OrdinaryDiffEqAlgorithm end
 struct RichardsonEuler <: OrdinaryDiffEqExtrapolationVarOrderVarStepAlgorithm
   max_order::Int
+  min_order::Int
   init_order::Int
 end
-RichardsonEuler(;max_order=9,init_order=1) = RichardsonEuler(max_order,init_order)
+RichardsonEuler(;max_order=4,min_order=1,init_order=1) = RichardsonEuler(max_order,min_order,init_order)
 
 struct RK46NL <: OrdinaryDiffEqAlgorithm end
 struct Heun <: OrdinaryDiffEqAdaptiveAlgorithm end
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -37,7 +37,7 @@ struct RichardsonEuler <: OrdinaryDiffEqExtrapolationVarOrderVarStepAlgorithm
   min_order::Int
   init_order::Int
 end
-RichardsonEuler(;max_order=9,min_order=1,init_order=9) = RichardsonEuler(max_order,min_order,init_order)
+RichardsonEuler(;max_order=9,min_order=1,init_order=5) = RichardsonEuler(max_order,min_order,init_order)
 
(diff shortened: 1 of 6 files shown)


src/caches/extrapolation_caches.jl
```diff
@@ -1,31 +1,53 @@
...
   fsalfirst = zero(rate_prototype)
-  tab = RichardsonEulerConstantCache()
-  T = fill(zeros(eltype(u), size(u)), (tab.m, tab.m))
-  RichardsonEulerCache(u,uprev,tmp,k,fsalfirst,tab,T)
+  order = 1
+  order_max = 9
+  dtproposeold = zero(dt)
+  dtpropose = zero(dt)
+  T = fill(zeros(eltype(u), size(u)), (order_max, order_max))
+  last_work = zero(dt)
```
ChrisRackauckas: prev. notice that the rest of the package uses `prev` for previous.

src/integrators/integrator_utils.jl
```diff
@@ -198,6 +199,23 @@ function step_reject_controller!(integrator,alg::JVODE)
   integrator.dt *= integrator.qold
 end
 
+function stepsize_controller!(integrator,alg::RichardsonEuler)
```
ChrisRackauckas: this is just a standard proportional controller? If so, no reason to just copy paste the code. Just change how it dispatches.

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -48,17 +48,22 @@ function initialize!(integrator,cache::RichardsonEulerConstantCache)
   integrator.fsallast = zero(integrator.fsalfirst)
   integrator.k[1] = integrator.fsalfirst
   integrator.k[2] = integrator.fsallast
+
+  cache.order = 3
```
ChrisRackauckas: ?

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -48,17 +48,22 @@ function initialize!(integrator,cache::RichardsonEulerConstantCache)
...
-  T = zeros(typeof(uprev), (m,m))
+  A = 1 + 1
+
   T[1,1] = u
+  dtpropose = dt #doubt but error_1 isn't defined
+  work = A/dtpropose
 
   halfdt = dt/2
-  for i in 2:m
+  for i in 2:order_max
```
ChrisRackauckas: order?

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -73,14 +78,96 @@ function perform_step!(integrator,cache::RichardsonEulerConstantCache,repeat_ste
...
+              integrator.k[2] = integrator.fsallast
+              return
+          else
+              if integrator.EEst > 4*(dt/halfdt)^4
+                  cache.order = order - 1
+                  cache.dtpropose = dtpropose
+                  return
+              end
+          end
+      elseif i == order
```
ChrisRackauckas: it looks like this only calculates if adaptive and i == order? That doesn't seem correct.

src/caches/extrapolation_caches.jl
```diff
@@ -1,31 +1,53 @@
...
 
-struct RichardsonEulerConstantCache <: OrdinaryDiffEqConstantCache
-  m::Int #Generalise type of m
-  # Can add different order of n_{i} here
-  function RichardsonEulerConstantCache()
-    m = 4
-    new(m)
-  end
+@cache mutable struct RichardsonEulerConstantCache{dtType,arrayType} <: OrdinaryDiffEqConstantCache
+  dtproposeold::dtType
```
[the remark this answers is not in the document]
saurabhkgp21: @ChrisRackauckas I had used dt argument of alg_cache() to initialise `prevdtpropose`. Shouldn't it match with `tType` by itself?

src/perform_step/extrapolation_perform_step.jl
saurabhkgp21: size of `T` is `(max_order, max_order)`, hence I had written something like this in my pseudo code. Now I had changed this with `size(T)[0]`. Is this fine?

src/perform_step/extrapolation_perform_step.jl
saurabhkgp21: @ChrisRackauckas  If step hasn't converged for `line == k - 1` it goes on to check for convergence in `line == k`. Here I have used order as `k`. Though now I have renamed order as cur_order for more clarity.
Am I missing something here?

src/integrators/integrator_utils.jl
saurabhkgp21: @ChrisRackauckas Here controller modifies `dt` according to calculated `dtpropose` in `perform_step`.

ChrisRackauckas: ahh it changed.

src/perform_step/extrapolation_perform_step.jl
ChrisRackauckas: Don't use `k` for order, it already has a meaning. Just name it `order`

src/algorithms.jl
```diff
@@ -31,7 +32,13 @@ ExplicitRK(;tableau=ODE_DEFAULT_TABLEAU) = ExplicitRK(tableau)
 @inline trivial_limiter!(u, f, t) = nothing
 
 struct Euler <: OrdinaryDiffEqAlgorithm end
-struct RichardsonEuler <: OrdinaryDiffEqAlgorithm end
+struct RichardsonEuler <: OrdinaryDiffEqExtrapolationVarOrderVarStepAlgorithm
+  max_order::Int
+  min_order::Int
+  init_order::Int
+end
+RichardsonEuler(;max_order=9,min_order=1,init_order=9) = RichardsonEuler(max_order,min_order,init_order)
```
ChrisRackauckas: init at order 9 might be high. I'd put it to 5

saurabhkgp21: In the initialising phase, work per step is checked for k = 1 to max_order. And the order was going as high as 7 in some cases. That's why I had put it as 9. If there can be other problems due to high init_order, I'll decrease this one.

ChrisRackauckas: The problem you were testing on was completely perfect for high order. Not all problems are like that.

saurabhkgp21: okay.. will set init_order as 5 and leave max_order as 9. Will it be fine?

ChrisRackauckas: yes

saurabhkgp21: done

ChrisRackauckas:
Yeah, I took a good look through it. It looks good. We will want a version that uses GBS to parallelize on though, since this method won't be quite as efficient. But as a method it looks complete. Is there a reason why this is still WIP?

saurabhkgp21:
Changed the title now. Yeah GBS will give better results upon parallelization than this one. We can also parallelize this one first since order and stepsize control is completed. After that, I can implement parallelize version of GBS.

PR #638, opened 2019-02-03
Update nlsolve

devmotion:
I tried to update Anderson acceleration to the same format as functional iteration and Newton method. Maybe it would be possible to simplify the code even more by exploiting the similar structure of all algorithms and replacing different parts (such as initialization and computation of new iterates) by function calls.

Some questions/suggestions:
- Is there a reason for using a common `NLSolverCache` instead of different types for each algorithm? (BTW, the names `AbstractNLsolveSolver` and `AbstractNLsolveCache` seem to be a bit inconsistent with the name `NLSolverCache`)
- Is there a reason for not including `alphas` and `residuals` in the cache for Anderson acceleration?
- Since StochasticDiffEq can be updated by copying and only replacing lines such as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/nlsolve/functional.jl#L30-L34 the non-linear solvers should better be moved to DiffEqBase (or another separate package), I guess. The relevant lines could be replaced by something like `f = nlsolve_f(integrator)` which then could be implemented differently in OrdinaryDiffEq and StochasticDiffEq

### DIFF
--- a/src/nlsolve/anderson.jl
+++ b/src/nlsolve/anderson.jl
@@ -1,22 +1,17 @@
-function (S::NLAnderson{false,<:NLSolverCache})(integrator)
+@muladd function (S::NLAnderson{false})(integrator)
   nlcache = S.cache
   @unpack t,dt,uprev,u,f,p = integrator
   @unpack z,tmp,W,κ,tol,c,γ,max_iter,min_iter,zs,gs = nlcache
   mass_matrix = integrator.f.mass_matrix
-  if typeof(integrator.f) <: SplitFunction
-    f = integrator.f.f1
-  else
-    f = integrator.f
-  end
+  f = integrator.f isa SplitFunction ? integrator.f.f1 : integrator.f
   # precalculations
   κtol = κ*tol
+  residuals = Array{eltype(z)}(undef, length(z), S.n + 1)
+  alphas = Array{eltype(z)}(undef, S.n)
+  ztmp = Array{eltype(z)}(undef, z isa Number ? 1 : size(z))
+  vecztmp = vec(ztmp)
 
-  # zs = zeros(S.n+1)
-  # gs = zeros(S.n+1)
-  residuals = fill(zero(eltype(z)), length(z), S.n+1)
-  alphas = fill(zero(eltype(z)), S.n)
-  # initial step of NLAnderson iteration
-  zs[1] = z
+  # initial step of fixed point iteration with Anderson acceleration
   iter = 1
   tstep = t + c*dt
   u = @. tmp + γ*z
@@ -26,27 +21,31 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
     mz = mass_matrix * z
     z₊ = dt .* f(u, p, tstep) .- mz .+ z
   end
+  zs[1] = z
   gs[1] = z₊
-  dz = z₊ .- z
-  ndz = integrator.opts.internalnorm(dz)
+
+  # circularly shift zs and gs and update zs[1]
   zptr = zs[S.n+1]
   gptr = gs[S.n+1]
-  for t = (S.n+1):-1:2
-    zs[t] = zs[t-1]
-    gs[t] = gs[t-1]
+  for t in S.n:-1:1
+    zs[t + 1] = zs[t]
+    gs[t + 1] = gs[t]
   end
-  zs[1] = zptr
-  gs[1] = gptr
-  # zs = circshift(zs, 1)
-  # gs = circshift(gs, 1)
-  z = z₊
-  zs[1] = z
-  η = nlcache.ηold
-  do_anderson = true
+  zs[1] = z₊
 
-  # anderson acceleration for fixed point iteration
-  fail_convergence = false
-  while (do_anderson || iter < min_iter) && iter < max_iter
+  # compute initial values for early stopping criterion
+  ndz = integrator.opts.internalnorm(z₊ .- z)
+
+  # update solution
+  z = z₊
+
+  # check stopping criterion for initial step
+  η = nlcache.ηold
+  do_anderson = true # TODO: this makes `min_iter` ≥ 2
+
+  # fixed point iteration with Anderson acceleration
+  while do_anderson && iter < max_iter
+    # compute next iterate
     iter += 1
     u = @. tmp + γ*z
     if mass_matrix == I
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
>Is there a reason for using a common NLSolverCache instead of different types for each algorithm? (BTW, the names AbstractNLsolveSolver and AbstractNLsolveCache seem to be a bit inconsistent with the name NLSolverCache)

They were common enough at first, but they have diverged. I agree that now it's kind of pushing it.

>Since StochasticDiffEq can be updated by copying and only replacing lines such as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/nlsolve/functional.jl#L30-L34 the non-linear solvers should better be moved to DiffEqBase (or another separate package), I guess. The relevant lines could be replaced by something like f = nlsolve_f(integrator) which then could be implemented differently in OrdinaryDiffEq and StochasticDiffEq

Yes, the next question I was going to have was about getting this over to StochasticDiffEq. Making it be part of DiffEqBase (or making a StarDiffEq for all the common utilities) is probably the best solution. I would like Anderson in DelayDiffEq's iterations too, but that would likely need heavy modification.

src/nlsolve/anderson.jl
```diff
@@ -58,156 +57,148 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
...
-    gs[1] = gptr
-    # zs = circshift(zs, 1)
-    # gs = circshift(gs, 1)
     zs[1] = z₊
+
+    # check early stopping criterion
     ndzprev = ndz
-    dz = z₊ .- z
-    ndz = integrator.opts.internalnorm(dz)
+    ndz = integrator.opts.internalnorm(z₊ .- z)
```
ChrisRackauckas: we should remove the temporary here. There must be something to write into?

ChrisRackauckas: nevermind, not in place

src/nlsolve/anderson.jl
```diff
@@ -58,156 +57,148 @@ function (S::NLAnderson{false,<:NLSolverCache})(integrator)
...
-    ztmp = zs[1] .- gs[1]
-    for i in 2:mk+1
-      @. residuals[:,i-1] = gs[i] - zs[i] + ztmp
-    end
-    resqr = qr(residuals[:,1:mk], Val(true))
-    if typeof(ztmp) <: AbstractArray
-      alphas[1:mk] = resqr \ ztmp
-    else
-      alphas[1:mk] = resqr \ [ztmp]
+    @. ztmp = zs[1] - gs[1]
```
ChrisRackauckas: shouldn't assume mutability when not in place

devmotion: Is it not OK even if `ztmp` is defined as an array in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/638/files/fe91244da8c13b9f29c7d62d339f253541bb25b9#diff-8118feaffe7fe662c953104e316305afR11?

ChrisRackauckas: That would work, but why not just keep it a type-matching value so that way SArrays work without a conversion to a vector here?

ChrisRackauckas: I think the only thing that truly needs to be an array is `residuals`?

ChrisRackauckas: Managing the case where `residuals` is an SArray` would be a good optimization to do some time in the future though.

devmotion:
>  Yes, the next question I was going to have was about getting this over to StochasticDiffEq. Making it be part of DiffEqBase (or making a StarDiffEq for all the common utilities) is probably the best solution. I would like Anderson in DelayDiffEq's iterations too, but that would likely need heavy modification.

Yes, my main motivation for this PR was getting more familiar with the implementation in OrdinaryDiffEq before adding Anderson to DelayDiffEq, and I thought cleaning the current implementation and adding it to StochasticDiffEq would be a good start.

src/nlsolve/anderson.jl
devmotion: The main issue I saw with the current implementation are lines such as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/638/files#diff-8118feaffe7fe662c953104e316305afL63, where it is implicitly assumed that `zs[i]`, `gs[i]` and `ztmp` are one-dimensional arrays (or scalars). The next thing I noticed was that since `residuals` is defined as an Array, the QR decomposition `qr` in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/638/files#diff-8118feaffe7fe662c953104e316305afL65 could be replaced by `qr!`, as in the in-place version. Similarly, the left division in line https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/638/files#diff-8118feaffe7fe662c953104e316305afL67 and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/638/files#diff-8118feaffe7fe662c953104e316305afL69 can be replaced by the in-place `ldiv!` as long as `ztmp` is guaranteed to be an Array, since also `alphas` is already defined as an Array. Hence I assumed the easiest way would be to define `ztmp` as an Array and just update it in every iteration.

devmotion:
> They were common enough at first, but they have diverged. I agree that now it's kind of pushing it.

I would like to address the introduction of separate caches in another PR since I want to keep the changes kind of limited.

> Yes, the next question I was going to have was about getting this over to StochasticDiffEq. Making it be part of DiffEqBase (or making a StarDiffEq for all the common utilities) is probably the best solution. I would like Anderson in DelayDiffEq's iterations too, but that would likely need heavy modification.

As soon as there's a decision on whether those common utilities should be moved to DiffEqBase or rather another package, I'll create a PR for moving nlsolve there (and some other common parts of StochasticDiffEq and OrdinaryDiffEq).

src/nlsolve/anderson.jl
devmotion: It seems that the standard implementation of `ldiv!` in LinearAlgebra doesn't work with a `ztmp` of type StaticArray due to https://github.com/JuliaArrays/StaticArrays.jl/issues/589. Otherwise it would be possible to use the implementation
```julia
ztmp = zs[1] .- gs[1]
vecztmp = _vec(ztmp)
for i in 2:(mk + 1)
    residuals[:, i - 1] .= _vec(gs[i]) .- _vec(zs[i]) .+ vecztmp
end
resqr = qr!(view(residuals, :, 1:mk), Val(true))
ldiv!(view(alphas, 1:mk), resqr, vecztmp)
for i in 1:mk
    z₊ = @. z₊ + alphas[i] * (gs[i+1] - gs[1])
end
```

YingboMa:
Before moving the current implementation to other packages, let's finish https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/633 with `SArray` compatibility in mind. That would resolve https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/638#discussion_r253300892, too.

PR #645, opened 2019-02-04
Drop units in `eigen_est`, since `internalnorm` drops units

YingboMa:


src/perform_step/low_order_rk_perform_step.jl
```diff
@@ -783,7 +783,7 @@ end
       ϱu += (k7[i] - k6[i])^2
       ϱd += (g7[i] - g6[i])^2
     end
-    integrator.eigen_est = sqrt(ϱu/ϱd)
+    integrator.eigen_est = ustrip(sqrt(ϱu/ϱd))
```
ChrisRackauckas: do we actually depend on unitful already?

devmotion:
Just noticed that https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/low_order_rk_perform_step.jl#L781 causes a type instability, as far as I can see. I guess it should be `ϱd = zero(eltype(g7))`.

src/perform_step/low_order_rk_perform_step.jl
```diff
@@ -778,12 +778,12 @@ end
     g6 = tmp
     g7 = u
     # Hairer II, page 22
-    ϱu, ϱd = zero(eltype(k7)), zero(eltype(k7))
+    ϱu, ϱd = zero(eltype(k7)), zero(eltype(g7))
```
devmotion: Sorry, my bad, I think both values should be squared as well.

src/perform_step/low_order_rk_perform_step.jl
```diff
@@ -778,12 +778,12 @@ end
...
     g7 = u
     # Hairer II, page 22
-    ϱu, ϱd = zero(eltype(k7)), zero(eltype(k7))
+    ϱu, ϱd = zero(eltype(k7))^2, zero(eltype(g7))^2
     @inbounds for i in eachindex(k7)
       ϱu += (k7[i] - k6[i])^2
       ϱd += (g7[i] - g6[i])^2
     end
-    integrator.eigen_est = sqrt(ϱu/ϱd)
+    integrator.eigen_est = sqrt(ϱu/ϱd)*oneunit(t)
```
ChrisRackauckas: why would this have units then?

YingboMa: Multiply by `oneunit(t)` cancels units.

YingboMa:
Bump

ChrisRackauckas:
Rebase first.

PR #649, opened 2019-02-06
Small fixes

devmotion:
Addresses comments in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/646 and removes superfluous `typeof` inside of `zero`. However, I'm not completely sure in which cases it is sufficient to use `similar` and in which cases `zero` is required, so I worked only on your remarks and similar occurrences so far.

### DIFF
--- a/src/caches/kencarp_kvaerno_caches.jl
+++ b/src/caches/kencarp_kvaerno_caches.jl
@@ -55,7 +55,7 @@ function alg_cache(alg::KenCarp3,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNo
     uf = DiffEqDiffTools.UJacobianWrapper(f,t,p)
   end
 
-  z₁ = similar(u); z₂ = similar(u); z₃ = similar(u); z₄ = z
+  z₁ = zero(u); z₂ = zero(u); z₃ = zero(u); z₄ = z
   atmp = similar(u,uEltypeNoUnits)
 
   KenCarp3Cache(u,uprev,du1,fsalfirst,k,z₁,z₂,z₃,z₄,k1,k2,k3,k4,dz,b,tmp,atmp,J,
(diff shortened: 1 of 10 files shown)


ChrisRackauckas:
>However, I'm not completely sure in which cases it is sufficient to use `similar` and in which cases `zero` is required, so I worked only on your remarks and similar occurrences so far.

zero is just safer. You can get random stuff if you don't zero. Technically the only way this should happen in a standard case is if we give it to a user to modify and they don't, expecting it to stay zero. If it's `similar`, that will be a non-zero derivative and blow up. But really, it's cheap and safe to always zero so we probably should.

src/caches/kencarp_kvaerno_caches.jl
```diff
@@ -128,10 +124,6 @@ function alg_cache(alg::KenCarp4,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNo
   tab = KenCarp4Tableau(real(uBottomEltypeNoUnits),real(tTypeNoUnits))
   γ, c = tab.γ, tab.c3
   @oopnlsolve
-
-  uprev3 = u
-  tprev2 = t
```
ChrisRackauckas: is that all of these?

devmotion:
> But really, it's cheap and safe to always zero so we probably should.

For all fields? I was wondering whether there are "safe" fields such as `atmp` that could be initialized with `similar` and how one could distinguish them from "unsafe" fields.

ChrisRackauckas:
The safe ones are the ones we write into and we know by broadcast that they will be overwritten before used. The unsafe ones are the ones where we expect the user to overwrite every input. There are examples where people don't, because they assume that inputs will have "zero" (so, if ... end don't do anything on `du[i]` otherwise because it's zero)

src/caches/kencarp_kvaerno_caches.jl
devmotion: Yes, at least that's all that I found

Issue #640, opened 2019-02-03
Minimizing and Testing Memory Usage in Low-Memory RK Methods


ChrisRackauckas:
@ranocha tasked us at first with fixing memory usage in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/178 . We now have all of the pieces in place to do it correctly (see https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/178#issuecomment-460039918 ). What remains to be done are some `calck` optimizations on the low-memory methods, and testing to make sure it works out. 

How to do it is already described, so let's talk about testing. Memory tracking makes sense in the asymptotic regime where `u0` is large. When `u0` is large, say 100,000, then, since the size of the integrator is constant, the size of the `uType` and `rateType` vectors will completely dominate. In that case, we can measure the integrator's size in order to see how many registers we are actually using. It's quite accurate:

```julia
using OrdinaryDiffEq
u0 = rand(1000,1000)
prob = ODEProblem((du,u,p,t)-> du .= u,u0,(0.0,1.0))
integ = init(prob,Tsit5(),save_start=false,save_end=false,
                          save_everystep = false) # this sets dense = calck = false
Base.summarysize(integ)/Base.summarysize(u0) # 13.00020737396313
```

and understands aliasing:

```julia
Base.summarysize([u0 for i in 1:7]) / Base.summarysize(u0) # 1.0000119999400003
```

So with this, our goal should be to get these methods down to exactly the amount they require. Some things to note are:

- When we first start out in `solve.jl`, we copy `u0`. We should add a new argument to the common interface for `copy_start = true`. True for safety, since if you change `u0` all sorts of wonky things will happen to people's code. But true performance purists may want us to re-use that array, so let's do it.
- Low-memory RK methods, when `calck` is false, don't need the `k` values. They can make the algorithm work out just by aliasing those to another array instead. Note that these are crucial to common use though, since `saveat` is a widely used thing, so we should try to see if we can do this in a way that eliminates allocations even when `calck=true` (which is the `saveat` case)
- Low-memory RK methods papers say how many registers it should be using, so we better be matching that integer. If we aren't, we should dig in and find out why.

ranocha:
[...]

PR #654, opened 2019-02-08
alias uprev=u if calck==false for some low storage methods

ranocha:
Cf. #640.

### DIFF
--- a/test/ode/ode_low_storage_rk_tests.jl
+++ b/test/ode/ode_low_storage_rk_tests.jl
@@ -32,7 +32,6 @@ test_problems_nonlinear = [prob_ode_nonlinear, prob_ode_nonlinear_inplace]
 # Note: The memory requirements of the 2N methods can be reduced if an assignment
 # of the form `tmp = A2end[i]*tmp + dt*f(u, p, t+c2end[i]*dt)` can be carried out
 # without saving `f(u, p, t+c2end[i]*dt)` as `k`.
-# Note: Additional memory can be saved for some schemes if `uprev` can be aliased with `u`.
 u0_large = rand(10^6)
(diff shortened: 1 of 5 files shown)


test/ode/ode_low_storage_rk_tests.jl
```diff
@@ -196,7 +195,7 @@ for prob in test_problems_nonlinear
   @test sim.𝒪est[:final] ≈ OrdinaryDiffEq.alg_order(alg) atol=testTol
 end
 integ = init(prob_ode_large, alg, dt=1.e-2, save_start=false, save_end=false, save_everystep=false)
-@test Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 5
+@test Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 4
```
ChrisRackauckas: what's the theoretical size that these methods should be at?

ranocha: The theoretically best size would be `2`. We have two additional arrays since we store `u0` as part of the problem and we don't support function evaluations of the form `u2 = a*u2 + b*dt*f(u1)` using only two arrays.

ChrisRackauckas: Got it. For the first, we should add a common interface argument for whether or not to re-use that `u0` as `u`. Defaulting to false, but this is a case where you'd usually want true. For the second part, I am inclined to say we just add that form to ODEFunction.

ranocha: That sounds good. A keyword such as `alies_u0`?

ChrisRackauckas: yup that's a good name

PR #655, opened 2019-02-08
add keyword `alias_u0` with default value `false`

ranocha:
Cf. https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/654#discussion_r255011597 and #640.

This is my first PR changing some keyword arguments. Hence, it should be reviewed carefully because I might have missed some connections to other parts.

### DIFF
--- a/src/integrators/type.jl
+++ b/src/integrators/type.jl
@@ -43,7 +43,6 @@ mutable struct DEOptions{absType,relType,QT,tType,F1,F2,F3,F4,F5,F6,tstopsType,d
   force_dtmin::Bool
   advance_to_tstop::Bool
   stop_at_next_tstop::Bool
-  alias_u0::Bool
 end
 
 """
--- a/src/integrators/type.jl
+++ b/src/integrators/type.jl
@@ -43,6 +43,7 @@ mutable struct DEOptions{absType,relType,QT,tType,F1,F2,F3,F4,F5,F6,tstopsType,d
   force_dtmin::Bool
   advance_to_tstop::Bool
   stop_at_next_tstop::Bool
+  alias_u0::Bool
 end
 
 """
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -56,7 +56,8 @@ function DiffEqBase.__init(
   progress_message = ODE_DEFAULT_PROG_MESSAGE,
   userdata=nothing,
   allow_extrapolation = alg_extrapolates(alg),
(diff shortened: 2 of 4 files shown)


src/integrators/type.jl
```diff
@@ -43,6 +43,7 @@ mutable struct DEOptions{absType,relType,QT,tType,F1,F2,F3,F4,F5,F6,tstopsType,d
   force_dtmin::Bool
   advance_to_tstop::Bool
   stop_at_next_tstop::Bool
+  alias_u0::Bool
```
ChrisRackauckas: I'm not sure this is actually necessary since it's not used for anything.

ChrisRackauckas:
This looks good, but we need to change the other solver packages as well to keep uniformity in the common interface.

src/integrators/type.jl
ranocha: I'll remove it.

ranocha:
Is this list of packages that have to be updated complete?

- [ ] https://github.com/JuliaDiffEq/StochasticDiffEq.jl
- [ ] https://github.com/JuliaDiffEq/DelayDiffEq.jl
- [ ] https://github.com/JuliaDiffEq/DASKR.jl
- [ ] https://github.com/JuliaDiffEq/ODE.jl
- [ ] https://github.com/JuliaDiffEq/ODEInterfaceDiffEq.jl
- [ ] https://github.com/JuliaDiffEq/Sundials.jl

ChrisRackauckas:
I think DelayDiffEq might be fine. GeometricIntegratorsDiffEq.jl and BridgeDiffEq.jl should get an update though tests are failing, and LSODA.jl should update.

ranocha:
Okay. I've missed LSODA.jl because it's not part of JuliaDiffEq.

ranocha:
I can't really see a possibility for aliasing for https://github.com/JuliaDiffEq/BridgeDiffEq.jl.

ChrisRackauckas:
Yes that just uses small immutables.

devmotion:
> I think DelayDiffEq might be fine.

Maybe it would be more intuitive to never alias the `ODEIntegrator` that is created in DelayDiffEq but depending on the keyword `alias_u0` change the `DDEIntegrator` in https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/solve.jl#L114-L117?

ChrisRackauckas:
True that would make sense.

PR #656, opened 2019-02-09
Fix vec in oop SDIRK

devmotion:
Fix test errors https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/487827113 due to non-existing method `vec(::Float64)`.


ChrisRackauckas:
The package manager broke, so test won't run until that's fixed.

PR #659, opened 2019-02-12
IRKC

deeepeshthakur:
@ChrisRackauckas @YingboMa Tried Implementing IRKC from [IRKC: An IMEX solver for stiff diffusion–reaction PDEs](https://www.sciencedirect.com/science/article/pii/S037704270500600X)


src/rkc_utils.jl
```diff
@@ -70,7 +79,7 @@ function maxeig!(integrator, cache::OrdinaryDiffEqConstantCache)
       z = uprev + quot*tmp
     else
       # An arbitrary change on `z`
-      cache.z = shuffle(cache.z)
```
deeepeshthakur: We have used this `maxeig!` for ROCK2, ROCK4, RKC. The `cache` here is `ConstantCache` and it doesn't have any field `z`, thus I'm getting an error. I tried changing `cache.z` to `z`, then I get an error that `shuffle not defined`.

src/integrators/integrator_utils.jl
```diff
@@ -253,7 +253,7 @@ function stepsize_controller!(integrator,alg::PredictiveControllerAlgs)
     q = inv(integrator.opts.qmax)
   else
     gamma = integrator.opts.gamma
-    if alg isa RKC
+    if alg isa Union{RKC,IRKC}
```
ChrisRackauckas: `<:`, not `isa`

src/rkc_utils.jl
```diff
@@ -1,15 +1,19 @@
 # This function calculates the largest eigenvalue
 # (absolute value wise) by power iteration.
-
+const RKCAlgs = Union{RKC,IRKC}
 function maxeig!(integrator, cache::OrdinaryDiffEqConstantCache)
   isfirst = integrator.iter == 1 || integrator.u_modified
   @unpack t, dt, uprev, u, f, p, fsalfirst = integrator
   maxiter = 50
-  safe = (integrator.alg isa RKC) ? 1.0 : 1.2
+  safe = (integrator.alg isa RKCAlgs) ? 1.0 : 1.2
```
ChrisRackauckas: <:

src/rkc_utils.jl
```diff
@@ -1,15 +1,19 @@
...
   # Initial guess for eigenvector `z`
   if isfirst
-    if integrator.alg isa RKC
-      z = fsalfirst
+    if integrator.alg isa RKCAlgs
```
ChrisRackauckas: <:

src/rkc_utils.jl
ChrisRackauckas: that's for sure going to break the other algorithms @YingboMa

src/integrators/integrator_utils.jl
deeepeshthakur: @ChrisRackauckas 
getting this error when changing this
```
ERROR: TypeError: in <:, expected Type, got 
IRKC{0,true,LinSolveFactorize{typeof(LinearAlgebra.lu!)},
NLNewton{true,OrdinaryDiffEq.NLNewtonCache{Nothing,Nothing,Nothing,
Nothing,Nothing,Nothing}},DataType,Nothing,Nothing,Float64,Int64,:Predictive}
```

src/rkc_utils.jl
deeepeshthakur: @ChrisRackauckas 
getting this error when changing this
```
ERROR: TypeError: in <:, expected Type, got 
IRKC{0,true,LinSolveFactorize{typeof(LinearAlgebra.lu!)},
NLNewton{true,OrdinaryDiffEq.NLNewtonCache{Nothing,Nothing,Nothing,
Nothing,Nothing,Nothing}},DataType,Nothing,Nothing,Float64,Int64,:Predictive}
```

ChrisRackauckas: Yes. Need to typeof

src/perform_step/rkc_perform_step.jl
```diff
@@ -460,3 +460,260 @@ end
...
+  maxeig!(integrator, cache)
+  nlsolve!, nlcache = nlsolve, nlsolve.cache
+
+  # The the number of degree for Chebyshev polynomial
+  maxm = max(2,Int(floor(sqrt(integrator.opts.internalnorm(integrator.opts.reltol,t)/(10.0*eps(integrator.opts.internalnorm(uprev,t)))))))
+  mdeg = 1 + Int(floor(sqrt(1.54*dt*integrator.eigen_est + 1.0)))
+  if mdeg >= maxm
+    mdeg = maxm
+  end
+  mdeg = (mdeg < 50) ? 50 : mdeg
```
deeepeshthakur: @ChrisRackauckas I increased the number of stages manually by setting a minimum, and the answer is getting close to the exact answer. I think the `integrator.eigen_est` is being under calculated.

ChrisRackauckas: good investigating. eigen_est is a very crude measure. Is there a safety factor on there to tweak?

src/rkc_utils.jl
YingboMa: Change it to `Base.shuffle`?

src/perform_step/rkc_perform_step.jl
YingboMa: It would be great if we have an Arnoldi based algorithm.

src/rkc_utils.jl
deeepeshthakur: I made these changes as a temporary alternative
```
nind = length(uprev)
if (nind != 1)
     ind = 1 + iter % nind
     z[ind] = uprev[ind] - (z[ind] - uprev[ind])
else
     z = -z
end
```

src/perform_step/rkc_perform_step.jl
deeepeshthakur: @ChrisRackauckas I don't think so! The problem is arising because the `eigen_est` is supposed to be calculated only for the non stiff part or `f.f2` in the `SplitODEFunction`, I tried to make those changes but It's not working.

ChrisRackauckas: Ahh yes, do a `if typeof(integrator.f) <: SplitFunction` for that.

deeepeshthakur:
@YingboMa You used this setup to test power iteration for RKC methods both for `inplace` and `outofplace` versions. 
```
  for i in 1:10, iip in [true, false], alg in [ROCK2(), ROCK4(), RKC()]
    A = randn(20,20)
    test_f(u,p,t) = A*u
    test_f(du,u,p,t) = mul!(du, A, u)
    prob = ODEProblem{iip}(test_f, randn(20), (0,1.))
    integrator = init(prob, alg)
    eigm = maximum(abs.(eigvals(A)))
    maxeig!(integrator, integrator.cache)
    eigest = integrator.eigen_est
    @test eigest ≈ eigm rtol=0.1eigm
  end
````
This is working but I tried a similar style and it's not. I don't understand why?
```
  for i in 1:10, iip in [true, false], alg in [IRKC()]
    A = randn(20,20)
    B = randn(20,20)
    test_f1 = (u,p,t) -> A*u
    test_f2 = (u,p,t) -> B*u
    test_f1 = (du,u,p,t) -> du .= A*u
    test_f2 = (du,u,p,t) -> du .= B*u
    ff_split = SplitFunction(test_f1,test_f2)
    prob = SplitODEProblem{iip}(ff_split, randn(20,1), (0.0,1.))
    integrator = init(prob, alg)
    eigm = maximum(abs.(eigvals(A)))
    maxeig!(integrator, integrator.cache)
    eigest = integrator.eigen_est
    @test eigest ≈ eigm rtol=0.1eigm
  end
```

deeepeshthakur:
@ChrisRackauckas I made all the necessary corrections I could think of. If this passes the CIs, it should be ready to merge.

deeepeshthakur:
@ChrisRackauckas I'm getting an `UndefVarError` and I didn't make any change in that file.

PR #661, opened 2019-02-13
Cache LU factorization for oop solvers

devmotion:
Naive approach to cache LU factorization of `W` for oop solvers if it's not a scalar.

### DIFF
--- a/src/nlsolve/utils.jl
+++ b/src/nlsolve/utils.jl
@@ -117,7 +117,7 @@ DiffEqBase.@def oopnlsolve begin
           J = jacobian_finitediff(uf, uprev, alg.diff_type)
         end
       end
-      W = J isa Number ? J : lu(J)
+      W = J isa Number ? J : lu(J; check=false)
     end
 
     nlcache = NLNewtonCache(κ,tol,min_iter,max_iter,10000,_nlcache.new_W,z,W,γ,c,ηold,dz,tmp,b,k)
--- a/src/nlsolve/utils.jl
+++ b/src/nlsolve/utils.jl
@@ -108,7 +108,16 @@ DiffEqBase.@def oopnlsolve begin
       end
       W = WOperator(f.mass_matrix, dt, J, false)
     else
-      W = typeof(u) <: Number ? u : Matrix{uEltypeNoUnits}(undef, 0, 0) # uEltype?
+      if DiffEqBase.has_jac(f)
+        J = f.jac(uprev, p, t)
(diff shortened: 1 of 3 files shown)


src/nlsolve/utils.jl
```diff
@@ -108,7 +108,16 @@ DiffEqBase.@def oopnlsolve begin
...
+      if DiffEqBase.has_jac(f)
+        J = f.jac(uprev, p, t)
+      else
+        if alg_autodiff(alg)
+          J = jacobian_autodiff(uf, uprev)
+        else
+          J = jacobian_finitediff(uf, uprev, alg.diff_type)
+        end
+      end
+      W = J isa Number ? J : lu(J; check=false)
```
ChrisRackauckas: why do we want to disable the check?

devmotion: Because of that test failure, I thought: https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/492972826
In this line, we're only interested in obtaining the correct type of `W` (otherwise lines such as `nlcache.W = calc_W!` lead to errors later on) but it seemed to me that the concrete values are not relevant at this point.

YingboMa: Maybe `LU{LinearAlgebra.lutype(eltype(J))}(J, collect(1:size(J, 2)), 0)`?

YingboMa: Or `LU{LinearAlgebra.lutype(eltype(J))}(J, zeros(LinearAlgebra.BlasInt, size(J, 2)), 0)`

devmotion: Would there be any advantage over just disabling the check? To me it seems those specialized constructions might not cover all cases and are more likely to fail with non-standard types?

YingboMa: That would save one lu factorization to compute. For now, LU factorization doesn't handle units anyway. I don't see how it would fail for non-standard types.

devmotion: I mean failing in the sense that it does not yield exactly the same type as `lu(J)`. For instance, if someone/some package implements `lu` for a non-standard array type, `lu` might not give you exactly the same type as `LU{....}(...)`?

YingboMa: Right. I agree.

ChrisRackauckas:
I guess that's fine, but we really shouldn't be doing an extra lu if we don't have to.

PR #663, opened 2019-02-15
Implementation of 2R+ schemes from Carpenter Kennedy Lewis(2000)

deeepeshthakur:
@ChrisRackauckas @ranocha I tried implementing LDDRK methods from [Kennedy, Carpenter, Lewis (2000)](https://www.sciencedirect.com/science/article/pii/S0168927499001415). I implemented a generalized perform step for van der Houwen 2R+ scheme used in this paper and one test algorithm which follows this scheme. I followed the naming scheme `[author's name]LDDRK[stage][order]_[embedded_order for error estimation]`. It looks good. If this generalized perform step works fine, then I can add the constants and tests for other similar schemes. 
I've added CKLLDDRK43_2 as a test for the perform_step! functions.
@ranocha I'm not able to figure out what will be the right tolerences for some tests. So presently I've just assigned the minimum with which they pass those tests.
@ChrisRackauckas Also it's failing one of the CIs
```
alg = CKLLDDRK43_2
dts = 1 ./ 2 .^(8:-1:4)
sim = test_convergence(dts, test_problem_linear[1], alg)
Test Failed at REPL[28]:3
  Expression: ≈(sim.𝒪est[:final], OrdinaryDiffEq.alg_order(alg), atol=testTol)
   Evaluated: 3.9825784370847237 ≈ 3 (atol=0.25) 
plot(sim)
```
but because I'm getting a pretty decent plot, I used `@test_broken` for this one.
image
The convergence plots for the other two test are:
```
alg = CKLLDDRK43_2
dts = 1 ./ 2 .^(8:-1:4)
sim = test_convergence(dts, test_problem_only_time[1], alg)
plot(sim)
```
image

```
alg = CKLLDDRK43_2
dts = 1 ./ 2 .^(8:-1:4)
sim = test_convergence(dts, test_problem_nonlinear[1], alg)
plot(sim)
```
image

The methods that will use this perform_step are:
1. CKLLDDRK43_2
1. CKLLDDRK54_3
1. CKLLDDRK95_4C (linear stability–error compromise)
1. CKLLDDRK95_4S (maximum linear stability)
1. CKLLDDRK95_4M (minimum truncation error)

Similar functions can be implemented for 3R+ and 4R+ versions
#493


test/ode/ode_low_storage_rk_tests.jl
```diff
@@ -312,6 +312,31 @@ sol_new = solve(new_prob_ode_nonlinear_inplace, alg, dt=1.e-4, save_everystep=fa
...
+  @test_broken sim.𝒪est[:final] ≈ OrdinaryDiffEq.alg_order(alg) atol=testTol
+end
+for prob in test_problems_nonlinear
+  sim = test_convergence(dts, prob, alg)
+  @test sim.𝒪est[:final] ≈ OrdinaryDiffEq.alg_order(alg) atol=testTol
+end
+integ = init(prob_ode_large, alg, dt=1.e-2, save_start=false, save_end=false, save_everystep=false)
+@test Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 7
+integ = init(prob_ode_large, alg, dt=1.e-2, save_start=false, save_end=false, save_everystep=false, alias_u0=true)
+@test Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 7
```
deeepeshthakur: @ranocha I was talking about these constants. I'm not able to figure out the optimum value

test/ode/ode_dense_tests.jl
```diff
@@ -211,6 +211,9 @@ regression_test(CFRLDDRK64(), 3.0e-5, 3.0e-5)
 # TSLDDRK74
 regression_test(TSLDDRK74(), 3.0e-5, 3.0e-5)
 
+# CKLLDDRK43_2
+regression_test(CKLLDDRK43_2(), 7.0e-5,3.5e-5)
```
deeepeshthakur: @ranocha Here I have chosen the minimum tolerances so that the method passes the tests. Could you suggest some more accurate values?

test/ode/ode_cache_tests.jl
```diff
@@ -4,6 +4,7 @@ CACHE_TEST_ALGS = [Euler(),Midpoint(),RK4(),SSPRK22(),SSPRK33(),SSPRK53(), SSPRK
   SSPRK63(),SSPRK73(),SSPRK83(),SSPRK432(),SSPRK932(),SSPRK54(),SSPRK104(),
   ORK256(), CarpenterKennedy2N54(), HSLDDRK64(), DGLDDRK73_C(), DGLDDRK84_C(), DGLDDRK84_F(), NDBLSRK124(), NDBLSRK134(), NDBLSRK144(),
   CFRLDDRK64(), TSLDDRK74(),
+  CKLLDDRK43_2(),
```
deeepeshthakur: ` @test_nowarn sol = solve(prob,alg,callback=callback,dt=1/2)`
This is reporting `ERROR: BoundsError` for `CKLLDDRK43_2`.

src/caches/low_storage_rk_caches.jl
```diff
@@ -1371,3 +1371,71 @@ end
...
+  B̂2 = convert(T, 0.09091523008632837)   #514528521746//5659431552419)
+  B̂3 = convert(T, 2.866496742725443)     #27030193851939//9429696342944)
+  B̂ᵢ = SVector(B̂1, B̂2, B̂3)
+
+  Bₗ = convert(T, -2.681109033041384)    #-101169746363290//37734290219643)
+  B̂ₗ = convert(T, -2.298093456892615)    #-69544964788955//30262026368149)
+
+  C1 = convert(T2, 0.32416573882874605)   #1017324711453//9774461848756)
+  C2 = convert(T2, 0.5570978645055429 + 0.10407986927510238)    #8237718856693//13685301971492)
+  C3 = convert(T2, -0.08605491431272755 + 0.10407986927510238 + 0.6019391368822611)    #57731312506979//19404895981398)
```
deeepeshthakur: @ChrisRackauckas The values of `c` are not explicitly mentioned in the paper so I had to calculate using Buther-Table.

ranocha:
Thanks for you effort, @deeepeshthakur! I will have a look at the implementation etc. soon.

ranocha:
```
alg = CKLLDDRK43_2
dts = 1 ./ 2 .^(8:-1:4)
sim = test_convergence(dts, test_problem_linear[1], alg)
Test Failed at REPL[28]:3
  Expression: ≈(sim.𝒪est[:final], OrdinaryDiffEq.alg_order(alg), atol=testTol)
   Evaluated: 3.9825784370847237 ≈ 3 (atol=0.25) 
plot(sim)
```
That's fine because the scheme has a linear order of 4, as described in section 3.3 of the article (I guess you've implemented RK3(2)4[2R+]C, correct?). You should test `@test sim.𝒪est[:final] ≈ OrdinaryDiffEq.alg_order(alg)+1 atol=testTol` and add a corresponding comment.

test/ode/ode_low_storage_rk_tests.jl
ranocha: - The schemes are 2R+ schemes, so they should have a storage of `2` for `adaptive==false` and `3` for `adaptive==true`. In order to enable less space for `adaptive==false`, you have to opt-in to this optimisation via  
 https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/d155eb474f5569fdfdcccb5e02583f9d28db86cb/src/alg_utils.jl#L416-L428
  This is done in 
 https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/d155eb474f5569fdfdcccb5e02583f9d28db86cb/src/solve.jl#L222-L228
  For these new algorithms, we have to change `uses_uprev(alg::OrdinaryDiffEqAlgorithm)` to `uses_uprev(alg::OrdinaryDiffEqAlgorithm, adaptive::Bool)`, I think.
- We have one additional storage because the storage optimised assignments of the form `u1 = f(t, u1)` are not implemented yet.
- If `alias_u0==false` (default), there is one additional storage place for the initial condition. However, setting `alias_u0=true` should reduce the required storage.

To sum up: It seems that the current implementation uses a bit more storage that necessary and we need to change some of the internal functions. I need to have a deeper look at the implementation in order to see whether there are some optimisations.

test/ode/ode_dense_tests.jl
ranocha: No, that approach is just fine. The tests in this file are some regression tests, so you should choose the tolerances exactly in the way you did - implement the algorithm, test it carefully, and choose the tolerances such that they are only a little bit bigger than absolutely necessary to let the tests pass.

src/caches/low_storage_rk_caches.jl
```diff
@@ -1371,3 +1371,71 @@ end
...
+  Bₗ::T
+  B̂ₗ::T
+  Bᵢ::SVector{N,T}
+  B̂ᵢ::SVector{N,T}
+  Cᵢ::SVector{N,T2}
+end
+
+
+function CKLLDDRK43_2ConstantCache(::Type{T},::Type{T2}) where {T,T2}
+  A1 = convert(T, 0.32416573882874605)   #11847461282814//36547543011857
```
ranocha: I am not completely sure how many of the rational coefficients given in that article are exact solutions of the order conditions. Nevertheless, you should implement the rational forms, I think. They should at least provide some additional accuracy compared to the `Float64` forms.

test/ode/ode_cache_tests.jl
ranocha: I think there is some problem with the cache, some variables are not `resize`d correctly. But @ChrisRackauckas will definitively know more about that.

src/perform_step/low_storage_rk_perform_step.jl
```diff
@@ -182,3 +182,90 @@ end
...
+  @. k  = fsalfirst
+  @. tmp = zero(u)
+
+  #stages 1 to s-1
+  for i in eachindex(A₁)
+    @. tmp = tmp + (Bᵢ[i] - B̂ᵢ[i])*dt*k
+    @. u   = Xᵢ + A₁[i]*dt*k
+    @. Xᵢ  = u + (Bᵢ[i] - A₁[i])*dt*k
+    f(k, u, p, t + Cᵢ[i]*dt)
+  end
```
ranocha: This part is a bit confusing for me because
- `A₁[i]` is added and subtracted again to get `Xᵢ` (which might add additional numerical errors).
- `Xᵢ` seems to store what will later be `u` while `u` stores intermediate values.

I know, the description in the artcile of Kennedy et al. is also a bit confusing because of the swapping of the registers. But there are nicer (for me) descriptions of these schemes, e.g. Algorithm 2 of [Ketcheson (2010)](http://www.davidketcheson.info/assets/papers/2010_LSRK_postprint.pdf). The adaptive part is missing there but can be added similar to what you did in this loop.

src/perform_step/low_storage_rk_perform_step.jl
```diff
@@ -182,3 +182,90 @@ end
...
+
+  #last stage
+  @. tmp = tmp + (Bₗ - B̂ₗ)*dt*k
+  @. u   = Xᵢ  + Bₗ*dt*k
+
+  #Error estimate
```
ranocha: If we want to reduce the memory requirements (and number of operations), the computation of the error estimator above (what is now called `tmp`) should be avoided if `adaptive==false`. In that case, the additional array could also be aliased.

ranocha:
@ChrisRackauckas should also have a look. As described in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/493#issuecomment-461000414, using two steps (1. correct and verified implementation, 2. optimisation) is a good approach.

src/caches/low_storage_rk_caches.jl
ChrisRackauckas: They are definitely helpful if anyone goes to higher order, since otherwise the float64 coefficients will just have accuracy dropoff

ChrisRackauckas: while sometimes not practical, that can be helpful in internal benchmarks and testing.

test/ode/ode_cache_tests.jl
ChrisRackauckas: Yeah, this is for testing how resize works, so you'd have to build based on the macro assumptions. I'll see I can spot the failure cause without digging in.

test/ode/ode_dense_tests.jl
ChrisRackauckas: Yup, these regression tests are just checking for if the interpolation goes wonky some time in the future, for example if someone accidentally aliases `k` when `calck=true`.

src/caches/low_storage_rk_caches.jl
```diff
@@ -1371,3 +1371,71 @@ end
...
   ParsaniKetchesonDeconinck3S205ConstantCache(real(uBottomEltypeNoUnits),real(tTypeNoUnits))
 end
+
+
+
+# 2R+ low storage methods introduced by van der Hauwen
+@cache struct LowStorageRK2RPCache{uType,rateType,uNoUnitsType,TabType} <: OrdinaryDiffEqMutableCache
+  u::uType
+  k::rateType
+  Xᵢ::rateType
```
ChrisRackauckas: this name is causing the resize test to fail, because `@cache` is looking for the standard names IIRC.

ranocha: Why is `Xᵢ::rateType`? In the step, it is used as `uType`, e.g. in `Xᵢ  = u + (Bᵢ[i] - A₁[i])*dt*k`.

src/caches/low_storage_rk_caches.jl
deeepeshthakur: The problem with using rational coefficients is that the values of `c` is not specifically mentioned, so I had to calculate `c` values by adding up row elements in the butcher table. But when adding rational coefficients, the numbers go beyond `Int64` range. So I converted all of them to `Float64`. and mentioned their rational values next to them.

src/caches/low_storage_rk_caches.jl
deeepeshthakur: @ChrisRackauckas Which names? `LowStorageRK2RPCache`?

deeepeshthakur: > Why is `Xᵢ::rateType`? In the step, it is used as `uType`, e.g. in `Xᵢ = u + (Bᵢ[i] - A₁[i])*dt*k`.

Oh I previously used it as rateType. I'll change it.

src/caches/low_storage_rk_caches.jl
ranocha: You could use bigger integer values, e.g. `Int128(11847461282814)//36547543011857`. Does that help?

test/ode/ode_low_storage_rk_tests.jl
deeepeshthakur: @ranocha I made the changes you suggested,
```
@test Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 5
@test Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 4
```
Now the tests work for these parameters

deeepeshthakur: I wrote a new uses_uprev function. But after this function 
```
@test Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 6
@test Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 5
```
memory usage ratio is 6 and 5

src/caches/low_storage_rk_caches.jl
deeepeshthakur: @ChrisRackauckas I changed the names of `Xᵢ` to `X` and others respectively still, I'm getting the same error.

deeepeshthakur:
@ranocha @ChrisRackauckas I think I've incorporated all the suggestions. The only issue now is the `BoundsError`

src/algorithms.jl
```diff
@@ -60,6 +60,7 @@ struct NDBLSRK134 <: OrdinaryDiffEqAlgorithm end
 struct NDBLSRK144 <: OrdinaryDiffEqAlgorithm end
 struct CFRLDDRK64 <: OrdinaryDiffEqAlgorithm end
 struct TSLDDRK74 <: OrdinaryDiffEqAlgorithm end
+struct CKLLDDRK43_2 <: OrdinaryDiffEqAdaptiveAlgorithm end
```
ranocha: I still think that this algorithm shouldn't be `CKLLDDRK43_2` because it is not LDD.

src/caches/low_storage_rk_caches.jl
```diff
@@ -1371,3 +1371,70 @@ end
 function alg_cache(alg::ParsaniKetchesonDeconinck3S205,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
   ParsaniKetchesonDeconinck3S205ConstantCache(real(uBottomEltypeNoUnits),real(tTypeNoUnits))
 end
+
+
+
+# 2R+ low storage methods introduced by van der Hauwen
```
ranocha: van der Houwen

src/perform_step/low_storage_rk_perform_step.jl
```diff
@@ -182,3 +182,86 @@ end
...
+  @unpack A,Bl,Bhl,B,Bh,C = cache.tab
+
+  @. k   = fsalfirst
+  integrator.opts.adaptive && (@. tmp = zero(uprev))
+
+  #stages 1 to s-1
+  for i in eachindex(A)
+    integrator.opts.adaptive && (@. tmp = tmp + (B[i] - Bh[i])*dt*k)
+    @. u = u + B[i]*dt*k
+    f(k, u + (A[i] - B[i])*dt*k, p, t + C[i]*dt)
```
ranocha: This line will create allocations during each call because of `u + (A[i] - B[i])*dt*k`. You should again pre-allocate some storage for that part.

test/ode/ode_low_storage_rk_tests.jl
```diff
@@ -312,6 +312,31 @@ sol_new = solve(new_prob_ode_nonlinear_inplace, alg, dt=1.e-4, save_everystep=fa
...
+end
+for prob in test_problems_linear
+  sim = test_convergence(dts, prob, alg)
+  @test sim.𝒪est[:final] ≈ OrdinaryDiffEq.alg_order(alg)+1 atol=testTol
+end
+for prob in test_problems_nonlinear
+  sim = test_convergence(dts, prob, alg)
+  @test sim.𝒪est[:final] ≈ OrdinaryDiffEq.alg_order(alg) atol=testTol
+end
+integ = init(prob_ode_large, alg, dt=1.e-2, save_start=false, save_end=false, save_everystep=false)
```
ranocha: You should also add `adaptive=false` if you want to get the minimal number of allocations. It might be good to test both cases (`adaptive=false` and `adaptive=true`). For `adaptive=false`, the storage requirement should be smaller.

src/algorithms.jl
deeepeshthakur: I think it is a low storage algorithm but I'm not exactly sure about the dispersion and dissipation part, although I implemented it assuming it was an `LDDRK`

src/perform_step/low_storage_rk_perform_step.jl
deeepeshthakur: But won't the compiler do exactly the same thing?

deeepeshthakur: Making a separate allocation for this increases the memory allocations.
```
integ = init(prob_ode_large, alg,dt=1.e-2, save_start=false, save_end=false, save_everystep=false)
  Expression: Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 6
   Evaluated: 7 <= 6

integ = init(prob_ode_large, alg, dt=1.e-2, save_start=false, save_end=false, save_everystep=false, alias_u0=true)
  Expression: Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 5
   Evaluated: 6 <= 5
```
Earlier these worked for 6 and 5 respectively

deeepeshthakur:
@ChrisRackauckas @ranocha 
I think the only things remaining are 
- [ ] `BoundsError` in the `cache_test`
- [ ] Name

src/algorithms.jl
ranocha: In their article, Carpenter, Kennedy and Lewis derive algorithms for the Navier-Stokes equation. Thus, they try to find a reasonable balance between different properties. Nevertheless, their `C` schemes such as RK3(2)4[2R+]C are optimised for error efficiency, not for dispersion & dissipation.

src/perform_step/low_storage_rk_perform_step.jl
ranocha: > But won't the compiler do exactly the same thing?

No. If you pre-allocate storage for that expression and store it in the cache, the allocation will only happen once during initialisation and the cache can be reused for every iteration. In its current state, the allocation will happen for every stage of every step. This will likely be slower.

deeepeshthakur:
@ranocha 
```
affect! = function (integrator)
  u = integrator.u
  resize!(integrator,length(u)+1)
  maxidx = findmax(u)[2]
  Θ = rand()
  u[maxidx] = Θ
  u[end] = 1-Θ
  nothing
end
```
Here we are resizing `integrator.u`, so will `integrator.uprev` be resized automatically ?

deeepeshthakur:
@ChrisRackauckas 
Added `CKLLS54_3C`
**CI Plots are:**
```
alg = CKLLS54_3C()
dts = 1 ./ 2 .^(8:-1:4)
prob = test_problems_only_time[1]
sim = test_convergence(dts, prob, alg)
plot(sim)
```
for this test `Evaluated: 1.0156948852266512 ≈ 4 (atol=0.25) ` but because the plot is linear so `@test sim.𝒪est[:final] ≈ OrdinaryDiffEq.alg_order(alg)-3 atol=testTol`
image

```
alg = CKLLS54_3C()
dts = 1 ./ 2 .^(8:-1:4)
prob = test_problems_linear[1]
sim = test_convergence(dts, prob, alg)
plot(sim)
```
image

```
alg = CKLLS54_3C()
dts = 1 ./ 2 .^(8:-1:4)
prob = test_problems_nonlinear[1]
sim = test_convergence(dts, prob, alg)
plot(sim)
```
image

deeepeshthakur:
@ChrisRackauckas @ranocha 
Added `CKLLSRK95_4S'
All the plots are linear but if you could re-check the order evaluation of first group of CIs.
```
alg = CKLLSRK95_4S()
dts = 1 ./ 2 .^(6:-1:2)
for prob in test_problems_only_time
  sim = test_convergence(dts, prob, alg)
  @test sim.𝒪est[:final] ≈ 1 atol=testTol
end
for prob in test_problems_linear
  sim = test_convergence(dts, prob, alg)
end
for prob in test_problems_nonlinear
  sim = test_convergence(dts, prob, alg)
end
```

deeepeshthakur:
@ChrisRackauckas @ranocha
Added `CKLLSRK95_4C'
All the plots are linear but the evaluated orders are coming out to be 1.
```
alg = CKLLSRK95_4C()
dts = 1 ./ 2 .^(8:-1:4)
for prob in test_problems_only_time
  sim = test_convergence(dts, prob, alg)
  @test sim.𝒪est[:final] ≈ 1 atol=testTol
end
for prob in test_problems_linear
  sim = test_convergence(dts, prob, alg)
  @test sim.𝒪est[:final] ≈ 1 atol=testTol
end
for prob in test_problems_nonlinear
  sim = test_convergence(dts, prob, alg)
  @test sim.𝒪est[:final] ≈ 1 atol=testTol
end
```
Also the `regression test` tolerences are very high as compared to the others
`regression_test(CKLLSRK95_4C(), 3.0e-3,5.5e-3)`

src/caches/low_storage_rk_caches.jl
```diff
@@ -1559,3 +1559,71 @@ end
 function alg_cache(alg::CKLLSRK95_4S,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
   CKLLSRK95_4SConstantCache(real(uBottomEltypeNoUnits),real(tTypeNoUnits))
 end
+
+
+function CKLLSRK95_4CConstantCache(::Type{T},::Type{T2}) where {T,T2}
+  A1  = convert(T, BigInt(1107026461565)//BigInt(5417078080134))
```
ranocha: This line should be deleted, shouldn't it?

ranocha:
When the experimental order of convergence is only 1, there seems to be some error in the coefficients. If I were you, I would a) check the coefficients in the paper and the implementation again and b) check the Runge-Kutta order conditions for the implemented coefficients.

deeepeshthakur:
@ranocha 
**These are the nine order conditions given for 5th order ERK**
All these should be 0.
```
τ1 = -6.310129403899615e-27
τ2 = -0.016972398104717748
τ3 = -0.011504011646363872
τ4 = -0.023894427685531305
τ5 = 0.19984713428097445
τ6 = -0.02168354204697044
τ7 = -0.013273852998868298
τ8 = -0.011735848502595453
τ9 = -0.0008873752338293298
τ10 = -0.009374593543047193
τ11 = -0.006631771897826025
τ12 = -0.011014351071663089
τ13 = -0.00450337741443769
τ14 = -0.010195137781596668
τ15 = -0.011265645427779447
τ16 = -0.0042194890046816985
τ17 = -0.0029865632119084238
```
BTW `c` values were not specifically mentioned so I calculated them using
```
C1 = A1
C2 = A2 + B1
C3 = A3 + B2 + B2
C4 = A4 + B3 + B3 + B3
C5 = A5 + B4 + B4 + B4 + B4
C6 = A6 + B5 + B5 + B5 + B5 + B5
C7 = A7 + B6 + B6 + B6 + B6 + B6 + B6
C8 = A8 + B7 + B7 + B7 + B7 + B7 + B7 + B7
C   = [0//1,C1,C2,C3,C4,C5,C6,C7,C8]
```
and the above parameters are calculated using these `c` values
This Buther Array is given in the paper:
image

ranocha:
So there's definitively something wrong with these coefficients, which explains the bad experimental order of convergence.

src/caches/low_storage_rk_caches.jl
```diff
@@ -1371,3 +1371,259 @@ end
...
+  B̂5  = convert(T, BigInt(1105885670474)//BigInt(4964345317203))
+  B̂6  = convert(T, BigInt(1408484642121)//BigInt(8758221613943))
+  B̂7  = convert(T, BigInt(1454774750537)//BigInt(11112645198328))
+  B̂8  = convert(T, BigInt(772137014323)//BigInt(4386814405182))
+  B̂ᵢ = SVector(B̂1, B̂2, B̂3, B̂4, B̂5, B̂6, B̂7, B̂8)
+
+  Bₗ = convert(T, BigInt(3559252274877)//BigInt(14424734981077))
+  B̂ₗ = convert(T, BigInt(277420604269)//BigInt(1857595682219))
+
+  C1  = convert(T2, BigInt(1107026461565)//BigInt(1107026461565))                                                                                                                                                                                   # A1
```
ranocha: These coefficients are all one.

ranocha: @deeepeshthakur Could you please re-calculate the coefficients `C1...C8` and test whether the experimental order of convergence/ order conditions are better?

deeepeshthakur: Getting Order 5 convergence in all CIs. Thank You. I messed up the regex.

deeepeshthakur:
@ChrisRackauckas @ranocha
Added `CKLLSRK95_4M'
All CI plots are linear and have order of convergence 5.

```
alg = CKLLSRK95_4M()
dts = 1 ./ 2 .^(6:-1:2)
for prob in test_problems_only_time
  sim = test_convergence(dts, prob, alg)
end
for prob in test_problems_linear
  sim = test_convergence(dts, prob, alg)
end
for prob in test_problems_nonlinear
  sim = test_convergence(dts, prob, alg)
end
```
This is the last of 2R+ schemes. There is something wrong with the coefficients of `CKLLSRK95_4C', the order is coming out to be 1 instead of 5, so I've marked it's tests as broken.

src/perform_step/low_storage_rk_perform_step.jl
ChrisRackauckas: This hasn't been addressed yet?

deeepeshthakur: It's done as directed. I don't know why it's not showing outdated.

src/perform_step/low_storage_rk_perform_step.jl
ChrisRackauckas: I meant to say that this one hasn't been done yet.

deeepeshthakur: It's Done!

PR #665, opened 2019-02-17
Fix early stopping condition

devmotion:
It seems the early stopping condition is incorrect due to [short-circuit evaluation](https://docs.julialang.org/en/v1.1/manual/control-flow/#Short-Circuit-Evaluation-1):
```julia
julia> false || true && println("Hello world")
Hello world

julia> true || false && println("Hello world")
true
```


ChrisRackauckas:
Does the Newton method have the same issue?

devmotion:
No, there it is correct.

PR #667, opened 2019-02-21
Add DEStat

YingboMa:
```julia
julia> using OrdinaryDiffEq

julia> prob = ODEProblem((du,u,p,t)->@.(du=0.01*u), ones(100), (0,100.));

julia> solve(prob, Tsit5()).destats
DiffEqBase.DEStats
Number of function evaluations: 39
Number of W matrix evaluations: 0
Number of linear solves:        0
Number of accepted steps:       6
Number of rejected steps:       0

julia> solve(prob, AutoTsit5(Rosenbrock23())).destats
DiffEqBase.DEStats
Number of function evaluations: 39
Number of W matrix evaluations: 0
Number of linear solves:        0
Number of accepted steps:       6
Number of rejected steps:       0
Maximum eigenvalue recorded:    0.010000000023693143

julia> solve(prob, ImplicitEuler()).destats
DiffEqBase.DEStats
Number of function evaluations: 265
Number of W matrix evaluations: 64
Number of linear solves:        72
Number of accepted steps:       39
Number of rejected steps:       25
```

### DIFF
--- a/src/nlsolve/newton.jl
+++ b/src/nlsolve/newton.jl
@@ -54,7 +54,7 @@ Equations II, Springer Series in Computational Mathematics. ISBN
   else
     dz = _reshape(W \ _vec(ztmp), axes(ztmp))
   end
-  integrator.destat.nsolve += 1
+  integrator.destats.nsolve += 1
 
   # compute initial values for early stopping criterion
   ndz = integrator.opts.internalnorm(dz,t)
@@ -81,7 +81,7 @@ Equations II, Springer Series in Computational Mathematics. ISBN
     else
       dz = _reshape(W \ _vec(ztmp), axes(ztmp))
     end
-    integrator.destat.nsolve += 1
+    integrator.destats.nsolve += 1
 
     # check early stopping criterion
     ndzprev = ndz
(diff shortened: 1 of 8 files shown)


ChrisRackauckas:
I don't think this counts the linear solves in a Rosenbrock method?

How is the number of function evaluations for AutoTsit5 so high there? It seems like it doesn't reset each time.

ChrisRackauckas:
Fails on ParameterizedFunctions, that's tricky.

src/nlsolve/newton.jl
```diff
@@ -54,6 +54,7 @@ Equations II, Springer Series in Computational Mathematics. ISBN
   else
     dz = _reshape(W \ _vec(ztmp), axes(ztmp))
   end
+  integrator.destats.nsolve += 1
```
kanav99: This is currently incrementing the linear solve counter on non-linear solves. I guess we should have two counters - each for linear and nonlinear.

kanav99: Oh my bad, still this doesn't counts the linear solves in the Rosenbrocks.

ChrisRackauckas: yeah, we can keep improving this

PR #672, opened 2019-02-24
Do not evaluate Jacobian nor compute LU factorization in oopnlsolve

YingboMa:
This PR will fix the CI errors in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/103

@devmotion consider that we don't even have unit aware stiff solvers. Trying to handle custom number types with `lu` defined on it seems to be unnecessary.

### DIFF
--- a/src/nlsolve/utils.jl
+++ b/src/nlsolve/utils.jl
@@ -108,16 +108,19 @@ DiffEqBase.@def oopnlsolve begin
       end
       W = WOperator(f.mass_matrix, dt, J, false)
     else
-      if DiffEqBase.has_jac(f)
-        J = f.jac(uprev, p, t)
-      else
-        if alg_autodiff(alg)
-          J = jacobian_autodiff(uf, uprev)
-        else
-          J = jacobian_finitediff(uf, uprev, alg.diff_type)
-        end
-      end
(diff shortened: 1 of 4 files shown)


src/nlsolve/utils.jl
```diff
@@ -108,16 +108,19 @@ DiffEqBase.@def oopnlsolve begin
...
+      #  J = f.jac(uprev, p, t)
+      #else
+      #  if alg_autodiff(alg)
+      #    J = jacobian_autodiff(uf, uprev)
+      #  else
+      #    J = jacobian_finitediff(uf, uprev, alg.diff_type)
+      #  end
+      #end
+      #W = J isa Number ? J : lu(J; check=false)
+      W = u isa Number ? u : LU{LinearAlgebra.lutype(eltype(u))}(Matrix{uEltypeNoUnits}(undef, 0, 0),
```
ChrisRackauckas: how will alternative linsolves and J types work here?

YingboMa: We don't have the option to switch linsolve on OOP solvers, yet.

ChrisRackauckas: oh yeah true. And there's not really a reason for it either.

PR #673, opened 2019-02-25
Restructure non-linear solvers and add updates of QR factorization

devmotion:
Add separate types for the algorithms, caches, and the common non-linear solver. Moreover, implements updates of the QR decomposition in the same way as https://github.com/JuliaNLSolvers/NLsolve.jl/pull/210. Currently still 4 convergence tests fail. The (unmodified) implementation of functional iteration with mass matrices solves not exactly the same fixed-point problem (mathematically it yields the same solution but contraction of the mapping might not be guaranteed anymore, it seems?), which might explain why I had to adjust the mass matrix tests in particular for Anderson acceleration.


src/nlsolve/functional.jl
```diff
@@ -1,154 +1,291 @@
...
+        ldiv!(Rcur, mul!(γscur, Qcur', vec(dz)))
+        integrator.destats.nsolve += 1
+
+        # update next iterate
+        for i in 1:history
+          z = @. z - γs[i] * Δz₊s[i]
+        end
+
+        # update norm of residuals
+        ndz = integrator.opts.internalnorm(z .- z₊ .+ dz, tstep)
```
ChrisRackauckas: do we need a temporary?

devmotion: I thought we don't need one in the out-of-place implementation.

ChrisRackauckas: oh true

ChrisRackauckas:
I like these changes a lot. This is a lot cleaner than before.

YingboMa:
I wonder if there is a way of letting `RadauIIA5` also use the existing nlsolve framework.

ChrisRackauckas:
I am not sure since it's not solving the same nonlinear equation as the others. We can punt on that for later.

ChrisRackauckas:
This needs to incorporate #659

devmotion:
This PR? Or a later fix?

It seems that reverting back to at least two functional iterations (i.e., disabling checking the stopping criterion in the initial step) seems to fix the convergence issues I mentioned above.

ChrisRackauckas:
>It seems that reverting back to at least two functional iterations (i.e., disabling checking the stopping criterion in the initial step) seems to fix the convergence issues I mentioned above.

Most hardcoded versions of functional iterations require at least two iterations. Allowing the user to choose that could be a nice feature, but yes requiring 2 (at least as a default) is sensible.

ChrisRackauckas:
>This PR? Or a later fix?

Just do to IRKC what's done on the other methods.

devmotion:
Tests pass now. I had to change the stopping criterion slightly and added an explicit check whether `iszero(ndz)` in the first iteration, so that even for functional iteration not always two iteration steps are performed.

ChrisRackauckas:
Is this no longer WIP?

devmotion:
No. I'm sure the PR can be improved further but now it contains the desired functionality and is in a working state.

ChrisRackauckas:
Thanks. I'll review the newest changes and try to merge today.

ChrisRackauckas:
Looks like just that one comment.

src/nlsolve/functional.jl
```diff
@@ -73,7 +73,7 @@ Equations II, Springer Series in Computational Mathematics. ISBN
     # check divergence (not in initial step)
     if iter > 1
       θ = ndz / ndzprev
-      if θ ≥ 1 || ndz * θ^(max_iter - iter) > κtol * (1 - θ)
+      if θ > 1 || ndz * θ^(max_iter - iter) > κtol * (1 - θ)
```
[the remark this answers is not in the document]
devmotion: It seemed a bit cleaner to me (I'm not sure how reliably equality tests are with floating point numbers), and the second condition should cover the case `θ = 1` anyway.

PR #676, opened 2019-02-28
Fix https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/674

devmotion:
Additionally I added an explicit test to ensure that for static arrays a static LU decomposition of `W` is created (and hence throughout the integration `W` is of that type). By the way, when adding this test I noticed that finite differencing (i.e., using `sol = solve(ode, ImplicitEuler(autodiff=false))`) does not work because of the assignment in line https://github.com/JuliaDiffEq/DiffEqDiffTools.jl/blob/master/src/jacobians.jl#L147 - is this a regression or just something that should be fixed at some point?

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,13 +2,13 @@ julia 1.0
 DiffEqBase 5.3.2
 DiffEqOperators 3.2.0
 Parameters 0.10.0
-ForwardDiff 0.7.0
+ForwardDiff 0.10.3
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Not a regression and it's known: https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/373 .

PR #678, opened 2019-02-28
Counts number of f evaluations

kanav99:
This counts the number of `f` evals in `integrator.destats.nf`. Had to count this manually, so took quite time.
I don't have nice computing power right now, lets wait for travis to complete the tests.

### DIFF
--- a/src/perform_step/bdf_perform_step.jl
+++ b/src/perform_step/bdf_perform_step.jl
@@ -157,7 +157,7 @@ function initialize!(integrator, cache::SBDFConstantCache)
   integrator.k = typeof(integrator.k)(undef, integrator.kshortsize)
   cache.du₁ = f1(uprev,p,t)
   cache.du₂ = f2(uprev,p,t)
-  integrator.destats.nf += 2
+  integrator.destats.nf += 1
   integrator.fsalfirst = cache.du₁ + cache.du₂
 
   # Avoid undefined entries if k is an array of arrays
@@ -208,7 +208,7 @@ function perform_step!(integrator,cache::SBDFConstantCache,repeat_step=false)
               ( cache.uprev2 = uprev;  cache.k₁ = du₂ )
   cache.du₁ = f1(u, p, t+dt)
   cache.du₂ = f2(u, p, t+dt)
-  integrator.destats.nf += 2
+  integrator.destats.nf += 1
   integrator.fsallast = cache.du₁ + cache.du₂
   integrator.k[1] = integrator.fsalfirst
   integrator.k[2] = integrator.fsallast
@@ -226,7 +226,7 @@ function initialize!(integrator, cache::SBDFCache)
   integrator.k[2] = integrator.fsallast
   f1(cache.du₁, uprev, p, t)
   f2(cache.du₂, uprev, p, t)
-  integrator.destats.nf += 2
+  integrator.destats.nf += 1
   @. integrator.fsalfirst = cache.du₁ + cache.du₂
 end
 
@@ -270,7 +270,7 @@ function perform_step!(integrator, cache::SBDFCache, repeat_step=false)
               ( cache.uprev2 .= uprev;  cache.k₁ .= du₂ )
   f1(du₁, u, p, t+dt)
   f2(du₂, u, p, t+dt)
-  integrator.destats.nf += 2
+  integrator.destats.nf += 1
   @. k = du₁ + du₂
 end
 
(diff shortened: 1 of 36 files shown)


src/perform_step/bdf_perform_step.jl
```diff
@@ -153,6 +157,7 @@ function initialize!(integrator, cache::SBDFConstantCache)
   integrator.k = typeof(integrator.k)(undef, integrator.kshortsize)
   cache.du₁ = f1(uprev,p,t)
   cache.du₂ = f2(uprev,p,t)
+  integrator.destats.nf += 2
```
ChrisRackauckas: I don't think we should double count the IMEX methods.

src/perform_step/split_perform_step.jl
```diff
@@ -15,6 +16,7 @@ end
   @unpack t,dt,uprev,u,f,p = integrator
   u = @. uprev + dt*integrator.fsalfirst
   integrator.fsallast = f.f1(u,p,t+dt) + f.f2(u,p,t+dt)  # For the interpolation, needs k at the updated point
+  integrator.destats.nf += 2
```
ChrisRackauckas: IMEX issue again

src/perform_step/kencarp_kvaerno_perform_step.jl
```diff
@@ -283,6 +289,7 @@ end
   if typeof(integrator.f) <: SplitFunction
     integrator.k[1] = integrator.fsalfirst
     integrator.fsallast = integrator.f(u, p, t+dt)
+    integrator.destats.nf += 1
```
ChrisRackauckas: IMEX issue

ChrisRackauckas:
This misses the function calls in the derivative routines.

src/perform_step/kencarp_kvaerno_perform_step.jl
kanav99: In KenCarp3, `f2` is calculated thrice and `f1` twice. So should I increment by 3?

kanav99:
For derivative routines, according to DiffEqDiffTools, should we do this:
```julia
tmp = length(x)
if typeof(fx) == Nothing
  tmp *= 2 # As we need to calculate both fx0 and fx
end
if eltype(df)<:Complex
  tmp *= 2 # For im and re part
end
integrator.destats.nf += tmp
```
But I doubt because `f`may assume `x` as a single element, so the statement `tmp = length(x)` might be wrong.

ChrisRackauckas:
The Jacobian isn't handled

ChrisRackauckas:
`tmp *= 2 # As we need to calculate both fx0 and fx`

I think forward differencing only calculates fx0 once?

ChrisRackauckas:
>if eltype(df)<:Complex

complex differentiation doesn't occur when eltype is complex.

kanav99:
For the complex issue, I was trying to count this evaluation: https://github.com/JuliaDiffEq/DiffEqDiffTools.jl/blob/master/src/gradients.jl#L183 We have calculated f(x) twice in case df is complex. Am I missing something?

kanav99:
This commit fails the test as algorithms like `NorsettEuler, ETDRK2, ETDRK3, ETDRK4, HochOst4` don't have the field diff_type. I am not sure how to proceed from here

ChrisRackauckas:
>This commit fails the test as algorithms like NorsettEuler, ETDRK2, ETDRK3, ETDRK4, HochOst4 don't have the field diff_type. I am not sure how to proceed from here. I will investigate a bit more.

Those are exponential integrators. You can just count the `f2` on those.

ChrisRackauckas:
>For the complex issue, I was trying to count this evaluation: https://github.com/JuliaDiffEq/DiffEqDiffTools.jl/blob/master/src/gradients.jl#L183 We have calculated f(x) twice in case df is complex. Am I missing something?

@dextorious I don't think it needs both `f` evaluations, right?

ChrisRackauckas:
I realized that we should add some really good tests for this. You can create a function which has a global ref `counter = Ref(0)` and then in the ODE function do `counter[] += 1`. That will give exactly how many function calls are done. Then you can run the algorithm and compare this to DEStats. Do this on the main algorithms from each class (excluding exponential integrator for now, since they will need some weird different counter), so I would test `BS3`, `Tsit5`, `Vern7/8/9`, `KenCarp4`, `Rosenbrock23`, `Rodas5`, and switching up the differentiation styles on them. Mark the ones that are broken with `@test_broken` and we can merge, and continue to fix more of them in further PRs.

kanav99:
Okay this sounds like a plan. I was actually calculating number for automatic differentiation which turned out to be one only. Only this case of autodiff is a bit peculiar. I will quickly write the tests and push them first.

kanav99:
In this commit: 
1. In autodiff, I have incremented by 1 only, though each call of `f` is a bit expensive
2. `autodiff=false` case doesn't give accurate results.
Observation: On changing `length(u0)` the deficit from actual value doesn't change. For eg: Currently length(u0) = 2 in the tests the deficit is 28 (255-227) but on `length(u0) = 1` deficit is still same. So bug is not in the `derivative` or `jacobian`.

ChrisRackauckas:
This is very strong start at least. Thanks!

PR #679, opened 2019-03-02
[WIP] SDIRK 4th order 4 stage method

saurabhkgp21:
@ChrisRackauckas 
- What should be the initial guess(`nlsolver.z`) for implicit method for each stage? 
- Since this method isn't completely singly diagonal implicit, I have used `update_W!` after changing `gamma` in third stage. Is it correct?

Threading will be added once this method is tested without threads.

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -177,7 +177,7 @@ module OrdinaryDiffEq
 
   export ImplicitEuler, ImplicitMidpoint, Trapezoid, TRBDF2, SDIRK2, Kvaerno3,
          KenCarp3, Cash4, Hairer4, Hairer42, SSPSDIRK2, Kvaerno4, Kvaerno5,
-         KenCarp4, KenCarp5
+         KenCarp4, KenCarp5, RK4Threaded
 
   export GenericImplicitEuler, GenericTrapezoid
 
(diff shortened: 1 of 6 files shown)


src/OrdinaryDiffEq.jl
```diff
@@ -177,7 +177,7 @@ module OrdinaryDiffEq
 
   export ImplicitEuler, ImplicitMidpoint, Trapezoid, TRBDF2, SDIRK2, Kvaerno3,
          KenCarp3, Cash4, Hairer4, Hairer42, SSPSDIRK2, Kvaerno4, Kvaerno5,
-         KenCarp4, KenCarp5
+         KenCarp4, KenCarp5, RK4Threaded
```
ChrisRackauckas: this is a bad name since it's not RK4 threaded...

ChrisRackauckas:
>What should be the initial guess(nlsolver.z) for implicit method for each stage?

For now do the zero guess. This is something that needs to be derived.

>Since this method isn't completely singly diagonal implicit, I have used update_W! after changing gamma in third stage. Is it correct?

yes

src/OrdinaryDiffEq.jl
saurabhkgp21: Will `DIRK44_2`(Diagonal Implicit RK 4th order, 4 stage, 2 parallel) be okay?

ChrisRackauckas: what's the name of the author the tableau comes from?

saurabhkgp21: Iserles and Nørsett, 1990

ChrisRackauckas: Yeah, so I'd call it like IserNor4

PR #682, opened 2019-03-03
Add warning message when ImplicitEuler or Trapezoid is used without saving

YingboMa:


ChrisRackauckas:
why don't we just fix the error estimator? These are really just useful in PDEs where we expect save_everystep=false

ChrisRackauckas:
Wait, why is this an issue? Aren't they used cached values for previous points?

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/sdirk_perform_step.jl#L296

YingboMa:
`uprev2` is from `integrator` https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/sdirk_perform_step.jl#L282, is that going to be an issue?

ChrisRackauckas:
That doesn't use `sol`. It's a thing that's handled like FSAL as part of the integrator, and only when `alg_extrapolates` is true.

YingboMa:
I see.

PR #683, opened 2019-03-03
Do not dispatch on controllers

YingboMa:


src/algorithms.jl
```diff
@@ -20,12 +20,6 @@ abstract type OrdinaryDiffEqExtrapolationVarOrderVarStepAlgorithm <: OrdinaryDif
 struct FunctionMap{scale_by_time} <: OrdinaryDiffEqAlgorithm end
 FunctionMap(;scale_by_time=false) = FunctionMap{scale_by_time}()
 
-@enum Controller begin
```
ChrisRackauckas: why not use enums?

src/integrators/integrator_utils.jl
```diff
@@ -152,35 +152,96 @@ function solution_endpoint_match_cur_integrator!(integrator)
...
-  # PI-controller
-  EEst,beta1,q11,qold,beta2 = integrator.EEst, integrator.opts.beta1, integrator.q11,integrator.qold,integrator.opts.beta2
-  if iszero(EEst)
-    q = inv(integrator.opts.qmax)
-  else
-    @fastmath q11 = EEst^beta1
-    @fastmath q = q11/(qold^beta2)
-    integrator.q11 = q11
-    @fastmath q = max(inv(integrator.opts.qmax),min(inv(integrator.opts.qmin),q/integrator.opts.gamma))
+  if ispredictive(alg)
```
ChrisRackauckas: this looks a bit harder to maintain, add to, and understand. Why not still have smaller functions that end up called, but inline them?

src/algorithms.jl
YingboMa: It will be harder for the users to change the controllers, and plus, it will be a breaking change.

PR #685, opened 2019-03-03
Parallel ERK method

saurabhkgp21:
A 5-parallel 2-processor ERK formula from Kutta’s 3-parameter family of 5th-order 6-stage schemes.

@ChrisRackauckas Which test should be added for threading?


src/alg_utils.jl
```diff
@@ -134,6 +134,7 @@ alg_order(alg::Exprb32) = 3
 alg_order(alg::Exprb43) = 4
 alg_order(alg::Anas5) = 5
 alg_order(alg::RK46NL) = 4
+alg_order(alg::PaRK2p5) = 5
```
ChrisRackauckas: KuttaPRK2p5

src/perform_step/park_perform_step.jl
```diff
@@ -0,0 +1,78 @@
...
+  @unpack α21,α31,α32,α41,α42,α43,α5_6 = cache.tab
+  @unpack β1,β3,β5,β6,c2,c3,c4,c5_6 = cache.tab
+
+  f( k1, uprev, p, t)
+  f( k2, uprev + dt*α21*k1, p, t + c2*dt)
+  f( k3, uprev + dt*(α31*k1 + α32*k2), p, t + c3*dt)
+  f( k4, uprev + dt*(α41*k1 + α42*k2 + α43*k3), p, t + c4*dt)
+
+  if integrator.alg.threading == false
+    f( k5_6[1], uprev + dt*(α5_6[1,1]*k1 + α5_6[1,2]*k2 + α5_6[1,3]*k3 + α5_6[1,4]*k4),
```
ChrisRackauckas: the `u` term here allocates a ton

src/perform_step/park_perform_step.jl
```diff
@@ -0,0 +1,78 @@
...
+@muladd function perform_step!(integrator,cache::PaRK2p5Cache,repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack k,k1,k2,k3,k4,k5_6,fsalfirst = cache
+  @unpack α21,α31,α32,α41,α42,α43,α5_6 = cache.tab
+  @unpack β1,β3,β5,β6,c2,c3,c4,c5_6 = cache.tab
+
+  f( k1, uprev, p, t)
+  f( k2, uprev + dt*α21*k1, p, t + c2*dt)
+  f( k3, uprev + dt*(α31*k1 + α32*k2), p, t + c3*dt)
+  f( k4, uprev + dt*(α41*k1 + α42*k2 + α43*k3), p, t + c4*dt)
```
ChrisRackauckas: allocating, and also probably fails on the + between arrays

test/ode/ode_convergence_tests.jl
```diff
@@ -181,6 +181,16 @@ for i = 1:2
   sim19 = test_convergence(dts,prob,Hairer4())
   @test sim19.𝒪est[:final] ≈ 4 atol=testTol
 
+  sim20 = test_convergence(dts,prob,PaRK2p5(threading=true))
```
ChrisRackauckas: put it with the ERK methods

src/perform_step/park_perform_step.jl
```diff
@@ -0,0 +1,78 @@
+function initialize!(integrator,cache::PaRK2p5ConstantCache)
```
ChrisRackauckas: just PRK for the file names

test/ode/ode_convergence_tests.jl
saurabhkgp21: Should I also add complex tests?

ChrisRackauckas: I don't think that's needed here? There's already complex tests?

saurabhkgp21: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/5390222e6dd85e8558c23e0cbed264d6e919e097/test/complex_tests.jl#L14
These explicit methods do have separate complex tests

src/perform_step/prk_perform_step.jl
```diff
@@ -57,19 +57,26 @@ end
...
+    @. u = uprev + dt*(α5_6[1,1]*k1 + α5_6[1,2]*k2 + α5_6[1,3]*k3 + α5_6[1,4]*k4)
+    f( k5_6[1], u, p, t + c5_6[1]*dt)
+
+    @. u = uprev + dt*(α5_6[2,1]*k1 + α5_6[2,2]*k2 + α5_6[2,3]*k3 + α5_6[2,4]*k4)
+    f( k5_6[2], u, p, t + c5_6[2]*dt)
   else
     Threads.@threads for i in [1,2]
-      f( k5_6[i], uprev + dt*(α5_6[i,1]*k1 + α5_6[i,2]*k2 + α5_6[i,3]*k3 + α5_6[i,4]*k4),
-           p, t + c5_6[i]*dt)
+      @. u = uprev + dt*(α5_6[i,1]*k1 + α5_6[i,2]*k2 + α5_6[i,3]*k3 + α5_6[i,4]*k4)
```
ChrisRackauckas: That's not thread safe

saurabhkgp21: yeah.. sorry I somehow missed it
Should I pre-allocate separate memory for both threads or let it allocate memory on the run? While pre-allocating, same memory can also be used for calculation in parent thread.

ChrisRackauckas: you'll need to make use of `tmp` for one thread and you can use `u` for the other. Then you can make it `tmps = (u,tmp)`, and grab from `tmps` dependent on `i`.

PR #686, opened 2019-03-03
4R+ and 5R+ methods from Carpenter Kennedy Lewis(2000)

deeepeshthakur:
@ChrisRackauckas @ranocha Implemented LowStorageRK methods from [Kennedy, Carpenter, Lewis (2000)](https://www.sciencedirect.com/science/article/pii/S0168927499001415). `[author's name]LDDRK[stage][order]_[embedded_order for error estimation]_[4 or 5]R`.
I'm not sure about the `CKLLSRK85_4FM_4R` scheme all the others are working fine. The first issue is that I think some constants are missing from the paper. I've taken them as zero. The other thing is that when I used the `BigFloat` version of all test problems the order is coming out fine otherwise I couldn't find any `dts` which give satisfactory results. Hence the test are marked broken. 
Also with this all the schemes mentioned in that paper are done.
**Methods Implemented:**
1. `CKLLSRK54_3N_4R`
2. `CKLLSRK54_3M_4R`
3. `CKLLSRK65_4M_4R`
4. `CKLLSRK85_4FM_4R`
5. `CKLLSRK75_4M_5R`

#493


ChrisRackauckas:
>The other thing is that when I used the BigFloat version of all test problems the order is coming out fine otherwise I couldn't find any dts which give satisfactory results.

So bigfloat is fine? Why not test with bigfloat then?

deeepeshthakur:
Then we'll have to make changes in `DiffEqProblemLibrary`. Linear test problems are imported. So, there we'll have to change `uType` to `BigFloat`.

ChrisRackauckas:
There's one in there with BigFloat.

ChrisRackauckas:
`remake` also works.

deeepeshthakur:
remake?

ChrisRackauckas:
http://docs.juliadiffeq.org/latest/basics/problem.html#Modification-of-problem-types-1

deeepeshthakur:
Oh, I'll do this. Thanks!

ChrisRackauckas:
But you should probably use the bigfloat problem since that will have an analytical solution setup for bigfloat IIRC

deeepeshthakur:
I changed for 16 methods. But yes most of them need. Almost all the 3R, 4R methods do.

ChrisRackauckas:
okay, let's see how Travis handles this :)

deeepeshthakur:
But yes. We already did the hard work, getting some `dts` that worked for each of them. If the compile time is large, then we can revert. Btw 2 out of 7 tests. Have uType as BigFLoat already.

Issue #609, opened 2019-01-19
Small allocation in in-place QNDF method


YingboMa:
The allocation is really minuscule, hence the low priority label. 

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/9f38e3b51c1a0d176637b39e85d0fb611bbade27/src/perform_step/bdf_perform_step.jl#L773

PR #688, opened 2019-03-04
Clean up and optimize bdf_utils and test QNDF

YingboMa:


kanav99:
I believe we can replace this line https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/commit/a5bfcdae704115916b939a4893bf3946c7d74748#diff-a47f650c5d66e06a667d8a85cec5c42eL782
with `rmul!(R, U)` to cover up the issue #609, can we?

YingboMa:
No, matrix-matrix multiplication cannot be done in-place. Though, one could do the LAPACK trick to LU factorize it, as triangular matrices multiply can be carried out in-place. But that would be completely unnecessary as it is only a 5x5 allocation.

PR #689, opened 2019-03-05
Update RichardsonEuler()

saurabhkgp21:
- Update name to be more specific (AitkenNeville).
- Method wasn't working due to change in calculate residuals. Passing t as argument in `calculate_residuals` and `internalnorm` solves this.

Should travis be passing even after failing tests?

### DIFF
--- a/test/ode/ode_extrapolation_tests.jl
+++ b/test/ode/ode_extrapolation_tests.jl
@@ -17,14 +17,14 @@ for i = 1:2
   prob = probArr[i]
   # Order Convergence test
   for j = 1:4
-      sim = test_convergence(dts,prob,RichardsonEuler(j,j,j))
+      sim = test_convergence(dts,prob,AitkenNeville(j,j,j))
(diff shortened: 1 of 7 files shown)


test/ode/ode_extrapolation_tests.jl
```diff
@@ -17,14 +17,14 @@ for i = 1:2
   prob = probArr[i]
   # Order Convergence test
   for j = 1:4
-      sim = test_convergence(dts,prob,RichardsonEuler(j,j,j))
```
ChrisRackauckas: Was this file ever added to runtests.jl? Might've gotten missed.

saurabhkgp21: My mistake. I missed it.
This test file has both regression tests and convergence tests. Which group should it belong?

ChrisRackauckas: put it into convergence tests III

PR #693, opened 2019-03-08
Fix regression in VCABM

mauro3:
As far as I can tell, in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/commit/a9119399c5ff88841eff2e3763eef47319c0214b a few refactors were missed.  Fixed here.  Maybe a test should be added but I don't have time, sorry!


ChrisRackauckas:
yes, this should probably get a test. But for now I'll make sure this gets updated anyways. Thanks for the PR!

PR #696, opened 2019-03-09
Incorporates changes in DEStats

kanav99:
Cross ref: JuliaDiffEq/DiffEqBase.jl#196


kanav99:
Currently not sure why the count is lagging by 1 for each method. Strange as for methods like `Vern`s, the files are untouched.

ChrisRackauckas:
Oh, everything defaults to `-1` now so that way it's easy to see what values were not overriden in the other software. We might want to manually create a DEStats that starts at 0 here.

PR #699, opened 2019-03-13
Parallel RichardsonEuler

saurabhkgp21:


src/perform_step/extrapolation_perform_step.jl
```diff
@@ -190,16 +190,16 @@ function prediction!(integrator,cache::RichardsonEulerConstantCache,temp_dt)
 end
 
 function prediction!(integrator,cache::RichardsonEulerCache,temp_dt)
```
ChrisRackauckas: This is definitely not thread-safe. Every thread is writing to the same cache.

saurabhkgp21: Every call of `prediction!` writes to `T[n,1]` only and for each thread, value of `n` will be different due to different `temp_dt`. Also an instance of `ktilde` is created to act as a dummy `k` in function. I am also not using `u` from cache to write intermediate value of `u` in starting. Though this will definetly increase allocation in this function.
Can there be any other problems in this function which I am not considering?

ChrisRackauckas: ooh, I see. I read the `@.` on the wrong side. Yes, it would be worthwhile to create a cache for that temporary, one for each thread. That's an optimization that can be handled in a future PR, but is a very necessary one.

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -250,8 +256,14 @@ end
...
-  prediction!(integrator,cache,dt/2)
+  if integrator.alg.threading
+    Threads.@threads for i in [1,2]
+      prediction!(integrator,cache,dt/i)
+    end
+  else
+    prediction!(integrator,cache,dt)
+    prediction!(integrator,cache,dt/2)
+  end
   T[2,2] = (2*T[2,1] - T[1,1])/(2-1)
```
ChrisRackauckas: why only extrapolate to the second order?

saurabhkgp21: This is classic Richardson extrapolation. AitkenNeville is based on this algorithm only which goes to higher order using recurrence formula. I think going to higher order will make it AitkenNeville.

src/perform_step/extrapolation_perform_step.jl
saurabhkgp21: I'll create a new PR with optimization for threading.

PR #701, opened 2019-03-18
add Rosenbrock-W method 6S4O(S)

huanglangwen:
I added the non-inplace version of the fixed-step 6S4O(S) Rosenbrock-W method (https://www.sciencedirect.com/science/article/pii/S0377042709006396). It passed the convergence test with order 4.


test/ode/ode_rosenbrock_tests.jl
```diff
@@ -200,6 +200,15 @@ sim = test_convergence(dts,prob,Ros4LStab())
...
 @test length(sol) < 20
 
+###RosenbrockW6S4OS
+prob = prob_ode_linear
+
+sim = test_convergence(dts,prob,RosenbrockW6S4OS())
+@test sim.𝒪est[:final] ≈ 4 atol=testTol
+
+sol = solve(prob,RosenbrockW6S4OS(),tstops=dts)
+@test length(sol) < 20
```
YingboMa: This test can be safely omitted, as `RosenbrockW6S4OS` is not adaptive.

src/alg_utils.jl
```diff
@@ -9,6 +9,7 @@ isfsal(alg::FunctionMap) = false
 isfsal(alg::Rodas4) = false
 isfsal(alg::Rodas42) = false
 isfsal(alg::Rodas4P) = false
+isfsal(alg::RosenbrockW6S4OS)=false
```
ChrisRackauckas: I don't see a good reason why this shouldn't use FSAL for a 3rd order interpolation?

huanglangwen: I thought it need `c6=1` in `du = f(u, p, t+c6*dt)` to have FSAL property?

YingboMa:
You can evaluate at c=1 anyway at the end of a step to make a “fake” FSAL
and get a third order dense output for free.

PR #712, opened 2019-04-04
[Do not merge] Experimenting Jacobian re-use strategy

YingboMa:


src/alg_utils.jl
```diff
@@ -383,6 +383,7 @@ gamma_default(alg::IRKC) = 8//10
 qsteady_min_default(alg::OrdinaryDiffEqAlgorithm) = 1
 qsteady_max_default(alg::OrdinaryDiffEqAlgorithm) = 1
 qsteady_max_default(alg::OrdinaryDiffEqAdaptiveImplicitAlgorithm) = 6//5
+qsteady_min_default(alg::OrdinaryDiffEqAdaptiveImplicitAlgorithm) = 5//6
```
ChrisRackauckas: this is risky below 1, right?

src/integrators/controllers.jl
```diff
@@ -61,13 +61,13 @@ end
 
 @inline function predictive_step_accept_controller!(integrator, alg, q)
   if q <= integrator.opts.qsteady_max && q >= integrator.opts.qsteady_min
-    q = one(q)
+    return integrator.dt
```
ChrisRackauckas: is this correct?

PR #714, opened 2019-04-06
Adaptive explicit extrapolation

AlthausKonstantin:
This are two implementations of the adaptive extrapolation of the explicit midpoint rule (Similar to `ODEX` and `DIFEX `).
I developed the original code for my bachelor's thesis and would like to add it now to the JuliaDiffEq ecosystem.

This is the first working version of the algorithms and I would love to get some feedback:
Are my additions ok? Is there something I have to change? Are there some essential options from the common interface I should implement?

Since it is the first time that I contribute to an open source project, I would also be grateful to get some tips, which tests I should implement.

Thanks!


src/alg_utils.jl
```diff
@@ -105,6 +105,11 @@ get_current_alg_order(alg::QNDF,cache) = cache.order
 get_current_adaptive_order(alg::QNDF,cache) = cache.order
 get_current_adaptive_order(alg::OrdinaryDiffEqExtrapolationVarOrderVarStepAlgorithm,cache) = cache.cur_order
 get_current_alg_order(alg::OrdinaryDiffEqExtrapolationVarOrderVarStepAlgorithm,cache) = cache.cur_order
+get_current_alg_order(alg::ExtrapolationMidpointDeuflhard,cache) = 2(cache.n_curr + 1)
+get_current_adaptive_order(alg::ExtrapolationMidpointDeuflhard,cache) = 2cache.n_curr + 1
```
ChrisRackauckas: The error estimator always uses an embedded method of order 1 less? Just double checking since I don't have the book on me for how Hairer defined it.

src/alg_utils.jl
```diff
@@ -339,16 +346,22 @@ beta2_default(alg::FunctionMap) = 0
 beta2_default(alg::DP8) = 0//1
 beta2_default(alg::DP5) = 4//100
 beta2_default(alg::DP5Threaded) = 4//100
+beta2_default(alg::ExtrapolationMidpointDeuflhard) = 0//1
```
ChrisRackauckas: Is there an issue with PI adaptivity here?

src/caches/extrapolation_caches.jl
```diff
@@ -47,3 +47,258 @@ function alg_cache(alg::RichardsonEuler,u,rate_prototype,uEltypeNoUnits,uBottomE
...
+  # Values that are mutated
+  Q::Vector{QType} # Storage for stepsize scaling factors. Q[n] contains information for extrapolation order (n + alg.n_min - 1)
+  n_curr::Int64 # Storage for the current extrapolation order
+  n_old::Int64 # Storage for the extrapolation order n_curr before perfom_step! changes the latter
+
+  # Constant values
+  subdividing_sequence::Array{BigInt,1}  # subdividing_sequence[n] is used for the (n -1)th internal discretisation
+  stage_number::Array{Int,1} # Stage_number[n] contains information for extrapolation order (n + alg.n_min - 1)
+  # Weights and Scaling factors for extrapolation operators
+  extrapolation_weights::Array{Rational{BigInt},2}
```
ChrisRackauckas: Do these need to be Rational? Maybe it doesn't matter all that much, but Bigs are allocating, so it would be worthwhile to profile if the calculations on these actually end up being costly, and whether just having them as Float64 may be numerically stable enough to get the job done. If Float64 is enough, then typing these to make `t` would make sense.

src/caches/extrapolation_caches.jl
```diff
@@ -47,3 +47,258 @@ function alg_cache(alg::RichardsonEuler,u,rate_prototype,uEltypeNoUnits,uBottomE
...
+  extrapolation_scalars::Array{Rational{BigInt},1}
+  # Weights and scaling factors for internal extrapolation operators (used for error estimate)
+  extrapolation_weights_2::Array{Rational{BigInt},2}
+  extrapolation_scalars_2::Array{Rational{BigInt},1}
+end
+
+function alg_cache(alg::ExtrapolationMidpointDeuflhard,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
+  @unpack n_min, n_init, n_max = alg
+
+  QType = tTypeNoUnits <: Integer ? typeof(qmin_default(alg)) : tTypeNoUnits # Cf. DiffEqBase.__init in solve.jl
```
ChrisRackauckas: peculiar that the type of this is needed. I'll see where this goes...

src/caches/extrapolation_caches.jl
```diff
@@ -47,3 +47,258 @@ function alg_cache(alg::RichardsonEuler,u,rate_prototype,uEltypeNoUnits,uBottomE
...
+  # Compute scaling factors for internal extrapolation operators
+  extrapolation_scalars_2 = ones(Rational{BigInt}, n_max)
+  extrapolation_scalars_2[1] = -nodes[2]
+  for n = 1:(n_max-1)
+      extrapolation_scalars_2[n+1] = -extrapolation_scalars_2[n] * nodes[n+2]
+  end
+
+  # Compute scaling factors for extrapolation operators
+  extrapolation_scalars = -nodes[1] * [BigInt(1); extrapolation_scalars_2]
+
```
ChrisRackauckas: Because the amount of pre-computations is based on `n_max`, it actually matters in small enough settings. That is something that may be worth remembering to document, or at least test a little bit. BTW, at when `n_max` does overflow start to occur in the sequences?

src/caches/extrapolation_caches.jl
```diff
@@ -47,3 +47,258 @@ function alg_cache(alg::RichardsonEuler,u,rate_prototype,uEltypeNoUnits,uBottomE
...
+  stage_number::Array{Int,1} # Stage_number[n] contains information for extrapolation order (n - 1)
+  # Weights and Scaling factors for extrapolation operators
+  extrapolation_weights::Array{Rational{BigInt},2}
+  extrapolation_scalars::Array{Rational{BigInt},1}
+  # Weights and scaling factors for internal extrapolation operators (used for error estimate)
+  extrapolation_weights_2::Array{Rational{BigInt},2}
+  extrapolation_scalars_2::Array{Rational{BigInt},1}
+end
+
+function alg_cache(alg::ExtrapolationMidpointHairerWanner,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
```
ChrisRackauckas: I feel like some refactoring can make this re-use a lot of code from the previous cache function. That's a second order issue that we can handle later though.

src/caches/extrapolation_caches.jl
```diff
@@ -47,3 +47,258 @@ function alg_cache(alg::RichardsonEuler,u,rate_prototype,uEltypeNoUnits,uBottomE
...
+  tmp = zero(u)
+  T = fill(zero(u), alg.n_max + 1)
+  res = uEltypeNoUnits.(zero(u))
+  fsalfirst = zero(rate_prototype)
+  k = zero(rate_prototype)
+
+  cc = alg_cache(alg,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,Val{false})
+
+  ExtrapolationMidpointDeuflhardCache(u_tilde, u_temp1, u_temp2, tmp, T, res, fsalfirst, k,
+      cc.Q, cc.n_curr, cc.n_old, cc.subdividing_sequence, cc.stage_number, cc.extrapolation_weights, cc.extrapolation_scalars, cc.extrapolation_weights_2, cc.extrapolation_scalars_2)
```
ChrisRackauckas: it might be easier to just stick the `cc` in here, using composition to improve code re-use. But again with the other comment, optimizing code simplicity and re-use can come in a later PR.

src/integrators/integrator_utils.jl
```diff
@@ -296,6 +296,187 @@ function step_reject_controller!(integrator,alg::PredictiveControllerAlgs)
...
+  # Standard stepsize controller
+  # Compute and save the stepsize scaling based on the latest error estimate of the current order
+  if iszero(integrator.EEst)
+    q = inv(integrator.opts.qmax)
+  else
+    # Update gamma and beta1
+    integrator.opts.beta1 = typeof(integrator.opts.beta1)(1 // (2integrator.cache.n_curr + 1))
+    integrator.opts.gamma = typeof(integrator.opts.gamma)(1 // 4)^integrator.opts.beta1
+    # Compute new stepsize scaling
+    qtmp = integrator.EEst^integrator.opts.beta1 / integrator.opts.gamma
```
ChrisRackauckas: I see why you got rid of beta2. Something that would be worthwhile later would be to try and make this use PI instead of P, since step stabilization will help a method based on explicit RKs take larger steps.

ChrisRackauckas: But hmm, yes anything based on a work estimate is hard to stabilize.

src/integrators/integrator_utils.jl
```diff
@@ -296,6 +296,187 @@ function step_reject_controller!(integrator,alg::PredictiveControllerAlgs)
...
+end
+
+function stepsize_predictor!(integrator,alg::ExtrapolationMidpointDeuflhard,n_new::Int64)
+  # Compute and save the stepsize scaling for order n_new based on the latest error estimate of the current order.
+  if iszero(integrator.EEst)
+    q = inv(integrator.opts.qmax)
+  else
+    # Initialize
+    @unpack t,EEst = integrator
+    tol = integrator.opts.internalnorm(integrator.opts.reltol,t) # Deuflhard's approach relies on EEstD ≈ ||relTol||
```
ChrisRackauckas: This might have weird interactions with element-wise tolerances, but does work.

ChrisRackauckas:
What's the reference you used for the stepping behaviors?

src/alg_utils.jl
AlthausKonstantin: No that is a mistake. The embedded method is of order `2cache.n_curr`. I will change that

src/caches/extrapolation_caches.jl
AlthausKonstantin: Ok, I will mention this in the documentation.

As far as I've tested the constructor, I have not been able to produce an overflow (for any sequence) if using `BigInt`. As I replaced `BigInt` by `Int`, I got an overflow for the Romberg sequence and `n_max = 6`

src/caches/extrapolation_caches.jl
AlthausKonstantin: I haven't played with Floats here so far.

The good thing about the`BigInt`s is that the weights are converted in `perform_step` to the type of `u`. That allows to adapt the precision of the problem.
I'd suspect that if one uses `Float64` weights and say a`BigFloat` representation for the problem, the approximation error will not get any better than `eps(Float64)`.

src/caches/extrapolation_caches.jl
AlthausKonstantin: That's true. Am I "allowed" define a function producing the weights etc. somewhere? If so where?

src/caches/extrapolation_caches.jl
AlthausKonstantin: Here I decided against that to make the call `@unpack subdividing_sequence = cache` and such possible for both caches.

Would you prefer an `if - else` statement before calling `@unpack` in order to check if a cache is mutable or not?

src/integrators/integrator_utils.jl
AlthausKonstantin: That is a problem I have not figured out yet. The thing is that in Deuflhard's approach some sort of notion of a relative tolerance (< 1 !)  is needed for the convergence monitor and stepsize prediction.

I guess this forces the user in the case of vector valued rel. and abs. tolerance to think of these vectors as scaling vectors multiplied by `tol`. 

So far I haven't played with element-wise tolerances and don't know what the effects are...

AlthausKonstantin:
> What's the reference you used for the stepping behaviors?

What do you mean exactly - the literature for the adaptivity?

src/caches/extrapolation_caches.jl
ChrisRackauckas: What about just converting to `u`'s type here?

src/caches/extrapolation_caches.jl
ChrisRackauckas: just do it in another part of this file, and then call it from in here.

ChrisRackauckas:
What's the literature you used for the implementation? That's what I meant.

AlthausKonstantin:
> What's the literature you used for the implementation? That's what I meant.

My two main sources have been the [book](https://doi.org/10.1007/978-3-540-78862-1) of Hairer, Wanner and Nørsett  (Solving Ordinary Differential Equations I) 
and  Deuflahrd's [paper](http://dx.doi.org/10.1007/BF01418332) (Order and stepsize control in extrapolation methods). 
I also sketched their reasoning - as I understood it and based my code on - in my thesis (I added it to my [repository](https://github.com/AlthausKonstantin/Extrapolation)). 
Does that help?

src/caches/extrapolation_caches.jl
AlthausKonstantin: Ah I see -  the caches are recreated for every call of `solve`, right? Yes, then the type of `u` should be an option.
I guess I will set up an experiment to check the stability of the weights!

BTW: Would it be possible to move the weights into the algorithm's `struct`? That way the computation of the tables for every `solve` call would be bypassed..

src/caches/extrapolation_caches.jl
AlthausKonstantin: Ok I will change that!

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -14,7 +14,7 @@ end
 
 function perform_step!(integrator,cache::RichardsonEulerCache,repeat_step=false)
   @unpack t,dt,uprev,u,f,p = integrator
-  @unpack k,fsalfirst,T,utilde,atmp,dtpropose,cur_order,A = cache
+  @unpack k,fsalfirst,T,u_tilde,atmp,dtpropose,cur_order,A = cache
```
ChrisRackauckas: no reason to change this unless it's changed everywhere else too.

ChrisRackauckas:
This needs some convergence tests. Set the minimum and maximum orders to be the same and add a few convergence simulations like is done with the other methods.

ChrisRackauckas:
Switch broadcasts to `@..` https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/716

AlthausKonstantin:
I added some convergence tests - most of them fail. 
I investigated the issue in an interactive .ipynb and it seems the convergence for sufficiently high orders is so fast, that the error is very small in the beginning and increases due to rounding errors (?) for even smaller step sizes. Thus the `test_convergence` function does not observe the desired asymptotical behavior of the error.

What is the appropriate way of sharing the notebook with you? Is just putting a .zip here ok?
[Convergence Simulation.ipynb.zip](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/files/3097844/Convergence.Simulation.ipynb.zip)

src/perform_step/extrapolation_perform_step.jl
AlthausKonstantin: Done

ChrisRackauckas:
Use a version of the problem where `u0` is BigFloat so that way you are using arbitrary precision arithmetic, and test convergence of high orders with that. That is also a good test that the implementation isn't relying on Float64 anywhere.

AlthausKonstantin:
> Switch broadcasts to `@..` #716
Does this also apply to expressions like `broadcast(+, A, B)`  or only to `@.`?

AlthausKonstantin:
> Use a version of the problem where `u0` is BigFloat so that way you are using arbitrary precision arithmetic, and test convergence of high orders with that. That is also a good test that the implementation isn't relying on Float64 anywhere.

Thank you! 
I adapted the tests. But curiously enough it seems to be vital for my methods that also the time span is given by two `BigFloat` if the subdividing sequence happens to be different from the Romberg sequence (See the convergence plots in the updated [notebook](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/files/3103205/190422.Convergence.Simulation.ipynb.zip)). 

Why this behavior depends on the chosen subdividing sequence is not really clear to me...

ChrisRackauckas:
>Does this also apply to expressions like broadcast(+, A, B) or only to @.?

All if possible. `@..` is just a faster form of broadcast since it removes some aliasing assumptions when the broadcast is supposed to match a map.

ChrisRackauckas:
I handled the merge conflicts. We may want to just merge the PR once I check out the convergence tests are all good, and then future updates can take place in follow up PRs. It's much easier to keep doing small PRs than it is to keep a branch from going stale.

ChrisRackauckas:
My guess is that something odd happens in a spot like `j_int = 2Int64(subdividing_sequence[i+1])`, but yeah that bigfloat subdividing sequence issue is a bit hard to track down. I'll open an issue on it. It's not the worst thing in the world since the majority users use Float64, but it would be nice to find out sometime in the future. 

`sum( broadcast(*, T[1:(i+1)], eltype(uprev).(extrapolation_weights[1:(i+1), (i+1)])) )` that expression allocates a bit and should be changed to using generators instead of broadcast, but that's probably not a big deal so it can be handled in a future PR.

I'll merge if tests pass and we can continue from there. Thanks! This is quite a nice PR, and we will want to continue down the path of https://arxiv.org/abs/1305.6165 . With Julia's new multithreading system this will be pretty exciting.

One contribution you may want to think about is the implicit methods. Since you have the sequences now, swapping out for the implicit parts shouldn't be too bad: just use the nonlinear solvers already in OrdinaryDiffEq. Doing this with the parallelism was attempted in Python, but of course Python got in the way: https://github.com/numerical-mathematics/extrapolation/issues/27 . @YingboMa and I have interesting test cases where BLAS multithreading doesn't speed up the calculation (and in fact slows it down), so taking better control of the parallelism can help on some 50ish ODE very implicit systems.

ChrisRackauckas:
This commit introduced a binary and needs to be deleted somehow: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/714/commits/6ddcfcb7183f954d47a4562da73e3fd468b98e36 , otherwise it will sit and clog and repo history.

ChrisRackauckas:
I cleaned it up and removed the binary from the history. If all went well, then tests will pass again and I'll merge.

AlthausKonstantin:
Okay, thank you so much for all the help!
After this merges I'd also like to contribute to the documentation. Should I just open another pull request in the DiffEqDocs with a reference to this PR?
Same for the Benchmarking, should I add this algorithms to the NonStiffODE notebooks?

ChrisRackauckas:
>After this merges I'd also like to contribute to the documentation. Should I just open another pull request in the DiffEqDocs with a reference to this PR?

Yes

>Same for the Benchmarking, should I add this algorithms to the NonStiffODE notebooks?

Everything is moving away from notebooks. We're in the middle of a benchmark overhaul. See https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl/pull/32

ChrisRackauckas:
##### `ExtrapolationMidpointDeuflhard` and `ExtrapolationMidpointHairerWanner`

`ExtrapolationMidpointDeuflhard` and `ExtrapolationMidpointHairerWanner` adapt order
and stepsize. Both extrapolate the explicit midpoint rule and differ only in the
implementation of the adaptive behavior.

###### Arguments for Initializing the Algorithms

Both methods accept the same set of arguments:
* `min_extrapolation_order` and `max_extrapolation_order`
 restrict the range of values the adaptively controlled extrapolation
 order can attain.
 Since the order control mechanism relies on information from the *previous*  time
 step, the order of extrapolation for the very first timestep is provided externally
 by `init_extrapolation_order`.

  - `ExtrapolationMidpointDeuflhard` uses by default an order range of `1:10` and
  starts with `init_extrapolation_order = 5`.
  -  `ExtrapolationMidpointHairerWanner` works similarly but uses a default range of
  `2:10` since it  requires
  `2 <= min_extrapolation_order <=  init_extrapolation_order < max_extrapolation_order`.

* `sequence_symbol` specifies the step-number sequences, also called the subdividing
 sequence, for both algorithms.
 Possible values for this flag are `:harmonic`, `:romberg` or `:bulirsch`.

To override the default values use keyword arguments like shown above for the
 `AitkenNevillie` algorithm.

If you specify values that do not make sense, e.g.
 `init_extrapolation_order = 3, min_extrapolation_order = 4 `or
 `sequence_symbol = :foo`, the constructor will correct the input and warn you
 it did so.

Finally note that the *extrapolation order* and the *order* of the method do not
 coincide. For both algorithms an extrapolation order of `n` translates to an error of
 of order `2(n+1)`.

###### Details on the Implementation

 As mentioned above, the explicit midpoint rule is the internal method which is
 extrapolated. Since it is not a one step method, it is started with one explicit
 Euler step.

The actual extrapolation is not achieved by the
[algorithm of Aitken and Neville](https://en.wikipedia.org/wiki/Neville%27s_algorithm)
 but by usage of the First Barycentric Formula (A short overview of Barycentric
Interpolation
is given in this [paper](https://doi.org/10.1137/S0036144502417715) of Berrut
 and Trefethen).
This means that scalars and weights necessary for carrying out the extrapolation are
precomputed and tabulated once at the beginning of the call of `solve`.
From there the
actual extrapolation performed in each timestep is achieved by computing a linear
combination of the internal discretizations that have been obtained by the explicit
midpoint rule.
Thus initializing the precomputed values requires time and storage of order
`max_extrapolation_order ^ 2` but the time computing the extrapolation for some timestep
scales only linearly in `max_extrapolation_order`.

Since the algorithms do not use the Aitken Neville scheme for extrapolation, they
also use a different embedded approximation than other extrapolation methods, e.g.
`ODEX`, do.
Using a notation common for the algorithm of Aitken Neville, let the internal
discretizations be given by `T[1:n, 1]`. Then `ODEX` would use the entry `T[n,n]` as
the new solution and estimate the local error with approximation given by `T[n,n-1]`
(cf. equation (9.23) in
[Solving Ordinary Differential Equations I](https://doi.org/10.1007/978-3-540-78862-1)
of Hairer, Wanner and Nørsett).
Since for the two algorithms presented here only the entries `T[i,i]` for `i = 1:n`
are accessible, they use the value `T[n-1,n-1]` for error estimation instead.

PR #717, opened 2019-04-13
Towards Common Perform step

kanav99:
This is a very rough implementation for common perform step for SSPRK22 and SSPRK33, which I am expanding to more algorithms.
This implementation is not at all feasible to be merged as it allocates in constant cache. I am learning the metaprogramming in Julia to know what all is possible for us.
Ideas I am thinking:
1. Make a custom `@unpack` which if an entry is not found in the object, assigns some specific default value. Like if `c0` doesn't exist in cache, it should be set to `1.0`.


src/caches/ssprk_caches.jl
```diff
@@ -8,7 +8,24 @@
   step_limiter!::StepLimiter
 end
 
-struct SSPRK22ConstantCache <: OrdinaryDiffEqConstantCache end
+struct SSPRK22ConstantCache{T, T2} <: OrdinaryDiffEqConstantCache
+  α::Array{T}
```
ChrisRackauckas: SArrays

src/caches/ssprk_caches.jl
```diff
@@ -34,7 +51,28 @@ alg_cache(alg::SSPRK22,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTyp
   step_limiter!::StepLimiter
 end
 
-struct SSPRK33ConstantCache <: OrdinaryDiffEqConstantCache end
```
ChrisRackauckas: you can probably make this cache and the previous the same?

ChrisRackauckas: we need to be careful about the ones that have a different interpolation though.

kanav99: Yeah we can have same caches but different constructors (I mean using inheritence).

kanav99:
Shifting to SArrays makes us very close to our previous performance stats, atleast for this small script:
```julia
using BenchmarkTools, OrdinaryDiffEq
f(u,p,t) = 1.01*u
u0=1/2
tspan = (0.0,1.0)
prob = ODEProblem(f,u0,tspan)
@btime sol = solve(prob,SSPRK33(),reltol=1e-8,abstol=1e-8,dt=0.01)
```
_Master_:
```
  15.479 μs (170 allocations: 15.48 KiB)
```
_PR_:
```
  15.690 μs (171 allocations: 15.75 KiB)
```

ChrisRackauckas:
Switch broadcasts to `@..`

ChrisRackauckas:
Ref: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/233

PR #719, opened 2019-04-17
WIP: Full GPU stiff ODE solvers

ChrisRackauckas:
With @YingboMa 

```julia
using OrdinaryDiffEq, CuArrays, LinearAlgebra
function lorenz(du,u,p,t)
    mul!(du,A,u)
end
A = cu(-rand(3,3))
u0 = cu([1.0;0.0;0.0])
tspan = (0.0,100.0)
sol = solve(prob,Tsit5())

sol = solve(prob,Rosenbrock23(linsolve=culinsolve))

function culinsolve(x,A,b,update_matrix=false)
    _qr = qr(A)
    _x = UpperTriangular(_qr.R) \ (_qr.Q' * reshape(b,length(b),1))
    x .= vec(_x)
end
function culinsolve(::Type{Val{:init}},f,u0_prototype)
    culinsolve
end
```

works with this PR. This means we have the following action items:

1) Define dense Jacobians in a generic way. Outer product works fine, but is a little hacky. If someone has a better solution then sure.
2) Linear solvers need to be defined for CuArrays. `qr(A)` works, but surprisingly using `\` on that doesn't. So there needs to be a PR to CuArrays.jl that makes lu or qr works. Default linsolve should be setup to handle this.
3) Do we need to zero out `J`? We can do `fill!(0,r .* r')` or something. What's written here has an alloc.

### DIFF
--- a/src/caches/firk_caches.jl
+++ b/src/caches/firk_caches.jl
@@ -82,7 +82,7 @@ function alg_cache(alg::RadauIIA5,u,rate_prototype,uEltypeNoUnits,uBottomEltypeN
   k = similar(rate_prototype); k2 = similar(rate_prototype); k3 = similar(rate_prototype)
   fw1 = similar(rate_prototype); fw2 = similar(rate_prototype); fw3 = similar(rate_prototype)
 
-  J = fill(zero(uEltypeNoUnits),length(u),length(u))
+  J = false .* rate_prototype .* rate_prototype'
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Newton methods work for freeee!

```julia
sol = solve(prob,KenCarp3(linsolve=culinsolve))
```

PR #725, opened 2019-04-26
Add Project.toml

YingboMa:


Project.toml
```diff
@@ -0,0 +1,35 @@
...
+julia = ">= 1.0"
+DiffEqBase = ">= 5.6.3"
+DiffEqOperators = ">= 3.2.0"
+Parameters = ">= 0.10.0"
+ForwardDiff = ">= 0.10.3"
+GenericSVD = ">= 0.0.2"
+NLsolve = ">= 0.14.1"
+RecursiveArrayTools = ">= 0.18.6"
+DiffEqDiffTools = ">= 0.4.0"
+Reexport
```
ChrisRackauckas: ?

Project.toml
```diff
@@ -0,0 +1,36 @@
+name = "OrdinaryDiffEq"
+uuid = "9c2ba320-684e-11e9-1325-35842128880e"
```
fredrikekre: This needs to be https://github.com/JuliaRegistries/General/blob/5779b981069587d8e933415ab7f8b71a51151c1b/O/OrdinaryDiffEq/Package.toml#L2, xref https://github.com/JuliaLang/julia/issues/31859

PR #727, opened 2019-05-01
Explicit extrapolation fix

AlthausKonstantin:
This Pull Request contains two fixes for the methods `ExtrapolationMidpointDeuflhard` and
`ExtrapolationMidpointHairerWanner`.

1) Order Selection process for  `ExtrapolationMidpointHairerWanner`.
  `n_curr` in `perform_step!` must be greater or equal to 3.
  Otherwise, if `n_min` attains its minimum 2,  there will be 0 passed as an index to an array:
  https://github.com/AlthausKonstantin/OrdinaryDiffEq.jl/blob/4ec7bda492b83c3925c9f3afb21618fea2cf4160/src/perform_step/extrapolation_perform_step.jl#L453.

2) Use of `sequence_symbol` in src/caches/extrapolation_caches.jl is deprecated.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -90,7 +90,7 @@ function ExtrapolationMidpointHairerWanner(;min_order=2,init_order=5, max_order=
   # Enforce 2 <=  min_order
   # and min_order + 1 <= init_order <= max_order - 1:
   n_min = max(2, min_order)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Thanks! Yes, I was wondering about that off-by-one thing that was going on. Thank you for clarifying it.

PR #729, opened 2019-05-04
ROCK2 Less Memory Usage

deeepeshthakur:
@ChrisRackauckas Removes extra array from `ROCK2Cache`


ChrisRackauckas:
Nice! This makes it a pretty low-memory stiff ODE solver. That would be interesting to demonstrate.

deeepeshthakur:
I did a few small benchmarks using Linear problems, but there was no significant improvement in performance.

ChrisRackauckas:
Reducing memory wouldn't cause a performance improvement, but it would allow for more scalability.

PR #731, opened 2019-05-08
Threading in AitkenNeville

saurabhkgp21:
@ChrisRackauckas Which tests are required for this PR?

### DIFF
--- a/src/caches/extrapolation_caches.jl
+++ b/src/caches/extrapolation_caches.jl
@@ -32,20 +32,35 @@ function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomElt
   fsalfirst = zero(rate_prototype)
   cur_order = max(alg.init_order, alg.min_order)
   dtpropose = zero(dt)
-  T = fill(zeros(eltype(u), size(u)), (alg.max_order, alg.max_order))
+  T = Array{typeof(u),2}(undef, alg.max_order, alg.max_order)
+  # Array of arrays of length equal to number of threads to store intermediate
+  # values of u and k. [Thread Safety]
+  u_tmps = Array{typeof(u),1}(undef, Threads.nthreads())
+  k_tmps = Array{typeof(k),1}(undef, Threads.nthreads())
+  # Initialize each element of u_tmps and k_tmps to different instance of
+  # zeros array similar to u and k respectively
+  for i=1:Threads.nthreads()
+      u_tmps[i] = zeros(size(u))
+      k_tmps[i] = zeros(size(k))
+  end
+  # Initialize each element of T to different instance of zeros array similar to u
+  for i=1:alg.max_order
+      for j=1:alg.max_order
+          T[i,j] = zeros(size(u))
+      end
+  end
   work = zero(dt)
   A = one(Int)
   atmp = similar(u,uEltypeNoUnits)
   step_no = zero(Int)
-  u_tmps = fill(u, Threads.nthreads())
-  k_tmps = fill(k, Threads.nthreads())
   AitkenNevilleCache(u,uprev,tmp,k,utilde,atmp,fsalfirst,dtpropose,T,cur_order,work,A,step_no,u_tmps,k_tmps)
 end
 
 function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
   dtpropose = zero(dt)
   cur_order = max(alg.init_order, alg.min_order)
-  T = fill(zero(eltype(u)), (alg.max_order, alg.max_order))
+  T = Array{typeof(u),2}(undef, alg.max_order, alg.max_order)
+  @. T = u
   work = zero(dt)
   A = one(Int)
   step_no = zero(Int)
--- a/src/caches/extrapolation_caches.jl
+++ b/src/caches/extrapolation_caches.jl
@@ -40,14 +40,12 @@ function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomElt
   # Initialize each element of u_tmps and k_tmps to different instance of
   # zeros array similar to u and k respectively
   for i=1:Threads.nthreads()
-      u_tmps[i] = zeros(size(u))
-      k_tmps[i] = zeros(size(k))
+      u_tmps[i] = zero(u)
+      k_tmps[i] = zero(rate_prototype)
   end
-  # Initialize each element of T to different instance of zeros array similar to u
+  # Initialize first column of T to different instance of zeros array similar to u
   for i=1:alg.max_order
-      for j=1:alg.max_order
-          T[i,j] = zeros(size(u))
-      end
+    T[i,1] = zero(u)
   end
   work = zero(dt)
   A = one(Int)
--- a/src/caches/extrapolation_caches.jl
+++ b/src/caches/extrapolation_caches.jl
@@ -43,9 +43,11 @@ function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomElt
       u_tmps[i] = zero(u)
       k_tmps[i] = zero(rate_prototype)
   end
-  # Initialize first column of T to different instance of zeros array similar to u
+  # Initialize lower triangle of T to different instance of zeros array similar to u
   for i=1:alg.max_order
-    T[i,1] = zero(u)
+    for j=1:i
+      T[i,j] = zero(u)
+    end
   end
   work = zero(dt)
   A = one(Int)
--- a/src/caches/extrapolation_caches.jl
+++ b/src/caches/extrapolation_caches.jl
@@ -1,4 +1,4 @@
-@cache mutable struct AitkenNevilleCache{uType,rateType,arrayType,dtType,uNoUnitsType,tmpType} <: OrdinaryDiffEqMutableCache
+@cache mutable struct AitkenNevilleCache{uType,rateType,arrayType,dtType,uNoUnitsType} <: OrdinaryDiffEqMutableCache
   u::uType
   uprev::uType
   tmp::uType
@@ -12,7 +12,8 @@
   work::dtType
   A::Int
   step_no::Int
-  tmps::tmpType
+  u_tmps::Array{uType,1}
+  k_tmps::Array{rateType,1}
 end
 
 @cache mutable struct AitkenNevilleConstantCache{dtType,arrayType} <: OrdinaryDiffEqConstantCache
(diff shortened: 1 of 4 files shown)


src/perform_step/extrapolation_perform_step.jl
```diff
@@ -15,30 +15,48 @@ end
...
     end
-    T[i,1] = copy(u)
-    # Richardson Extrapolation
-    for j in 2:i
-      T[i,j] = ((2^(j-1))*T[i,j-1] - T[i-1,j-1])/((2^(j-1)) - 1)
+  else
+    Threads.@threads for i in 2:size(T)[1]
+      dt_temp = dt/(2^(i-1))
+      # Solve using Euler method
+      @muladd @.. tmps[i] = uprev + dt_temp*fsalfirst
```
ChrisRackauckas: Instead of making this per `i`, should be per thread. You can use `threadid` to know which thread it is, and `nthreads` to know how many threads you have available for when you preallocate `tmps`.

src/caches/extrapolation_caches.jl
```diff
@@ -35,7 +36,8 @@ function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomElt
   A = one(Int)
   atmp = similar(u,uEltypeNoUnits)
   step_no = zero(Int)
-  AitkenNevilleCache(u,uprev,tmp,k,utilde,atmp,fsalfirst,dtpropose,T,cur_order,work,A,step_no)
+  tmps = fill(u, size(T)[1])
```
ChrisRackauckas: You want `Threads.nthreads()` number of tmps

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -15,30 +15,48 @@ end
...
+      f(k, tmps[i], p, t+dt_temp)
```
ChrisRackauckas: This `k` will be overwritten by each thread.

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -15,30 +15,48 @@ end
...
+      for j in 2:i
+        T[i,j] = ((2^(j-1))*T[i,j-1] - T[i-1,j-1])/((2^(j-1)) - 1)
+      end
     end
-    T[i,1] = copy(u)
-    # Richardson Extrapolation
-    for j in 2:i
-      T[i,j] = ((2^(j-1))*T[i,j-1] - T[i-1,j-1])/((2^(j-1)) - 1)
+  else
+    Threads.@threads for i in 2:size(T)[1]
```
ChrisRackauckas: For this PR, let's do this. In the future though, we will have to break this down a bit. this assumes that Each row of T requires the same amount of work to calculate. But is that true?

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -15,30 +15,48 @@ end
...
+      integrator.destats.nf += 1
+      for j in 2:2^(i-1)
+        @muladd @.. tmps[i] = u + dt_temp*k
+        f(k, tmps[i], p, t+j*dt_temp)
+        integrator.destats.nf += 1
+      end
+      T[i,1] = copy(tmps[i])
+      # Richardson Extrapolation
+      for j in 2:i
+        T[i,j] = ((2^(j-1))*T[i,j-1] - T[i-1,j-1])/((2^(j-1)) - 1)
```
ChrisRackauckas: notice this part requires i-1, so it cannot be done in parallel. Move it out of the parallel loop.

src/perform_step/extrapolation_perform_step.jl
saurabhkgp21: No. To balance the work done by threads we can create `n/2` threads where n is equal to the number of rows in `T`. It will balance the work of threads by computing `T[1,1]` and `T[n,1]` on one thread, `T[2,1]` and `T[n-1,1]` on another thread and so on.

ChrisRackauckas: `size(T)[1]` is the same as `length(T)`.

ChrisRackauckas: >To balance the work done by threads we can create n/2 threads where n is equal to the number of rows in T. It will balance the work of threads by computing T[1,1] and T[n,1] on one thread, T[2,1] and T[n-1,1] on another thread and so on.

Yes that's a good plan.

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -16,29 +16,50 @@ end
...
+      end
+    end
+  else
+    Threads.@threads for i in 2:size(T)[1]
+      dt_temp = dt/(2^(i-1))
+      # Solve using Euler method
+      @muladd @.. u_tmps[Threads.threadid()] = uprev + dt_temp*fsalfirst
+      f(k_tmps[Threads.threadid()], u_tmps[Threads.threadid()], p, t+dt_temp)
+      integrator.destats.nf += 1
```
saurabhkgp21: What should be the value of `integrator.destats.nf`. This doesn't seem to be thread-safe.

ChrisRackauckas: You can calculate that outside of the loop since you know before hand that in one step it's (order *(order +1 ))/2 f-evaluations (that stands for `number of f evaluations`)

src/caches/extrapolation_caches.jl
```diff
@@ -30,18 +32,35 @@ function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomElt
   fsalfirst = zero(rate_prototype)
   cur_order = max(alg.init_order, alg.min_order)
   dtpropose = zero(dt)
-  T = fill(zeros(eltype(u), size(u)), (alg.max_order, alg.max_order))
+  T = Array{typeof(u),2}(undef, alg.max_order, alg.max_order)
```
ChrisRackauckas: typeof(rate_prototype)

src/caches/extrapolation_caches.jl
```diff
@@ -30,18 +32,35 @@ function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomElt
...
+  # Array of arrays of length equal to number of threads to store intermediate
+  # values of u and k. [Thread Safety]
+  u_tmps = Array{typeof(u),1}(undef, Threads.nthreads())
+  k_tmps = Array{typeof(k),1}(undef, Threads.nthreads())
+  # Initialize each element of u_tmps and k_tmps to different instance of
+  # zeros array similar to u and k respectively
+  for i=1:Threads.nthreads()
+      u_tmps[i] = zeros(size(u))
```
ChrisRackauckas: zero(u)

src/caches/extrapolation_caches.jl
```diff
@@ -30,18 +32,35 @@ function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomElt
...
+      k_tmps[i] = zeros(size(k))
```
ChrisRackauckas: zero(rate_prototype)

src/caches/extrapolation_caches.jl
```diff
@@ -30,18 +32,35 @@ function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomElt
...
+  end
+  # Initialize each element of T to different instance of zeros array similar to u
+  for i=1:alg.max_order
+      for j=1:alg.max_order
+          T[i,j] = zeros(size(u))
```
ChrisRackauckas: zero(rate_prototype)

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -16,29 +16,48 @@ end
...
+      # Solve using Euler method
+      @muladd @.. u = uprev + dt_temp*fsalfirst
+      f(k, u, p, t+dt_temp)
       integrator.destats.nf += 1
+      for j in 2:2^(i-1)
+        @muladd @.. u = u + dt_temp*k
+        f(k, u, p, t+j*dt_temp)
+        integrator.destats.nf += 1
+      end
+      T[i,1] = copy(u)
```
ChrisRackauckas: `T[i,1] .= u`

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -16,29 +16,48 @@ end
...
+    Threads.@threads for i in 2:min(size(T)[1],cur_order+1)
+      dt_temp = dt/(2^(i-1))
+      # Solve using Euler method
+      @muladd @.. u_tmps[Threads.threadid()] = uprev + dt_temp*fsalfirst
+      f(k_tmps[Threads.threadid()], u_tmps[Threads.threadid()], p, t+dt_temp)
+      for j in 2:2^(i-1)
+        @muladd @.. u_tmps[Threads.threadid()] = u_tmps[Threads.threadid()] + dt_temp*k_tmps[Threads.threadid()]
+        f(k_tmps[Threads.threadid()], u_tmps[Threads.threadid()], p, t+j*dt_temp)
+      end
+      T[i,1] = copy(u_tmps[Threads.threadid()])
```
ChrisRackauckas: `.=` without the copy.

src/caches/extrapolation_caches.jl
ChrisRackauckas: actually, zero(u)

ChrisRackauckas: And you don't need to fill this entire matrix.

src/caches/extrapolation_caches.jl
```diff
@@ -30,18 +32,33 @@ function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomElt
...
+  k_tmps = Array{typeof(k),1}(undef, Threads.nthreads())
+  # Initialize each element of u_tmps and k_tmps to different instance of
+  # zeros array similar to u and k respectively
+  for i=1:Threads.nthreads()
+      u_tmps[i] = zero(u)
+      k_tmps[i] = zero(rate_prototype)
+  end
+  # Initialize first column of T to different instance of zeros array similar to u
+  for i=1:alg.max_order
+    T[i,1] = zero(u)
```
ChrisRackauckas: You need to initialize the full triangle

src/caches/extrapolation_caches.jl
```diff
@@ -30,18 +32,35 @@ function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomElt
...
-  AitkenNevilleCache(u,uprev,tmp,k,utilde,atmp,fsalfirst,dtpropose,T,cur_order,work,A,step_no)
+  AitkenNevilleCache(u,uprev,tmp,k,utilde,atmp,fsalfirst,dtpropose,T,cur_order,work,A,step_no,u_tmps,k_tmps)
 end
 
 function alg_cache(alg::AitkenNeville,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
   dtpropose = zero(dt)
   cur_order = max(alg.init_order, alg.min_order)
-  T = fill(zero(eltype(u)), (alg.max_order, alg.max_order))
+  T = Array{typeof(u),2}(undef, alg.max_order, alg.max_order)
+  @. T = u
```
ChrisRackauckas:
```suggestion
  @.. T = u
```

ChrisRackauckas:
Fantastic work!

src/caches/extrapolation_caches.jl
ChrisRackauckas: Make a PR with this.

PR #733, opened 2019-05-11
IRKC adaptive fix

deeepeshthakur:
@ChrisRackauckas


src/perform_step/rkc_perform_step.jl
```diff
@@ -728,7 +728,11 @@ function perform_step!(integrator, cache::IRKCCache, repeat_step=false)
   # error estimate
   if isnewton(nlsolver) && integrator.opts.adaptive
     update_W!(integrator, cache, dt, false)
-    mul!(tmp, get_W(nlsolver), @.. dt*(0.5*(du₂ - f2ⱼ₋₁) + (0.5 - μs₁)*(du₁ - f1ⱼ₋₁)))
+    if DiffEqBase.has_invW(f)
+        mul!(vec(tmp),get_W(nlsolver),vec(@.. dt*(0.5*(du₂ - f2ⱼ₋₁) + (0.5 - μs₁)*(du₁ - f1ⱼ₋₁))))
+      else
+        cache.linsolve(vec(tmp),get_W(nlsolver),vec(@.. dt*(0.5*(du₂ - f2ⱼ₋₁) + (0.5 - μs₁)*(du₁ - f1ⱼ₋₁))),false)
```
ChrisRackauckas: This still allocates BTW and should have the right part write into a temporary

deeepeshthakur: Then I'll have add another array. Because We cannot do
`mul!(vec(tmp),get_w(nlsolver),vec(tmp))`

PR #741, opened 2019-05-20
CleanUp ROCK Methods and Memory Usage ROCK4

deeepeshthakur:
@ChrisRackauckas 
Reduces memory usage in `ROCK4`.
Makes the code more readable and Implementation style now resembles the paper more for both `ROCK2` and `ROCK4`
Also makes one small correction in `perform_step`

### DIFF
--- a/src/perform_step/rkc_perform_step.jl
+++ b/src/perform_step/rkc_perform_step.jl
@@ -178,7 +178,6 @@ end
   for i in 2:cache.mdeg
     μ, κ = recf[cache.start + (i - 2)*2 + 1], recf[cache.start + (i - 2)*2 + 2]
     ν = - 1 - κ
-    dtμ = dt*μ
     u = f(uᵢ₋₁, p,tᵢ₋₁)
     tᵢ₋₁ = dt*μ - ν*tᵢ₋₂ - κ*tᵢ₋₃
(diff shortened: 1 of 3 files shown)


src/perform_step/rkc_perform_step.jl
```diff
@@ -153,77 +162,87 @@ end
...
-  temp1 = dt * fpa[cache.mdeg][1]
-  gprev = f(u, p, ci1)
+  uᵢ₋₁ = f(u, p, tᵢ₋₁)
   integrator.destats.nf += 1
-  gprev3 = u + temp1 * gprev
+  uᵢ₋₂ = u + a₃₁*uᵢ₋₁
+  uᵢ₋₃ = u + a₄₁*uᵢ₋₁
+  u    += B₁*uᵢ₋₁
+  integrator.opts.adaptive && (tmp = B̂₁*uᵢ₋₁)
+  uᵢ₋₁ = u + (a₂₁ - B₁)*uᵢ₋₁
```
deeepeshthakur: This is addition subtraction of the same thing but the only way to avoid it is to add another array. Also both `a21` and `B1` are constants so I think doing this should not be an issue.

ChrisRackauckas: yeah this should be fine.

PR #743, opened 2019-05-21
Adds Benchmark Suite

kanav99:
@ChrisRackauckas  server is complete, atleast for basic functionality. The node server is https://github.com/kanav99/DiffEqBenchmarkServer and a small package as its helper is https://github.com/kanav99/DiffEqDiagrams . It is currently configured to work with my fork of OrdinaryDiffEq, but can be shifted effortlessly. It only has one time setup then configuring it is very easy. It is hosted at http://f6df067e.ngrok.io and you can trigger it by making a pr on my fork and commenting `@DiffEqBot runbenchmarks`.

### DIFF
--- a/benchmark/REQUIRE
+++ b/benchmark/REQUIRE
@@ -1,0 +1,1 @@
+BenchmarkTools
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,0 +1,14 @@
+using BenchmarkTools, OrdinaryDiffEq
+f(u,p,t) = u
+prob = ODEProblem(f,1.0,(0.0,1.0))
+
+SUITE = BenchmarkGroup()
+SUITE["algs"] = BenchmarkGroup()
+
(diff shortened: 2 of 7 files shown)


ChrisRackauckas:
Lets get those repos migrated over to DiffEq, and we should get a few other benchmarks (like stiff) running. We should get at least one method per method class benchmarked.

kanav99:
Okay I will transfer the ownership of the repos to DiffEq. We need to specify which problems we should solve to be added to this suit.

kanav99:
I cannot make a repo in DiffEq right now, so we would need new repositories instead for both of them.

kanav99:
@DiffEqBot abort

benchmark/REQUIRE
```diff
@@ -0,0 +1 @@
+BenchmarkTools
```
ChrisRackauckas: does this still use REQUIRE?

kanav99: Yes PkgBenchmark still uses REQUIRE to work

PR #744, opened 2019-05-21
Thread scheduling for AitkenNeville

saurabhkgp21:


src/perform_step/extrapolation_perform_step.jl
```diff
@@ -142,19 +143,34 @@ function perform_step!(integrator,cache::AitkenNevilleConstantCache,repeat_step=
...
-      for j in 2:2^(i-1)
-        @muladd u_temp = @.. u_temp + dt_temp*k_temp
-        k_temp = f(u_temp, p, t+j*dt_temp)
+    # Balance workload of threads by computing T[1,1] with T[max_order,1] on
+    # same thread, T[2,1] with T[max_order-1,1] on same thread. Similarly fill
+    # first column of T matrix
+    Threads.@threads for i in 1:ceil(max_order/2)
+      # indices array hold row index of T matrix to be computed in this
+      # execution of thread
+      indices = [Int(i)]
```
ChrisRackauckas: this will allocate more than you want. Re-structure to not have this array alloc.

saurabhkgp21: Should we do the calculation of `T[i,1]` and `T[max_order+1-i,1]` by copy-pasting the same code twice? That will reduce the memory allocation though. Is there any other way around?

ChrisRackauckas: You can make a tuple, but remember that its length is then type information, so always make it length 2 and check for equality.

ChrisRackauckas:
that looks good to me. Is this complete then?

PR #745, opened 2019-05-21
Limiting Step_Size ROCK methods

deeepeshthakur:
@ChrisRackauckas 
[Current performance with Stiff probems](https://htmlpreview.github.io/?https://github.com/deeepeshthakur/Urls/blob/master/ROCK_Hires.html)


deeepeshthakur:
Controller Choice is justified [here](https://github.com/deeepeshthakur/OrdinaryDiffEq.jl/wiki/Controller-Setups-for-ROCK2-and-ROCK4-using-Stiff-and-NonStiff-Setups.)

PR #754, opened 2019-05-30
Changes for VectorContinuousCallback

kanav99:
Ref: JuliaDIffEq/DiffEqBase.jl#221


src/integrators/type.jl
```diff
@@ -101,8 +101,10 @@ mutable struct ODEIntegrator{algType<:OrdinaryDiffEqAlgorithm,IIP,uType,tType,pT
   saveiter::Int
   saveiter_dense::Int
   cache::CacheType
-  callback_cache::Array{uType}
-  previous_condition::Array{uType}
+  callback_cache::CallbackCacheType
```
ChrisRackauckas: can we just make a callback cache that each package instantiates? Seems easier than keeping track of a bunch of fields.

src/integrators/type.jl
```diff
@@ -101,8 +101,10 @@ mutable struct ODEIntegrator{algType<:OrdinaryDiffEqAlgorithm,IIP,uType,tType,pT
...
+  previous_condition::CallbackCacheType
+  callback_next_sign::CallbackCacheType
```
ChrisRackauckas: This can be a bitarray?

src/integrators/type.jl
kanav99: I don't think so we can do that. All these fields are used in DiffEqBase, so they have to be common. Also they need to be in integrator because of thread safety as pointed out here https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/221#issue-280346404

src/integrators/type.jl
kanav99: Yeah this can be. This was defaulted to float previously thats why I chose float. Lets change this to BitArray.

src/integrators/type.jl
ChrisRackauckas: Yes, but there can be a `CallbackCache` that the integrator calls that makes all of these arrays appear.

src/solve.jl
```diff
@@ -156,6 +156,13 @@ function DiffEqBase.__init(
 
   callbacks_internal = CallbackSet(callback,prob.callback)
 
+  max_len = DiffEqBase.max_vector_callback_length(callbacks_internal)
+  if max_len !== -1
```
ChrisRackauckas: not a huge deal, but breaks inference. Can we instead just search for the existence of a VectorContinuousCallback?

kanav99: how does it break inference, return type is always Int64? I would need the max length out of all VectorContinuousCallback to make the CallbackCache.

ChrisRackauckas: `max_len` cannot be inferred since it's a value, and then which of the branches is taken isn't inferred.

ChrisRackauckas:
Put a lower bound on DiffEqBase v5.9.0

src/solve.jl
kanav99: how do you propose to do this? I think `Val{}` might not be of use as well

ChrisRackauckas: Look for VectorContinuousCallback. That is type information.

kanav99:
I have made another PR, please tag after that is approved and merge.

kanav99:
Oh god this is bad. Making fixes.

ChrisRackauckas:
That last test is fixed by https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/238

PR #756, opened 2019-05-31
[WIP] Generic Rosenbrock algorithm using macros

huanglangwen:
I'm writing a generic Rosenbrock algorithm generated from the tableau using macros.

- [x] non-inplace and fixed step
- [x] inplace with fixed step
- [x] adaptive step


src/generic_rosenbrock.jl
```diff
@@ -0,0 +1,59 @@
+struct RosenbrockTableau
+    a::Array{Real,2}
```
YingboMa: Parameterize the eltype instead of using Real, so that the compiler can specialize in it.

src/generic_rosenbrock.jl
```diff
@@ -0,0 +1,59 @@
...
+        push!(cacheExpr.args,:($(Symbol("C$(cind[1])$(cind[2])"))=convert(T,$(tab.C[Cind]))))
+    end
+=#
+#---------------Constant Cache--------------
+macro RosenbrockW6S4OSConstantCache
+
+end
+
+function RosenbrockW6S4OSTableau()
+    a=[0                  0                  0                  0                  0;
```
ChrisRackauckas: This should probably be a static array so the length can be used at compile time. Probably an optimization for later.

src/generic_rosenbrock.jl
```diff
@@ -0,0 +1,59 @@
+struct RosenbrockTableau
+    a::Array{Real,2}
+    C::Array{Real,2}
+    b::Array{Real,1}
+    gamma::Real
+    d::Array{Real,1}
+    c::Array{Real,1}
+end
+
+macro _genConstantCache(inds,name,T)
```
YingboMa: This is very minor, but I would like to point out that it is recommended to use lowercase letters for functions https://docs.julialang.org/en/v1/manual/variables/#Stylistic-Conventions-1

src/generic_rosenbrock.jl
huanglangwen: Emmm, I thought functions called inside the macros are evaluated at the compile time.

src/generic_rosenbrock.jl
```diff
@@ -0,0 +1,62 @@
...
+    b::Array{T,1}
+    gamma::T2
+    d::Array{T,1}
+    c::Array{T2,1}
+end
+
+macro _gen_constant_cache(inds,name,T)
+    quote
+        for ind in $inds
+            push!(cacheExpr.args,:($(Symbol($name,"$(ind[1])$(ind[2])"))=convert($T,$(tab.$name[ind]))))
```
YingboMa: I don’t think you want to have a macro that mutates cacheExpr.

src/generic_rosenbrock.jl
```diff
@@ -0,0 +1,62 @@
...
+    ainds=findall(!iszero,tab.a)
+    Cinds=findall(!iszero,tab.C)
+    cacheExpr=quote end
+    @_gen_constant_cache ainds a :T
+    @_gen_constant_cache Cinds C :T
+    @_gen_constant_cache eachindex(tab.b) b :T
+    push!(cacheExpr.args,:(gamma=convert(T2,$(tab.gamma))))
+    @_gen_constant_cache eachindex(tab.d) d :T
+    @_gen_constant_cache eachindex(tab.c) c :T2
+
```
YingboMa: This functions returns the last Expr that you pushed and I am not sure about why cacheExpr is needed.

src/generic_rosenbrock.jl
ChrisRackauckas: They are. So I guess gen_constant_cache will be ran from a macro?

huanglangwen: yes

src/generic_rosenbrock.jl
huanglangwen: I'm working on this function. Eventually, it will return cacheExpr which is used inside the macro.

src/generic_rosenbrock.jl
```diff
@@ -0,0 +1,941 @@
...
+end
+
+"""
+    Ros4dummyTableau()
+
+Generate a dummy tableau for ROS4 methods. It can be considered as performing elementwise OR to the masks 
+of those sepecific tableaus: `Ros4dummyTableau()==_masktab(RosShamp4Tableau()) OR _masktab(Veldd4Tableau()) OR ...`
+ROS4 methods have the property of a4j==a3j so a is a 3*3 matrix instead of a 4*4 matrix and c is a 1*3 vector instead of a 1*4 vector.
+"""
+function Ros4dummyTableau()#create a tabmask for all ROS4 methods where false->0,true->non-0
```
YingboMa: I wonder if the masking can be done at runtime. For instance, add `map(!iszero, xxx)` to the `RosenbrockAdaptiveTableau` constructor and fields to hold that information. Of course, this is not a priority.

src/generic_rosenbrock.jl
```diff
@@ -0,0 +1,941 @@
...
+            else
+                @.. du1 = dtC41*k1 + dtC42*k2 + dtC43*k3
+                mul!(du2,mass_matrix,du1)
+                @.. linsolve_tmp = du + dtd4*dT + du2
+            end
+        end
+        constperformstepexpr=gen_constant_perform_step(tabmask,constcachename,n_normalstep,specialstepconst)
+        performstepexpr=gen_perform_step(tabmask,cachename,n_normalstep,specialstep)
+        return esc(quote $([constperformstepexpr,performstepexpr]...) end)
+    end
```
YingboMa: Could you add an `else` case that throws an error?

src/generic_rosenbrock.jl
```diff
@@ -0,0 +1,941 @@
...
+        push!(cacheexprs,gen_algcache(cacheexpr,cachename,constcachename,:ROS34PW3,ROS34PW3tabname))
+        return esc(quote $(cacheexprs...) end)
+    elseif part.value==:init
+        return esc(gen_initialize(cachename,constcachename))
+    elseif part.value==:performstep
+        performstepexprs=Array{Expr,1}()
+        push!(performstepexprs,gen_constant_perform_step(tabmask,constcachename,n_normalstep))
+        push!(performstepexprs,gen_perform_step(tabmask,cachename,n_normalstep))
+        return esc(quote $(performstepexprs...) end)
+    end
```
YingboMa: Same here.

YingboMa:
https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/546828997#L1001

The CI failure seems unrelated.

src/generic_rosenbrock.jl
huanglangwen: Yes, that can be done by `_masktab`. The point is, I prefer to specify a generic tab for a group of methods for places where generic characteristics are needed (like perform_step, tableau struct and cache struct) while not emphasizing any specific tableau.

PR #762, opened 2019-06-04
Restructure tests

ChrisRackauckas:
Restructures the tests into proper folders and absorbs OrdinaryDiffEqExtendedTests.jl


src/perform_step/kencarp_kvaerno_perform_step.jl
```diff
@@ -49,7 +49,7 @@ end
   nlsolver.z = z₂ = z₁
 
   nlsolver.tmp = uprev + γ*z₁
-  nlsolver.c = 2γ
+  nlsolver.c = γ
```
ChrisRackauckas: @YingboMa notice that new tests found this error in time-dependent derivatives of ESDIRKs, so if tests had any of this before, this is why.

PR #770, opened 2019-06-17
Fix `destats` when using cached forward difference

YingboMa:
Ref: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/769, https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/756#issuecomment-502785908


ChrisRackauckas:
lol yay, we did get better.

PR #780, opened 2019-06-22
Optimization for extrapolation algorithms

### LINKED ISSUE #777: let blocks in constant cache extrapolation and all AitkenNeville
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/776 shows that the `let` blocks make the multithreaded extrapolation methods pretty top notch at low tolerances. However, they need to be applied to the other perform steps. In the constant cache version, you need to return from the `let` block, and then save that to the variable again outside the `let` block, which is a little tedious but would make this tremendously faster.

saurabhkgp21:
fixes #777

### DIFF
--- a/src/perform_step/extrapolation_perform_step.jl
+++ b/src/perform_step/extrapolation_perform_step.jl
@@ -39,22 +39,25 @@ function perform_step!(integrator,cache::AitkenNevilleCache,repeat_step=false)
       end
     end
   else
-    # Balance workload of threads by computing T[1,1] with T[max_order,1] on
-    # same thread, T[2,1] with T[max_order-1,1] on same thread. Similarly fill
-    # first column of T matrix
-    Threads.@threads for i in 1:2
-      startIndex = (i == 1) ? 1 : max_order
-      endIndex = (i == 1) ? max_order - 1 : max_order
-      for index in startIndex:endIndex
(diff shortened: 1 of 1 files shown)


src/perform_step/extrapolation_perform_step.jl
```diff
@@ -479,38 +484,46 @@ function perform_step!(integrator,cache::ExtrapolationMidpointDeuflhardConstantC
...
+          endIndex = (i == 1) ? n_curr - 1 : n_curr
+          for index = startIndex : endIndex
+            j_int_temp = 2Int64(subdividing_sequence[index+1])
+            dt_int_temp = dt / (2j_int_temp) # Stepsize of the ith internal discretisation
+            u_temp4 = uprev
+            u_temp3 = u_temp4 + dt_int_temp*integrator.fsalfirst # Euler starting step
+            for j = 2 : 2j_int_temp
+              T[index+1] = u_temp4 + 2dt_int_temp*f(u_temp3, p, t + (j-1)dt_int_temp) # Explicit Midpoint rule
+              u_temp4 = u_temp3
+              u_temp3 = T[index+1]
```
ChrisRackauckas: actually, if T is all we need out of here then it's mutated so that doesn't need to be returned.

PR #781, opened 2019-06-23
[WIP] : Parallel 4th order Rosenbrock Method

saurabhkgp21:
Paper :- https://link.springer.com/article/10.1007/s40819-014-0002-x

### DIFF
--- a/src/perform_step/rosenbrock_perform_step.jl
+++ b/src/perform_step/rosenbrock_perform_step.jl
@@ -254,6 +254,49 @@ end
   integrator.u = u
 end
 
+function initialize!(integrator, cache::NPROS4ConstantCache)
+  integrator.kshortsize = 4
+  integrator.k = typeof(integrator.k)(undef, integrator.kshortsize)
+  integrator.fsalfirst = integrator.f(integrator.uprev, integrator.p, integrator.t)
+  integrator.destats.nf += 1
+
+  # Avoid undefined entries if k is an array of arrays
+  integrator.fsallast = zero(integrator.fsalfirst)
+  integrator.k[1] = zero(integrator.fsalfirst)
+  integrator.k[2] = zero(integrator.fsalfirst)
+  integrator.k[3] = zero(integrator.fsalfirst)
+  integrator.k[4] = zero(integrator.fsalfirst)
+end
+
+@muladd function perform_step!(integrator, cache::NPROS4ConstantCache, repeat_step=false)
+  @unpack t,dt,uprev,u,f,p,k = integrator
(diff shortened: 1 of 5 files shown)


src/perform_step/rosenbrock_perform_step.jl
```diff
@@ -254,6 +254,49 @@ end
...
+@muladd function perform_step!(integrator, cache::NPROS4ConstantCache, repeat_step=false)
+  @unpack t,dt,uprev,u,f,p,k = integrator
+  @unpack α21,α31,α32,α41,α42,α43,b1,b2,b3,b4,γ11,γ21,γ22,γ31,γ32,γ33,γ41,γ42,γ43,γ44 = cache
+
+  W = calc_W!(integrator, cache, dt*γ11, repeat_step)
+  k1 = _reshape(W\-_vec(dt*integrator.fsalfirst), axes(uprev))
+
+  W = calc_W!(integrator, cache, dt*γ22, repeat_step)
+  is_compos = integrator.alg isa CompositeAlgorithm
+  J = calc_J(integrator, cache, is_compos)
```
saurabhkgp21: @ChrisRackauckas can we calculate `J` like this?

ChrisRackauckas: You shouldn't need to build J. W = I - gamma*J

ChrisRackauckas: Hmm, actually that's really odd. @Yingboma take a look at https://link.springer.com/article/10.1007/s40819-014-0002-x . Instead of using a time-gradient here, it uses a Jacobian.

ChrisRackauckas: And yes, that's how you calculate J.

ChrisRackauckas: It might just be an error in the paper...

ChrisRackauckas: there's some other errors in the paper. Changing to use past values of k doesn't guarantee the same order of convergence.

ChrisRackauckas: Actually, it looks like it falls in the class defined by MODIFIED PARALLEL ROSENBROCK METHODS FOR STIFF DIFFERENTIAL EQUATIONS

saurabhkgp21: If we aren't using the past values of `k`, will it be possible to parallelize this method?

ChrisRackauckas: Yes. You can factorize the two different W's on different threads. You really should be factorizing twice and re-using them here. I think we didn't before in the constant cache versions because StaticArrays didn't have a factorization method, but we should probably change that.

PR #783, opened 2019-06-23
Kwarg Verification Hotfix

kanav99:
This is the idea I was mentioning on slack. I couldn't see anything wrong in this way, but feel free to close it if it has some compatibilty issues with some other package.


ChrisRackauckas:
I'm scared of the upstream consequences of this. Try out DiffEqParamEstim and DiffEqSensitivity and see if tests pass.

kanav99:
Yeah it does throw a lot of warnings in DiffEqParamEstim, though this could be solved by passing `allow_extra_args = true` there, but not a good way to solve this problem. Should I close this then?

PR #784, opened 2019-06-23
Delete DP5threaded

saurabhkgp21:


ChrisRackauckas:
Note that this is not a breaking change because this method was never actually documented.

PR #791, opened 2019-06-26
Add sparse diff

huanglangwen:


test/interface/sparsediff_tests.jl
```diff
@@ -0,0 +1,40 @@
...
+    j-i==0 && (A[i,j]=-2)
+  end
+  A
+end
+  
+
+colors = repeat(1:3,10)
+odefun_sp = ODEFunction(f,colorvec=colors)
+prob_sp = ODEProblem(odefun_sp,[1.,2.,3,4,5,4,3,2,1],(0.,10.))
+prob_std = ODEProblem(f,[1.,2.,3,4,5,4,3,2,1],(0.,10.))
```
ChrisRackauckas: should test with jac_prototype a sparse matrix as well

ChrisRackauckas:
Waiting on https://github.com/JuliaDiffEq/SparseDiffTools.jl/pull/35

src/derivative_wrappers.jl
```diff
@@ -40,55 +40,104 @@ function jacobian(f, x, integrator)
     alg = unwrap_alg(integrator, true)
     local tmp
     if alg_autodiff(alg)
-      J = jacobian_autodiff(f, x)
-      tmp = 1
+      if DiffEqBase.has_colorvec(integrator.f)
```
ChrisRackauckas: I don't think we need to branch here. We can always take the colorvec one. Let's discuss this.

src/derivative_wrappers.jl
```diff
@@ -40,55 +40,104 @@ function jacobian(f, x, integrator)
...
       integrator.destats.nf += 1
     else
       isforward = alg.diff_type === Val{:forward}
       if isforward
         forwardcache = get_tmp_cache(integrator, alg, unwrap_cache(integrator, true))[2]
         f(forwardcache, x)
         integrator.destats.nf += 1
-        DiffEqDiffTools.finite_difference_jacobian!(J, f, x, jac_config, forwardcache, dir = diffdir(integrator))
+        if DiffEqBase.has_colorvec(integrator.f)
+          tmp=jacobian_finitediff_forward!(J, f, x, jac_config, forwardcache, integrator, integrator.f.colorvec)
```
ChrisRackauckas: can get rid of the branch and just dispatch on `::Nothing`

src/derivative_wrappers.jl
```diff
@@ -40,55 +40,104 @@ function jacobian(f, x, integrator)
...
     end
     integrator.destats.nf += tmp
     J
 end
 
-jacobian_autodiff(f, x) = ForwardDiff.derivative(f,x)
-jacobian_autodiff(f, x::AbstractArray) = ForwardDiff.jacobian(f, x)
+jacobian_autodiff(f, x) = (ForwardDiff.derivative(f,x),1)
+jacobian_autodiff(f, x::AbstractArray) = (ForwardDiff.jacobian(f, x),1)
+function jacobian_autodiff(f, x::AbstractArray, integrator)
```
YingboMa: Maybe we could to do the function evaluation counting properly, e.g. `nf = length(u) \div chunk_size`. That could be a separate PR.

ChrisRackauckas:
Only minor things. We can merge once tests are passing.

ChrisRackauckas:
Test failure is unrelated and due to a very strict test. https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/293 . Merging.

src/derivative_wrappers.jl
huanglangwen: I'm not sure that every f has colorvec.

src/derivative_wrappers.jl
huanglangwen: Which branch?

ChrisRackauckas:
This needs to happen to StochasticDiffEq as well.

tkoolen:
It looks like this PR added 16 packages as dependencies (recursively via SparseDiffTools). I'm getting slightly worried about bloat; it's just so much more code that can break. For example, I already ran into https://github.com/FluxML/Zygote.jl/issues/260 and https://github.com/JuliaMath/Calculus.jl/issues/133.

ChrisRackauckas:
SparseDiffTools in general is necessary, but the difficulty is Zygote. Indeed Zygote adds all of that IRTools etc. It's wanted in SparseDiffTools though for sparse reverse mode and Hessians. Do you think we should Requires.jl Zygote until it matures?

PR #792, opened 2019-06-26
Add `is_mass_matrix_alg`

YingboMa:


ChrisRackauckas:
ESDIRK is not though? Needs a linsolve of just the mass matrix in the first step IIRC

PR #797, opened 2019-06-29
Implicit Euler Extrapolation method

saurabhkgp21:


src/perform_step/extrapolation_perform_step.jl
```diff
@@ -230,6 +230,101 @@ function perform_step!(integrator,cache::AitkenNevilleConstantCache,repeat_step=
...
+  alg = unwrap_alg(integrator, true)
+  max_order = min(size(T)[1], cur_order+1)
+
+  for i in 1:max_order
+    dt_temp = dt/(2^(i-1)) # Romberg sequence
+    u = uprev
+    fsal_first = integrator.fsalfirst
+    for j in 1:2^(i-1)
+        integrator.dt = dt_temp
+        update_W!(integrator, cache, dt_temp, repeat_step)
```
ChrisRackauckas: This only needs to be called once per dt, since it uses the Jacobian at time `t`, not `t+dt`, and it's `I - dt*J`.

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -230,6 +230,101 @@ function perform_step!(integrator,cache::AitkenNevilleConstantCache,repeat_step=
...
+
+        # initial guess
+        if alg.extrapolant == :linear
+          nlsolver.z = dt_temp*fsal_first
```
ChrisRackauckas: We should get much better interpolants in most cases here. For example, if the `j=1` case has been calculated, we can just use a linear interpolation between t and t+dt values.

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -230,6 +230,193 @@ function perform_step!(integrator,cache::AitkenNevilleConstantCache,repeat_step=
...
+  @unpack u_tmps, k_tmps, nlsolver, z = cache
+
+  alg = unwrap_alg(integrator, true)
+
+  max_order = min(size(T)[1],cur_order+1)
+
+  for i in 1:max_order
+    dt_temp = dt/(2^(i-1)) # Romberg sequence
+    u_tmp = copy(uprev)
+    fsal_first = copy(integrator.fsalfirst)
```
ChrisRackauckas: there's no temporary array to use?

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -230,6 +230,193 @@ function perform_step!(integrator,cache::AitkenNevilleConstantCache,repeat_step=
...
+  @unpack k,T,utilde,atmp,dtpropose,cur_order,A = cache
+  @unpack u_tmps, k_tmps, nlsolver, z = cache
+
+  alg = unwrap_alg(integrator, true)
+
+  max_order = min(size(T)[1],cur_order+1)
+
+  for i in 1:max_order
+    dt_temp = dt/(2^(i-1)) # Romberg sequence
+    u_tmp = copy(uprev)
```
ChrisRackauckas: use a temporary from cache for this

saurabhkgp21: cache optimization still needs to be done. I am trying to converge this method for order > 1 for now.

ChrisRackauckas: oh okay, no worries

ChrisRackauckas:
Code looks good, just needs tests.

test/algconvergence/ode_extrapolation_tests.jl
```diff
@@ -70,6 +70,20 @@ testTol = 0.2
   end
 end # AitkenNeville
 
+@testset "Testing ImplicitEulerExtrapolation" begin
+  for prob in problem_array
+    global dts
+
+    #  Convergence test
+    for j = 1:4
+      sim = test_convergence(dts,prob,AitkenNeville(max_order = j,
```
ChrisRackauckas: ImplicitEulerExtrapolation

PR #798, opened 2019-06-30
Add W_transform for Newton methods

ChrisRackauckas:


ChrisRackauckas:
@YingboMa is this unfinished for a reason? Does it need a DiffEqBase change?

PR #799, opened 2019-06-30
allow tuple-defined u0

ChrisRackauckas:
Requires https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/255 to not be breaking on dynamical problems


ChrisRackauckas:
Next version of DiffEqBase will require this for DynamicalODEProblem

PR #812, opened 2019-07-05
[WIP] Resize remaining fields

kanav99:
Highly WIP PR, do not merge. Apologies for completing this a bit late. I will write intensive tests for this first.

### DIFF
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -153,14 +153,29 @@ end
 
 get_chunksize(jac_config::ForwardDiff.JacobianConfig{T,V,N,D}) where {T,V,N,D} = N
 
-function DiffEqBase.resize_jac_config!(jac_config::ForwardDiff.JacobianConfig, uf, du1, z, alg, i)
-  prev_chunksize = get_chunksize(jac_config)
-  ForwardDiff.JacobianConfig(uf,du1,z,ForwardDiff.Chunk{prev_chunksize}())
+function DiffEqBase.resize_jac_config!(jac_config::ForwardDiff.JacobianConfig, i)
+  for j in eachindex(jac_config.duals)
+    resize!(jac_config.duals[j], i)
+  end
+  jac_config
 end
 
-function DiffEqBase.resize_jac_config!(jac_config::DiffEqDiffTools.JacobianCache, uf, du1, z, alg, i)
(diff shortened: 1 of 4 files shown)


kanav99:
All looks good to me, I might be specifying the problem wrong but the last problem's `u` pass even further than what it should actually stop at. Rest tests pass. Ready for review @ChrisRackauckas

src/derivative_wrappers.jl
```diff
@@ -102,6 +102,18 @@ function DiffEqBase.build_jac_config(alg::OrdinaryDiffEqAlgorithm,f,uf,du1,uprev
   jac_config
 end
 
+get_chunksize(jac_config::ForwardDiff.JacobianConfig{T,V,N,D}) where {T,V,N,D} = N
+
+function DiffEqBase.resize_jac_config!(jac_config::ForwardDiff.JacobianConfig, uf, du1, z, alg, i)
+  prev_chunksize = get_chunksize(jac_config)
+  ForwardDiff.JacobianConfig(uf,du1,z,ForwardDiff.Chunk{prev_chunksize}())
```
ChrisRackauckas: did you resize these things elsewhere?

kanav99: Oh yes I forgot to tell about something. https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/misc_utils.jl#L59
This line fails when `autodiff=false` as we don't have duals in `DiffEqDiffTools.JacobainCache`. `full_cache` is used in `resize!` so it fails for `Rosenbrock23(;autodiff=false)` on master. Now that I resize them explicitly, should I remove that from `cache` macro?

ChrisRackauckas: Yes

kanav99:
I will hold on this PR for now because if we remove the fields from cache because of this - https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/122#issuecomment-509880155 we would need to change something more.

kanav99:
Needs small changes in DiffEqBase (https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/297) then I will add `resize_tests.jl` to runtests.jl

ChrisRackauckas:
oh, I missed that the new tests weren't added to runtests.jl and thought they passed 🤷‍♂

kanav99:
Oh don't worry they pass anyway. I also need to make some more tests to be double sure.

PR #815, opened 2019-07-07
[WIP] Parallel 4th order DIRK 

kanav99:
Implementing algorithm in https://doi.org/10.1093/imanum/10.4.463
This is my first algorithm implementation, I will optimize this slowly. Looks like we need more changes in NLSolver for this to happen.

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -84,6 +84,21 @@ function calc_J!(integrator, cache::OrdinaryDiffEqMutableCache, is_compos)
   is_compos && (integrator.eigen_est = opnorm(J, Inf))
 end
 
+function indexed_calc_J!(integrator, cache::OrdinaryDiffEqMutableCache, is_compos, idx)
+  @unpack t,dt,uprev,u,f,p = integrator
+  J = cache.J[idx]
+  if DiffEqBase.has_jac(f)
+    f.jac(J, uprev, p, t)
+  else
+    nlsolver = cache.nlsolver[idx]
+    @unpack du1,uf,jac_config = nlsolver
+    uf.t = t
+    uf.p = p
+    jacobian!(J, uf, uprev, du1, integrator, jac_config)
+  end
+  integrator.destats.njacs += 1
+  is_compos && (integrator.eigen_est = opnorm(J, Inf))
+end
 """
     WOperator(mass_matrix,gamma,J[;transform=false])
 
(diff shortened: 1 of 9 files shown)


kanav99:
I have made special indexed versions of previous functions so that I don't break original functions by mistake. Newton methods failed to converge right now. I am wondering how do you make a first guess for Newton solvers?

ChrisRackauckas:
for now, initialize by setting `z=0`.

kanav99:
@ChrisRackauckas I dont know why this one diverges at the first nlsolve of first step itself. Can you see what I am doing wrong? I must be doing something silly

kanav99:
Non threaded version is now fine and getting nice results now. Threaded version is still a bit problem. And I get random results, that means some sort of threading mistake. I fixed the `u` issue in the nlsolver by temporarily making a new local variable.

kanav99:
First in-method threaded newton method works now! Though this takes more time than non threaded right now, I will optimize it now.

src/perform_step/pdirk_perform_step.jl
```diff
@@ -11,38 +11,36 @@ end
   @unpack nlsolver,k1,k2 = cache
   if alg.threading == true
     Threads.@threads for i in 1:2
```
ChrisRackauckas: you should put `let` blocks around this, otherwise you lose type-stability. It's a bug with the captured variables from some anonymous functions. See the other multithreaded methods for examples

src/derivative_utils.jl
```diff
@@ -396,5 +411,73 @@ function update_W!(nlsolver::NLSolver, integrator, cache::OrdinaryDiffEqConstant
...
+function indexed_update_W!(nlsolver::NLSolver, integrator, cache::OrdinaryDiffEqMutableCache, dt, idx, repeat_step)
+  if isnewton(nlsolver)
+    indexed_calc_W!(nlsolver, integrator, cache, dt, idx, repeat_step, true)
+  end
+  nothing
+end
+
+function indexed_calc_W!(nlsolver, integrator, cache::OrdinaryDiffEqMutableCache, dtgamma, idx, repeat_step, W_transform=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack J,W = cache
```
ChrisRackauckas: if you just had the `nlsolver` use the `W` within it, it should be fine without having a whole new function?

src/perform_step/pdirk_perform_step.jl
```diff
@@ -0,0 +1,79 @@
...
+        nlsolver[i].γ = 2//3
+        nlsolver[i].c = 2//3
+      end
+      k1[i] .= DiffEqBase.nlsolve!(nlsolver[i], nlsolver[i].cache, integrator)
+    end
+    nlsolvefail(nlsolver[1]) && return
+    nlsolvefail(nlsolver[2]) && return
+    Threads.@threads for i in 1:2
+      if i == 1
+        indexed_update_W!(integrator, cache, dt/2, 1, repeat_step)
```
ChrisRackauckas: you shouldn't need to update this again

src/derivative_utils.jl
kanav99: Yup it should be, but it's a workaround while W and J are in alg cache. I will remove them from these indexed versions.

kanav99:
Threaded and Non-threaded are now at the same benchmarks.

src/derivative_utils.jl
kanav99: Ah we don't have a J in NLNewtonCache right now. That would be a breaking and big change

kanav99:
I guess this is complete now, check if this needs any more changes when you are free @ChrisRackauckas @YingboMa. Though for simple `f(u,p,t) = u` there is no significant gain, this might be good when `f` takes long to evaluate. I would need an example to check.

ChrisRackauckas:
yeah, you need a large enough Jacobian for it to matter.

PR #817, opened 2019-07-09
[WIP] Eliminate p

devmotion:
I started to pass `integrator` instead of `p` when calling `f` (see https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/116). I tried to be consistent with the method signatures used for continuous callbacks (http://docs.juliadiffeq.org/latest/features/callback_functions.html#ContinuousCallbacks-1), and hence replaced `f([du, ]u, p, t)` with `f([du, ]u, t, integrator)`. Depending on whether the user wants to unpack the parameters of `integrator` (which would be default) or not, this call would be forwarded to `f.f([du, ], u, get_param(integrator), t)` or `f.f([du, ], u, t, integrator)`. [It would be possible to just forward the arguments to `f.f` if the last argument is not a `DEIntegrator` - not sure whether that would be useful but then it would still be possible for a user to evaluate `f` with a fixed set of parameters.]

However, even at this early stage of the implementation I noticed some (possible) issues:
- As mentioned in https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/116#issuecomment-508266533, I don't know how to avoid passing `p` in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L399 since that call happens before the integrator is created.
- Similarly, the ODE solution is created before the integrator exists. That seems problematic due to https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/solutions/ode_solutions.jl#L15-L16: the solution will always compute an interpolation based on the parameters of the ODE problem and not on the integrator.

The primary use case I had in mind (not knowing how well this will work in the end) was passing around the integrator while solving DDEs, to be able to access the history function more easily/directly. In this case it seems strange/insufficient if the solution would only pass around the parameters.

The problem regarding the solution could maybe be avoided if the solution is not part of the integrator, and e.g. only created in `solve!(integrator)` (which could forward to `solve!!(sol, integrator)` to allow passing an existing solution) and then passed to `loopfooter!` and `postamble!`. However, since in that case the solution could not be accessed as a field of `integrator`, it's not obvious anymore that the whole setup would be beneficial for solving DDEs.

Since I confused myself quite a bit now, I'm wondering:
- What would be a typical scenario and function `f` in an ODE setting, in which we want to pass around `integrator`?
- In such a scenario, is it reasonable to assume that the analytic solution does not depend on the integrator?
- Which other properties can we expect to not depend on the integrator but only on the problem parameters? Or could everything be dependent on `integrator`?

### DIFF
--- a/src/initdt.jl
+++ b/src/initdt.jl
@@ -1,7 +1,6 @@
 @muladd function ode_determine_initdt(u0,t,tdir,dtmax,abstol,reltol,internalnorm,prob::DiffEqBase.AbstractODEProblem{uType,tType,true},integrator) where {tType,uType}
   _tType = eltype(tType)
   f = prob.f
-  p = integrator.p
   oneunit_tType = oneunit(_tType)
   dtmax_tdir = tdir*dtmax
 
@@ -16,11 +15,11 @@
   if get_current_isfsal(integrator.alg, integrator.cache) && typeof(integrator) <: ODEIntegrator
     # Right now DelayDiffEq has issues with fsallast not being initialized
     f₀ = integrator.fsallast
-    f(f₀,u0,p,t)
+    f(f₀, u0, t, integrator)
   else
     # TODO: use more caches
     f₀ = zero.(u0./t)
-    f(f₀,u0,p,t)
+    f(f₀, u0, t, integrator)
   end
 
   # TODO: use more caches
@@ -95,7 +94,7 @@
   u₁ = similar(u0) # required by DEDataArray
   @.. u₁ = u0 + dt₀_tdir*f₀
   f₁ = similar(f₀)
-  f(f₁,u₁,p,t+dt₀_tdir)
+  f(f₁, u₁, t+dt₀_tdir, integrator)
 
   if prob.f.mass_matrix != I
     integrator.alg.linsolve(ftmp, prob.f.mass_matrix, f₁, false)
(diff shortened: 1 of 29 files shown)


devmotion:
I used the following script for a quick test:
```julia
using DiffEqBase, OrdinaryDiffEq, Test

function f_params_iip(du, u, p, t)
    du[1] = p * u[1]

    nothing
end

function f_integrator_iip(du, u, t, integrator)
    du[1] = get_params(integrator) * u[1]

    nothing
end

function test(alg)
    prob_params = ODEProblem(ODEFunction{true,true,true}(f_params_iip), [1.0], (0.0, 10.0), 1.01)
    sol_params = solve(prob_params, alg)

    prob_params2 = ODEProblem(f_params_iip, [1.0], (0.0, 10.0), 1.01)
    sol_params2 = solve(prob_params2, alg)

    @test sol_params.t == sol_params2.t && sol_params.u[1, :] == sol_params2.u[1, :]

    prob_integrator = ODEProblem(ODEFunction{true,true,false}(f_integrator_iip), [1.0], (0.0, 10.0), 1.01)
    sol_integrator = solve(prob_integrator, alg)

    @test sol_params.t == sol_integrator.t && sol_params.u[1, :] == sol_integrator.u[1, :]

    @test sol_params(0:0.1:10) == sol_integrator(0:0.1:10)

    nothing
end
```

Together with https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/287, `test(Tsit5())` runs successfully, but both `test(Vern6())` and `test(ImplicitEuler())` fail due to the problems with dense interpolations and the cache of the nonlinear solvers described above.

devmotion:
Regarding the passing of parameters when interpolating the solution: would it be reasonable to pass the whole solution object and unpack the parameters `p` if desired, similar to the approach with the integrator? Feels at least somewhat reasonable and in that way one could evaluate the solution at earlier time points if needed when evaluating `f`.

PR #826, opened 2019-07-11
Add iip vs oop test

ChrisRackauckas:


ChrisRackauckas:
@devmotion just letting you know this exists, but has some tests marked broken. This may be where to target some of the implicit changes for getting iip and oop the same for delaydiffeq.

PR #829, opened 2019-07-12
Add _savevalues! and _postamble!

devmotion:
This PR is needed for https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/124.


devmotion:
Similar to the previous addition of `_loopfooter!`, this breaks DelayDiffEq.

PR #830, opened 2019-07-13
[WIP] Remove redundant NLSolve fields from alg cache

kanav99:
Follow up for JuliaDiffEq/DiffEqBase.jl#288, tests pass after DiffEqBase is tagged. Now I need to remove all the redundant fields from alg cache.

### DIFF
--- a/src/caches/bdf_caches.jl
+++ b/src/caches/bdf_caches.jl
@@ -66,7 +66,7 @@ end
   du₂::rateType
 end
 
-@cache mutable struct SBDFCache{uType,rateType,N,F} <: OrdinaryDiffEqMutableCache
+@cache mutable struct SBDFCache{uType,rateType,N} <: OrdinaryDiffEqMutableCache
   cnt::Int
   u::uType
   uprev::uType
--- a/src/caches/bdf_caches.jl
+++ b/src/caches/bdf_caches.jl
@@ -1,5 +1,4 @@
-@cache mutable struct ABDF2ConstantCache{F,N,dtType,rate_prototype} <: OrdinaryDiffEqConstantCache
-  uf::F
+@cache mutable struct ABDF2ConstantCache{N,dtType,rate_prototype} <: OrdinaryDiffEqConstantCache
   nlsolver::N
   eulercache::ImplicitEulerConstantCache
   dtₙ₋₁::dtType
@@ -8,40 +7,25 @@ end
 
 function alg_cache(alg::ABDF2,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,
                    uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
-  γ, c = 1//1, 1
+  γ, c = 2//3, 1
   J, W = oop_generate_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits)
   nlsolver = oopnlsolve(alg,u,uprev,p,t,dt,f,W,J,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,γ,c)
-  @getoopnlsolvefields
-  eulercache = ImplicitEulerConstantCache(uf,nlsolver)
-
-  nlsolver.γ = 2//3
-  nlsolver.c = 1
+  eulercache = ImplicitEulerConstantCache(nlsolver)
 
   dtₙ₋₁ = one(dt)
   fsalfirstprev = rate_prototype
 
-  ABDF2ConstantCache(uf, nlsolver, eulercache, dtₙ₋₁, fsalfirstprev)
+  ABDF2ConstantCache(nlsolver, eulercache, dtₙ₋₁, fsalfirstprev)
 end
 
-@cache mutable struct ABDF2Cache{uType,rateType,uNoUnitsType,JType,WType,UF,JC,N,F,dtType} <: OrdinaryDiffEqMutableCache
+@cache mutable struct ABDF2Cache{uType,rateType,uNoUnitsType,N,dtType} <: OrdinaryDiffEqMutableCache
   uₙ::uType
   uₙ₋₁::uType
   uₙ₋₂::uType
-  du1::rateType
   fsalfirst::rateType
   fsalfirstprev::rateType
-  k::rateType
-  z::uType
   zₙ₋₁::uType
-  dz::uType
-  b::uType
-  tmp::uType
   atmp::uNoUnitsType
-  J::JType
-  W::WType
-  uf::UF
-  jac_config::JC
-  linsolve::F
   nlsolver::N
   eulercache::ImplicitEulerCache
   dtₙ₋₁::dtType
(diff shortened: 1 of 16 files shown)


src/derivative_utils.jl
```diff
@@ -70,6 +82,31 @@ either ForwardDiff or finite difference will be used depending on the
...
+  if isdefined(cache, :nlsolver)
+    calc_J!(cache.nlsolver, integrator, cache, is_compos)
+  elseif isdefined(cache, :J)
+    calc_J_in_cache!(integrator, cache, is_compos)
+  else
+    error("No J found in the cache")
+  end
+end
+
+function calc_J!(nlsolver::NLSolver, integrator, cache::OrdinaryDiffEqMutableCache, is_compos)
```
ChrisRackauckas: why not just pass J and have 1 function?

src/derivative_utils.jl
```diff
@@ -70,6 +82,31 @@ either ForwardDiff or finite difference will be used depending on the
 `jac_config` of the cache.
 """
 function calc_J!(integrator, cache::OrdinaryDiffEqMutableCache, is_compos)
+  if isdefined(cache, :nlsolver)
```
ChrisRackauckas: 1 function, just an `if` to get the `J`?

src/derivative_utils.jl
kanav99: That's because we unpack uf,du1,jac_config inside these functions, passing them in function too will make a lot of arguments.

kanav99:
ABDF2 don't converge right now for some reason.

src/perform_step/bdf_perform_step.jl
```diff
@@ -95,8 +95,7 @@ end
     cache.dtₙ₋₁ = dtₙ
     perform_step!(integrator, cache.eulercache, repeat_step)
     cache.fsalfirstprev .= integrator.fsalfirst
-    # why?
-    # nlsolver.tmp = tmp
+    nlsolver.tmp = tmp
```
ChrisRackauckas: Should be `.=`

kanav99: why? In fact why are we doing this? tmp is already alias of nlsolver.tmp

ChrisRackauckas: if it's already an alias then yeah, it doesn't need to be there. but I thought this was the difference between `tmp` and `cache.tmp`.

ChrisRackauckas:
Note that these changes need to be done in StochasticDiffEq too

kanav99:
See the last commit, that was the culprit :stuck_out_tongue:

kanav99:
Wow this PR actually found great bugs.

src/derivative_utils.jl
```diff
@@ -371,6 +446,35 @@ function calc_W!(integrator, cache::OrdinaryDiffEqConstantCache, dtgamma, repeat
...
+  @unpack uf = nlsolver
+  mass_matrix = integrator.f.mass_matrix
+  isarray = typeof(uprev) <: AbstractArray
+  # calculate W
+  uf.t = t
+  is_compos = typeof(integrator.alg) <: CompositeAlgorithm
+  if (f isa ODEFunction && islinear(f.f)) || (f isa SplitFunction && islinear(f.f1.f))
+    J = f.f1.f
+    W = WOperator(mass_matrix, dtgamma, J, false; transform=W_transform)
+  elseif DiffEqBase.has_jac(f)
```
ChrisRackauckas: @devmotion SDIRK tests won't pass until this matches the mutable cache one, which should be a separate PR so that we can get this one correct first.

src/perform_step/bdf_perform_step.jl
```diff
@@ -788,7 +793,7 @@ function perform_step!(integrator,cache::QNDFCache,repeat_step=false)
     @.. ϕ += γₖ[i]*D[i]
   end
   @.. ϕ *= γ
-  tm = fill!(cache.b, zero(eltype(u)))
+  tm = fill!(nlsolver.ztmp, zero(eltype(u)))
```
ChrisRackauckas: what's this about?

src/perform_step/sdirk_perform_step.jl
```diff
@@ -367,23 +373,20 @@ end
 
   # TODO: Add extrapolation
   @.. zᵧ = zprev
-  nlsolver.z = zᵧ
-
+  z .= zᵧ
   @.. tmp = uprev + d*zprev
   nlsolver.c = γ
-  zᵧ = nlsolve!(integrator, cache)
+  zᵧ .= nlsolve!(integrator, cache)
```
ChrisRackauckas: interesting. Was there an accidental harmless aliasing here before?

src/caches/bdf_caches.jl
```diff
@@ -8,73 +7,54 @@ end
...
-  W = oop_generate_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits)
-  nlsolver = oopnlsolve(alg,u,uprev,p,t,dt,f,W,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,γ,c)
-  @getoopnlsolvefields
-  eulercache = ImplicitEulerConstantCache(uf,nlsolver)
-
-  nlsolver.γ = 2//3
-  nlsolver.c = 1
+  γ, c = 2//3, 1
+  J, W = oop_generate_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits)
+  nlsolver = oopnlsolve(alg,u,uprev,p,t,dt,f,W,J,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,γ,c)
```
ChrisRackauckas: bugs me that this is W,J but I don't want to break that right now.

ChrisRackauckas:
@YingboMa take a look if you have time, this is a big one.

src/perform_step/bdf_perform_step.jl
kanav99: The field `b` of cache was actually aliased to be `nlsovler.ztmp` everywhere

src/perform_step/sdirk_perform_step.jl
kanav99: I think there were some pointer swapping things before which were dependent on the fact that we store two references to the same array therefore they worked. Now we have single reference, I would have to see if I can re-do those tricks

src/caches/bdf_caches.jl
kanav99: yeah the order should be fixed, I will fix this after the PR

src/derivative_utils.jl
devmotion: Speaking of arguments, is there any reason for passing around `is_compos`? Its value should only depend on `integrator`, so it seems a bit redundant?

kanav99:
Any reason you can see why PSOS tests fail in push but pass in PR?

ChrisRackauckas:
That's a little random. I wouldn't worry about it. Every once in awhile `push` acts weird (maybe cache issue).

devmotion:
Just noticed: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_interface.jl#L110-L111 should be updated as well, I guess

PR #831, opened 2019-07-14
Add failing GPU test with mass matrix

ranjanan:
For the Rosenbrock23 method


test/gpu/simple_gpu.jl
```diff
@@ -33,3 +33,17 @@ u0 = [1.0;0.0;0.0]
 tspan = (0.0,100.0)
 prob_num = ODEProblem(ff2,u0,tspan)
 sol = solve(prob_num,Rosenbrock23(linsolve=LinSolveGPUFactorize()))
+
+# Mass matrix example 
+CuArrays.allowscalar(false)
+function g(du, u, p, t)
+    du[1] = -p[1]*u[1] + p[2]*u[2]*u[3]
```
ChrisRackauckas: This is going to fail because of scalar indexing.

PR #835, opened 2019-07-20
Broadcast Rosenbrock solvers

YingboMa:


src/perform_step/rosenbrock_perform_step.jl
```diff
@@ -1089,7 +1089,7 @@ end
   if mass_matrix == I
     @.. linsolve_tmp = du + (dtC71*k1 + dtC72*k2 + dtC73*k3 + dtC74*k4 + dtC75*k5 + dtC76*k6)
   else
-    @.. du1 =dtC72*k2 + dtC73*k3 + dtC74*k4 + dtC75*k5 + dtC76*k6
+    @.. du1 = dtC71*k1 + dtC72*k2 + dtC73*k3 + dtC74*k4 + dtC75*k5 + dtC76*k6
```
devmotion: Are there any other issues with Rodas5? I just redid some DelayDiffEq benchmarks some days ago, and the results are almost exactly as the [ones online from end of April](http://juliadiffeq.org/DiffEqBenchmarks.jl/html/StiffDDE/QuorumSensing.html) apart from some issues and regressions with Rodas5, if I remember correctly. If I find the time, I'll create a PR to DiffEqBenchmarks later today.

PR #836, opened 2019-07-20
fix incorrect usage of f.jac

huanglangwen:


ChrisRackauckas:
We should get test coverage of this, but it's such a rare case it's not the biggest deal

PR #837, opened 2019-07-21
fast path for structured matices

ChrisRackauckas:
```julia
using LinearAlgebra, DifferentialEquations
f = (du,u,p,t) -> du .= u ./ t
jac = (J,u,p,t) -> (J[1,1] = 1/t; J[2,2] = 1/t; J)
jp = Diagonal(zeros(2))
fun = ODEFunction(f; jac=jac, jac_prototype=jp)
prob = ODEProblem(fun,ones(2),(1.0,10.0))
solve(prob,Rosenbrock23(autodiff=false))
```

Needs more tests


ChrisRackauckas:
Doesn't handle `W` correctly: what if M is dense? For now I think we just check `M isa UniformScaling`.

ChrisRackauckas:
https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/300 required

PR #840, opened 2019-07-25
Add test dependencies to Project.toml

devmotion:
Tests did not run on my computer with Julia 1.3.


devmotion:
I didn't manage to run the whole test suite before breakfast, so might still be missing a dependency from the standard library.

PR #844, opened 2019-07-26
Fix small dt change criterion

YingboMa:


YingboMa:
@devmotion could you take a look? I think this is right.

ChrisRackauckas:
Yes, I've been waiting for a benchmark

PR #850, opened 2019-08-02
Hairer Wanner regression test

saurabhkgp21:


src/integrators/controllers.jl
```diff
@@ -268,8 +268,8 @@ function step_accept_controller!(integrator,alg::ExtrapolationMidpointHairerWann
   s = integrator.cache.stage_number
 
   # Compute new order based on available quantities
-  win_min_old = n_old - 1 # cf. win_min in perfom_step! of the last step
-  tmp = win_min_old:(n_curr + 1) # Index range for the new order
+  win_min_old = min(n_old, n_curr) - 1 # cf. win_min in perfom_step! of the last step
```
ChrisRackauckas: is this the bug you were talking about in the benchmarks?

saurabhkgp21: Yes

PR #858, opened 2019-08-03
some more SDIRK GPU compatibility

ChrisRackauckas:


src/perform_step/sdirk_perform_step.jl
```diff
@@ -278,11 +278,9 @@ end
       r = c*dt^3/2 # by mean value theorem 3rd DD equals y'''(s)/6 for some s
 
       # @.. tmp = r*abs(((u - uprev)/dt1 - (uprev - uprev2)/dt2) - ((uprev - uprev2)/dt3 - (uprev2 - uprev3)/dt4)/dt5)
-      @inbounds for i in eachindex(u)
-        DD31 = (u[i] - uprev[i])/dt1 - (uprev[i] - uprev2[i])/dt2
-        DD30 = (uprev[i] - uprev2[i])/dt3 - (uprev2[i] - uprev3[i])/dt4
-        tmp[i] = r*integrator.opts.internalnorm((DD31 - DD30)/dt5,t)
-      end
+      @.. DD31 = (u - uprev)/dt1 - (uprev- uprev2)/dt2
+      @.. DD30 = (uprev - uprev2)/dt3 - (uprev2 - uprev3)/dt4
```
YingboMa: `DD31` and `DD30` are not arrays here.

src/perform_step/sdirk_perform_step.jl
```diff
@@ -278,11 +278,9 @@ end
...
+      @.. tmp = r*integrator.opts.internalnorm((DD31 - DD30)/dt5,t)
```
YingboMa:
```suggestion
      @.. tmp = r*integrator.opts.internalnorm((
          ((u - uprev)/dt1 - (uprev- uprev2)/dt2) #DD31 
          - ((uprev - uprev2)/dt3 - (uprev2 - uprev3)/dt4) #DD30
      )/dt5,t)
```

ChrisRackauckas:
KenCarp4 converges with order 4 on non-split but 3 with split non- autonomous
KenCarp5 converges with order 2 with non-autonomous non-split

@YingboMa this might be what the bad results in the Leucine model was about since it's non-autonomous.

src/nlsolve/nlsolve.jl
```diff
@@ -27,7 +27,7 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator)
       θ = ndz / ndzprev
 
       # divergence
-      if θ > 1
+      if θ > 1.2
```
ChrisRackauckas: why this?

src/nlsolve/nlsolve.jl
```diff
@@ -42,6 +42,10 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator)
     apply_step!(nlsolver, integrator)
 
     # check for convergence
+    if iter == 1 && ndz < 1e-5
```
ChrisRackauckas: was lower needed for this value?

src/nlsolve/nlsolve.jl
YingboMa: https://github.com/LLNL/sundials/blob/38aa03b990e616db52acd053730c31fcc3eb91f9/src/cvode/cvode_nls.c#L27-L31

Sundials actually uses `2`, should we bump this?

ChrisRackauckas: Yeah, we should probably adopt their "just keep trying a bit". Convergence of Newton is generally non-monotonic so assuming something close to it is doomed to extra factorizations.

src/nlsolve/nlsolve.jl
YingboMa: Lowering this value makes all convergence tests pass.

YingboMa: We can tune this value in subsequent PRs.

test/algconvergence/non-autonomous_convergence_tests.jl
```diff
@@ -0,0 +1,73 @@
...
+                                        nonauto2,
+                              analytic=analytic),
+                              u0,tspan)
+
+testTol = 0.2
+
+kwargs = (reltol=1e-16, abstol=1e-16)
+for prob in [prob1, prob2, prob3, prob4]
+  dts = 1 .//2 .^(12:-1:8)
+  sim = test_convergence(dts,prob,KenCarp3(); kwargs...)
```
ChrisRackauckas: is it not possible to get it passing without this?

YingboMa: Yes, it is necessary.

ChrisRackauckas: no changes to the nlsolver could do it?

YingboMa: It must need the tolerance information to converge to the correct accuracy. I don't see there is anything that nlsolver could do.

ChrisRackauckas: Oh yes, I see what you mean. It needs the tolerance so that when the fixed dt is small it actually converges enough. Interesting.

PR #859, opened 2019-08-05
Different SDIRK2 method

### LINKED ISSUE #842: Different SDIRK2
Not much to it. This would be a good training method.

[TPWRD.2008.923397.pdf](https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/files/3435425/TPWRD.2008.923397.pdf)

NikStoyanov:
Add a different SDIRK2 method. Closes #842.

```julia
using OrdinaryDiffEq, Plots
f(u,p,t) = 1.01*u
u0=1/2
tspan = (0.0,1.0)
prob = ODEProblem(f,u0,tspan)
sol = solve(prob,SDIRK22())
plot(sol, xaxis="Time (t)", yaxis="u(t)", label="SDIRK22")
plot!(sol.t, t->0.5*exp(1.01t), line=3, ls=:dash, label="Analytical")
```

Gives:

image


src/caches/sdirk_caches.jl
```diff
@@ -176,6 +176,64 @@ function alg_cache(alg::SDIRK2,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUn
...
+  uhold::U
+  rhs::R
+  nl_rhs::NR
+  tab::Tab
+end
+
+function alg_cache(alg::SDIRK22,u,rate_prototype,uEltypeNoUnits,tTypeNoUnits,uBottomEltypeNoUnits,
+                   uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{false}})
+  # 1-1/√2
+  a = 0.29289321881345254
```
ChrisRackauckas: we should keep it using the analytical values and generate the tableau values using `eltype(u)`, so that way it support precision higher than Float64

src/perform_step/sdirk_perform_step.jl
```diff
@@ -512,6 +514,138 @@ end
...
+  rhs.t = t
+  rhs.dt = dt
+  rhs.a = dt
+  nlres = alg.nlsolve(nl_rhs,uhold)
+  uhold[1] = nlres[1]
+  integrator.fsallast = f(uhold[1],p,t+dt)
+  u = uhold[1]
+
+  if integrator.opts.adaptive && integrator.success_iter > 0
+    # local truncation error (LTE) bound by dt^2/2*max|y''(t)|
```
ChrisRackauckas: does this error estimator make sense for this method?

src/caches/sdirk_caches.jl
NikStoyanov: Ok, I will change to analytical values and use ```eltype(u)``` for the type.

src/perform_step/sdirk_perform_step.jl
NikStoyanov: I think the method can be seen as two consecutive implicit Euler calculations, so I looked at how it was done for ```GenericImplicitEuler``` here:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/abaf387213e1b1b020b795242cf4bd5ca2a950d8/src/perform_step/generic_implicit_perform_step.jl#L68

Is there a better way to do this?

ChrisRackauckas: Is that the full leading truncation error term?

NikStoyanov: I see, the method is 2nd order so the leading error term is

```julia
1/12 dt^3 max|y'''(t)|
```

I will update to reflect this.

ChrisRackauckas: Take a look at the error estimator of the Trapezoidal method for inspiration. It's a little unstable to estimate the third derivative like that, but without a better error estimate that's about the best you can do.

ChrisRackauckas:
Test failure looks real. Let me know if you'd like some help!

src/caches/sdirk_caches.jl
```diff
@@ -176,6 +176,64 @@ function alg_cache(alg::SDIRK2,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUn
...
+  # -√2
+  α = -1.4142135623730951
+  # 1+√2
+  β = 2.414213562373095
+  tab = (a,α,β)
+
+  tmp = similar(u); atmp = similar(u,uEltypeNoUnits)
+  k = zero(rate_prototype)
+  dual_cache = DiffCache(u,Val{determine_chunksize(u,get_chunksize(alg.nlsolve))})
+  rhs = ImplicitRHS(f,tmp,t,t,t,dual_cache,p)
```
ChrisRackauckas: I think you looked at the wrong thing for inspiration. Take a look at Trapezoid, not GenericXXX

NikStoyanov: I just had a look at how the newton method is implemented and it makes sense now. I will follow the Trapezoid method and update the code when I get some time. Thanks for the help!

NikStoyanov:
Rebased to keep the history clean and pushed the modified algorithm. Local convergence test are ok, just waiting for the full test set on travis now.

src/perform_step/sdirk_perform_step.jl
```diff
@@ -513,6 +515,151 @@ end
...
+  γdt = γ*dt
+  update_W!(integrator, cache, γdt, repeat_step)
+  nlsolver.tmp = uprev + γdt*integrator.fsalfirst
+  z = nlsolve!(integrator, cache)
+  nlsolvefail(nlsolver) && return
+  u = nlsolver.tmp
+
+  if integrator.opts.adaptive
+    if integrator.iter > 2
+      # local truncation error (LTE) bound by dt^3/12*max|y'''(t)|
```
ChrisRackauckas: I'm not sure this LTE is correct, since that's the one for Trapezoid? What's the true LTE? That would change one of the constants in here, which isn't a big deal but is probably worth mentioning.

ChrisRackauckas:
Thanks!

PR #862, opened 2019-08-06
Add DAE support for outplace version of ROSW, ROS4, RODAS and ROS34/44 class solvers

huanglangwen:


huanglangwen:
Rosenbrock23 class is a bit confusing, I'm not quite sure where to add mass_matrix.

ChrisRackauckas:
Add these to the mass matrix tests

ChrisRackauckas:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/test/interface/mass_matrix_tests.jl

For Rosenbrock23, the in-place one is done so that should make it more clear. It's not W-transformed so it looks a little different (we should W-transform it though,...)

PR #863, opened 2019-08-08
DAEProblem Support

kanav99:
Very unoptimized implementation of DImplicitEuler for now, we might need a general `nlsolver` or maybe NLsolve.jl might suffice for these use cases.


src/derivative_wrappers.jl
```diff
@@ -129,7 +129,7 @@ function jacobian!(J::AbstractMatrix{<:Number}, f, x::AbstractArray{<:Number}, f
     nothing
 end
 
-function DiffEqBase.build_jac_config(alg::OrdinaryDiffEqAlgorithm,f,uf,du1,uprev,u,tmp,du2)
+function DiffEqBase.build_jac_config(alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm},f,uf,du1,uprev,u,tmp,du2)
```
ChrisRackauckas: I could imagine this having to be different for DAEs?

src/perform_step/dae_perform_step.jl
```diff
@@ -0,0 +1,36 @@
...
+  alg = unwrap_alg(integrator, true)
+
+  # initial guess
+  guess = tmp
+  guess .= zero(eltype(u))
+
+  zero_func! = (out, x) -> begin
+                f(out, (x .- uprev) ./ dt, x, p, t)
+              end
+  u .= NLsolve.nlsolve(zero_func!, guess).zero
```
ChrisRackauckas: The hard part will be trying to make this use our same nlsolver setup.

ChrisRackauckas:
I like it, clean. Let's try to get this working and through, and in another PR try to handle the Jacobians and nlsolvers better.

src/derivative_wrappers.jl
kanav99: currently useless, we dont even use them in DAEs

src/perform_step/dae_perform_step.jl
kanav99: yeah, some new overload of `nlsolve!` is required right now because the current setup expects explicit form very hard.

src/alg_utils.jl
```diff
@@ -1,8 +1,8 @@
-isautodifferentiable(alg::OrdinaryDiffEqAlgorithm) = true
+isautodifferentiable(alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm}) = true
```
devmotion: Would it make sense to rename `OrdinaryDiffEqAlgorithm` to `ODEAlgorithm` for consistency? And introduce an alias for `Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm}`?

src/perform_step/dae_perform_step.jl
```diff
@@ -0,0 +1,36 @@
...
+@muladd function perform_step!(integrator, cache::DImplicitEulerCache, repeat_step=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack atmp,tmp = cache
+  alg = unwrap_alg(integrator, true)
+
+  # initial guess
+  guess = tmp
+  guess .= zero(eltype(u))
+
+  zero_func! = (out, x) -> begin
```
devmotion: Not sure if it's worth adjusting this if the nlsolver setup is changed in the near future, but this closure could be fixed using the let block trick.

src/alg_utils.jl
kanav99: Agreed. What should we call the Union?

devmotion: If `ODEAlgorithm` is introduced, maybe just `OrdinaryDiffEqAlgorithm` since `ODEAlgorithm` and `DAEAlgorithm` are defined by OrdinaryDiffEq?

src/perform_step/dae_perform_step.jl
ChrisRackauckas: yeah I wouldn't worry about this. The goal will be to quickly get this working with the standard Jacobian/W/nlsolver setup.

kanav99:
Now the inplace version is complete. Out-of-place solving is not supported by NLsolve.jl so I guess it needs our own nlsolver to work. I am working on a DiffEqBase PR for new routines.

PR #868, opened 2019-08-12
Fixes for changes of non-linear solvers

devmotion:
This PR contains fixes for upstream changes in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/315.

It should be easy to remove the duplicated implementations of `calc_J` etc in `derivative_utils.jl` by just passing the ODE cache or the NLSolver cache but I think it's better to do that in a separate PR.

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -40,7 +40,7 @@ module OrdinaryDiffEq
 
   using DiffEqBase: check_error!, @def, @.. , _vec, _reshape
 
-  using DiffEqBase: nlsolvefail, isnewton, set_new_W!, get_W, iipnlsolve, oopnlsolve, @getiipnlsolvefields, @getoopnlsolvefields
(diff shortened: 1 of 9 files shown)


ChrisRackauckas:
looks fine, we just need to do the bounding correctly.

devmotion:
> looks fine, we just need to do the bounding correctly.

It seems we should upper-bound the version of DiffEqBase in OrdinaryDiffEq (and StochasticDiffEq), release a new version of OrdinaryDiffEq (and StochasticDiffEq), merge the PR to DiffEqBase, release a new version of DiffEqBase, and then add a lower bound for that version to OrdinaryDiffEq (and StochasticDiffEq)? 

Can/should we upper-bound DiffEqBase for previous releases in the Registry?

ChrisRackauckas:
>Can/should we upper-bound DiffEqBase for previous releases in the Registry?

That's hard... can the other dispatch just exist for making the transition easier, and then lower bound?

devmotion:
Even if we keep a dispatch, it will be breaking since we moved fields from the solver to the cache. There's no way that any old version will work with a new DiffEqBase release.

ChrisRackauckas:
true, okay.

PR #869, opened 2019-08-12
Add unitless cache for Rosenbrock methods

devmotion:
See https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/315#issuecomment-520432589.


huanglangwen:
Need to update generic Rosebrock methods as well.

PR #872, opened 2019-08-14
ImplictiEulerExtrapolation parallel

saurabhkgp21:


src/caches/extrapolation_caches.jl
```diff
@@ -70,10 +70,12 @@ end
 @cache mutable struct ImplicitEulerExtrapolationCache{uType,rateType,arrayType,dtType,JType,WType,F,JCType,GCType,uNoUnitsType,TFType,UFType} <: OrdinaryDiffEqMutableCache
   uprev::uType
   u_tmp::uType
+  u_tmps::Array{uType,1}
   utilde::uType
   tmp::uType
   atmp::uNoUnitsType
   k_tmp::rateType
```
ChrisRackauckas: do you also need this other one then?

src/caches/extrapolation_caches.jl
```diff
@@ -70,10 +70,12 @@ end
 @cache mutable struct ImplicitEulerExtrapolationCache{uType,rateType,arrayType,dtType,JType,WType,F,JCType,GCType,uNoUnitsType,TFType,UFType} <: OrdinaryDiffEqMutableCache
   uprev::uType
   u_tmp::uType
```
ChrisRackauckas: ?

src/caches/extrapolation_caches.jl
```diff
@@ -89,7 +91,8 @@ end
   tf::TFType
   uf::UFType
   linsolve_tmp::rateType
```
ChrisRackauckas: ?

src/derivative_utils.jl
```diff
@@ -392,6 +392,57 @@ function calc_W!(integrator, cache::OrdinaryDiffEqMutableCache, dtgamma, repeat_
   return nothing
 end
 
+function calc_W!(integrator, cache::OrdinaryDiffEqMutableCache, dtgamma, repeat_step, W_index::Int, W_transform=false)
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack J,W = cache
```
ChrisRackauckas: this is just silly. Why not just pass in `W` into these routines? It would fix these problems.

ChrisRackauckas:
I'll merge because this is correct, but it should become a lot cleaner. However, @kanav99 is going through the W calculations right now for DAEs, so he might want to handle that transition to passing in `W` instead of indexing it from a cache.

src/perform_step/extrapolation_perform_step.jl
```diff
@@ -332,23 +344,33 @@ function perform_step!(integrator,cache::ImplicitEulerExtrapolationConstantCache
...
-        integrator.destats.nsolve += 1
-        u_tmp = u_tmp + k
-        k_copy = f(u_tmp, p, t+j*dt_temp)
+  let max_order=max_order, dt=dt, integrator=integrator, cache=cache, repeat_step=repeat_step,
+    uprev=uprev, T=T
+    Threads.@threads for i in 1:2
+      startIndex = (i==1) ? 1 : max_order
+      endIndex = (i==1) ? max_order-1 : max_order
+      for index in startIndex:endIndex
+        dt_temp = dt/(2^(index-1)) # Romberg sequence
```
ChrisRackauckas: why Romberg?

src/caches/extrapolation_caches.jl
```diff
@@ -121,10 +124,21 @@ end
 
 function alg_cache(alg::ImplicitEulerExtrapolation,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Type{Val{true}})
   u_tmp = similar(u)
+  u_tmps = Array{typeof(u_tmp),1}(undef, Threads.nthreads())
+
+  for i=1:Threads.nthreads()
+    u_tmps[i] = zero(u_tmp)
```
ChrisRackauckas: one improvement that here is just to make the first ones in the arrays equal, i.e. `u_tmps[1] = u_tmp`. Then the algorithm would be fine. Right now you have an extra thing for each one, so you have an extra Jacobian, and that's really bad memory-wise.

src/derivative_utils.jl
ChrisRackauckas: @kanav99 @huanglangwen this is something we should follow up on. It's because our original `calc_W!` was never intended for thread-safety, and so to make it thread-safe there is this code duplication. We should assume this case can happen in the next refactor.

ChrisRackauckas:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/875 explains the cache cleanup I was mentioning with some code.

src/derivative_utils.jl
huanglangwen: Does it mean `calc_W!` should have an additional parameter of *readonly* `W` ?

ChrisRackauckas: it should have `J and `W` I think, where it writes into `W` for a given `J`, instead of pulling them from the cache (since there may be more than 1)

huanglangwen: But it won't work on OOP.

ChrisRackauckas: For OOP it will need to return W.

kanav99: I think we should have a pseudo-nlsolver kind of thing for algorithms which don't have an actual nlsolver like Rosenbrocks, the struct should have aliases to the `J` and `W` and similar stuff, should be `<: NLSolver` so that we have a common implementation of the functions. This way it won't make us change the original derivative utilities everytime we make a new algorithm just for an overload.

kanav99: Note that algorithms with nlsolver struct are already well handled for parallel applications.

PR #873, opened 2019-08-14
add J, W cache and linsolve interface for rosenbrock oop methods and add J cache for firk oop

huanglangwen:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/871

### DIFF
--- a/src/caches/rosenbrock_caches.jl
+++ b/src/caches/rosenbrock_caches.jl
@@ -67,7 +67,7 @@ function alg_cache(alg::Rosenbrock23,u,rate_prototype,uEltypeNoUnits,uBottomElty
   J,W = iip_generate_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits)
   tmp = zero(rate_prototype)
   atmp = similar(u, uEltypeNoUnits)
-  tab = Ros23ConstantCache(real(uBottomEltypeNoUnits))
+  tab = Rosenbrock23Tableau(real(uBottomEltypeNoUnits))
   tf = DiffEqDiffTools.TimeGradientWrapper(f,uprev,p)
   uf = DiffEqDiffTools.UJacobianWrapper(f,t,p)
   linsolve_tmp = zero(rate_prototype)
@@ -95,7 +95,7 @@ function alg_cache(alg::Rosenbrock32,u,rate_prototype,uEltypeNoUnits,uBottomElty
   J,W = iip_generate_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits)
   tmp = zero(rate_prototype)
   atmp = similar(u, uEltypeNoUnits)
-  tab = Ros32ConstantCache(real(uBottomEltypeNoUnits))
+  tab = Rosenbrock32Tableau(real(uBottomEltypeNoUnits))
 
   tf = DiffEqDiffTools.TimeGradientWrapper(f,uprev,p)
   uf = DiffEqDiffTools.UJacobianWrapper(f,t,p)
@@ -117,7 +117,7 @@ struct Rosenbrock23ConstantCache{T,TF,UF,JType,WType,F} <: OrdinaryDiffEqConstan
 end
 
 function Rosenbrock23ConstantCache(T::Type,tf,uf,J,W,linsolve)
-  tab = Ros23ConstantCache(T)
+  tab = Rosenbrock23Tableau(T)
   Rosenbrock23ConstantCache(tab.c₃₂,tab.d,tf,uf,J,W,linsolve)
 end
 
@@ -140,7 +140,7 @@ struct Rosenbrock32ConstantCache{T,TF,UF,JType,WType,F} <: OrdinaryDiffEqConstan
 end
 
 function Rosenbrock32ConstantCache(T::Type,tf,uf,J,W,linsolve)
-  tab = Ros32ConstantCache(T)
+  tab = Rosenbrock32Tableau(T)
   Rosenbrock32ConstantCache(tab.c₃₂,tab.d,tf,uf,J,W,linsolve)
 end
 
--- a/src/caches/rosenbrock_caches.jl
+++ b/src/caches/rosenbrock_caches.jl
@@ -67,7 +67,7 @@ function alg_cache(alg::Rosenbrock23,u,rate_prototype,uEltypeNoUnits,uBottomElty
   J,W = iip_generate_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits)
   tmp = zero(rate_prototype)
   atmp = similar(u, uEltypeNoUnits)
-  tab = Rosenbrock23ConstantCache(real(uBottomEltypeNoUnits),identity,identity,false,false)
+  tab = Rosenbrock23ConstantCache(real(uBottomEltypeNoUnits),identity,identity,nothing,nothing,nothing)
   tf = DiffEqDiffTools.TimeGradientWrapper(f,uprev,p)
   uf = DiffEqDiffTools.UJacobianWrapper(f,t,p)
   linsolve_tmp = zero(rate_prototype)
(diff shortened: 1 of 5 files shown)


devmotion:
I think we should not introduce an additional implementation for generating `J` (and `W`) but rather use/modify the existing
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/cd45039e49eefdc5d60aa72b54f91e84851eb747/src/integrators/integrator_utils.jl#L386
and
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/cd45039e49eefdc5d60aa72b54f91e84851eb747/src/integrators/integrator_utils.jl#L412

huanglangwen:
It seems the oopgenW is broken when `u` is a StaticArray. Not sure how to fix it.
https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/571800687

huanglangwen:
It seems the oopgenW is broken when `u` is a StaticArray. Not sure how to fix it.
https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/571800687

devmotion:
I guess it's the first time that someone actually tests the whole logic of `oop_generate_W`. I think
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/cd45039e49eefdc5d60aa72b54f91e84851eb747/src/integrators/integrator_utils.jl#L426-L432
is incorrect, since column vectors with only one entry end up in the second branch. I assume it should be changed to
```julia
J = if ndims(u) == 2 # `u` is already a matrix
        u
      else # `u` is a vector
        hcat(u, u)
      end
```
Can you check if that solves the issue?

To be honest, it still feels very crude to me, the units are incorrect and will fail if the dimension of `u` is greater than 2.

ChrisRackauckas:
Aren't we already vecing `u` when applying the linear operator? If not, we should, and in that case `false .* rate_prototype .* rate_prototype'` should work, ever for `u` a Number.

devmotion:
> Aren't we already vecing u when applying the linear operator? If not, we should, and in that case false .* rate_prototype .* rate_prototype' should work, ever for u a Number.

I don't see how this is related to the test failure? But I agree that probably we should just replace the snippet above and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/cd45039e49eefdc5d60aa72b54f91e84851eb747/src/integrators/integrator_utils.jl#L438 with
```julia
J = false .* _vec(u) .* _vec(u)'
```
That would be much cleaner.

Then we could just branch depending on the type of `u` for the computation of `W` (if we do not want to compute one LU factorization, see https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/672).

ChrisRackauckas:
The real difficulty is that we don't have an OOP linsolve interface, which we need to get in order to really complete this I think?

devmotion:
And the fact that `lu(J)` will fail since `J` is singular, as I just noticed...

huanglangwen:
> 
> 
> And the fact that `lu(J)` will fail since `J` is singular, as I just noticed...

Your patch works anyway. Shall I push that?

devmotion:
> And the fact that lu(J) will fail since J is singular, as I just noticed...

Actually, it seems `lu(@SMatrix zeros(2,2))` works (and only `lu(zeros(2,2))` fails), so I think you could even write
```julia
J = false .* _vec(u) .* _vec(u)'
W = if u isa StaticArray
  lu(J)
  elseif u isa Number
    u
  else
    LU{LinearAlgebra.lutype(uEltypeNoUnits)}(Matrix{uEltypeNoUnits}(undef, 0, 0),
                                                 Vector{LinearAlgebra.BlasInt}(undef, 0),
                                                 zero(LinearAlgebra.BlasInt))
  end
```
instead of https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/1330fa7a01626a4c334d171a619802475be8d036/src/integrators/integrator_utils.jl#L432-L448

huanglangwen:
> 
> 
> The real difficulty is that we don't have an OOP linsolve interface, which we need to get in order to really complete this I think?

What's the motivation to move from `k₃ = _reshape(W\-_vec(linsolve_tmp), axes(uprev))`?

YingboMa:
`W\b` is wasteful to do at every stage. A better strategy is to factorize it first and reuse the factorized object. Hence we need to add an OOP linsolve interface.

YingboMa:
> ~W\b is wasteful to do at every stage. A better strategy is to factorize it first and reuse the factorized object.~

Never mind. We want to use factorization methods other than `lu`, so we need it.

src/integrators/integrator_utils.jl
```diff
@@ -384,59 +384,63 @@ DiffEqBase.nlsolve_f(integrator::ODEIntegrator) =
...
-      if ArrayInterface.isstructured(f.jac_prototype) || f.jac_prototype isa SparseMatrixCSC
-        J = similar(f.jac_prototype)
-        W = similar(J)
-      elseif DiffEqBase.has_jac(f) && !DiffEqBase.has_Wfact(f) && f.jac_prototype !== nothing
-        J = nothing
-        W = WOperator(f, dt, true)
-      else
-        J = false .* vec(u) .* vec(u)'
-        W = similar(J)
+  if isdefined(alg,:nlsolve)
```
YingboMa:
```suggestion
  if alg isa NewtonAlgorithm
```

src/caches/rosenbrock_caches.jl
```diff
@@ -64,19 +64,10 @@ function alg_cache(alg::Rosenbrock23,u,rate_prototype,uEltypeNoUnits,uBottomElty
...
-    J = nothing # is J = W.J better?
-  else
-    J = false .* vec(rate_prototype) .* vec(rate_prototype)' # uEltype?
-    W = similar(J)
-  end
+  J,W = iip_generate_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits)
   tmp = zero(rate_prototype)
   atmp = similar(u, uEltypeNoUnits)
-  tab = Rosenbrock23ConstantCache(real(uBottomEltypeNoUnits),identity,identity)
+  tab = Rosenbrock23ConstantCache(real(uBottomEltypeNoUnits),identity,identity,nothing,nothing,nothing)
```
devmotion: I always thought that we shouldn't misuse the oop cache as tableau but rather use separate tableau structs - and even more so with the new fields for `J` and `W`.

ChrisRackauckas: yeah, it's really just a short hand that always worked... until it didn't.

huanglangwen:
Is this pr good for merge? I want to start a new pr for unifying oop and iip in `derivative_utils.jl` .

src/caches/rosenbrock_caches.jl
```diff
@@ -67,7 +67,7 @@ function alg_cache(alg::Rosenbrock23,u,rate_prototype,uEltypeNoUnits,uBottomElty
   J,W = iip_generate_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits)
   tmp = zero(rate_prototype)
   atmp = similar(u, uEltypeNoUnits)
-  tab = Rosenbrock23ConstantCache(real(uBottomEltypeNoUnits),identity,identity,nothing,nothing,nothing)
+  tab = Ros23ConstantCache(real(uBottomEltypeNoUnits))
```
devmotion: I think I would prefer the name `Rosenbrock23Tableau`.

huanglangwen: I agree, but then we need to correct all the tab struct in `\tableaus` as well as places calling the tabs in `\caches`

src/caches/rosenbrock_caches.jl
```diff
@@ -95,7 +95,7 @@ function alg_cache(alg::Rosenbrock32,u,rate_prototype,uEltypeNoUnits,uBottomElty
   J,W = iip_generate_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits)
   tmp = zero(rate_prototype)
   atmp = similar(u, uEltypeNoUnits)
-  tab = Rosenbrock32ConstantCache(real(uBottomEltypeNoUnits),identity,identity,nothing,nothing,nothing)
+  tab = Ros32ConstantCache(real(uBottomEltypeNoUnits))
```
devmotion: Similar to my comment above, I think we should use something a bit more meaningful than `Ros32ConstantCache`.

src/caches/rosenbrock_caches.jl
devmotion: Sure, there might be some other structs which probably should be renamed as well but that can be done in a different PR, I guess. For now, IMO it's sufficient to name the new struct in a reasonable way :smiley:

ChrisRackauckas:
Let's continue the renames as another PR.  Also note that in the sphere of influence for this should be @saurabhkgp21's newest implicit extrapolation methods. For those methods, we need a safe-hatch so that way Jacobian reuse is just generally turned off. In other cases, having an easy switch on it is nice anyways, so it's something to keep in mind.

PR #876, opened 2019-08-16
[WIP] Simplify `derivative_utils.jl`

huanglangwen:
Continuing from https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/873 and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/871 . The goal is to simplify four `calc_W!` functions into two groups: one for rosenbrock, one for nlnewton. `calc_W!` should behave the same for OOP and IIP.

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -51,6 +51,8 @@ function calc_J(integrator, cache::OrdinaryDiffEqConstantCache, is_compos)
   if DiffEqBase.has_jac(f)
     J = f.jac(uprev, p, t)
   else
+    cache.uf.t = t
+    cache.uf.p = p
     J = jacobian(cache.uf,uprev,integrator)
   end
   integrator.destats.njacs += 1
@@ -63,6 +65,8 @@ function calc_J(nlsolver, integrator, cache::OrdinaryDiffEqConstantCache, is_com
   if DiffEqBase.has_jac(f)
     J = f.jac(uprev, p, t)
   else
+    nlsolver.uf.t = t
+    nlsolver.uf.p = p
     J = jacobian(nlsolver.uf,uprev,integrator)
   end
   integrator.destats.njacs += 1
@@ -107,14 +111,7 @@ function calc_J!(nlsolver::NLSolver, integrator, cache::OrdinaryDiffEqMutableCac
 end
 
 function calc_J!(nlsolver::NLSolver, integrator, cache::OrdinaryDiffEqConstantCache, is_compos)
-  @unpack t,dt,uprev,u,f,p = integrator
-  if DiffEqBase.has_jac(f)
-    nlsolver.cache.J = f.jac(uprev, p, t)
-  else
-    nlsolver.cache.J = jacobian(nlsolver.uf,uprev,integrator)
-  end
-  integrator.destats.njacs += 1
-  is_compos && (integrator.eigen_est = opnorm(nlsolver.cache.J, Inf))
+  nlsolver.cache.J = calc_J(nlsolver,integrator,cache,is_compos)
 end
 
 function calc_J_in_cache!(integrator, cache::OrdinaryDiffEqMutableCache, is_compos)
@@ -133,14 +130,7 @@ function calc_J_in_cache!(integrator, cache::OrdinaryDiffEqMutableCache, is_comp
 end
 
 function calc_J_in_cache!(integrator, cache::OrdinaryDiffEqConstantCache, is_compos)
-  @unpack t,dt,uprev,u,f,p = integrator
-  if DiffEqBase.has_jac(f)
-    cache.J = f.jac(uprev, p, t)
-  else
-    cache.J = jacobian(cache.uf,uprev,integrator)
-  end
-  integrator.destats.njacs += 1
-  is_compos && (integrator.eigen_est = opnorm(cache.J, Inf))
+  cache.J = calc_J(integrator,cache,is_compos)
 end
 
 """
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
IMO one change we should make while doing this is that the W we want to calculate on should be passed in. See https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/872 for what happens when doing multithreaded implicit methods. That extra method should just be deleted when this is fixed.

devmotion:
As soon as the algorithm-specific fields of the non-linear solvers are moved to the cache, there's no need anymore for separate `calc_J` and `calc_J!` methods (see https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/868). It's not possible right now, but that's something we should keep in mind.

src/derivative_utils.jl
```diff
@@ -106,6 +106,17 @@ function calc_J!(nlsolver::NLSolver, integrator, cache::OrdinaryDiffEqMutableCac
   is_compos && (integrator.eigen_est = opnorm(J, Inf))
 end
 
+function calc_J!(nlsolver::NLSolver, integrator, cache::OrdinaryDiffEqConstantCache, is_compos)
```
devmotion: There already exists an implementation for out-of-place function, it's called `calc_J`. So if we need an in-place variant `calc_J!` (which I'm not sure about) we could just define `calc_J!` as `cache.J = calc_J(...)`

src/derivative_utils.jl
```diff
@@ -106,6 +106,17 @@ function calc_J!(nlsolver::NLSolver, integrator, cache::OrdinaryDiffEqMutableCac
...
+  @unpack t,dt,uprev,u,f,p = integrator
+  if DiffEqBase.has_jac(f)
+    nlsolver.cache.J = f.jac(uprev, p, t)
+  else
+    nlsolver.cache.J = jacobian(nlsolver.uf,uprev,integrator)
```
devmotion: We should update `uf` first, similar to the other implementations of `calc_J!`. This is currently missing in the out-of-place implementations `calc_J`.

src/derivative_utils.jl
```diff
@@ -121,6 +132,17 @@ function calc_J_in_cache!(integrator, cache::OrdinaryDiffEqMutableCache, is_comp
   is_compos && (integrator.eigen_est = opnorm(J, Inf))
 end
 
+function calc_J_in_cache!(integrator, cache::OrdinaryDiffEqConstantCache, is_compos)
```
devmotion: As I said, I'm not sure we want to do this.

src/derivative_utils.jl
huanglangwen: But why does `calc_J` produce the right answer without updating `uf` ?

devmotion: Since it's updated in `calc_W!` in lines such as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/dcff0bd6809adc640ce910ddbeeba8ed05091431/src/derivative_utils.jl#L525 (but not for `p`, so that's still not correct). I think we should remove those updates in `calc_W!` since the update should only happen if we actually need `uf`, which is only the case in `calc_J`.

src/derivative_utils.jl
```diff
@@ -51,6 +51,8 @@ function calc_J(integrator, cache::OrdinaryDiffEqConstantCache, is_compos)
   if DiffEqBase.has_jac(f)
     J = f.jac(uprev, p, t)
   else
+    cache.uf.t = t
+    cache.uf.p = p
```
devmotion: Great! :+1:  Can you remove the updates of `uf` in the oop version of `calc_W!`?

huanglangwen: I'm thinking of delete oop version of `cache_W!` directly.

huanglangwen:
I'm a bit confused by the `calc_W!`. Rosenbrock and extrapolation methods call `calc_W!(integrator, cache, dtgamma, repeat_step, W_transform)` while methods with nlnewton call `calc_W!(nlsolver, integrator, cache, dt, repeat_step, true)`. So why the first `calc_W!` has so many logic for newton methods like `W_dt = isnewton ? cache.nlsolver.cache.W_dt : dt` ?

src/derivative_utils.jl
devmotion: And only use `update_W!`? Currently I don't like that `calc_W!` is called `calc_W!` both for in-place and out-of-place functions but updates `W` only for the in-place functions and returns a new `W` for the out-of-place functions. So I thought maybe it could be done similar to `calc_J` and `calc_J!`, i.e., we could have an out-of-place variant `calc_W` and an updating variant `calc_W!`.

Moreover, I'm not happy with the fact that at the moment there's a lot of reuse strategies hidden in the implementation of `calc_W!`. Maybe it would be cleaner if `calc_W!` would just calculate `W`, without considering any reuse strategies, similar to what we do in `calc_J`. Ideally then we would have a separate function (such as `update_W!`) that first figures out if `J` should be updated (and then calls `calc_J!`) and then if `W` should be updated (and then calls `calc_W!`).

devmotion:
It was one implementation in the beginning and then just copied to a new function when the non-linear solvers were changed. I'm sure many parts can be improved.

src/derivative_utils.jl
huanglangwen: Agree.

huanglangwen:
Why is `f.Wfact` called every time when `calc_W!` is called?

kanav99:
Check out #878 if that helps us out for this patch

huanglangwen:
I guess I'll wait for https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/878 . @kanav99 has done most of the work and we just need to append some oop functions to it.

ChrisRackauckas:
and this intersects with https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/325 . @devmotion might want to take note of all of these and make sure that works with these 3 planned changes.

devmotion:
The non-linear solver changes will break the accesses `nlsolver.uf` but there's a bunch of things that we have to fix after the DiffEqBase update, so I would go on and merge this PR in it's current state. Tests pass and oop functions can be added in a separate PR, IMO.

devmotion:
OK, let's merge this and continue improving derivative_utils.jl in separate PRs.

PR #877, opened 2019-08-16
Rename tableau struct for rosenbrock methods

huanglangwen:
Following https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/873. Methods other than Rosenbrock use tableau struct as constant cache, so no changes are made.

### DIFF
--- a/src/caches/rosenbrock_caches.jl
+++ b/src/caches/rosenbrock_caches.jl
@@ -205,7 +205,7 @@ function alg_cache(alg::ROS3P,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
   J,W = iip_generate_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits)
   tmp = zero(rate_prototype)
   atmp = similar(u, uEltypeNoUnits)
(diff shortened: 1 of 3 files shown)


devmotion:
Wouldn't it be cleaner to have a tableau struct and a constant cache struct, where the second one potentially only has the tableau as single field?

huanglangwen:
I'm happy with that for now, because I'm not sure the impact of changing cache struct. We need at least change the `perform_step!` .

devmotion:
I guess, it should be quite straightforward and basically we would only have to unpack the tableau in the same way in the oop and iip implementations of `perform_step!` and `add_steps!`. But I can do that in a separate PR.

ChrisRackauckas:
Something introduced a test failure?

devmotion:
Maybe the interpolations dispatch incorrectly, now that the constant caches were renamed?

huanglangwen:
I got no error on my computer. Perhaps it's a random one, and I guess it's QDNF2's problem.

devmotion:
Tests pass now on Travis as well, so I merge. Thanks!

PR #878, opened 2019-08-16
[Experimental] Uses PseudoNLSolver

kanav99:
Just a small patch to show how we can remove redundant code which is due to Rosenbrocks and other methods where we don't need nlsolver, but need jacobian stuff. (Ref https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/872#discussion_r314663684) 
This introduces `PseudoNLSolver` which only acts as a container of certain variables. It is not used for solving non-linear equations. 
If this is approved, I can move rosenbrocks to this struct too and then we can delete a lot of code from `derivative_utils.jl`


devmotion:
But the separate struct would only be useful for threaded algorithms where we might have multiple Jacobians etc, right? I mean, for the Rosenbrock methods we can just pass the Rosenbrock cache itself, since they already contain these fields. That's what I wanted to try, at least, to get rid of all these duplicated methods. Basically, I thought it would only be necessary to pass either the cache of the non-linear solver (after we're done with the PR to DiffEqBase) or the Rosenbrock etc cache to the `calc_J!` etc methods.

ChrisRackauckas:
If we go this route, just call it SemiImplicit which is the name in the literature.

kanav99:
Agreed @devmotion. That would fix the threading problem too I guess. Should I close this?

ChrisRackauckas:
I say go for this, since handling it in an nlsolver might make it easier to do W-reuse? Though we need a switch for that since the extrapolation methods and Rosenbrock non-W shouldn't do W-reuse at all, given that they have to factorize for the different orders. @huanglangwen will need to make note of that.

kanav99:
Okay. I think it wont be tough to make further changes for jacobian reuse stuff as we use `isnewton` flags which already handle such functions.
I strongly want all of the algorithms to have a same interface and this PR does that.
I will make further changes now including renaming the struct to `SemiImplicitNLSolver`.

kanav99:
Depends on JuliaDiffEq/DiffEqBase.jl#315 to work, resize fails for now. After this nlsolver changes are done, resize need to be made general (to work for all algos)

src/derivative_utils.jl
```diff
@@ -1,3 +1,26 @@
+mutable struct SemiImplicitNLSolver{WType,JType,du1Type,ufType,jcType}
```
devmotion: I guess, `NLSemiImplicitCache` (or something along those lines) would be more consistent with DiffEqBase. It's not an alternative to `NLSolver` but just a stripped down version of `NLNewtonCache`, it seems?

devmotion:
There's some test error, it seems.

src/derivative_utils.jl
kanav99: It is basically just a cache, just a container, but it needs to be saved by the name `nlsolver` in cache because it is used as a nlsolver in `derivative_utils.jl`. Its like making `derivative_utils.jl` think like this is an actual nlsolver so it fetches the required fields in the same way it would fetch from a nlsolver.

kanav99:
The error is due to resize, I would need a better resize and luckily the resize changes in JuliaDiffEq/DiffEqBase.jl#315 is all what I need.

src/derivative_utils.jl
devmotion: I think we should make it a cache instead, `derivative_utils.jl` shouldn't have to work with the solvers at all after the changes in DiffEqBase. There's nothing contained anymore in `nlsolver` that could be of interest to `derivative_utils.jl` apart from the cache, I think?

kanav99: No, problem lies for threaded algorithms. We can have multiple nlsolvers so specifying nlsolver is needed.

ChrisRackauckas:
This is somewhat weird because, unlike any of the other nonlinear solvers, you cannot actually swap this around. The algorithms which are semi-implicit are not necessarily higher order if you use Newton instead, and the Newton algorithms are not necessarily higher order if you use semi-implicit forms. So while the nonlinear solvers are an option, you wouldn't want anyone to choose this option in the methods that have nonlinear solver options, and for the semi-implicit methods you wouldn't even want to expose the option and instead hardcode it to be this. That makes me think something is going wrong here and there might be a nicer way to achieve the parallelism. For example, why not have the `generate_W` etc. functions just require the user to pass `J` and optionally `W` for in-place?

kanav99:
What do you mean by exposing? We are not allowing user to interact with SemiImplicitNLSolver, just like NLSolver. In NLSolver we allow the interaction with the algorithm only (NLNewton, NLFunctional, NLAnderson). For SemiImplicitNLSolver, we don't have any algorithm. So it's basically hardcoded.
For the case of handling threaded algorithms, we would need atleast one extra parameter, be it W or some index.

src/derivative_utils.jl
devmotion: Hmm, I don't see why you.would.want to pass around the (Pseudo)NLsolver structs. To me it seems passing multiple caches should.be sufficient.

PR #891, opened 2019-08-22
Upper bound DiffEqBase and ExponentialUtilities

devmotion:
Add semver compatible upper bounds for DiffEqBase and ExponentialUtilities (see https://julialang.github.io/Pkg.jl/v1/compatibility/#Version-specifier-format-1).

Should we do the same even for dependencies with 0.x version numbers? And should we update the registry for old releases?


ChrisRackauckas:
Past versions need the upper bound too. @stefankarpinski what's the right way to do that with the new registry setup?

StefanKarpinski:
Make a PR to the registry editing the Compat.toml files there by hand.

devmotion:
Should we also drop support for old versions of StaticArrays and DataStructures?

ChrisRackauckas:
I think so. At whatever point they put the 1.0 we should go with that, but until then we just stay up to date.

devmotion:
@ChrisRackauckas Good to merge now?

ChrisRackauckas:
Yes, but without editing the Compat.toml we cannot tag.

PR #896, opened 2019-08-23
Remove Int64

devmotion:
Fixes https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/895.

### DIFF
--- a/src/caches/extrapolation_caches.jl
+++ b/src/caches/extrapolation_caches.jl
@@ -220,11 +220,11 @@ function create_extrapolation_coefficients(::Type{T},alg::algType) where {T,algT
 
   # Initialize subdividing_sequence:
   if sequence == :harmonic
-      subdividing_sequence = [BigInt(n+1) for n = 0:n_max]
+      subdividing_sequence = BigInt.(1:(n_max + 1))
   elseif sequence == :romberg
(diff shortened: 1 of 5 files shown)


src/caches/extrapolation_caches.jl
```diff
@@ -547,11 +568,11 @@ end
...
   Q::Vector{QType} # Storage for stepsize scaling factors. Q[n] contains information for extrapolation order (n - 1)
-  n_curr::Int64 # Storage for the current extrapolation order
-  n_old::Int64 # Storage for the extrapolation order n_curr before perfom_step! changes the latter
+  n_curr::Int # Storage for the current extrapolation order
+  n_old::Int # Storage for the extrapolation order n_curr before perfom_step! changes the latter
   coefficients::extrapolation_coefficients
-  stage_number::Vector{Int64} # stage_number[n] contains information for extrapolation order (n - 1)
-  sigma::Rational{Int64} # Parameter for order selection
+  stage_number::Vector{Int} # stage_number[n] contains information for extrapolation order (n - 1)
+  sigma::Rational{Int} # Parameter for order selection
```
ChrisRackauckas: if this shows up in profiles then it should just be changed to a floating point type

PR #899, opened 2019-08-24
Set seed before loading problems

devmotion:
From time to time we see spurious test errors, e.g., in https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/575695972. Hopefully we can get rid of them by setting the random seed before loading the ODE test problems (the initial condition of the 2D problem is [chosen randomly](https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/blob/master/src/ode/ode_linear_prob.jl#L59-L61)).


devmotion:
Gitlab CI fails due to the incorrectly configured extrapolation tests. That's unrelated to this PR.

PR #903, opened 2019-08-25
Commit message CI options

kanav99:
I just made a small utility where you can tell Travis what to do with tests using commit messages.

1. If commit message starts with `base:<name-of-branch> commit message`, it will checkout a branch of DiffEqBase and then the tests will run. Forks are still a question. Though there is a way out I know of.
2. If commit message starts with `notest:` the tests will not run, saving CI time.

Might come handy at a cost of having ugly commit messages.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -5,6 +5,19 @@ const GROUP = get(ENV, "GROUP", "All")
 const is_APPVEYOR = Sys.iswindows() && haskey(ENV,"APPVEYOR")
 const is_TRAVIS = haskey(ENV,"TRAVIS")
 
+using LibGit2
+r = LibGit2.GitRepo("..")
+commit = LibGit2.peel(LibGit2.GitCommit, LibGit2.head(r))
+msg = LibGit2.message(commit)
+
+if msg[1:5] == "base:"
+  using Pkg
+  lastidx = findfirst(isequal(' '),msg) - 1
+  branch = msg[6:lastidx]
+  Pkg.add(PackageSpec(name="DiffEqBase", rev=branch))
+elseif msg[1:7] == "notest:"
+  exit(0)
+end
 #Start Test Script
 
 @time begin


devmotion:
>     If commit message starts with notest: the tests will not run, saving CI time.
>
> Might come handy at a cost of having ugly commit messages.

Isn't that exactly what `[skip ci]` is for? That's supported by both [Travis](https://docs.travis-ci.com/user/customizing-the-build#skipping-a-build) and [Appveyor](https://www.appveyor.com/docs/how-to/filtering-commits/#skip-directive-in-commit-message), and allows one to skip running tests.

devmotion:
> If commit message starts with base:<name-of-branch> commit message, it will checkout a branch of DiffEqBase and then the tests will run. Forks are still a question. Though there is a way out I know of.

I'm interested in what's the motivation for doing that?

Recently I started to believe that we really should just stick to having a proper dependency management. If we introduce a breaking change in `DiffEqBase` or any other package, we should just update the version from `x.y.z` to `(x+1).0.0`; if it's a new feature in a backward-compatible manner the update should instead be to `x.(y+1).0`, and if it's a bugfix release it should be `x.y.(z+1)`. The `0.x.y` versions are slightly different, according to [semver](https://semver.org/) breaking changes may be introduced at any point and the public API should not considered to be stable. I guess, most consistent with Pkg's handling of caret specifiers would be to use `0.(x+1).0` for any release that's not a bugfix release.

If the compatible versions in OrdinaryDiffEq and other packages are specified in a reasonable way by not allowing all future versions of the dependencies, then breaking changes can be introduced in all our dependencies without breaking OrdinaryDiffEq (BTW, the same thing should go for OrdinaryDiffEq and DelayDiffEq - any breaking change in OrdinaryDiffEq requires a new major release since otherwise we're constantly chasing bugs in DelayDiffEq that might already appear in production environments without us knowing about it). Then we can take the time to review if the breaking changes affect our package, provide fixes, or just update the compatibility bounds.

In that sense, IMO we shouldn't run CI tests with DiffEqBase master since regular users will never use OrdinaryDiffEq together with DiffEqBase master. I think, if a PR requires a new feature in DiffEqBase, we should just release a new version of DiffEqBase that we can test against (and, of course, update the lowest compatible version of DiffEqBase).

kanav99:
The rationale behind this feature is that we have to release DiffEqBase just to test OrdinaryDiffEq in case when we have two parallel PRs in both of the repositories. I get that we should have a stable way to release, but having a way to test with the PR you are working on the other repository is good as you don't have to wait for it getting merged and tagged. Plus good old commit messages work right too.
I personally have worked with lot of breaking changes this summer and this might an end to that for me

devmotion:
> The rationale behind this feature is that we have to release DiffEqBase just to test OrdinaryDiffEq in case when we have two parallel PRs in both of the repositories. I get that we should have a stable way to release, but having a way to test with the PR you are working on the other repository is good as you don't have to wait for it getting merged and tagged. Plus good old commit messages work right too.
I personally have worked with lot of breaking changes this summer and this might an end to that for me

I agree, this problem is a bit annoying. I had to deal with it also quite regularly, often PRs in DelayDiffEq depend on some changes in OrdinaryDiffEq or DiffEqBase. Usually I tend to test the relevant parts locally (mostly focused on things that could be affected by the changes) and just reopen the dependent PR after the upstream change is merged to rerun the CI tests. It's not ideal but at least it enforces tests against released versions.

Another idea would be to add a Manifest.toml that points to the desired branch of the test dependency in a separate commit, and revert that commit before the PR is merged. Then the CI tests should run with the desired upstream changes, and it wouldn't pollute the commit messages (although you would mess a bit with the git history). It seems with your suggested changes, I would have to add `base:...` to every commit in the PR that I want to test?

test/runtests.jl
```diff
@@ -5,6 +5,38 @@ const GROUP = get(ENV, "GROUP", "All")
...
+r = LibGit2.GitRepo("..")
+commit = LibGit2.peel(LibGit2.GitCommit, LibGit2.head(r))
+msg = LibGit2.message(commit)
+
+@show msg[1:6]
+@show msg[7:18]
+@show msg[7:12]
+@show msg[15:18]
+
+if length(msg) >= 5 && msg[1:5] == "base:"
```
devmotion: It's generally not a good idea to base indexing of a String on `length`. As mentioned in [the docs](https://docs.julialang.org/en/v1/manual/strings/index.html#String-Basics-1)
> `lastindex(str)` is not in general the same as `length(str)` for a string, because some Unicode characters can occupy multiple "code units".

In such cases [regular indexing might fail](https://docs.julialang.org/en/v1/manual/strings/index.html#Unicode-and-UTF-8-1).

In general, I guess `split` and `startswith` might be helpful for what you want to check.

kanav99:
Sorry I missed your messages. I have adressed the issue of non-uniform indexing due to unicode characters. I used regex to match and that might solve the issue.
> Another idea would be to add a Manifest.toml that points to the desired branch of the test dependency in a separate commit, and revert that commit before the PR is merged. Then the CI tests should run with the desired upstream changes, and it wouldn't pollute the commit messages (although you would mess a bit with the git history). 

Only problem with this is accidental merging of the Manifest file. Also the Manifest file is gitignored so we might need to have our own custom toml file for this case.

Also before merging this, we need to see why LibGit2 is not availible on Gitlab runner. It is a stdlib right?

devmotion:
It has to be added to the dependencies. That's not a Gitlab specific issue but it only shows up there since only on Gitlab we test with Julia 1.2 which requires that all used stdlibs are also specified as dependencies.

kanav99:
Funny error and even funnier suggested fix:
```
ERROR: LoadError: ArgumentError: Package Pkg not found in current path:
- Run `import Pkg; Pkg.add("Pkg")` to install the Pkg package.
```

kanav99:
Help needed @devmotion !
```
ERROR: Unsatisfiable requirements detected for package Pkg [fe1c5a76]:
 Pkg [fe1c5a76] log:
 ├─Pkg [fe1c5a76] has no known versions!
 └─restricted to versions * by an explicit requirement — no versions left
```

devmotion:
The UUID seems to be incorrect :open_mouth: At least we use a different one in DiffEqBase in https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/Project.toml#L43 (where we also run `using Pkg` in runtests.jl).

kanav99:
I ran this:
```julia-repl
julia> Pkg.METADATA_compatible_uuid("Pkg")
UUID("fe1c5a76-5840-53d2-82f9-288dd83ce2ce")
```

devmotion:
> I ran this:

I think the correct way to add a dependency (and especially the stdlibs, it seems) is to activate the checked out branch and to run `] add Pkg`. That will update Project.toml automatically.

kanav99:
All done. Requesting final reviews @devmotion @ChrisRackauckas.

devmotion:
I'm still not convinced we should actually do this. I see that it might be annoying sometimes if you depend on upstream changes, but I think the proper way is to either first include the changes in the downstream packages and move them upstream at a later point if they seem to work fine or just test changes locally and wait for an upstream release before running the CI tests.

I'm just worried that testing against unreleased branches might make the whole dependency mess that we currently have even worse. I think we should really try to specify SemVer-based version bounds for all dependencies - otherwise we will not get rid of the problem that old releases break all the time due to changes in one of their dependencies. That happened, e.g., multiple times with the recent OrdinaryDiffEq releases ([see my comment in this PR](https://github.com/JuliaRegistries/General/pull/2883)) since both dependencies had no compatibility bounds in OrdinaryDiffEq but also dependencies such as DiffEqBase or DiffEqProblemLibrary did not properly increase their version numbers and/or specify which packages they are compatible with. I think we should start getting rid of these issues by at least improving the situation in the JuliaDiffEq packages, and I'm pretty sure that people whose packages depend on OrdinaryDiffEq or other DiffEq packages will be very happy if breaking changes are properly reflected by the version numbers.

Put it shortly, my gut feeling is that this way of "manipulating" the CI tests is not really in line with that goal.

kanav99:
Agreed. I just think this could be helpful to run the "one last test" before merging. We should definitely have a smooth transition using deprecations and nice versioning systems.

PR #913, opened 2019-09-01
Relax the bound of Parameters.jl

YingboMa:
https://github.com/JuliaNLSolvers/Optim.jl/blob/40e939b1d5ab6c07dd0352c009c90983bf32d02b/Project.toml#L29


Project.toml
```diff
@@ -34,7 +34,7 @@ GenericSVD = "0.2"
 MacroTools = "0.5"
 MuladdMacro = "0.2.1"
 NLsolve = "4"
-Parameters = "0.11"
+Parameters = ">= 0.10.0"
```
devmotion: I'm not a fan of `>=` bounds anymore, since any breaking updates might break our packages then. I'd suggest 
```suggestion
Parameters = "0.10, 0.11"
```

PR #915, opened 2019-09-01
comment on nlsolvers

ChrisRackauckas:
This is essentially a comment on the nlsolvers and what needs to be done. This kind of stuff should get merged into https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/912 . Generally, there is absolutely no reason to have separate branches of everything for nlsolving vs everything else: it's a ridiculous amount of code duplication that is fixed by just passing in one value in almost every single function. 

Also, the linearity handling is messed up: it should just use the fact that matrix-free operators are a `WOperator`, and know that there is another form of linearity where things are arrays. Currently it always assumes linear means `WOperator`. The go-to statements are then unnecessary.

However, when dealing with linearity, the first Jacobian update never occurs, so it needs to be tracked so that way it's done in iteration 1.

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -46,28 +46,14 @@ jacobian update function, then it will be called for the update. Otherwise,
 either ForwardDiff or finite difference will be used depending on the
 `jac_config` of the cache.
 """
-function calc_J(integrator, cache::OrdinaryDiffEqConstantCache, is_compos)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Solved by https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/927

PR #917, opened 2019-09-03
Fix downstream ambiguities

ChrisRackauckas:


ChrisRackauckas:
Should fix https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/497

src/caches/basic_caches.jl
```diff
@@ -10,9 +10,16 @@ struct ODEChunkCache{CS} <: OrdinaryDiffEqConstantCache end
   current::Int
 end
 
-function alg_cache(alg::CompositeAlgorithm,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,iip)
+function alg_cache(alg::CompositeAlgorithm,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Val{false})
```
devmotion: Would
```suggestion
function alg_cache(alg::A,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,iip) where {A<:CompositeAlgorithm}
```
fix the issue as well? Then the code duplication wouldn't be needed.

ChrisRackauckas: That's probably still ambiguous, since other algorithms are dispatching on the second part.

PR #919, opened 2019-09-05
no more callback in prob

ChrisRackauckas:


ChrisRackauckas:
Failing test was the downstream test which is expected until this 6.0 update propagates.

PR #921, opened 2019-09-07
[ci skip] Benchmark

ChrisRackauckas:
@DiffEqBot runbenchmarks

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,8 @@ platform:
   #- x86 # 32-bit
   - x64 # 64-bit
 
+
+
 # Uncomment the following lines to allow failures on nightly julia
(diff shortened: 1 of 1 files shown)


kanav99:
@DiffEqBot runbenchmarks

kanav99:
DiffEqBot doesn't run on descriptions.

devmotion:
On master TRBDF2 also shows regressions compared with June 😞

ChrisRackauckas:
@yingboma I think mentioned a few things about how some of the Jacobian reuse improvements hurt small matrix benchmarks, and this must be hitting that. At least it looks like it's not this PR. We need to really take another look at that whole reuse logic anyways.

PR #922, opened 2019-09-07
Correctly recognize failed nonlinear solves

devmotion:
This PR fixes a possible issue of the current implementation in which `nlsolvefail(nlsolver) = false` might occur even if `fail_convergence = true`, since `nlsolver.status` was only set in case of convergence, divergence, and very slow convergence but kept unchanged if the maximum number of iterations was exceeded.

In my tests with the Jacobian and mass matrix tests (see https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/920#issue-314477109), the errors and statistics are not affected by these changes. This change might lead to a early exit of `perform_step!` now (as intended), but that would not be captured by these simple tests. For other problems and algorithms it might even change the statistics and number of accepted/rejected steps.

### DIFF
--- a/src/nlsolve/nlsolve.jl
+++ b/src/nlsolve/nlsolve.jl
@@ -12,12 +12,10 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator)
 
   initialize!(nlsolver, integrator)
   η = initial_η(nlsolver, integrator)
+  nlsolver.status = SlowConvergence
 
   local ndz
-  fail_convergence = true
-  iter = 0
-  while iter < maxiters
-    iter += 1
+  for iter in 1:maxiters
     nlsolver.iter = iter
 
     # compute next step and calculate norm of residuals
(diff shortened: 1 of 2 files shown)


src/nlsolve/nlsolve.jl
```diff
@@ -12,12 +12,10 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator)
 
   initialize!(nlsolver, integrator)
   η = initial_η(nlsolver, integrator)
+  nlsolver.status = SlowConvergence
```
devmotion: Should `SlowConvergence` be renamed?

ChrisRackauckas:
Before we do this, let's tune the performance regression benchmarks. I assume we are going to have 15 or so of these PRs and we need to see what's happening. The basic report should add TRBDF2. Then for the work precision diagrams we should really be mostly tracking Group1: Rosenbrock23, TRBDF2, CVODE_BDF Group2: Rodas4, Rodas5, KenCarp4, RadauIIA, and CVODE_BDF (as a baseline) on Orego, Pollution, and HIRES. Maybe throw in an LSODA baseline as well.

@kanav99 can we get that setup? Then we can just put PRs in and throw them to the benchmarks we really care about. It would basically automate a lot of the process @YingboMa did in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/713

kanav99:
@DiffEqBot runbenchmarks

devmotion:
@DiffEqBot runbenchmarks

devmotion:
I can't spot any regressions in the benchmarks compared with master: https://github.com/DiffEqBot/Reports/blob/master/OrdinaryDiffEq.jl/925/79ad1f5d6100e62ed9f8b1993d2298940199c763/1568063863895.md

PR #925, opened 2019-09-09
[ci skip] benchmark

kanav99:


kanav99:
@DiffEqBot runbenchmarks

ChrisRackauckas:
This looks good, though I would make one change. I would get rid of lsoda from Pollu, and keep a comment in there that it is because it's too slow and messes with the scaling of the plot.

devmotion:
Was this intended to be merged at all? I thought this PR was just opened for running the benchmarks.

ChrisRackauckas:
Oops, I thought this was the same one that changed the benchmark methods. Anyways, it's not a big deal this merged.

PR #927, opened 2019-09-13
Only use WOperator on things which aren't arrays

ChrisRackauckas:
Don't merge, this is just to show something


src/derivative_utils.jl
```diff
@@ -462,7 +462,9 @@ function calc_W(integrator, cache, dtgamma, repeat_step, W_transform=false)
   if (f isa ODEFunction && islinear(f.f)) || (f isa SplitFunction && islinear(f.f1.f))
     J = f.f1.f
```
devmotion: BTW this line is wrong (noticed some time ago but forgot to change it)

src/derivative_utils.jl
```diff
@@ -462,7 +462,9 @@ function calc_W(integrator, cache, dtgamma, repeat_step, W_transform=false)
   if (f isa ODEFunction && islinear(f.f)) || (f isa SplitFunction && islinear(f.f1.f))
```
devmotion: Should a check for `SplitAlgorithm`s be added, similar to the `calc_W!` version?

devmotion:
The test error indicates that the initial definition of `W` (in `build_J_W`, I guess) is not consistent anymore with the update logic in `calc_W!`.

ChrisRackauckas:
Yeah I won't be able to dig into this for a bit.

src/derivative_utils.jl
ChrisRackauckas: Checking for SplitFunction is probably better.

devmotion: In `calc_W!` (and some other places, if I remember correctly) we check for both

src/derivative_utils.jl
```diff
@@ -287,12 +303,12 @@ function do_newJ(integrator, alg::T, cache, repeat_step)::Bool where T # any cha
   return !fastconvergence
 end
 
-function do_newW(integrator, nlsolver::T, new_jac, W_dt)::Bool where T # any changes here need to be reflected in FIRK
-  integrator.iter <= 1 && return true
+function do_newW(integrator, nlsolver, new_jac, W_dt)::Bool # any changes here need to be reflected in FIRK
+  nlsolver === nothing && return true
```
ChrisRackauckas: I guess we ignore RosW for now.

src/derivative_utils.jl
```diff
@@ -482,7 +446,8 @@ end
 
 function calc_rosenbrock_differentiation!(integrator, cache, dtd1, dtgamma, repeat_step, W_transform)
   calc_tderivative!(integrator, cache, dtd1, repeat_step)
-  calc_W!(cache.W, integrator, cache, dtgamma, repeat_step, W_transform)
+  nlsolver = nothing
+  calc_W!(cache.W, integrator, nlsolver, cache, dtgamma, repeat_step, W_transform)
```
ChrisRackauckas: can we delete a method from this change as well?

src/derivative_utils.jl
```diff
@@ -521,7 +486,10 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{true})
       return nothing, nothing
     end
   end
-  if ArrayInterface.isstructured(f.jac_prototype) || f.jac_prototype isa SparseMatrixCSC
+  n = length(u)
+  if f.jac_prototype !== nothing && size(f.jac_prototype) != (n, n) ||
```
ChrisRackauckas: This is what we need to think more carefully about. Right now it's:

If it's some interesting array type, similar it
Otherwise if it has a Jacobian function on it and the jac_prototype is given, then it must be some weird array type so let's go with that operator
Otherwise it's the basic case.

But the middle case: do you need a Jacobian function to be a proper WOperator equation?

src/derivative_utils.jl
```diff
@@ -521,7 +486,10 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{true})
       return nothing, nothing
```
ChrisRackauckas: why would we treat Newton completely differently in this function? The same linearity rules and all of that would apply to split Rosenbrock. This seems completely unnecessary.

src/derivative_utils.jl
```diff
@@ -521,7 +486,10 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{true})
...
     end
   end
-  if ArrayInterface.isstructured(f.jac_prototype) || f.jac_prototype isa SparseMatrixCSC
+  n = length(u)
+  if f.jac_prototype !== nothing && size(f.jac_prototype) != (n, n) ||
+                                    ArrayInterface.isstructured(f.jac_prototype) ||
+                                    f.jac_prototype isa SparseMatrixCSC
     J = similar(f.jac_prototype)
     W = similar(J)
```
ChrisRackauckas: Anything happening in this function should get reflected in the OOP one

ChrisRackauckas:
@kanav99 should join this conversation as well. We need to make sure we finally get this part correct.

src/derivative_utils.jl
```diff
@@ -287,12 +300,12 @@ function do_newJ(integrator, alg::T, cache, repeat_step)::Bool where T # any cha
...
 
-function do_newW(integrator, nlsolver::T, new_jac, W_dt)::Bool where T # any changes here need to be reflected in FIRK
-  integrator.iter <= 1 && return true
+function do_newW(integrator, nlsolver, new_jac, W_dt)::Bool # any changes here need to be reflected in FIRK
+  nlsolver === nothing && return true
   new_jac && return true
   # reuse W when the change in stepsize is small enough
   dt = integrator.dt
-  new_W_dt_cutoff = T <: NLSolver ? nlsolver.cache.new_W_dt_cutoff : #= FIRK =# nlsolver.new_W_dt_cutoff
+  new_W_dt_cutoff = nlsolver isa NLSolver ? nlsolver.cache.new_W_dt_cutoff : #= FIRK =# nlsolver.new_W_dt_cutoff
```
ChrisRackauckas: this is a little odd as a way to find FIRK, but that can get cleaned up later.

src/derivative_utils.jl
```diff
@@ -372,78 +387,18 @@ function calc_W!(W, integrator, cache, dtgamma, repeat_step, W_transform=false)
...
-  if (f isa ODEFunction && islinear(f.f)) || (integrator.alg isa SplitAlgorithms && f isa SplitFunction && islinear(f.f1.f))
-    new_jac = false
-    @goto J2W # Jump to W calculation directly, because we already have J
-  end
-
   # check if we need to update J or W
-  W_dt = isnewton ? cache.nlsolver.cache.W_dt : dt # TODO: RosW
-  new_jac = isnewton ? do_newJ(integrator, alg, cache, repeat_step) : true
-  new_W = isnewton ? do_newW(integrator, cache.nlsolver, new_jac, W_dt) : true
+  W_dt = nlsolver === nothing ? dt : nlsolver.cache.W_dt # TODO: RosW
```
ChrisRackauckas: I get `nlsolver === nothing` is a check for Rosenbrock, but the code might read better if we somehow formalize this notion to semi-implicit nlsolver. FIRK is another special case. We should follow up with @kanav99's there in a future PR. CC @devmotion

src/derivative_utils.jl
```diff
@@ -456,11 +411,12 @@ end
...
   # calculate W
-  is_compos = typeof(integrator.alg) <: CompositeAlgorithm
-  if (f isa ODEFunction && islinear(f.f)) || (f isa SplitFunction && islinear(f.f1.f))
-    J = f.f1.f
+  is_compos = integrator.alg isa CompositeAlgorithm
+  islin, isode = islinearfunction(integrator)
+  if islin
+    J = isode ? f.f : f.f1.f # unwrap the Jacobian accordingly
     W = WOperator(mass_matrix, dtgamma, J, false; transform=W_transform)
   elseif DiffEqBase.has_jac(f)
```
ChrisRackauckas: missing `if Wfact` and OOP Wfact

src/derivative_utils.jl
```diff
@@ -351,14 +364,16 @@ end
...
-  @unpack J = cache
+  lcache = nlsolver === nothing ? cache : nlsolver.cache
+  @unpack J = lcache
   alg = unwrap_alg(integrator, true)
   mass_matrix = integrator.f.mass_matrix
   is_compos = integrator.alg isa CompositeAlgorithm
   isnewton = alg isa NewtonAlgorithm
 
+  # handle Wfact
   if W_transform && DiffEqBase.has_Wfact_t(f)
```
ChrisRackauckas: Do these even if a new `W` might not be needed? I guess the assumption is that it's cheap enough so that if it exists, then you might as well skip jacobian reuse, but that's not true with DiffEqGPU which wants to use `Wfact` but also have Jacobian reuse in its large factorizations.

ChrisRackauckas:
This is a straightforward improvement, so I am merging based on that, but some of those threads I would like to complete.

PR #931, opened 2019-09-18
Remove DiffEqBase git magic

devmotion:
Fixes https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/929.


devmotion:
Tests fail since currently ModellingToolkit is broken: the call in https://github.com/JuliaDiffEq/ModelingToolkit.jl/blob/9b8e9c44db17f9cfa7acdca1d1acfd46e6514942/src/systems/diffeqs/diffeqsystem.jl#L237 causes an error in https://github.com/JuliaDiffEq/ModelingToolkit.jl/blob/9b8e9c44db17f9cfa7acdca1d1acfd46e6514942/src/systems/diffeqs/diffeqsystem.jl#L184.

kanav99:
Can we do some kind of try-catch construct instead?

devmotion:
I'm not sure if it's worth trying to fix the current implementation. IMO using a dedicated Manifest.toml file is the safer, more general, and more future-proof approach.

devmotion:
Tests are still broken and require ModelingToolkit 0.7.2 with https://github.com/JuliaDiffEq/ModelingToolkit.jl/pull/179

devmotion:
Tests pass now, so hopefully no additional regressions are present in ModelingToolkit and/or OrdinaryDiffEq.

test/interface/jacobian_tests.jl
```diff
@@ -46,7 +46,7 @@ end
 
 prob = ODEProblem(lotka,[1.0,1.0],(0.0,1.0),[1.5,1.0,3.0,1.0])
 de = ModelingToolkit.modelingtoolkitize(prob)
-prob2 = remake(prob, f=ODEFunction(de..., Val(false), Wfact=true))
+prob2 = remake(prob, f=ODEFunction(de...; Wfact=true))
```
ChrisRackauckas: should add `jac=true`

devmotion:
Gitlab errors are unrelated and should be fixed by https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/932

devmotion:
Any additional comments or suggestions?

ChrisRackauckas:
nope looks good

PR #932, opened 2019-09-19
Fix calculation of W in ImplicitEulerExtrapolation method

devmotion:


devmotion:
This change affects only the Gitlab tests, the test errors are due to the changes in ModelingToolkit and fixed by https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/931.

PR #933, opened 2019-09-19
Single Broadcast Loop in Low Storage 2N methods

kanav99:
In the loop inside each time step, broadcast calls are reduced to one.

Issue: Stricter limitation on function `f`, for every index of `u`, it needs exactly one `setindex!` call i.e setting `du[i]` only once.

### DIFF
--- a/src/perform_step/low_storage_rk_perform_step.jl
+++ b/src/perform_step/low_storage_rk_perform_step.jl
@@ -60,10 +60,11 @@ end
   end
   # other stages
   for i in eachindex(A2end)
-    @.. tmp = A2end[i]*tmp
     if williamson_condition
+      wrapper.coefficient = A2end[i]
       f(wrapper, u, p, t+c2end[i]*dt)
     else
+      @.. tmp = A2end[i]*tmp
       f(k, u, p, t+c2end[i]*dt)
       @.. tmp += dt * k
     end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
I think that limitation makes sense. Make sure we get that documented.

src/perform_step/low_storage_rk_perform_step.jl
```diff
@@ -60,10 +60,11 @@ end
   end
   # other stages
   for i in eachindex(A2end)
-    @.. tmp = A2end[i]*tmp
```
ChrisRackauckas: this change needs to be done for every low storage method?

src/perform_step/low_storage_rk_perform_step.jl
```diff
@@ -76,14 +77,15 @@ end
...
 
-mutable struct WilliamsonWrapper{kType, dtType}
+mutable struct WilliamsonWrapper{kType, dtType, coffType}
   kref::kType
   dt::dtType
+  coefficient::coffType
 end
 
-@inline Base.setindex!(a::WilliamsonWrapper{kType, dtType}, b::bType, c::cType) where {kType, dtType, bType, cType} = (a.kref[c] += a.dt * b)
+@inline Base.setindex!(a::WilliamsonWrapper{kType, dtType}, b::bType, c::cType) where {kType, dtType, bType, cType} = (a.kref[c] = a.coefficient * a.kref[c] + a.dt * b)
```
devmotion: What's the purpose of all these type annotations and where clauses here? It seems only `WilliamsonWrapper` is needed?

src/perform_step/low_storage_rk_perform_step.jl
```diff
@@ -76,14 +77,15 @@ end
...
 @inline Base.size(a::WilliamsonWrapper{kType, dtType}) where {kType, dtType} = size(a.kref)
-@inline Base.copyto!(a::WilliamsonWrapper{kType, dtType}, b::bType) where {kType, dtType, bType} = @.. a.kref += a.dt * b
+@inline Base.copyto!(a::WilliamsonWrapper{kType, dtType}, b::bType) where {kType, dtType, bType} = @.. a.kref = a.coefficient * a.kref + a.dt * b
```
devmotion: Same here and the line above

src/perform_step/low_storage_rk_perform_step.jl
kanav99: We have WilliamsonWrapper working for all 2N methods, we would need similar wrappers for other methods.

kanav99:
Can I in any way check the performance gain/loss somewhere for GPU use case? Can I do it in Gitlab?

ChrisRackauckas:
This is MPI+GPU, so we'd need to do a bit more.

PR #948, opened 2019-10-17
Double the small ODE solver speed and remove usage of libm

ChrisRackauckas:
For small enough ODEs, the `^` operation in the PI-controller actually dominates the runtime. You can see it in the profiles. This is why that was done with `@fastmath` which is the only portion of the ODE solver that does that, and thus reduces reproducibility. This replaces the fast `^` with an actual `^` approximation, and the effect is essentially a 2x speedup on small equations:

```julia
using OrdinaryDiffEq, BenchmarkTools
function lorenz(du,u,p,t)
 du[1] = 10.0(u[2]-u[1])
end
u0 = [1.0;0.0;0.0]
tspan = (0.0,100.0)
@btime sol = solve(prob,Tsit5())

#=
Fast:
Length: 1318
400.499 μs (66 allocations: 6.27 KiB)

^:
Length: 1287
729.701 μs (66 allocations: 6.27 KiB)
=#

using Profile
Profile.clear()
@profile for i in 1:100 sol = solve(prob,Tsit5(),save_everystep=false) end
Juno.profiler()
```

The question I have is whether it's accurate enough to be relied on (https://discourse.julialang.org/t/slow-arbitrary-base-exponentiation-a-b/25386/20) but since it's just for an error estimate calculation which already is very hand-wavy it really don't need to be that accurate (just "good enough")

### DIFF
--- a/src/integrators/controllers.jl
+++ b/src/integrators/controllers.jl
@@ -35,14 +35,39 @@ end
   q
 end
 
+# https://github.com/etheory/fastapprox/blob/master/fastapprox/src/fastlog.h
+function fastlog2(x::Float32)::Float32
+    y = Float32(reinterpret(Int32, x))
+    y *= 1.1920928955078125f-7
+    y - 126.94269504f0
+end
+function fastlog2(x::Float64)::Float32
+   fastlog2(Float32(x))
+end
+
+# https://github.com/etheory/fastapprox/blob/master/fastapprox/src/fastexp.h
+function fastpow2(x::Float32)::Float32
+    clipp = ifelse(x < -126.0f0,-126.0f0,x)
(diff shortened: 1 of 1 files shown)


YingboMa:
percent_error

```julia
julia> using Plots

julia> percent_error(x, y) = (res=x^y; (abs(abs(fastpow(x,y))-res)/res)*100)
percent_error (generic function with 1 method)

julia> plt = contour(0.001:0.001:1, 0.08:0.001:0.5, percent_error, aspect_ratio=1, xscale=:log10, color=:viridis, dpi=200)

julia> savefig("$(homedir())/Desktop/percent_error.png")
```
The error seems too high. In some places, the error is over 100%.

PR #956, opened 2019-11-07
catch instant convergence

ChrisRackauckas:


YingboMa:
Umm, maybe this is not a good idea.
https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/608703787#L836

```julia
KenCarp3 = OrdinaryDiffEq.KenCarp3

Time derivative Tests: Test Failed at /home/travis/build/JuliaDiffEq/OrdinaryDiffEq.jl/test/regression/time_derivative_test.jl:33

  Expression: sol.errors[:final] < 1.0e-10

   Evaluated: 0.00010675204042853714 < 1.0e-10
```

src/nlsolve/nlsolve.jl
```diff
@@ -42,6 +42,10 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator)
     apply_step!(nlsolver, integrator)
 
     # check for convergence
+    if iter == 1 && ndz < 1e-3
```
devmotion: What's the reason for using the cutoff `1e-3` here? It seems a bit arbitrary and unfortunately leads to a bunch of test errors.

Couldn't this be covered by the existing check below? E.g., by changing it to
```
if η * ndz < κ && !(isnewton(nlsolver) && iszero(integrator.success_iter))
```
? It seems the check `iszero(ndz)` is not needed anyway?

ChrisRackauckas:
KenCarp is currently using the incorrect gamma so it fails on non autonomous problems in a very bad manner (which is why we should probably even just merge that PR as is and continue it later)

src/nlsolve/nlsolve.jl
ChrisRackauckas: The cutoff is somewhat arbitrary, but it is already scaled by tolerance so it's just a relative parameter. The difficulty here is that if it doesn't exist, then really good guesses cause a convergence failure because then the rate is essentially zero

YingboMa:
Closed by #858

PR #959, opened 2019-11-11
fix resizing functionality on composite algorithms

ChrisRackauckas:


ChrisRackauckas:
Fixes https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/958

PR #960, opened 2019-11-12
Sarray hotfix

ChrisRackauckas:


ChrisRackauckas:
@huanglangwen is there a reason why the number of function calls would increase?

huanglangwen:
I have already fixed that in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/937 .

PR #967, opened 2019-11-18
Use CompatHelper to update [compat] entries

devmotion:


ChrisRackauckas:
ModelingToolkit might be most in need of this

PR #968, opened 2019-11-18
Fix undefined variable in test

devmotion:
This PR specifies an undefined variable that was not copied from https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/965. Should fix current test errors (see, e.g., https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/613599420).


ChrisRackauckas:
🤦‍♂

PR #969, opened 2019-11-18
Fix extrapolation methods

devmotion:
This PR fixes the `calc_W!` calls in the in-place implementation of the extrapolation methods. Should fix current test errors (see, e.g., https://gitlab.com/juliadiffeq/OrdinaryDiffEq-jl/-/jobs/354216471).


devmotion:
Relevant tests pass now: https://gitlab.com/juliadiffeq/OrdinaryDiffEq-jl/-/jobs/354236321

The other GPU test error (a previously broken test which works now) is already fixed in DiffEqBase master and just requires a new release of DiffEqBase.

ChrisRackauckas:
looks like we have a GPU pass as well?

PR #984, opened 2019-12-02
Parameterize RKC constant cache

YingboMa:


ChrisRackauckas:
uhh, that was type unstable before?

YingboMa:
Yeah...

ChrisRackauckas:
Does this fix https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/980

PR #986, opened 2019-12-04
some inbounds and common interpolation performance improvements

ChrisRackauckas:
Works around some issues in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/985#issuecomment-561423105 in common cases. I assume StaticArrays broadcast will be fixed sooner or later, so I didn't spend the time to do the derivative of every single method, but at least common cases should be more optimized.


devmotion:
With this PR the test times on Travis are reduced quite a bit for the Regression and Integrators_I testset (compared to the last tests on master). I'm wondering if that's due to these changes or just noise.

devmotion:
Should also improve DelayDiffEq, I assume.

ChrisRackauckas:
Those timing improvements seem to be real.

PR #988, opened 2019-12-04
Improve interpolation speeds for fallback interpolations

ChrisRackauckas:


ChrisRackauckas:
```julia
using OrdinaryDiffEq, StaticArrays, BenchmarkTools
function lorenz_static(u,p,t)
    @inbounds begin
        dx = p[1]*(u[2]-u[1])
        dy = u[1]*(p[2]-u[3]) - u[2]
        dz = u[1]*u[2] - p[3]*u[3]
    end
    @SVector [dx,dy,dz]
end
u0 = @SVector [1.0,0.0,0.0]
p  = @SVector [10.0,28.0,8/3]
tspan = (0.0,100.0)
prob = ODEProblem(lorenz_static,u0,tspan,p)
sol = solve(prob,Tsit5(),saveat=0.1)
@btime sol(1.55)
@btime sol(1.5)
```

Before:

```julia
  1.450 μs (38 allocations: 1.14 KiB)
  930.000 ns (28 allocations: 976 bytes)
```

After:

```julia
514.057 ns (22 allocations: 912 bytes)
564.566 ns (22 allocations: 912 bytes)
```

ChrisRackauckas:
A few things to note:

- `avoid_constant_ends`  was ForwardDiff only, and was unsafe with any other AD. While in theory it's good because it potentially it removes an O(n) calculation if interpolating exactly at a data time point, in practice it actually made things slower unless `n` was "sufficiently large", so removing it is better timing-wise unless we're talking about larger arrays. If we are talking about larger arrays, the allocation time dominates over the fused broadcast interpolation call anyways, so it's not a major gain. So we can leave it to the user to do this if they want, and overall this is a speed win and a safety win. The other safety aspect is that returning an array directly from `u` instead of some copy was suspect... that was the only speed benefit really.
- Not erroring on extrapolating is the majority of the speed gain. So... is it worth it? Sometimes it's nice to extrapolate anyways.
- Linear interpolation with broadcast works just fine without performance losses, so nothing to see there.
- Hermite needs to split, so it uses `Val{true}` instead of `Val(true)` because the former doesn't allocate and is thus much faster.
- This new `_searchsortedfirst` is a bit faster as well.

ChrisRackauckas:
Supersceded by #994

PR #999, opened 2019-12-22
DAE support for OrdinaryDiffEq

kanav99:
Milestones:

- [x] Jacobian Wrapper - Pass the function and required temporary arrays to it and pass the wrapper to `jacobian!` function
- [x] Add a resType to the alg_cache call
- [x] Handle `calc_J` and `calc_W` functions to accept DAE
- [x] NLNewton `compute_step`
- [ ] Jacobian Reuse (partial issues, ultra slow convergence)
- [x] OOP support
- [ ] `dae_init_dt`
- [x] NLFunctional `compute_step`
- [x] NLAnderson `compute_step`
- [ ] Tests

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -458,6 +458,17 @@ function update_W!(nlsolver::AbstractNLSolver, integrator, cache, dtgamma, repea
   nothing
 end
 
+function dae_calculate_J!(nlsolver::AbstractNLSolver, integrator, cache::OrdinaryDiffEqMutableCache)
+  @unpack f, uprev = integrator
+  @unpack uf, J, jac_config, du1 = nlsolver.cache
+  if DiffEqBase.has_jac(f)
+    # todo
+  else
+    uf.γ = nlsolver.γ * inv(integrator.dt)
+    ForwardDiff.jacobian!(J, uf, du1, uprev, jac_config)
+  end
+end
+
 function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{IIP}) where IIP
   islin, isode = islinearfunction(f, alg)
   if f.jac_prototype isa DiffEqBase.AbstractDiffEqLinearOperator
@@ -491,7 +502,7 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{IIP}) where IIP
   return J, W
 end
 
-build_uf(alg::ßOrdinaryDiffEqAlgorithm,nf,t,p,::Val{true}) =
+build_uf(alg::OrdinaryDiffEqAlgorithm,nf,t,p,::Val{true}) =
   DiffEqDiffTools.UJacobianWrapper(nf,t,p)
 build_uf(alg::OrdinaryDiffEqAlgorithm,nf,t,p,::Val{false}) =
   DiffEqDiffTools.UDerivativeWrapper(nf,t,p)
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -358,8 +358,14 @@ function calc_W!(W, integrator, nlsolver::Union{Nothing,AbstractNLSolver}, cache
   @unpack t,dt,uprev,u,f,p = integrator
   lcache = nlsolver === nothing ? cache : nlsolver.cache
   @unpack J = lcache
+  isdae = integrator.alg isa DAEAlgorithm
   alg = unwrap_alg(integrator, true)
-  mass_matrix = integrator.f.mass_matrix
+  if isdae
+    # just alias W to be passed in calc_J!
+    J = lcache.W
+  else
+    mass_matrix = integrator.f.mass_matrix
+  end
   is_compos = integrator.alg isa CompositeAlgorithm
 
   # handle Wfact
@@ -458,14 +464,14 @@ function update_W!(nlsolver::AbstractNLSolver, integrator, cache, dtgamma, repea
   nothing
 end
 
-function dae_calculate_J!(nlsolver::AbstractNLSolver, integrator, cache::OrdinaryDiffEqMutableCache)
+function dae_calc_W!(nlsolver::AbstractNLSolver, integrator, cache::OrdinaryDiffEqMutableCache)
   @unpack f, uprev = integrator
-  @unpack uf, J, jac_config, du1 = nlsolver.cache
+  @unpack uf, W, jac_config, du1 = nlsolver.cache
   if DiffEqBase.has_jac(f)
     # todo
   else
     uf.γ = nlsolver.γ * inv(integrator.dt)
-    ForwardDiff.jacobian!(J, uf, du1, uprev, jac_config)
+    ForwardDiff.jacobian!(W, uf, du1, uprev, jac_config)
   end
 end
 
@@ -485,7 +491,10 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{IIP}) where IIP
     W = WOperator(f.mass_matrix, dt, J, IIP)
   else
     J = false .* _vec(u) .* _vec(u)'
-    W = if IIP
+    isdae = alg isa DAEAlgorithm
+    W = if isdae
+      J
+    elseif IIP
       similar(J)
     else
       W = if u isa StaticArray
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -491,7 +491,7 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{IIP}) where IIP
   return J, W
 end
 
-build_uf(alg::Union{DAEAlgorithm,OrdinaryDiffEqAlgorithm},nf,t,p,::Val{true}) =
+build_uf(alg::ßOrdinaryDiffEqAlgorithm,nf,t,p,::Val{true}) =
   DiffEqDiffTools.UJacobianWrapper(nf,t,p)
-build_uf(alg::Union{DAEAlgorithm,OrdinaryDiffEqAlgorithm},nf,t,p,::Val{false}) =
+build_uf(alg::OrdinaryDiffEqAlgorithm,nf,t,p,::Val{false}) =
   DiffEqDiffTools.UDerivativeWrapper(nf,t,p)
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -47,22 +47,32 @@ cache.
 function calc_J(integrator, cache)
   @unpack t,uprev,f,p,alg = integrator
 
-  if DiffEqBase.has_jac(f)
-    J = f.jac(uprev, p, t)
+  if alg isa DAEAlgorithm
+    if DiffEqBase.has_jac(f)
+      J = f.jac(duprev, uprev, p, t)
+    else
+      @unpack uf = cache
+      x = zero(uprev)
+      J = jacobian(uf, x, integrator)
+    end
   else
-    @unpack uf = cache
+    if DiffEqBase.has_jac(f)
+      J = f.jac(uprev, p, t)
+    else
+      @unpack uf = cache
 
-    uf.f = nlsolve_f(f, alg)
-    uf.p = p
-    uf.t = t
+      uf.f = nlsolve_f(f, alg)
+      uf.p = p
+      uf.t = t
 
-    J = jacobian(uf, uprev, integrator)
-  end
+      J = jacobian(uf, uprev, integrator)
+    end
 
-  integrator.destats.njacs += 1
+    integrator.destats.njacs += 1
 
-  if alg isa CompositeAlgorithm
-    integrator.eigen_est = opnorm(J, Inf)
+    if alg isa CompositeAlgorithm
+      integrator.eigen_est = opnorm(J, Inf)
+    end
   end
 
   J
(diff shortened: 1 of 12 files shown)


kanav99:
@ChrisRackauckas @devmotion @YingboMa requesting your comments on the interface. Basic call with an initial dt works now atleast. Need to fine tune the API now

kanav99:
image

This sort of equation is needed to be solved for `yn`. Our solver solves for `dy = yn - yn-1` instead. What should we set `γ` instead? I think it should never depend on `dt`. So for DAEs, I think we should be dividing by `dt` inside the solver instead. Hence I feel the current implementation inside the `compute_step` is already correct. What do you think @ChrisRackauckas ?

ChrisRackauckas:
For a DAE, then doesn't it make sense for `gamma = beta`? Then

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/nlsolve/newton.jl#L7

is exactly the quantity you are looking for, and corresponds exactly to the same one in the ODE case. The transformed nonlinear solver version and the DAE version are exactly the same in terms of gamma handling!

Let's be clear about what we're doing here. It's already in residual form, so it should be solving:

It solves
```math
0 = f((α₀*z + α'uhist) * invγdt, z, p, t + c⋅dt)
```
by iterating
```math
zᵏ⁺¹ = zᵏ - W\f((α₀*z + α'uhist) * invγdt, z, p, t + c⋅dt)
```

src/nlsolve/newton.jl
```diff
@@ -136,8 +157,10 @@ end
   # Diagonally Implicit Runge-Kutta Methods for Ordinary Differential
   # Equations. A Review, by Christopher A. Kennedy and Mark H. Carpenter
   # page 54.
-  γdt = γ * dt
-  !(W_γdt ≈ γdt) && (rmul!(dz, 2/(1 + γdt / W_γdt)))
+  if !isdae
```
ChrisRackauckas: we should similarly relax the DAE Newton

src/derivative_wrappers.jl
```diff
@@ -146,6 +146,20 @@ function DiffEqBase.build_jac_config(alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgor
   jac_config
 end
 
+function DiffEqBase.build_jac_config(alg::DAEAlgorithm,f,uprev,du1)
+  if !DiffEqBase.has_jac(f)
+    if alg_autodiff(alg)
+      jac_config = ForwardDiff.JacobianConfig(nothing, du1, uprev)
```
ChrisRackauckas: don't use ForwardDiff directly

src/nlsolve/utils.jl
```diff
@@ -120,6 +120,92 @@ function build_nlsolver(alg,nlalg::Union{NLFunctional,NLAnderson,NLNewton},u,upr
...
+  @. m.tmp_du = m.γ * x + m.tmp
+  @. m.tmp_u = x + m.uprev
+  m.f(out, m.tmp_du, m.tmp_u, m.p, m.t)
+end
+
+DiffEqBase.has_jac(f::SuperDAEJacobianWrapper) = DiffEqBase.has_jac(f.f)
+DiffEqBase.has_Wfact(f::SuperDAEJacobianWrapper) = DiffEqBase.has_Wfact(f.f)
+DiffEqBase.has_Wfact_t(f::SuperDAEJacobianWrapper) = DiffEqBase.has_Wfact_t(f.f)
+
+function build_nlsolver(alg::DAEAlgorithm,nlalg::Union{NLFunctional,NLAnderson,NLNewton},u,uprev,p,t,dt,
```
ChrisRackauckas: This is 99% a copy-paste of the other function, why?

src/nlsolve/utils.jl
```diff
@@ -120,6 +120,92 @@ function build_nlsolver(alg,nlalg::Union{NLFunctional,NLAnderson,NLNewton},u,upr
...
+    jac_config = build_jac_config(alg,f,uprev,du1)
+    linsolve = alg.linsolve(Val{:init},uf,u)
+
+    # TODO: check if the solver is iterative
+    weight = similar(u)
+
+    tType = typeof(t)
+    invγdt = inv(oneunit(t) * one(uTolType))
+
+    J = false .* _vec(u) .* _vec(u)'
```
ChrisRackauckas: this is not always correct. Why not use the existing function, but with a marker for whether to produce 1 output?

src/perform_step/dae_perform_step.jl
```diff
@@ -40,17 +40,14 @@ end
...
 
-  # initial guess
-  guess = tmp
-  guess .= zero(eltype(u))
-
-  zero_func! = (out, x) -> begin
-                f(out, (x .- uprev) ./ dt, x, p, t)
-              end
-  u .= NLsolve.nlsolve(zero_func!, guess).zero
+  @. nlsolver.tmp = 0
```
ChrisRackauckas: false

src/solve.jl
```diff
@@ -61,6 +70,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
                            allow_extrapolation = alg_extrapolates(alg),
                            initialize_integrator = true,
                            alias_u0 = false,
+                           alias_du0 = false,
```
ChrisRackauckas: we need to work out a more general system

src/nlsolve/utils.jl
```diff
@@ -120,6 +120,92 @@ function build_nlsolver(alg,nlalg::Union{NLFunctional,NLAnderson,NLNewton},u,upr
...
+  function SuperDAEJacobianWrapper(f,p,γ,tmp,uprev,t)
+    X = eltype(uprev)
+    N = ForwardDiff.chunksize(ForwardDiff.Chunk(uprev))
+    tmp_du = similar(uprev, ForwardDiff.Dual{Nothing,X,N})
+    tmp_u  = similar(uprev, ForwardDiff.Dual{Nothing,X,N})
+    new{typeof(f),typeof(p),typeof(tmp_du),typeof(tmp_u),typeof(γ),typeof(tmp),typeof(uprev),typeof(t)}(f,p,tmp_du,tmp_u,γ,tmp,uprev,t)
+  end
+end
+
+function (m::SuperDAEJacobianWrapper)(out,x)
```
ChrisRackauckas: This would be a million times more clear if it just had a good name. `DAEResidualJacobianWrapper` is what it is: it embeds all previous information to give the Jacobian for the residual calculation. I'd add a comment as to what this is actually calculating, i.e. why this corresponds to `f(x) = 0`

src/solve.jl
```diff
@@ -6,6 +6,15 @@ function DiffEqBase.__solve(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase
   integrator.sol
 end
 
+"""
+    DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.AbstractDAEProblem}, args...; kwargs...)
```
ChrisRackauckas: no. Just do this at the DiffEqBase level. This will just make things harder to maintain.

src/nlsolve/nlsolve.jl
```diff
@@ -11,7 +11,11 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator, cache=nothing, repeat_
   @label REDO
   if isnewton(nlsolver)
     cache isa Nothing && throw(ArgumentError("cache is not passed to `nlsolve!` when using NLNewton"))
-    update_W!(nlsolver, integrator, cache, nlsolver.γ*integrator.dt, repeat_step)
+    if integrator.alg isa DAEAlgorithm
+      dae_calculate_J!(nlsolver, integrator, cache)
+    else
+      update_W!(nlsolver, integrator, cache, nlsolver.γ*integrator.dt, repeat_step)
```
ChrisRackauckas: seems like this should just be `update_W!` for the DAE...

src/nlsolve/newton.jl
```diff
@@ -104,30 +104,51 @@ end
...
+      mul!(vec(ztmp), mass_matrix, vec(z))
+      @.. ztmp = (dt * k - ztmp) * invγdt
+    end
+    b = vec(ztmp)
   end
 
-  # update W
-  if W isa DiffEqBase.AbstractDiffEqLinearOperator
-    update_coefficients!(W, ustep, p, tstep)
+  if isdae
```
ChrisRackauckas: This is a mess, but should get replaced by the `jvp` mechanism anyways so I'll let it slide.

ChrisRackauckas:
Let's put the scope of the PR to just be handling what it currently has, and continue to the next pieces in a subsequent PR.

The main comments to address are:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/999#discussion_r361459935
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/999#discussion_r361460656

@YingboMa do you have any comments? I think that if that big code duplication gets fixed and calc_J/calc_W stuff is handled well then we're good to go.

This also needs tests. Specifically a convergence test on DImplicitEuler.

(BTW, does it have to be `DImplicitEuler`, or could it just be `ImplicitEuler`?)

kanav99:
> For a DAE, then doesn't it make sense for gamma = beta?

This seems a nice idea. We can make use of `invγdt` then.

> It solves
>
>0 = f((α₀*(uprev+z) + α'uhist) * invγdt, uprev + z, p, t + c⋅dt)
>by iterating
>
>zᵏ⁺¹ = zᵏ - W\f((α₀*(uprev+z) + α'uhist) * invγdt, uprev + z, p, t + c⋅dt)
>Correct?

Yes. For this reason, I need to multiply the inverse of dt over there. Now I can probably use `invγdt` instead.

I saw through your review comments. The PR was in WIP state and just wanted to show that it works and probably look for any logical errors. I am trying to understand Jacobian reuse and SparseDiffTools and working on the PR. Currently the PR is far from complete 😛 

> BTW, does it have to be DImplicitEuler, or could it just be ImplicitEuler

I can try, but we might need to remove the initial check which checks for incompatible `prob` and `alg`

ChrisRackauckas:
Or we just need a whitelist of algs for DAEs

kanav99:
I am currently unable to find why residuals are calculated to be very huge. Redundancy is very minimal now.

kanav99:
Worth pointing out: the `REDO` logic in `nlsolve!` is prone to infinite loops.

ChrisRackauckas:
@yingboma take a look at some of the nlsolver stuff.

kanav99:
Convergence for the non-adaptive case is very bad in here. Residuals seem okay.

kanav99:
Turned off Jacobian Reuse and continuing to complete the PR without it

kanav99:
I should get rid of the branches soon.

src/nlsolve/utils.jl
kanav99: I have aliased J and W to the same matrix now. Is it nice idea? Possible issues: `resize`

src/solve.jl
```diff
@@ -160,6 +182,15 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
   end
   rateType = typeof(rate_prototype) ## Can be different if united
 
+  if isdae
+    if uBottomEltype == uBottomEltypeNoUnits
+      res_prototype = u
+    else
+      res_prototype = one(u)
```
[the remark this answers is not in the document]
kanav99: Now that I am using SparseDiffTools, is this still an issue?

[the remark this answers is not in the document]
ChrisRackauckas: should be fine.

src/nlsolve/utils.jl
YingboMa: Don't alias J and W for now. Jacobian and W matrix use different reuse strategies.

src/derivative_utils.jl
```diff
@@ -47,22 +47,32 @@ cache.
...
 
-  if DiffEqBase.has_jac(f)
-    J = f.jac(uprev, p, t)
+  if alg isa DAEAlgorithm
+    if DiffEqBase.has_jac(f)
+      J = f.jac(duprev, uprev, p, t)
+    else
+      @unpack uf = cache
+      x = zero(uprev)
+      J = jacobian(uf, x, integrator)
```
YingboMa: This looks a little weird.

kanav99:
The aim of the changes inside the nlsolver was to make it able to solve the equations of form

```
g(x) = f((tmp + α*x) * invdt, uprev + x, p, t) = 0
```

For that, we needed the Jacobian of `g` at `x=0`. W is exactly that value. J is just the alias of W. Its not used anywhere.

kanav99:
Marked the PR ready for review. cc @ChrisRackauckas @YingboMa . Some topics left unticked are for me to handle in the coming PR.

Also, I have not yet tested `NLFunctional` correctly right now. I don't have an example to test, simple cases don't converge. Do you have an example which should work?

ChrisRackauckas:
>Also, I have not yet tested NLFunctional correctly right now. I don't have an example to test, simple cases don't converge. Do you have an example which should work?

It may be almost impossible for this to work on DAEs since it requires non-stiff.

YingboMa:
> For that, we needed the Jacobian of g at x=0. W is exactly that value. J is just the alias of W. Its not used anywhere.

I see. Makes sense.

ChrisRackauckas:
I'll merge, but this needs convergence tests.

PR #1002, opened 2019-12-29
fix in-place interpolants with out of place problems

ChrisRackauckas:


ChrisRackauckas:
@devmotion my guess is that this is the missing piece between the oop and iip of DelayDiffEq

devmotion:
I just reran all DelayDiffEq tests locally, and it seems they are unaffected by this PR.

PR #1012, opened 2020-01-23
General Fixes

kanav99:
## Replaces DiffEqDiffTools with FiniteDiff
Similar should be done in DiffEqBase.

## Fixes FiniteDiff

While integrating DAE with FiniteDiff, I found that `nlsolver.tmp` was passed in the `JacobianCache`, due to which `jacobian!` call changed its value. That should never happen because `nlsolver.tmp` is used as a parameter to nlsolve call. Did we ever experience issues with finite differentiation before?

## Initial dt set to 1e-6 if its not specified
It was discussed on slack to just set the initial `dt` to 1e-6

## Changes in DAEResidualJacobianWrapper to support FiniteDiff
We need changes probably here right now, the code there is not good IMO


kanav99:
Oops where did the other wrappers go

kanav99:
Needs JuliaDiffEq/DiffEqBase.jl#426

PR #1013, opened 2020-01-23
Fix BoundsError in interpolation

devmotion:
This PR fixes https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/168.

### DIFF
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -156,7 +156,7 @@ function ode_interpolation(tvals,id,idxs,deriv,p,continuity::Symbol=:left)
       # we have i₋ = i₊ = 1 if t = ts[1], i₊ = i₋ + 1 = lastindex(ts) if t > ts[end],
       # and otherwise i₋ and i₊ satisfy ts[i₋] < t ≤ ts[i₊]
       i₊ = min(lastindex(ts), _searchsortedfirst(ts,t,i₊,tdir > 0))
-      i₋ = i₊ > 1 ? i₊ - 1 : 1
+      i₋ = i₊ > 1 ? i₊ - 1 : i₊
     else
       # we have i₋ = i₊ - 1 = 1 if t < ts[1], i₊ = i₋ = lastindex(ts) if t = ts[end],
       # and otherwise i₋ and i₊ satisfy ts[i₋] ≤ t < ts[i₊]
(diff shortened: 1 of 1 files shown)


devmotion:
If I run DelayDiffEq tests locally against this PR, the BoundsError is gone, but the event tests fail for continuous callbacks:
```julia
continuous: Test Failed at /home/david/.julia/dev/DelayDiffEq/test/integrators/events.jl:19
  Expression: sol3.errors[:L2] < 0.0041
   Evaluated: 0.010774044572046462 < 0.0041
Stacktrace:
 [1] top-level scope at /home/david/.julia/dev/DelayDiffEq/test/integrators/events.jl:19
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
 [3] top-level scope at /home/david/.julia/dev/DelayDiffEq/test/integrators/events.jl:11
continuous: Test Failed at /home/david/.julia/dev/DelayDiffEq/test/integrators/events.jl:20
  Expression: sol3.errors[:L∞] < 0.013
   Evaluated: 0.03142380053553678 < 0.013
Stacktrace:
 [1] top-level scope at /home/david/.julia/dev/DelayDiffEq/test/integrators/events.jl:20
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
 [3] top-level scope at /home/david/.julia/dev/DelayDiffEq/test/integrators/events.jl:11
Test Summary:        | Pass  Fail  Total
Event Tests          |    5     2      7
  continuous         |          2      2
  discrete           |    2            2
  save discontinuity |    3            3
```
I'm wondering if the errors are caused by this PR or due to some other changes introduced after 5.27.0.

devmotion:
Hmm might be due to this PR since the ForwardDiff tests seem to fail. I still think that these lines should be removed, it is not very surprising that they might lead to `BoundsError`s.

devmotion:
I also reran the DelayDiffEq tests against the updated PR, apart from the two test errors mentioned above everything passes. Maybe we should just adjust the tolerances? Or could it point to an additional problem?

ChrisRackauckas:
That's just a weird test. The interepolation wouldn't effect the event location here, so it's not like the interpolation is finding a huge difference here. Sounds like other stepping behavior changes cause this difference. Likely interpolation fixes increased the stepsize so now ironically the test fails.

Issue #864, opened 2019-08-10
Explicit RK for starting singularity


ChrisRackauckas:
https://www.sciencedirect.com/science/article/abs/pii/S0096300319301511

also known as the Lane–Emden problem

PR #1027, opened 2020-02-04
Explicit RK Oliver (6 stages, 5th order) for start…

frankschae:
…ing singularity #864. I am very unsure whether I have set correctly integrator.destats.nf += 6 and integrator.kshortsize = 6.. Unfortunately, the tests fail (tolerance must be significantly increased) -- it would be great if someone could have a quick look and point me to possible problems :)  I have written out the Butcher table. I could push it to DivEqDevTools if that helps (?)

### DIFF
--- a/test/algconvergence/ode_convergence_tests.jl
+++ b/test/algconvergence/ode_convergence_tests.jl
@@ -33,6 +33,9 @@ testTol = 0.2
   sim3 = test_convergence(dts2,prob,KuttaPRK2p5(threading=false))
   @test sim3.𝒪est[:l∞] ≈ 5 atol=testTol
 
+  sim3 = test_convergence(dts2,prob,RKO65())
+  @test sim3.𝒪est[:l∞] ≈ 5 atol=testTol
+
   sim4 = test_convergence(dts,prob,BS3())
   @test sim4.𝒪est[:l2] ≈ 3 atol=testTol
   sim5 = test_convergence(dts, prob, AB3())
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -174,7 +174,7 @@ module OrdinaryDiffEq
 
   export FunctionMap, Euler, Heun, Ralston, Midpoint, RK4, ExplicitRK, OwrenZen3, OwrenZen4, OwrenZen5,
          BS3, BS5, RK46NL, DP5, Tsit5, DP8, Vern6, Vern7, Vern8, TanYam7, TsitPap8,
-         Vern9, Feagin10, Feagin12, Feagin14, CompositeAlgorithm, Anas5
+         Vern9, Feagin10, Feagin12, Feagin14, CompositeAlgorithm, Anas5, RKO65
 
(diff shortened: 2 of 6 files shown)


test/algconvergence/ode_convergence_tests.jl
```diff
@@ -33,6 +33,9 @@ testTol = 0.2
   sim3 = test_convergence(dts2,prob,KuttaPRK2p5(threading=false))
   @test sim3.𝒪est[:l∞] ≈ 5 atol=testTol
 
+  sim3 = test_convergence(dts2,prob,RKO65())
```
ChrisRackauckas: what does `plot(sim3)` look like?

test/algconvergence/ode_convergence_tests.jl
```diff
@@ -33,6 +33,9 @@ testTol = 0.2
...
+  @test sim3.𝒪est[:l∞] ≈ 5 atol=testTol
```
ChrisRackauckas: what result are you getting here?

frankschae: Expression: ≈(sim3.𝒪est[:l∞], 5, atol=testTol)
   Evaluated: 0.9645038959802423 ≈ 5 (atol=0.2)

Expression: ≈(sim3.𝒪est[:l∞], 5, atol=testTol)
   Evaluated: 0.9645038959807267 ≈ 5 (atol=0.2)

test/algconvergence/ode_convergence_tests.jl
frankschae: Foe the 2D inlace: 

prob_ode_2Dlinear

frankschae: For the 1D out of place

prob_ode_linear

ChrisRackauckas:
Looks like perfect first order convergence. This means there's a coefficients that's incorrect somewhere. Does each row of the A tableau sum to the corresponding c_i?

frankschae:
Do you mean A*e =c? They explicitly say that this condition is dropped in Section II.

ChrisRackauckas:
I don't have access to the paper, but how do they get higher convergence without it?

frankschae:
it holds only for indices greater than 1.. so they set c1 != 0  while at the same time they set b1 =0. But I just spotted that indeed also c6 != sum[A[6, I]] if I sum up the values in the table..:
1 != 0 + 123/300 + 5/2 + 3/20 - 75/38 = 516/475

test/algconvergence/ode_convergence_tests.jl
frankschae: With the updated value I get: 

Expression: ≈(sim3.𝒪est[:l∞], 5, atol=testTol)
   Evaluated: 4.9872356619789535 ≈ 5 (atol=0.2)


Expression: ≈(sim3.𝒪est[:l∞], 5, atol=testTol)
   Evaluated: 4.998263889697244 ≈ 5 (atol=0.2)

ChrisRackauckas:
Thanks!

PR #1031, opened 2020-02-13
Drop partials when computing the eigenvalue estimation

YingboMa:
Fixes https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/1018


src/solve.jl
```diff
@@ -218,8 +218,12 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
   alg_choice = Int[]
 
   if !adaptive && save_everystep && tspan[2]-tspan[1] != Inf
-    dt == 0 ? steps = length(tstops) :
-              steps = ceil(Int,internalnorm((tspan[2]-tspan[1])/dt,tspan[1]))
+    if dt == 0
+      steps = length(tstops)
+    else
+      abs(dt) < dtmin && throw(ArgumentError("Suplided dt is smaller than dtmin"))
```
ChrisRackauckas:
```suggestion
      abs(dt) < dtmin && throw(ArgumentError("Supplied dt is smaller than dtmin"))
```

src/solve.jl
```diff
@@ -6,6 +6,17 @@ function DiffEqBase.__solve(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase
   integrator.sol
 end
 
+tspan2dtmin(tspan) = tspan2dtmin(tspan, one(eltype(tspan)))
```
ChrisRackauckas: should this go to DiffEqBase and be uniformly applied to *DiffEq?

YingboMa: https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/438

PR #1037, opened 2020-02-17
Basic DAE Initialization

kanav99:
https://cse.cs.ucsb.edu/sites/default/files/publications/SCE001495.pdf

Using NLsolve right now, I will try to remove that out


src/solve.jl
```diff
@@ -382,6 +383,10 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
                              isout,reeval_fsal,
                              u_modified,opts,destats)
   if initialize_integrator
+    if isdae
+      initialize_dae!(integrator, u, du, prob.differential_vars, initializealg, Val(isinplace(prob)))
+      @show du
```
ChrisRackauckas:
```suggestion
```

src/initialize_dae.jl
```diff
@@ -0,0 +1,54 @@
...
+		else
+			u[i] = r.zero[i]
+		end
+	end
+end
+
+function initialize_dae!(integrator, u, du, differential_vars, alg::BrownFullBasicInit, ::Val{true})
+	@unpack p, t, f = integrator
+
+	nlequation = (x) -> begin
```
ChrisRackauckas:
```suggestion
	nlequation = (dx,x) -> begin
```

src/initialize_dae.jl
```diff
@@ -0,0 +1,54 @@
...
+	for i in 1:length(u)
+		if differential_vars[i]
+			du[i] = r.zero[i]
+		else
+			u[i] = r.zero[i]
+		end
+	end
+end
+
+function initialize_dae!(integrator, u, du, differential_vars, alg::BrownFullBasicInit, ::Val{true})
```
ChrisRackauckas:
```suggestion
function initialize_dae!(integrator, u, du, differential_vars, alg::BrownFullBasicInit, ::Val{false})
```

src/initialize_dae.jl
```diff
@@ -0,0 +1,54 @@
...
+
+	nlequation = (x) -> begin
+		for i in 1:length(x)
+			if differential_vars[i]
+				du[i] = x[i]
+			else
+				u[i] = x[i]
+			end
+		end
+		f(du, u, p, t)
```
ChrisRackauckas:
```suggestion
		du .= f(u, p, t)
```

ChrisRackauckas:
Needs a test.

kanav99:
Added some basic test. There is a failing test (non index 1 DAE), that means we know whats missing. I guess we need an interface to define what all is known to the user, and what all the user wants us to initialize

ChrisRackauckas:
>I guess we need an interface to define what all is known to the user, and what all the user wants us to initialize

That can come from the initialization method choice?

src/solve.jl
```diff
@@ -382,6 +383,9 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
                              isout,reeval_fsal,
                              u_modified,opts,destats)
   if initialize_integrator
+    if isdae
```
ChrisRackauckas: evaluate to see if initialization is required first? Or do that in initialize?

PR #1045, opened 2020-02-20
Less restrictive dtmin

YingboMa:
Needs https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/441

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEq"
 uuid = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "5.29.0"
+version = "5.29.1"
 
 [deps]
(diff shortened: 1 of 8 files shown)


ChrisRackauckas:
We should similarly do StochasticDiffEq and make sure DelayDiffEq is good with this.

ChrisRackauckas:
Some dispatches are missing: https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/653226807#L766

YingboMa:
https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/653272862#L878
broken test due to StaticArrays.jl's behavior change.
```julia
julia> (@SMatrix [1 2; 3 4]) * ones(2)
2-element SArray{Tuple{2},Float64,1,2} with indices SOneTo(2):
 3.0
 7.0
```

ChrisRackauckas:
The new dtmin doesn't seem to respect units 🤷‍♂

Issue #867, opened 2019-08-10
Zero-dissipation ERK 6(5) and 8(7) for oscillatory problems


ChrisRackauckas:
https://www.researchgate.net/publication/325858674_Fitted_modifications_of_Runge-Kutta_pairs_of_orders_65

https://www.researchgate.net/publication/326363860_New_phase-fitted_Runge-Kutta_pairs_of_orders_87

PR #1047, opened 2020-02-23
Phase Fitted RK8(7) with zero dissipation

utkarsh530:
Please review this PR for zero dissipation RK8(7) #867 . I have implemented [this](https://www.sciencedirect.com/science/article/pii/S037704271730095X), required coefficients taken from [here](https://www.sciencedirect.com/science/article/pii/0771050X81900103). I had checked for Bessel Function, the solutions are overlapping but with a warning: `┌Warning: Interrupted. Larger maxiters is needed.
└ @ DiffEqBase C:\Users\Utkarsh\.julia\packages\DiffEqBase\T0uMB\src\integrator_interface.jl:319
retcode: MaxIters`. Also for convergence tests, which set of problems to use since I saw that for higher-order we are not using `ode_convergence_tests.jl`?. Thanks.


ChrisRackauckas:
> Also for convergence tests, which set of problems to use since I saw that for higher-order we are not using ode_convergence_tests.jl?

Higher order methods are in that file testing on the same problem.

ChrisRackauckas:
>@ DiffEqBase C:\Users\Utkarsh\.julia\packages\DiffEqBase\T0uMB\src\integrator_interface.jl:319 retcode: MaxIters

If convergence tests are passing and you see this, it's likely an issue with the error estimator.

utkarsh530:
It seems that the convergence tests are failing too :disappointed: . I checked multiple times for Butcher table entries and they seem correct to me. I used the a_i's from PD8(7) and fitted from new paper. However, I haven't changed b_i's and c_i's and they remain same as the PD8(7) ones. How do I debug it? @ChrisRackauckas @kanav99

ChrisRackauckas:
What does the plot look like?

utkarsh530:
fig1
`sim3.�est[:l∞] = 1.0591448783186939`
fig2
`sim3.�est[:l∞] = 1.0591448783187083`

ChrisRackauckas:
that's clearly first order convergence, so there's something very wrong.

utkarsh530:
It seems that some relations are incorrectly mentioned in the paper. How to find out relations between `a_i's` like in the case which solely depend on `a0807`?

utkarsh530:
@ChrisRackauckas please review. 
Plots: 
fig3
`println(sim3.�est[:l∞]) = 8.525649617002895`
fig4
`println(sim3.�est[:l∞]) = 8.209409454655237`

PR #1049, opened 2020-02-28
Fix inability to use sparse Jacobian with colorvec for OOP

mauro3:
As discussed on slack, this allows to use a sparse jacobian for out-of-place objective functions.  This works for my case but probably could be done more general.

The solve time of https://gist.github.com/mauro3/e85c85a4257a46ed73999b08f9b80ec2 (a FD PDE code) improves from >100s to <1s with this PR.

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -538,23 +538,16 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{IIP}) where IIP
     elseif IIP
       similar(J)
     else
-      W = if u isa StaticArray
-      lu(J)
+      if u isa StaticArray
+        lu(J)
       elseif u isa Number
         u
+      elseif f.jac_prototype===nothing
+        LU{LinearAlgebra.lutype(uEltypeNoUnits)}(Matrix{uEltypeNoUnits}(undef, 0, 0),
+                                                 Vector{LinearAlgebra.BlasInt}(undef, 0),
+                                                 zero(LinearAlgebra.BlasInt))
       else
-        # TODO: make this more general, maybe by running calc_W and use its returned value
-        if f.jac_prototype isa SparseMatrixCSC
-          SuiteSparse.UMFPACK.UmfpackLU(Ptr{Cvoid}(), Ptr{Cvoid}(), 1, 1,
-                                        f.jac_prototype.colptr[1:1],
-                                        f.jac_prototype.rowval[1:1],
-                                        f.jac_prototype.nzval[1:1],
-                                        0)
-        else
-          LU{LinearAlgebra.lutype(uEltypeNoUnits)}(Matrix{uEltypeNoUnits}(undef, 0, 0),
-                                                   Vector{LinearAlgebra.BlasInt}(undef, 0),
-                                                   zero(LinearAlgebra.BlasInt))
-        end
+        ArrayInterface.lu_instance(f.jac_prototype)
       end
     end # end W
   end
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -531,7 +531,11 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{IIP}) where IIP
     end
     W = WOperator(f.mass_matrix, dt, J, IIP)
   else
-    J = false .* _vec(u) .* _vec(u)'
+    J = if f.jac_prototype === nothing
+      false .* _vec(u) .* _vec(u)'
+    else
+      deepcopy(f.jac_prototype)
+    end
     isdae = alg isa DAEAlgorithm
     W = if isdae
       J
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -543,9 +543,18 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{IIP}) where IIP
       elseif u isa Number
         u
       else
-        LU{LinearAlgebra.lutype(uEltypeNoUnits)}(Matrix{uEltypeNoUnits}(undef, 0, 0),
-                                                     Vector{LinearAlgebra.BlasInt}(undef, 0),
-                                                     zero(LinearAlgebra.BlasInt))
+        # TODO: make this more general, maybe by running calc_W and use its returned value
+        if f.jac_prototype isa SparseMatrixCSC
+          SuiteSparse.UMFPACK.UmfpackLU(Ptr{Cvoid}(), Ptr{Cvoid}(), 1, 1,
+                                        f.jac_prototype.colptr[1:1],
+                                        f.jac_prototype.rowval[1:1],
+                                        f.jac_prototype.nzval[1:1],
+                                        0)
+        else
+          LU{LinearAlgebra.lutype(uEltypeNoUnits)}(Matrix{uEltypeNoUnits}(undef, 0, 0),
+                                                   Vector{LinearAlgebra.BlasInt}(undef, 0),
+                                                   zero(LinearAlgebra.BlasInt))
+        end
       end
     end # end W
   end
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -542,18 +542,8 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{IIP}) where IIP
     elseif IIP
       similar(J)
     else
-      if u isa StaticArray
-        lu(J)
-      elseif u isa Number
-        u
-      elseif f.jac_prototype===nothing
-        LU{LinearAlgebra.lutype(uEltypeNoUnits)}(Matrix{uEltypeNoUnits}(undef, 0, 0),
-                                                 Vector{LinearAlgebra.BlasInt}(undef, 0),
-                                                 zero(LinearAlgebra.BlasInt))
-      else
-        ArrayInterface.lu_instance(f.jac_prototype)
-      end
-    end # end W
+      ArrayInterface.lu_instance(J)
+    end
   end
   return J, W
 end
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -39,7 +39,7 @@ end
 jacobian_autodiff(f, x, odefun) = (ForwardDiff.derivative(f,x),1)
 function jacobian_autodiff(f, x::AbstractArray, odefun)
   colorvec = DiffEqBase.has_colorvec(odefun) ? odefun.colorvec : 1:length(x)
-  sparsity = odefun.jac_prototype
+  sparsity = odefun.sparsity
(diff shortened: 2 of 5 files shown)


ChrisRackauckas:
Thanks! Can we get a test on this?

mauro3:
Which file to add the test?

ChrisRackauckas:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/9ce1fe0a673ada3167ea6f4762964c28b3e89ea7/test/interface/sparsediff_tests.jl

ChrisRackauckas:
>Thanks a lot! Maybe we should add the interface factorization(A) -> FactorizationType in ArrayInterface.

Yeah, that's what would need to happen.

mauro3:
I'll add some tests and update the PR to use Yingbo's ArrayInterface changes.

test/interface/sparsediff_tests.jl
```diff
@@ -34,22 +34,27 @@ jac_sp = sparse(generate_sparsity_pattern(10))
...
-@test length(sol_sp.t)==length(sol.t)
\ No newline at end of file
+for ad in [true, false]
+    for tol in [nothing, 1e-10]
+        sol_std=solve(prob_std,Rodas5(autodiff=ad),reltol=tol,abstol=tol)
+        @test sol_std.retcode==:Success
+        for prob in map(f->ODEProblem(f,u0,tspan),
+                        [ODEFunction(f,colorvec=colors,jac_prototype=jac_sp),
+                         ODEFunction(f,jac_prototype=jac_sp),
+                         #ODEFunction(f,colorvec=colors) # this one fails both the u[end] and length tests
```
mauro3: Any idea why the commented test would behave differently?

mauro3:
This will now fail tests because it needs https://github.com/JuliaDiffEq/ArrayInterface.jl/pull/39.

I added some tests but I don't know to test the type of the used Jacobian.

Oh, and I should add OOP tests...

test/interface/sparsediff_tests.jl
```diff
@@ -34,22 +47,43 @@ jac_sp = sparse(generate_sparsity_pattern(10))
...
+          @show i
+          # TODO: not sure why these test with Trapezoid fail
+          if i==2 && f===f_oop && ad==true && Solver==Trapezoid
+            @show "skipping"
+            continue
+          end
+          if i==1 && f===f_oop && ad==false && Solver==Trapezoid
+            @show "skipping"
+            continue
+          end
```
mauro3: These skipped tests fail with:
` MethodError: no method matching SuiteSparse.UMFPACK.UmfpackLU{Float64,Int64}(::LU{Float64,Array{Float64,2}})`. I'll look into it tomorrow.

test/interface/sparsediff_tests.jl
```diff
@@ -34,22 +47,43 @@ jac_sp = sparse(generate_sparsity_pattern(10))
...
-@test sol_sp.u[end]≈sol.u[end]
-@test length(sol_sp.t)==length(sol.t)
\ No newline at end of file
+
+for f in [f_ip, f_oop]
+  odefun_std = ODEFunction(f)
+  prob_std = ODEProblem(odefun_std,u0,tspan)
+
+  for ad in [true, false]
+    for Solver in [Rodas5, Trapezoid] # TODO: add solvers from yet other families?
```
mauro3: Seeing how `Trapezoid` fails some tests, should I add other solvers to this list?  Which ones?

test/interface/sparsediff_tests.jl
mauro3: Any hints appreciated.

test/interface/sparsediff_tests.jl
YingboMa: Could you try `KenCarp4`?

test/interface/sparsediff_tests.jl
```diff
@@ -34,22 +47,54 @@ jac_sp = sparse(generate_sparsity_pattern(10))
...
+                         ]))
+          isbroken = i==3 && (
+            (f, ad, tol) == (f_oop, true, nothing) ||
+            (f, ad, Solver, tol) == (f_oop, false, Trapezoid, nothing) ||
+            (f, Solver, tol) == (f_ip, Trapezoid, nothing) ||
+            (f, Solver) == (f_oop, Rodas5) ||
+            (f, Solver) == (f_ip, Rodas5) ||
+            (f, Solver) == (f_oop, KenCarp4) ||
+            (f, Solver) == (f_ip, KenCarp4)
+          )
```
mauro3: These failing tests need investigation.

mauro3:
The tests failed because the Jacobians sometimes got converted to dense arrays.  I updated all the instances where I think that occurred. However, I do not fully understand the code, so this needs a careful review!

Also note that with https://github.com/JuliaDiffEq/ArrayInterface.jl/pull/39, CI should now pass. (Although I haven't run the full test suite locally...)

test/interface/sparsediff_tests.jl
ChrisRackauckas: I wouldn't expect colors without sparsity to work.

test/interface/sparsediff_tests.jl
```diff
@@ -34,22 +47,54 @@ jac_sp = sparse(generate_sparsity_pattern(10))
...
+        # @show f,ad,Solver,tol
+        sol_std=solve(prob_std,Solver(autodiff=ad),reltol=tol,abstol=tol)
+        @test sol_std.retcode==:Success
+        for (i,prob) in enumerate(map(f->ODEProblem(f,u0,tspan),
+                        [ODEFunction(f,colorvec=colors,jac_prototype=jac_sp),
+                         ODEFunction(f,jac_prototype=jac_sp),
+                         ODEFunction(f,colorvec=colors)
+                         ]))
+          isbroken = i==3 && (
+            (f, ad, tol) == (f_oop, true, nothing) ||
```
ChrisRackauckas: all with AD fail?

mauro3: Only those with `ODEFunction(f,colorvec=colors)`, and Trapezoid passes there too.  Note that they are not completely wrong, just about a factor 10 larger (different) error.

test/interface/sparsediff_tests.jl
mauro3: IMO, the color stuff should be independent of the matrix type of J.

mauro3:
It looks like that this touches a lot of the code from PR https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/937 of @huanglangwen.  Maybe @huanglangwen has time to look at this?

test/interface/sparsediff_tests.jl
huanglangwen: Because one must have sparsity to get sparse jacobian work appropriately.

src/derivative_wrappers.jl
```diff
@@ -38,15 +38,9 @@ end
...
-    sparsity = odefun.jac_prototype
-    jac_prototype = nothing
-  else
-    colorvec = 1:length(x)
-    sparsity = nothing
-    jac_prototype = odefun.jac_prototype
-  end
+  colorvec = DiffEqBase.has_colorvec(odefun) ? odefun.colorvec : 1:length(x)
+  sparsity = odefun.jac_prototype
+  jac_prototype = odefun.jac_prototype
```
huanglangwen: Sadly we chose to use `jac_prototype` to represent both sparsity and the real *jac_prototype*, so we got a dilemma that someone may have a sparseCSC for sparsity but still wish to keep a dense jacobian, while others want both sparsity and jacobian to be sparse matrix.

ChrisRackauckas: maybe in the DiffEqFunction we should just default to sparsity=jac_prototype, but allow it to be changed independently?

huanglangwen: Makes sense, just like those in oop colored jacobian.

ChrisRackauckas: it would definitely not be the most used combination, but there are times when I can see it viable (I wanted it last week, so that means it does show up!)

mauro3: Thanks to @huanglangwen, this should now be resolved.

test/interface/sparsediff_tests.jl
mauro3: These are still "broken".

mauro3:
Tests will fail because it now needs https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/449.

src/derivative_utils.jl
```diff
@@ -531,21 +531,27 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{IIP}) where IIP
...
+    end
     isdae = alg isa DAEAlgorithm
     W = if isdae
       J
     elseif IIP
       similar(J)
     else
-      W = if u isa StaticArray
-      lu(J)
+      if u isa StaticArray
```
YingboMa: Those branches can be removed with `lu_instance`.

src/derivative_utils.jl
```diff
@@ -531,21 +531,27 @@ function build_J_W(alg,u,uprev,p,t,dt,f,uEltypeNoUnits,::Val{IIP}) where IIP
...
         u
-      else
+      elseif f.jac_prototype===nothing
         LU{LinearAlgebra.lutype(uEltypeNoUnits)}(Matrix{uEltypeNoUnits}(undef, 0, 0),
-                                                     Vector{LinearAlgebra.BlasInt}(undef, 0),
-                                                     zero(LinearAlgebra.BlasInt))
+                                                 Vector{LinearAlgebra.BlasInt}(undef, 0),
+                                                 zero(LinearAlgebra.BlasInt))
+      else
+        ArrayInterface.lu_instance(f.jac_prototype)
```
YingboMa:
```suggestion
        ArrayInterface.lu_instance(J)
```

test/interface/sparsediff_tests.jl
mauro3: I had a closer look at the broken tests:`maximum(sol_std.u[end].-sol.u[end])` is around 1e-5 for `tol=1e-10` and around `1e-3` for `tol=nothing`.

test/interface/sparsediff_tests.jl
mauro3: Is this still the case now with our update?  Because some tests still behave erratically.

src/derivative_utils.jl
mauro3: Done

mauro3:
The tests pass (apart from https://travis-ci.org/JuliaDiffEq/OrdinaryDiffEq.jl/jobs/658621217?utm_medium=notification&utm_source=github_status, which is due to  JuliaDiffEq/DiffEqBase.jl#449 needing changes in DelayDiffEq) and I think this is ready to be merged. 

For the broken tests, I would suggest to create an issue to not forget about them.

PR #1051, opened 2020-03-01
Implemented the Stanescu, Habashi 5stage 4th order method

kamalojasv181:


src/OrdinaryDiffEq.jl
```diff
@@ -180,7 +180,7 @@ module OrdinaryDiffEq
          SSPRKMSVS32, SSPRKMSVS43, SSPRK932, SSPRK54, SSPRK104
 
   export ORK256, CarpenterKennedy2N54, HSLDDRK64, DGLDDRK73_C, DGLDDRK84_C, DGLDDRK84_F, NDBLSRK124, NDBLSRK134, NDBLSRK144,
-         CFRLDDRK64, TSLDDRK74,
+         CFRLDDRK64, TSLDDRK74,LDDRK54
```
ChrisRackauckas: name it SHLDDRK54

kamalojasv181:
@ChrisRackauckas I think the error in `Integrators_I`  and `Regression` is not due to my commit as the kernal shows error due to  `CKLLSRK43_2` which is something that I never touched. Although error in AlgConvergence-II is i think due to my commit. The error message is 
  `LoadError: MethodError: no method matching alg_cache(::OrdinaryDiffEq.SHLDDRK54, ::Array{Float64,1}, ::Array{Float64,1}, ::Type{Float64}, ::Type{Float64}, ::Type{Float64}, ::Array{Float64,1}, ::Array{Float64,1}, ::DiffEqBase.ODEFunction{true,Main.##484.var"#5#6",LinearAlgebra.UniformScaling{Bool},Main.##484.var"#7#8",Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing}, ::Float64, ::Rational{Int64}, ::Float64, ::DiffEqBase.NullParameters, ::Bool, ::Val{true})`
. I am not able to tell why as I have defined a function named alg_cache.

ChrisRackauckas:
>I am not able to tell why as I have defined a function named alg_cache.

That is telling you that you should've defined an `alg_cache` method like is described in the developer documentation.

https://devdocs.juliadiffeq.org/latest/contributing/adding_algorithms/

src/OrdinaryDiffEq.jl
```diff
@@ -180,7 +180,7 @@ module OrdinaryDiffEq
          SSPRKMSVS32, SSPRKMSVS43, SSPRK932, SSPRK54, SSPRK104
 
   export ORK256, CarpenterKennedy2N54, HSLDDRK64, DGLDDRK73_C, DGLDDRK84_C, DGLDDRK84_F, NDBLSRK124, NDBLSRK134, NDBLSRK144,
-         CFRLDDRK64, TSLDDRK74,
+         CFRLDDRK64, TSLDDRK74,SHLDDRK54
```
```suggestion
         CFRLDDRK64, TSLDDRK74,SHLDDRK54,
```

ChrisRackauckas: that will fix the tests

test/algconvergence/ode_ssprk_tests.jl
```diff
@@ -77,6 +77,20 @@ integ = init(prob_ode_large, alg, dt=1.e-2, save_start=false, save_end=false, sa
 integ = init(prob_ode_large, alg, dt=1.e-2, save_start=false, save_end=false, save_everystep=false, alias_u0=true)
 @test Base.summarysize(integ) ÷ Base.summarysize(u0_large) <= 4
 
+println("SHLDDRK54")
+alg = SHLDDRK54()
+for prob in test_problems_only_time
+  sim = test_convergence(dts, prob, alg)
+  @test_broken sim.𝒪est[:final] ≈ OrdinaryDiffEq.alg_order(alg) atol=testTol
```
ChrisRackauckas: what do these convergence plots look like?

ChrisRackauckas: Do `plot(sim)`.

kamalojasv181:
@ChrisRackauckas the convergence plots were always good but with the problems that I tried. It took me good time to figure out that I had not added methods and cache for inplace problems. Now my package is passing all the tests on my machine. I think now it should not show any errors now.

kamalojasv181:
@ChrisRackauckas I think we can update the example code in the documentation. We can include a cache and perform_step for inplace updates so that next time someone else doesn't get confused. What say?

ChrisRackauckas:
Makes sense. Looks like the convergence plots are all `@test_broken` still?

kamalojasv181:
Following are the convergence plots for some of the problems.

Screenshot from 2020-03-03 06-09-18

Screenshot from 2020-03-03 06-09-41

Screenshot from 2020-03-03 06-09-52

ChrisRackauckas:
Looks like there's an error in the 6th or 7th digit of one of the tableau terms.

kamalojasv181:
Are you implying that some value in the paper are incorrect?

ChrisRackauckas:
Either that or the translation is incorrect. You'l need to dig in and check that order constraints to find which term is wrong.

kamalojasv181:
I am surprised how it passed the tests earlier.

kamalojasv181:
@ChrisRackauckas I tried verifying the consistency of all the α's, β's, γ's, a's, b's and c's and found them to be correct.
Screenshot from 2020-03-04 23-48-54

Screenshot from 2020-03-04 23-49-16

Screenshot from 2020-03-04 23-49-21

Also I checked convergence plot of a the 6 stage method(of the same paper which is already merged) and my 5 stage method for the same problem and found their convergence plots to be respectively
Screenshot from 2020-03-05 05-58-36

Screenshot from 2020-03-05 06-00-36

kamalojasv181:
@ChrisRackauckas do you think there is still room for improvement in the algorithm?

ChrisRackauckas:
The tests still seem to fail.

kamalojasv181:
@ChrisRackauckas , I tested the convergence by changing the dts(powers of 1/1.37) and I got the plot

Screenshot from 2020-03-07 00-58-37

I also tested some of other 2nd order algorithms and the error was of same order. Any inputs?

kamalojasv181:
@ChrisRackauckas , I was thinking maybe, we should update documentation https://devdocs.juliadiffeq.org/latest/contributing/adding_algorithms/ including the inplace functions?

ChrisRackauckas:
That would probably be a good idea.

PR #1054, opened 2020-03-04
Fixes chunksize

### LINKED ISSUE #1053: `chunk_size` not honoured
My understanding is that `chunk_size` will fix the `chunksize` used with `ForwardDiff`.  If this is so, then it's not working:
```julia
using OrdinaryDiffEq, ForwardDiff
const chunksize = 2

function rober(du,u,p,t)
  if eltype(u) <: ForwardDiff.Dual
    if ForwardDiff.npartials(u[1])!=chunksize
      @show chunksize
      @show ForwardDiff.npartials(u[1])
      error()
    end
  end

  y₁,y₂,y₃ = u
  k₁,k₂,k₃ = p
  du[1] = -k₁*y₁+k₃*y₂*y₃
  du[2] =  k₁*y₁-k₂*y₂^2-k₃*y₂*y₃
  du[3] =  k₂*y₂^2
  nothing
end
prob = ODEProblem(rober,[1.0,0.0,0.0],(0.0,1e5),(0.04,3e7,1e4))
sol = solve(prob, TRBDF2(chunk_size=chunksize))
```
prints and errors:
```
chunksize = 2
ForwardDiff.npartials(u[1]) = 3
ERROR: LoadError: 
```

I came across this because I wanted to use `DiffEqBase.dualcache` with a given `chunksize` but it doesn't work.

*Edit*: update due to Chris' comment `chunksize` 55->2

mauro3:
Fixes #1053. 

This is on top of PR #1049, so before that is not merged the diff will be messy.

Tnx to @huanglangwen, I only did the typing.

### DIFF
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -36,6 +36,19 @@ function derivative(f, x::Union{Number,AbstractArray{<:Number}},
     end
 end
 
+jacobian_autodiff(f, x, odefun, alg) = (ForwardDiff.derivative(f,x),1, alg)
+function jacobian_autodiff(f, x::AbstractArray, odefun, alg)
+  colorvec = DiffEqBase.has_colorvec(odefun) ? odefun.colorvec : 1:length(x)
+  sparsity = odefun.sparsity
+  jac_prototype = odefun.jac_prototype
+  maxcolor = maximum(colorvec)
+  chunk_size = get_chunksize(alg)==0 ? nothing : get_chunksize(alg) # SparseDiffEq uses different convection...
+  num_of_chunks = chunk_size==nothing ? Int(ceil(maxcolor / getsize(default_chunk_size(maxcolor)))) :
+                                        Int(ceil(maxcolor / chunk_size))
+  (forwarddiff_color_jacobian(f,x,colorvec = colorvec, sparsity = sparsity,
+                              jac_prototype = jac_prototype, chunksize=chunk_size),
+   num_of_chunks)
+end
 
 function _nfcount(N,diff_type)
   if diff_type==Val{:complex}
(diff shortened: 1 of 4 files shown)


src/derivative_wrappers.jl
```diff
@@ -37,19 +37,13 @@ function derivative(f, x::Union{Number,AbstractArray{<:Number}},
...
-    jac_prototype = odefun.jac_prototype
-  end
+function jacobian_autodiff(f, x::AbstractArray, odefun, chunk_size)
+  colorvec = DiffEqBase.has_colorvec(odefun) ? odefun.colorvec : 1:length(x)
+  sparsity = odefun.sparsity
+  jac_prototype = odefun.jac_prototype
   maxcolor = maximum(colorvec)
-  chunksize = getsize(default_chunk_size(maxcolor))
-  num_of_chunks = Int(ceil(maxcolor / chunksize))
+  chunk_size = chunk_size==0 ? getsize(default_chunk_size(maxcolor)) : chunk_size
```
ChrisRackauckas: this accepts Val?

src/derivative_wrappers.jl
```diff
@@ -37,19 +37,13 @@ function derivative(f, x::Union{Number,AbstractArray{<:Number}},
...
+  num_of_chunks = Int(ceil(maxcolor / chunk_size))
```
mauro3: @huanglangwen I'm a bit confused here: shouldn't the chunksize be passed to `forward_color_jacobian`?

huanglangwen: ah, sorry forgot this one.
https://github.com/JuliaDiff/SparseDiffTools.jl/blob/36978ea873c4a530f1decf01c99c901d33afbcb4/src/differentiation/compute_jacobian_ad.jl#L76

We should instead build a jaccache here like iip one.

src/derivative_wrappers.jl
huanglangwen: Yeah, need to wrap it with getsize

src/derivative_wrappers.jl
mauro3: Problem is that the OOP cache types don't have a field for `jac_cache`, thus as it stands this does not work.  Should this field be added or not? If so, I might need some help getting the plumbing right.

mauro3:
See https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/1054#discussion_r388575183 for latest news.

src/derivative_wrappers.jl
ChrisRackauckas: with OOP the assumption is that the cache should really just be local or type information to elide allocations.

mauro3:
This should be good now but it needs https://github.com/JuliaDiff/SparseDiffTools.jl/pull/96.

src/derivative_wrappers.jl
mauro3: I don't follow here. Is the code as it stands now ok?

src/derivative_wrappers.jl
mauro3: Ok, I hope the implementation as it stands now is good enough.

ChrisRackauckas:
Thanks for this good work. We'll need to figure out what's going on with StochasticDiffEq since this set of PRs had a few low level changes we need to shake out, but this is all some very good fixes.

mauro3:
Ok, Travis tests are green except the expected one related to DiffEqSensitivity and https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/449.  AppV test failed, not sure why.

mauro3:
Aright, hold-on with merging this.  My personal test case doesn't work yet and I want to investigate first.  Probably tomorrow.

mauro3:
The bug was in my script (I passed the chunksize as `Type{Val{12}}`, as is used by the `dualcache` function).  

Also the local, broken test case reported above https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/1054#issuecomment-595860471, seems to fail also fail on master.

So, from my point of view, this is good for merging.

ChrisRackauckas:
>Also the local, broken test case reported above #1054 (comment), seems to fail also fail on master.

Looks like it's processor dependent, given how much is optimized on, for what the convergence value is. We can make that test a little better, but its not a major concern.

ChrisRackauckas:
I'm handling the DiffEqSensitivity stuff today. Cheers!

mauro3:
And thanks again to @huanglangwen!

PR #1058, opened 2020-03-08
WIP: BDF Fixes

utkarsh530:
Hi, I am trying to fix random errors in BDF popping in travis build. I removed some `similar` allocations which I found to be suspicious. Testing more fixes.
@ChrisRackauckas @kanav99

### DIFF
--- a/src/caches/bdf_caches.jl
+++ b/src/caches/bdf_caches.jl
@@ -1,4 +1,4 @@
-  @cache mutable struct ABDF2ConstantCache{N,dtType,rate_prototype} <: OrdinaryDiffEqConstantCache
+@cache mutable struct ABDF2ConstantCache{N,dtType,rate_prototype} <: OrdinaryDiffEqConstantCache
   nlsolver::N
   eulercache::ImplicitEulerConstantCache
   dtₙ₋₁::dtType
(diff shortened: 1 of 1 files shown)


src/caches/bdf_caches.jl
```diff
@@ -1,4 +1,4 @@
-@cache mutable struct ABDF2ConstantCache{N,dtType,rate_prototype} <: OrdinaryDiffEqConstantCache
+  @cache mutable struct ABDF2ConstantCache{N,dtType,rate_prototype} <: OrdinaryDiffEqConstantCache
```
ChrisRackauckas: intentional?

ChrisRackauckas:
Those are definitely suspicious! Good catch.

src/caches/bdf_caches.jl
ChrisRackauckas:
```suggestion
@cache mutable struct ABDF2ConstantCache{N,dtType,rate_prototype} <: OrdinaryDiffEqConstantCache
```

ChrisRackauckas:
It's at least an improvement, so I'm merging.

utkarsh530:
Thanks! 😃

PR #1059, opened 2020-03-08
More stable Anas5 convergence testing

ChrisRackauckas:
It was hitting floating point zero error... so that effected it. Pulling the dts back revealed the true behavior which is that it achieves 6th order on problems where it knows the periodicity.


ChrisRackauckas:
@mauro3 thanks!

PR #1061, opened 2020-03-12
[WIP] Bugfixing approaches for midpointsplitting 

Biswajitghosh98:
I uncommented the tests for midpointsplitting, so that I can use its prototype for implementing a Midpoint Magnus Method approach, as suggested by @ChrisRackauckas 
1.While running the tests for "linear_method_tests.jl", the second parameter for DiffEqArrayOperator threw an error, treating 'f' as a variable, rather than a function. 
``` 
Linear Methods Tests: Error During Test at /home/biswajit/.julia/packages/SafeTestsets/A83XK/src/SafeTestsets.jl:25
  LoadError: MethodError: no method matching DiffEqBase.DiffEqArrayOperator(::Array{Int64,2}, ::Main.##422.var"#1#2")
  Closest candidates are:
    DiffEqBase.DiffEqArrayOperator(::AType; update_func) where AType at /home/biswajit/.julia/packages/DiffEqBase/y3Q1t/src/operators/basic_operators.jl:90
  Stacktrace: 
```

2. So, to fix that, I mapped f to the matrix. Then I got the error prompting that argument names of the function p are not unique. I saw that not typecasting p can bypass this error. So, just to see what happens next, I moved forward with it

```
Linear Methods Tests: Error During Test at /home/biswajit/.julia/packages/SafeTestsets/A83XK/src/SafeTestsets.jl:25
  LoadError: syntax: function argument names not unique
  Stacktrace:

``` 
3. Now, I am prompted the same error which I encountered in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/1055 , because I had used @. for broadcasting in perform_step, whereas using @.. solved the issue. 
I tried working on similar lines here, but that didn't quite work out.

```
Linear Methods Tests: Error During Test at /home/biswajit/.julia/packages/SafeTestsets/A83XK/src/SafeTestsets.jl:25
  LoadError: MethodError: no method matching alg_cache(::OrdinaryDiffEq.MidpointSplitting, ::Array{Float64,1}, ::Array{Float64,1}, ::Type{Float64}, ::Type{Float64}, ::Type{Float64}, ::Array{Float64,1}, ::Array{Float64,1}, ::DiffEqBase.ODEFunction{true,DiffEqBase.AffineDiffEqOperator{Float64,Tuple{Array{Float64,2},Array{Float64,2}},Tuple{},Array{Float64,1}},LinearAlgebra.UniformScaling{Bool},Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing}, ::Float64, ::Float64, ::Float64, ::DiffEqBase.NullParameters, ::Bool, ::Val{true})
  Closest candidates are:
    alg_cache(!Matched::OrdinaryDiffEq.SplitEuler, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Val{true}) at /home/biswajit/Desktop/GSOC/OrdinaryDiffEq.jl/src/caches/low_order_rk_caches.jl:18
    alg_cache(!Matched::OrdinaryDiffEq.Euler, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Val{true}) at /home/biswajit/Desktop/GSOC/OrdinaryDiffEq.jl/src/caches/low_order_rk_caches.jl:26

```

Any pointers on how to move forward with it will be very helpful !
Thank You !


test/algconvergence/linear_method_tests.jl
```diff
@@ -17,47 +17,47 @@ A = DiffEqArrayOperator([2.0 -1.0; -1.0 2.0])
...
+A = map(f,[0 1 0 0
+           0 0 0 0
+           0 0 0 0
+           0 0 0 0])
 
-# B = DiffEqArrayOperator([0 0 0 0
-#                          0 0 1 0
-#                          0 0 0 0
-#                          0 0 0 0], f)
+B = map(f,[0 0 0 0
```
ChrisRackauckas: This is not the same: this is not a time-dependent operator!

PR #1070, opened 2020-03-13
Singular MM's initialization

ChrisRackauckas:
Fixes https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/1039
Fixes https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/1042

@YingboMa can you add the singular mass-matrix initialization?

@kanav99 what should we do about `du`? Re-compute it on demand when we reinitialize?

P.S. need a doc update that describes the intializealgs

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,7 +26,7 @@ UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 [compat]
 ArrayInterface = "1.1, 2.0"
 DataStructures = "0.17"
-DiffEqBase = "6.19"
+DiffEqBase = "6.21"
 ExponentialUtilities = "1.2"
 FiniteDiff = "2"
 ForwardDiff = "0.10.3"
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -62,7 +62,6 @@ module OrdinaryDiffEq
   include("misc_utils.jl")
   include("algorithms.jl")
   include("alg_utils.jl")
-  include("initialize_dae.jl")
 
   include("nlsolve/type.jl")
   include("nlsolve/utils.jl")
@@ -112,6 +111,7 @@ module OrdinaryDiffEq
   include("integrators/integrator_utils.jl")
   include("cache_utils.jl")
   include("integrators/integrator_interface.jl")
+  include("initialize_dae.jl")
 
   include("perform_step/fixed_timestep_perform_step.jl")
   include("perform_step/symplectic_perform_step.jl")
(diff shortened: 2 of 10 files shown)


kanav99:
When all could the user ask for `du`? inside callbacks I guess and after each `step!` - any other case? In full initialization, `du` will always be set automatically. Otherwise in most BDFs, we have a linear formula of `du` we can have a separate function for each BDF. Note that `du` is never used internally and is not important for the algorithm to work.

ChrisRackauckas:
I guess it only shows up in initialize, but with this method we need a `du` to reinitialize after a callback, so we somehow have to pick one.

ChrisRackauckas:
I added initialization for singular mass matrices, but it assumes semi-explicit form.

ChrisRackauckas:
Open an issue on semi-explicit form in DAE initialization

ChrisRackauckas:
Open an issue on tolerances of DAE initialization

ChrisRackauckas:
Talking with Yingbo, I realized that things that commonly use a non semiexplicit form are index 2, since they are relating differential variables to differential variables as a constraint. Seems like an issue to punt for now.

kanav99:
> Open an issue on semi-explicit form in DAE initialization
We don't have a problem form for Semi Explicit DAEs. Can I work on it next?

dkarrasch:
This PR made the following problem fail, because it now calls `lu` on a `Diagonal`, which is not defined in `LinearAlgebra`:
```julia
using OrdinaryDiffEq, DiffEqDevTools, SparseArrays, LinearAlgebra

u0 = sin.(range(0, pi, length = 100))
tspan = (0.0, 1.0)

M = 0.5Diagonal(ones(100))
A = spdiagm(0 => ones(100))
update_func = (_A, u, p, t) -> _A.nzval .= t
f = DiffEqArrayOperator(A; update_func=update_func)

prob = ODEProblem(ODEFunction(f; mass_matrix=M), u0, tspan)
solve(prob, ImplicitEuler(linsolve=LinSolveFactorize(lu)))
```
gives
```julia
ERROR: MethodError: no method matching lu!(::Diagonal{Float64,Array{Float64,1}}, ::Val{true}; check=false)
```
Should there be specializations of `issingular` for structured matrices? Or maybe some keyword option, that allows to set `issingular` to `true`, but does an `lu` check by default?

ChrisRackauckas:
>Should there be specializations of issingular for structured matrices? Or maybe some keyword option, that allows to set issingular to true, but does an lu check by default?

Yes, we should specialize this on operator types and matrices.

ChrisRackauckas:
Sorry if this is taking awhile, but I want to make sure we do this right. I'm taking this all the way to ArrayInterface to make `issingular` a thing: https://github.com/JuliaDiffEq/ArrayInterface.jl/pull/42 and then we'll just be using that.

PR #1073, opened 2020-03-13
Implemented 2N storage Stanescu, Habashi RK method

kamalojasv181:
Screenshot from 2020-03-13 12-17-03
Convergence for dts in powers of 1.35


ChrisRackauckas:
Sounds like some coefficients are incorrect.

kamalojasv181:
@ChrisRackauckas , I checked all the coefficients from the paper, I have copied them all correctly. The problem is similar as in my last Pr to a different method of the same issue, when I tweak the dts a little more, I get a straight plot.
Screenshot from 2020-03-13 14-59-58

kamalojasv181:
I think this is related to floating point precision.

src/caches/ssprk_caches.jl
```diff
@@ -169,6 +169,97 @@ function alg_cache(alg::SHLDDRK52,u,rate_prototype,uEltypeNoUnits,uBottomEltypeN
...
+  c32::T2
+  c42::T2
+  c52::T2
+  c62::T2
+
+  step::Int
+end
+
+function SHLDDRK_2NConstantCache(T1,T2)
+  α21 = T1(-0.6051226)
```
ChrisRackauckas: these coefficients are not high accuracy. Do they not provide more digits?

ChrisRackauckas:
The precision ends where the coefficients end. But that plot where you have a straight line is capturing the convergence rate, is the rate the right value? If not, you need to check the order conditions are satisfied.

src/caches/ssprk_caches.jl
kamalojasv181: There are only these many digits provided.

kamalojasv181:
Screenshot from 2020-03-17 05-34-21
My alg is passing all the tests

kamalojasv181:
Although after using a high tolerence

test/algconvergence/ode_ssprk_tests.jl
```diff
@@ -92,6 +92,22 @@ for prob in test_problems_nonlinear
   @test_broken sim.𝒪est[:final] ≈ OrdinaryDiffEq.alg_order(alg) atol=testTol
 end
 
+println("SHLDDRK_2N")
+dts_SHLDDRK_2N  = (1/2) .^ (0:3)
+alg = SHLDDRK_2N()
+for prob in test_problems_only_time
+  sim = test_convergence(dts_SHLDDRK_2N, prob, alg)
+  @test sim.𝒪est[:final] ≈ 4 atol=1
```
ChrisRackauckas: this tolerance is way too high to trust.

kamalojasv181:
Testing it against 5 will fail tests which give value around 4. And there are such tests

kamalojasv181:
Sorry , I mistakenly read 4.46 as 4.64 and since it was larger and tests were passing, therefore I never checked it again

kamalojasv181:
is 0.46 high as well?

ChrisRackauckas:
there is still an atol of 1 in here.

kamalojasv181:
Yes that's because one otest is 3.3 and other 3

PR #1090, opened 2020-03-18
[WIP]Lie euler

Biswajitghosh98:


Biswajitghosh98:
@YingboMa 
No, the two are different, as 
Y(n) = e^(hL)*Y(n-1) + e^(hL)*hN(Y(n-1)) is the expression for Lawson Euler, (https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/exponential_rk_perform_step.jl L54-60)
and 
Y(n) = e^(hF(Y(n-1)))*Y(n-1) for Lie Euler are differently implemented (https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/1090/files linear_perform_step.jl L50-55).

I may be wrong. Please rectify me if that's the case. Thank You ! :)

ChrisRackauckas:
Indeed, it's different. One is semilinear and a possibly dependent operator, but this is just for a quasi-linear system.

PR #1092, opened 2020-03-20
[WIP] MagnusGauss4

Biswajitghosh98:
Issue : It fails convergence tests as of now.
Expected : 4
Result : 1

### DIFF
--- a/src/perform_step/linear_perform_step.jl
+++ b/src/perform_step/linear_perform_step.jl
@@ -54,10 +54,10 @@ function perform_step!(integrator, cache::MagnusGauss4Cache, repeat_step=false)
   # println("Here b $B")
 
   if integrator.alg.krylov
-    u .= (expv(dt,(A+B)/2 + dt*sqrt(3)*(B*A-A*B)/12, u; m=min(alg.m, size(L1,1)), opnorm=integrator.opts.internalopnorm, iop=alg.iop)
+    u .= expv(dt,(A+B) ./ 2 + (dt*sqrt(3)) .* (B*A-A*B) ./ 12, u; m=min(alg.m, size(L1,1)), opnorm=integrator.opts.internalopnorm, iop=alg.iop)
   else
     #A = Matrix(L) #size(L) == () ? convert(Number, L) : convert(AbstractMatrix, L)
-    u .= exp((dt/2)*(A+B)+((dt^2)*(sqrt(3)/12))*(B*A-A*B)) * u
+    u .= exp((dt/2) .* (A+B)+((dt^2)*(sqrt(3)/12)) .* (B*A-A*B)) * u
   end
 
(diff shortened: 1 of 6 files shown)


src/perform_step/linear_perform_step.jl
```diff
@@ -28,6 +28,42 @@ function perform_step!(integrator, cache::MagnusMidpointCache, repeat_step=false
...
+end
+
+function perform_step!(integrator, cache::MagnusGauss4Cache, repeat_step=false)
+  @unpack t,dt,uprev,u,p,alg = integrator
+  @unpack W,k,tmp = cache
+  mass_matrix = integrator.f.mass_matrix
+
+  L1 = integrator.f.f
+  update_coefficients!(L1,u,p,t+dt*(1/2+sqrt(3)/6))
+  L2 = integrator.f.f
```
ChrisRackauckas: This isn't going to work: it's going to alias the two operators.

src/perform_step/linear_perform_step.jl
```diff
@@ -28,6 +28,42 @@ function perform_step!(integrator, cache::MagnusMidpointCache, repeat_step=false
...
+  update_coefficients!(L2,u,p,t+dt*(1/2-sqrt(3)/6))
+  A = Matrix(L1)
```
ChrisRackauckas: I don't know if you want to allocate, but for now sure

src/perform_step/linear_perform_step.jl
```diff
@@ -28,6 +28,42 @@ function perform_step!(integrator, cache::MagnusMidpointCache, repeat_step=false
...
+  B = Matrix(L2)
+  # println("Here a $A")
+  # println("Here b $B")
+
+  if integrator.alg.krylov
+    u .= expv(dt,(A+B) ./ 2 + (dt*sqrt(3)) .* (B*A-A*B) ./ 12, u; m=min(alg.m, size(L1,1)), opnorm=integrator.opts.internalopnorm, iop=alg.iop)
```
ChrisRackauckas: I think we want to do successive application, though doing the commutator will be the issue...

src/perform_step/linear_perform_step.jl
Biswajitghosh98: @ChrisRackauckas So, if i allocate A = Matrix(L), then update L again and then assign B=Matrix(L), should it work then ?

ChrisRackauckas: yes it should. We won't want to do this for performance reasons, but do it for now so we can start with a working version to then optimize.

Biswajitghosh98: Doing it increased the order to 2.

ChrisRackauckas:
Test with the non-Krylov method for now

Issue #1077, opened 2020-03-15
QNDFs fail mass matrix tests while QBDFs pass


ChrisRackauckas:
Curiously, QNDF2 is fine. From the mass_matrix_tests.jl:

```julia

function make_mm_probs(mm_A, ::Val{iip}) where iip
  # iip
  function mm_f(du,u,p,t)
    update_coefficients(mm_A,u,p,t)
    mm_b = vec(sum(mm_A; dims=2))
    mul!(du,mm_A,u)
    du .+= t * mm_b
    nothing
  end
  mm_g(du,u,p,t) = (@. du = u + t; nothing)

  # oop
  mm_f(u,p,t) = mm_A * (u .+ t)
  mm_g(u,p,t) = u .+ t

  mm_analytic(u0, p, t) = @. 2 * u0 * exp(t) - t - 1

  u0 = ones(3)
  tspan = (0.0, 1.0)

  prob = ODEProblem(ODEFunction{iip,true}(mm_f, analytic=mm_analytic, mass_matrix=mm_A), u0, tspan)
  prob2 = ODEProblem(ODEFunction{iip,true}(mm_g, analytic=mm_analytic), u0, tspan)

  prob, prob2
end

function _norm_dsol(alg,prob,prob2,dt=1/10)
  sol = solve(prob,  alg,dt=dt,adaptive=false)
  sol2 = solve(prob2,alg,dt=dt,adaptive=false)
  norm(sol .- sol2)
end

almost_I = Matrix{Float64}(1.01I, 3, 3)

prob, prob2 = make_mm_probs(almost_I, Val(true))
@show _norm_dsol(QBDF1(),prob,prob2)
@show _norm_dsol(QNDF1(),prob,prob2)
@show _norm_dsol(QBDF2(),prob,prob2)
@show _norm_dsol(QNDF2(),prob,prob2)
@show _norm_dsol(QBDF(),prob,prob2)
@show _norm_dsol(ABDF2(),prob,prob2)
@show _norm_dsol(QNDF(),prob,prob2)
```

```
_norm_dsol(QBDF1(), prob, prob2) = 9.42055475210265e-16
_norm_dsol(QNDF1(), prob, prob2) = 2.1123576160648838e-5
_norm_dsol(QBDF(), prob, prob2) = 9.42055475210265e-16
_norm_dsol(ABDF2(), prob, prob2) = 0.0
[...]

PR #1094, opened 2020-03-26
Fix: QNDF mass matrix tests

utkarsh530:
Fixes All BDF and NDF methods for mass matrix tests #1077.
@ChrisRackauckas please review.

### DIFF
--- a/src/caches/bdf_caches.jl
+++ b/src/caches/bdf_caches.jl
@@ -327,7 +327,7 @@ function alg_cache(alg::QNDF,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnit
 
   max_order = 5
   atmp = similar(u,uEltypeNoUnits)
-  utilde = nlsolver.tmp
+  utilde = similar(u)
 
   QNDFCache(fsalfirst,D,D2,R,U,1,max_order,udiff,dts,atmp,utilde,nlsolver,h,0)
 end
--- a/src/caches/bdf_caches.jl
+++ b/src/caches/bdf_caches.jl
@@ -172,7 +172,7 @@ function alg_cache(alg::QNDF1,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
   U!(1,U)
 
   atmp = similar(u,uEltypeNoUnits)
-  utilde = similar(u)
+  utilde = zero(u)
   uprev2 = zero(u)
   dtₙ₋₁ = zero(dt)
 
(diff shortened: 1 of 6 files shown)


kanav99:
Split Regression test set into two sets

src/perform_step/bdf_perform_step.jl
```diff
@@ -653,8 +653,6 @@ function perform_step!(integrator,cache::QNDFConstantCache,repeat_step=false)
       end
       backward_diff!(cache,D,D2,k)
     end
-  else
```
ChrisRackauckas: what happens during the else now?

src/caches/bdf_caches.jl
```diff
@@ -327,7 +327,7 @@ function alg_cache(alg::QNDF,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnit
 
   max_order = 5
   atmp = similar(u,uEltypeNoUnits)
-  utilde = nlsolver.tmp
+  utilde = similar(u)
```
ChrisRackauckas:
```suggestion
  utilde = zero(u)
```

src/perform_step/bdf_perform_step.jl
utkarsh530: The gamma value should not be updated before `cnt>k`, the code was doing that previously. You can see the same results in QNDF1. (Current Implementation gives 1st order estimate only which needs to be checked)

src/caches/bdf_caches.jl
utkarsh530: Okay

utkarsh530:
@ChrisRackauckas made the requested changes.

ChrisRackauckas:
Tests fail

utkarsh530:
Rebuiliding Travis

kanav99:
If you compare the times of all jobs in last master successful build [here](https://travis-ci.org/github/SciML/OrdinaryDiffEq.jl/builds/664163754?utm_medium=notification&utm_source=github_status) and this build, it seems like Travis has cut down it's resources recently and every job has its time increased. I think the best bet is to split up the Regression tests.
I am restarting the master tests if my claim is right.

utkarsh530:
Alright @kanav99 thanks for the help. I will update the PR with split regression tests.

utkarsh530:
@ChrisRackauckas please have a look.

ChrisRackauckas:
These fixes will do more than just mass matrices: check the benchmarks and see if they improved.

PR #1095, opened 2020-03-28
implemented KYK SSPRK42

d-netto:


ChrisRackauckas:
Which paper does this correspond to?

src/perform_step/ssprk_perform_step.jl
```diff
@@ -51,6 +51,66 @@ end
...
+  @unpack α10, α20, α21, α30, α32, α40, α43, β10, β21, β30, β32, β40, β43 = cache.tab
+
+  # u1 -> stored as u
+  @. u = α10*uprev + dt*β10*integrator.fsalfirst
+  f(u, p, t)
+  # u2
+  @. u₂ = α20*uprev + α21*u + dt*β21*k
+  f(u₂, p, t)
+  # u3
+  @. tmp = α30*uprev + α32*u2 + dt*β30*integrator.fsalfirst + dt*β32*k
```
ChrisRackauckas: you might want to `@..` this line and the next one.

ChrisRackauckas: It's a special internal macro we made which is more optimized than `@.` on long expressions.

ChrisRackauckas:
LGTM. Will merge after tests are passing, but we should change that use of `@.` to `@..`

src/perform_step/ssprk_perform_step.jl
d-netto: Ok. I'll change it to @..

The paper is Kubatko, Yeager, Ketcheson (2014)

ChrisRackauckas:
make Radau in a different PR and revert that. This is good to merge. Also, use a different branch when you create PRs: doing PRs from master always becomes painful sometime or another.

d-netto:
ok, gonna revert it

ChrisRackauckas:
Looks like the convergence rate is 1 causing a failure.

d-netto:
> Looks like the convergence rate is 1 causing a failure.

Do you know what could be causing it?

ChrisRackauckas:
Check the conditions on the coefficients for one. Narrow it down: is it the in-place version or the out of place? If it's in-place, check for possible aliasing, etc. If you figure out what exact case it is but can't figure out the error, then let me know and I can look for something that could cause that behavior.

d-netto:
Out of place test gives a convergence rate of 1. In place test gives a Method Error

ChrisRackauckas:
Looks like there's still a test failure. This causes the tests to stall?

d-netto:
not sure what is causing the test failure here. The versions before adding the @.. macro passed

d-netto:
Oh, just saw your comment in the slack, nvm

d-netto:
it passed now. I'll keep working in the Radau method

ChrisRackauckas:
Awesome.

PR #1097, opened 2020-03-29
Make `nextfloat(integrator.opts.dtmin)` the lower bound of initdt

YingboMa:
This prevents erroring immediately after `initdt`.


ChrisRackauckas:
Can you do the same to StochasticDiffEq?

PR #1101, opened 2020-04-03
Relax iip vs oop tests

YingboMa:


YingboMa:
For high order integrators, the error estimate is so low (~ 1e-13) that we are going to lose digits in computing step sizes. So, there is going to be a relatively big discrepancy in IIP vs OOP tests in those methods.

devmotion:
That's unfortunate since without adaptivity it's not checked anymore that the step sizes and hence the error estimates are roughly the same. Is there a way to avoid these issues by modifying or changing the problem instead?

YingboMa:
Maybe change it to `BigFloat`?

PR #1105, opened 2020-04-08
Fixing DAEs

kanav99:
- [x] FSAL
- [ ] Callbacks
- [ ] Interpolation

### DIFF
--- a/src/integrators/integrator_utils.jl
+++ b/src/integrators/integrator_utils.jl
@@ -368,6 +368,11 @@ end
 
 function reset_fsal!(integrator)
   # Under these condtions, these algorithms are not FSAL anymore
+  if integrator.f isa DAEFunction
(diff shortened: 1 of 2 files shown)


src/integrators/integrator_utils.jl
```diff
@@ -368,6 +368,11 @@ end
 
 function reset_fsal!(integrator)
   # Under these condtions, these algorithms are not FSAL anymore
+  if integrator.f isa DAEFunction
+    set_current_du!(integrator, integrator.cache)
```
ChrisRackauckas, line 372: This can't be done because historical values are invalid after a discontinuity.

TobiasFranzBRB:
What is the status of "DABDF2 fails with discrete callbacks"?

ChrisRackauckas:
Needs someone to fix it.

PR #1106, opened 2020-04-10
Improvements in adaptivity of QNDF

utkarsh530:
I had found some inconsistencies which I am addressing here:

1. QNDF is VSVO(variable step variable order). Currently, the MATLAB suite paper didn't mention on how to change order. So while looking on PRs, I found this paper [here](https://www.sciencedirect.com/science/article/pii/S1877050914002683) which states how to reject steps and select orders.
2. Although the solution obtaining with current QNDF was accurate (I checked with graphs and extreme values, any full-proof method on how to compare the solution with likes of RODAS5 will be appreciated), the no. of steps it took to reach t_final was mammoth (around 600k array dimension). 
3. What I anticipate was happening that, from setting integrator.EEst, the cache.h changes in the next step. However, the current code was directly making changes to cache.h, without changing integrator.EEst adhering to pass/fail, although the correct step was being calculated, but was superseded by previously setting integrator.EEst. Hence made required changes as in the PR.

Results:

Before fixing: 
```
@time sol = solve(prob,QNDF(),abstol=1/10^14,reltol=1/10^14)
141.594930 seconds (38.52 M allocations: 1.798 GiB, 0.94% gc time)
retcode: Success
t: 614872-element Array{Float64,1}
```
After fixing:
```
@time sol = solve(prob,QNDF(),abstol=1/10^14,reltol=1/10^14)
  3.439076 seconds (1.32 M allocations: 61.825 MiB)
retcode: Success
t: 1715-element Array{Float64,1}
```

Contrast to RODAS5 which takes 1243 steps and ode15s takes 1210 steps.

For consistency of solution:
```
sol = solve(prob, QNDF())
sol1 = solve(prob, MATLABDiffEq.ode15s())
sol2 = solve(prob, Rodas5())
julia> plot(sol,linewidth = 3,tspan=(0.0,5.0))
julia> plot!(sol3,lw = 3,ls = :dash,tspan=(0.0,5.0))
```
qndf

@ChrisRackauckas @kanav99 please review.

### DIFF
--- a/src/perform_step/bdf_perform_step.jl
+++ b/src/perform_step/bdf_perform_step.jl
@@ -527,10 +527,10 @@ function perform_step!(integrator,cache::QNDF2Cache,repeat_step=false)
   alg = unwrap_alg(integrator, true)
   cnt = integrator.iter
   k = 2
-  if cnt == 1 || cnt == 2
+  if cnt <=2
     κ = zero(alg.kappa)
     γ₁ = 1//1
-    γ₂ = 1//1
+    γ₂ = 1//1 + 1//2
   elseif dtₙ₋₁ != dtₙ₋₂
     κ = alg.kappa
     γ₁ = 1//1
@@ -794,9 +794,9 @@ function perform_step!(integrator,cache::QNDFCache,repeat_step=false)
     end
   else
     κ = zero(integrator.alg.kappa[k])
-    for i = 1:k
-      γₖ[i] = 1//1
-    end
+    # for i = 1:k
+    #   γₖ[i] = 1//1
+    # end
   end
   γ = inv((1-κ)*γₖ[k])
   nlsolver.γ = γ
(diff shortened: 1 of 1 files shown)


src/perform_step/bdf_perform_step.jl
```diff
@@ -822,18 +822,18 @@ function perform_step!(integrator,cache::QNDFCache,repeat_step=false)
...
 
   if integrator.opts.adaptive
-    if cnt == 1
+    if cnt > k
       integrator.EEst = one(integrator.EEst)
-    elseif cnt == 2
-      @.. utilde = (u - uprev) - (udiff[1] * dt/dts[1])
-      calculate_residuals!(atmp, utilde, uprev, u, integrator.opts.abstol, integrator.opts.reltol, integrator.opts.internalnorm, t)
-      integrator.EEst = integrator.opts.internalnorm(atmp,t)
+    # elseif cnt == 2
```
ChrisRackauckas, line 827: agreed, this is weird...

src/perform_step/bdf_perform_step.jl
```diff
@@ -822,18 +822,18 @@ function perform_step!(integrator,cache::QNDFCache,repeat_step=false)
 
 
   if integrator.opts.adaptive
-    if cnt == 1
+    if cnt > k
```
ChrisRackauckas, line 825: why >k?

utkarsh530: Yeah, I guess there's some rectification required. (Was done because 1st step is calculated using BDF (cnt = 1), and then NDF is used. That's why set it EEst to 1 when cnt>k).

ChrisRackauckas: but after k steps it should be safe to be changing around the stepsize?

utkarsh530: Are you talking about `integrator.EEst`?  I assume it must be safe. But let me look into it further to make adaptivity more logical. Can you suggest something?

utkarsh530:
Another weird issue: 
After applying changes to the out-of-place method (not pushed currently), the algorithm starts giving wrong results. (Although the same problem works perfectly in in-place)

MWE:

```
g(u,p,t) = 10 * u

u0 = 1.01
tspan = (0.0, 1.0)
prob1 = ODEProblem(g, u0, tspan)

sol = solve(prob1, QNDF())

function test!(du,u,p,t)
     du[1] = 10*u[1]
     du[2] = 10*u[2]
end

u0 = [1.01;1.01]
tspan = (0.0, 1.0)
prob = ODEProblem(test!, u0, tspan)

sol = solve(prob, QNDF())
```
Can anybody help me with this?

src/perform_step/bdf_perform_step.jl
kanav99: this basically means stepsize won't change after 5 iterations, that's basically non-adaptive stepsizing

kanav99: > The ode15s is not allowed to change either the order or the step size until a minimum of
(k + 2) consecutive steps are given using the same order formula and the same step size. If one
step is unsuccessful, the order of the method or the step size are reduced. When the compulsory
(k + 2) successful steps are given, it is possible to change the order and the step size. In this
case, the step sizes which correspond to orders (k − 1) (for k > 1), k and (k + 1) (whenever the
maximum order defined has not been reached) are calculated.

This is from the paper, this doesn't look consistent with your implementation

Issue #97, opened 2017-08-04
Implicit SSP Methods


ChrisRackauckas:
While on the Implicit RK rampage, I plan to pick up methods specific for hyperbolic PDEs.

http://www.sciencedirect.com/science/article/pii/S0168927407001651
http://people.maths.ox.ac.uk/macdonald/imssp.pdf
Maybe even https://arxiv.org/pdf/1510.04303.pdf

SDIRK will follow the setup that was created by the IE/Trapezoid/TRBDF2 methods.

ChrisRackauckas:
Of these the ones which seem like they could be efficient are the higher order SDIRK

http://www.sciencedirect.com/science/article/pii/S0168927407001651

PR #1113, opened 2020-04-20
Added SFSDIRK 4th Order 5,6,7,8 stage methods

kamalojasv181:


kamalojasv181:
https://github.com/SciML/OrdinaryDiffEq.jl/issues/97 
http://www.sciencedirect.com/science/article/pii/S0168927407001651

Issue #1093, opened 2020-03-21
Linearly Implicit SDIRK


ChrisRackauckas:
https://www.sciencedirect.com/science/article/pii/S0168927400000611

kamalojasv181:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/1115 , https://github.com/SciML/OrdinaryDiffEq.jl/pull/1114

PR #1114, opened 2020-04-20
Cfnlirk3

kamalojasv181:


kamalojasv181:
https://github.com/SciML/OrdinaryDiffEq.jl/issues/1093

src/perform_step/kencarp_kvaerno_perform_step.jl
```diff
@@ -412,6 +414,213 @@ end
...
+  ################################## Solve Step 3
+
+  if typeof(integrator.f) <: SplitFunction
+    z₃ .= z₂
+    @.. u = tmp + γ*z₂
+    f2(k2, u, p, t + c2*dt); k2 .*= dt
+    integrator.destats.nf2 += 1
+    @.. tmp = uprev + a31*z₁ + a32*z₂ + ea31*k1 + ea32*k2
+  else
+    # Guess is from Hermite derivative on z₁ and z₂
```
ChrisRackauckas, line 582: how is this from Hermite?

kamalojasv181: Oops!. Removed the comment.

ChrisRackauckas:
Does the paper describe the stage predictors? How did you make the choice here?

kamalojasv181:
Nope. It doesn't .  I just thought initializing it to a close value might help it converge it faster

ChrisRackauckas:
Yes, that's known as a minimum correction predictor

kamalojasv181:
So, I just take the previous stage values as predictors. Is there anything better I could have done?

ChrisRackauckas:
It's good enough for now. We should open an issue afterwards to investigate other stage predictors for this. You can do things like derive interpolation estimates for a `c_i` ever decreases, since now you're not extrapolating. Stuff like that can improve the implementation quite a bit.

kamalojasv181:
Can you please give me a relevant pointer? I really need to catch up before opening an issue.

kamalojasv181:
Anything else besides the predictors?

ChrisRackauckas:
If you haven't read https://www.sciencedirect.com/science/article/abs/pii/0168927495001158 it's a good overview to implementations of stiff ODE solvers.

kamalojasv181:
I think I should open up a new issue. Because I have done the same in bunch of other places as well. Alright?

ChrisRackauckas:
Just in the SDIRK methods. Yeah, we should get a list of the ones you did that might need a second look at the stage predictors.

Looks like the tests are failing here though?

kamalojasv181:
Nope. Algconvergence 2 is failing. I havent added any tests that are a part of Algconvergence 2

ChrisRackauckas:
Looks like compile times may have increased when it updated to Julia v1.4. I restarted the test, but ping me tomorrow if it doesn't pass. I think it can just be broken up or some print statements can be added to stop Travis from timing out.

ChrisRackauckas:
Fixed on master. Rebase.

Issue #1068, opened 2020-03-13
Fer and Cayley Expansion methods for time-dependent ODEs


ChrisRackauckas:
Like Magnus. https://arxiv.org/pdf/0810.5488.pdf

- [ ] Fer4
- [ ] Fer6
- [ ] Cayley4
- [ ] Cayley6

PR #1116, opened 2020-04-22
[WIP] Implement CayleyEuler

SebastianM-C:
I have begun implementing the CayleyEuler method form https://github.com/SciML/OrdinaryDiffEq.jl/issues/1068 (more precisely from [1]). I used the LieEuler PR as a model for this one. The method was developed to solve the isospectral flow equation
```
Y'=[B(Y),Y],
```
where Y is a real symmetric matrix (Sn) and B is a function that maps Sn to real skew-symmetric matrices. It has been show that solving the above equation is equivalent to solving
```
Q'=B(Y)Q,   Q(0)=I
```
with `Y(t)=Q(t)*Y0*transpose(Q(t))`, i.e. we solve a linear problem in with different coordinates and impose the orthogonality of Y by construction.
Because of this reason, I implemented the method as solving the linear problem in auxiliary coordinates together with the orthogonality relation for Y.
In order to easily compare methods, I also taken into account the `SplitODEProblem` formulation, but in that case I only look at the linear operator (`f1`). I think this should be more clearly communicated to the user, as this method does not solve a general `SplitODEProblem`.

The method is very similar to the one quoted in [2], but there a modification to the implicit midpoint method is described. The method in [1] is an explicit method.
I have reproduced the error plot in the paper (fig. 1.1), but I'm having trouble with the order estimation. In [1] it says that this scheme "has the same order of accuracy" as the Forward Euler, but in my case the convergence plot looks quite strange:
image

and the order estimate for `:l2` is 1.56.

Regarding the implementation details
- I wanted to ask if it's ok to put the method together with the linear ones. 
- I am not sure where the implementation if the Cayley transformation should live. It would be helpful if the user could give a custom `cay`, since there are some special cases like the SO(3) group, for which the Rodrigues relation can be used to simplify computations.
- Also regarding the Cayley transformation, there are ways to compute it for SO(N) via the Rodrigues formula [3], but I'm not sure if they are computationally efficient. There are also iterative methods, but I did not investigate this. In the current implementation I did not optimize the Cayley transform.
- I think the structure of the cache could be simplified. (I'm not sure if `W` is relevant)
- I still have to implement the out-of-place version
- In [2] there is a discussion about interpolations, but I'm not familiar with the topic yet and I don't know if it's relevant in this case.
_______________
[1] Iserles, Arieh, Hans Z. Munthe-Kaas, Syvert P. Nørsett, and Antonella Zanna. “Lie-Group Methods.” Acta Numerica 9 (January 2000): 215–365. https://doi.org/10.1017/S0962492900002154.

[2] Calvo, Mari, Arieh Iserles, and Antonella Zanna. “Numerical Solution of Isospectral Flows.” Mathematics of Computation 66, no. 220 (1997): 1461–86. https://doi.org/10.1090/S0025-5718-97-00902-2.

[3] Andrica, Dorin, and Oana Liliana Chender. “Rodrigues Formula for the Cayley Transform of Groups SO(n) and SE(n).” Studia Universitatis Babeş-Bolyai Mathematica 60, no. 1 (2015): 31--38. http://www.cs.ubbcluj.ro/~studia-m/2015-1/2015_1/04-andrica-chender-final.pdf.

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -198,7 +198,7 @@ module OrdinaryDiffEq
          Kvaerno3, KenCarp3, Cash4, Hairer4, Hairer42, SSPSDIRK2, Kvaerno4,
          Kvaerno5, KenCarp4, KenCarp5, ESDIRK54I8L2SA, SFSDIRK4
 
-  export MagnusMidpoint, LinearExponential, MagnusLeapfrog, LieEuler
+  export MagnusMidpoint, LinearExponential, MagnusLeapfrog, LieEuler, CayleyEuler
 
   export Rosenbrock23, Rosenbrock32, RosShamp4, Veldd4, Velds4, GRK4T, GRK4A,
          Ros4LStab, ROS3P, Rodas3, Rodas4, Rodas42, Rodas4P, Rodas5,
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -49,7 +49,7 @@ isimplicit(alg::CompositeAlgorithm) = any(isimplicit.(alg.algs))
 
 isdtchangeable(alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm}) = true
 isdtchangeable(alg::CompositeAlgorithm) = all(isdtchangeable.(alg.algs))
-isdtchangeable(alg::Union{LawsonEuler,NorsettEuler,LieEuler,ETDRK2,ETDRK3,ETDRK4,HochOst4,ETD2}) = false # due to caching
+isdtchangeable(alg::Union{LawsonEuler,NorsettEuler,LieEuler,CayleyEuler,ETDRK2,ETDRK3,ETDRK4,HochOst4,ETD2}) = false # due to caching
 
 ismultistep(alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm}) = false
 ismultistep(alg::CompositeAlgorithm) = any(ismultistep.(alg.algs))
@@ -149,6 +149,7 @@ alg_order(alg::Ralston) = 2
 alg_order(alg::LawsonEuler) = 1
 alg_order(alg::NorsettEuler) = 1
 alg_order(alg::LieEuler) = 1
+alg_order(alg::CayleyEuler) = 2
 alg_order(alg::ETDRK2) = 2
 alg_order(alg::ETDRK3) = 3
 alg_order(alg::ETDRK4) = 4
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -662,6 +662,8 @@ struct LinearExponential <: OrdinaryDiffEqExponentialAlgorithm
 end
 LinearExponential(;krylov=:off, m=10, iop=0) = LinearExponential(krylov, m, iop)
 
+struct CayleyEuler <: OrdinaryDiffEqAlgorithm end  
+
 ################################################################################
 
 # FIRK Methods
--- a/src/caches/linear_caches.jl
+++ b/src/caches/linear_caches.jl
@@ -48,6 +48,31 @@ function alg_cache(alg::LieEuler,u,rate_prototype,uEltypeNoUnits,
   LieEulerConstantCache()
 end
 
+@cache struct CayleyEulerCache{uType,rateType,WType} <: OrdinaryDiffEqMutableCache
+  u::uType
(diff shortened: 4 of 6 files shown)


SebastianM-C:
The code to reproduce fig. 1.1 mentioned above is
```julia
using OrdinaryDiffEq, Test
using LinearAlgebra

function B(y::AbstractMatrix)
    b = similar(y)
    N = size(b, 1)
    for l=1:N
        for k=1:l
            if k < l
                b[k,l] = y[k, l-1] - y[k+1, l]
            else
                b[k,l] = 0
            end
        end
    end
    for l=1:N
        for k=l+1:N
            @inbounds b[k,l] = -b[l,k]
        end
    end

    return b
end

commutator(A,B) = A*B - B*A
function update_func(A, u, p, t)
    A .= B(u)
    return nothing
end

η = diagm([1,2,3,4,5.])
A = DiffEqArrayOperator(Matrix{eltype(η)}(I(size(η,1))), update_func=update_func)
tspan = (0., 20.)

f = SplitFunction(A, (du,u,p,t)->du.=-u*B(u), _func_cache=similar(η))
prob = SplitODEProblem(f, η, tspan)

@time sol_e = solve(prob, Euler(), dt=1/10)
@time sol = solve(prob, CayleyEuler(), dt=1/10)

using Plots

t_err(sol) = [norm(B(sol[i])) for i in eachindex(sol)]
eig_err(sol) = [norm(eigvals(sol[i])-eigvals(η)) for i in eachindex(sol)]
plot(ylims=(10^-16, 10^2), yaxis=:log10);
plot!([eig_err(sol)[2:end] eig_err(sol_e)[2:end]], label=["Eigenvalue error (CFE)" "Eigenvalue (FE)"])
plot!([t_err(sol)[2:end] t_err(sol_e)[2:end]], label=["Toeplitz error (CFE)" "Toeplitz error (FE)"])
```
and it produces
image

ChrisRackauckas:
Cut out that largest dt to improve the convergence estimate. It's an outlier.

ChrisRackauckas:
>In [2] there is a discussion about interpolations, but I'm not familiar with the topic yet and I don't know if it's relevant in this case.

It would be nice to have, but an issue can be opened to add it later.

ChrisRackauckas:
>I am not sure where the implementation if the Cayley transformation should live. It would be helpful if the user could give a custom cay, since there are some special cases like the SO(3) group, for which the Rodrigues relation can be used to simplify computations.

I think cay should be a function associated with the DiffEqOperator interface, which can then allow operators to define functions to override the default

>I'm not sure if W is relevant

Doesn't look like you use it.

SebastianM-C:
> Cut out that largest dt to improve the convergence estimate. It's an outlier.

Was the problem due to the stability region of the method?

Regarding the cache, I saw that a lot of methods have `k` in cache, but some don't use it. Should I also remove it from the `CayleyEuler` cache? I wasn't sure if it's not implied to be there by some other parts of the code-base.

ChrisRackauckas:
>Was the problem due to the stability region of the method?

Probably, it's hard to tell since all proofs are only for dt->0, so large dt is just not covered by the theory.

>Regarding the cache, I saw that a lot of methods have k in cache, but some don't use it. Should I also remove it from the CayleyEuler cache? I wasn't sure if it's not implied to be there by some other parts of the code-base.

Usually it's there for the interpolation.

SebastianM-C:
> Usually it's there for the interpolation.
I, see. In this case I'll leave in case interpolation will be added in the future.

SebastianM-C:
Another filed I'm not using is `uprev2`. Is there a list somewhere with what the different fields in the caches are used for? (Such that I don't ask you for every one of them)

ChrisRackauckas:
That's just for two-step methods, i.e. methods which use u_n and u_n-1 to compute u_n+1.

SebastianM-C:
If I remove it from `alg_cache` should I also remove it from the method signature or just the cache struct?

ChrisRackauckas:
just from the cache struct.

ChrisRackauckas:
Still WIP?

SebastianM-C:
The remaining part would be to use `DiffEqArrayOperator`s in the Cayley transformation like you said, but I'm not sure yet how to implement it. I'll try to look at how it's done for `expv` (though it's not quite the same thing).

Also I wanted to ask if you think this would be something that should leave in a separate package (if so, which). I looked for Cayley transformation related stuff in Julia, but didn't find anything yet.

PR #1119, opened 2020-04-23
Fixed the merge conflicts for CFNLIRK4

kamalojasv181:


kamalojasv181:
Actually I altered the dts as my method was not converging for very low dt. Due to that some other 3rd order algorithm gave an otest value of 3.154 . So I changed atol from 0.15 to 0.17 to be safe.

ChrisRackauckas:
Why was it not converging for low dt?

kamalojasv181:
I think due to the precision

ChrisRackauckas:
why do you think that?

kamalojasv181:
I didn't think anything else was possibly wrong. Let me see it again

kamalojasv181:
I didn't think anything else was possibly wrong. Let me see it again

ChrisRackauckas:
what does the convergence plot look like? You can't just say it's precision without justifying it...

kamalojasv181:
Screenshot from 2020-04-23 06-59-17
Here is the convergence plot. With this abrupt change for lowest, I couldn't think of anything else that might have gone wrong and all other tests passing.

kamalojasv181:
I think that I did a sloppy job here. Sorry about that. Will check and report it.

PR #1120, opened 2020-04-25
[WIP]Implemented two step explicit RK method

kamalojasv181:
https://arxiv.org/pdf/1106.3626.pdf . It has problem in convergence.

Screenshot from 2020-04-25 05-37-29

It can be seen that convergence plot is a straight line but otest is around 1 where it should be 5. 

Regarding the starting procedure for this method, the author of the paper suggests :

Screenshot from 2020-04-25 05-41-46

Therefore I chose to go with the SSPRK54 for the first h/4 part of the step 1.

Kindly pour in your thoughts @ChrisRackauckas , @ranocha


kamalojasv181:
Please ignore the changed the tolerence commit message. It was meant for some other branch so I wrote the wrong commit message for something else that I did in this branch

ranocha:
I didn't have enough time to dive deeper into that issue yet. It seems that you get first order convergence in that plot, don't you? Which problem did you use to get that plot?
If you start with something simple without explicit dependence on time, e.g. u'(t) = u(t), your calculation of the cᵢ shouldn't matter.
Another idea to start the debugging is to use exact starting values to just test the TSRK coefficients. If that works, you know that there is an issue with the starting procedure.

PR #1121, opened 2020-04-25
Don't differentiate the step controller

ChrisRackauckas:
Odd little issue. The PI-controller in the stepper is not actually differentiable, so we need to keep the ties in the differentiation in the dts and t, but when we hit the qold gain on the integral control, we want to drop the AD there since we want to just have the chosen dt there. This fixes a fairly obscured bug that was found in one case of Pumas

>okay, tlag is fixed, and I also fixed the nan-safeness I think, and that also could fix the estimation robustness. Essentially, when differentiating something that also involves differentiating a callback, if a step rejected twice in a row where the error amount was less than qmax, and some of the states being differentiated have a zero derivative w.r.t. one of the parameters. I.e., this is the edge case-iest of edge cases, so it's hard to tell if you ever hit it in the optimization, but now you won't.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEq"
 uuid = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "5.34.1"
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Requires SciML/DiffEqBase.jl#492 to fix that test case.

PR #1123, opened 2020-04-27
[WIP]Better interpolation

kamalojasv181:


src/tableaus/sdirk_tableaus.jl
```diff
@@ -913,6 +917,12 @@ function CFNLIRK3Tableau(T,T2)
   a43 = convert(T,-0.64436317068447276)
   c3 = (1+γ)/2
   c2 = γ
+  θ = c3/c2
```
ChrisRackauckas, line 920: why these choices?

kamalojasv181:
I initially did 

IMG_20200427_202828

But this was not working, then I did, what kencarp3 did

Initially the predictions were this without interpolating:

Screenshot from 2020-04-27 09-11-26

When I did this thing:
Screenshot from 2020-04-27 09-28-26

Since it was working but I was unsure why that's why I decided to push it. I was about to ask about it in the dm.

PR #1125, opened 2020-04-30
Failed step handling in NDF

utkarsh530:
Hi,

Previously failed step in NDF was not working properly. There were some possible bugs in the code like simple returning the function without the value which was happening in the `stepsize_and_order` function and simply returning the `perform_step` function was not calculating at that `t` again.
I hope to convert in into controller because that would be a neat implementation. (Trying to find what is the current problem in the step-sizing, I have got some leads and will implement them iteratively). 
@ChrisRackauckas @kanav99 could you please have a look?

### DIFF
--- a/src/bdf_utils.jl
+++ b/src/bdf_utils.jl
@@ -139,7 +139,7 @@ function stepsize_and_order!(cache, est, estₖ₋₁, estₖ₊₁, h, k)
     if cache.c >= 1  # postfail
       cache.h = h/2
       cache.order = k
-      return
+      return false
     end
     if 1.2 < z <= 10
       hₖ = F * h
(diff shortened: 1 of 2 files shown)


src/bdf_utils.jl
```diff
@@ -139,7 +139,7 @@ function stepsize_and_order!(cache, est, estₖ₋₁, estₖ₊₁, h, k)
     if cache.c >= 1  # postfail
       cache.h = h/2
       cache.order = k
-      return
+      return false
```
ChrisRackauckas, line 142: that's a big difference haha

ChrisRackauckas:
Definitely an improvement. Nice.

>Trying to find what is the current problem in the step-sizing, I have got some leads and will implement them iteratively

What are some of your current leads?

utkarsh530:
I tried to read the `ode15s.m` file, these were some of my observations:
1. The method keeps the `order` and `dt` same for at least `k+2` (k= order) iterations. (Which is definitely not happening).
2. The step-size controller mentioned [here](https://www.sciencedirect.com/science/article/pii/S1877050914002683) might be incomplete. I am verifying it with directly with the `ode15s.m` which is quite a serious task. I am facing some difficulties in isolating the code-blocks and checking whether currently it is implemented or not.
2. The `init_dt` calculation is quite off from our codebase. I guess that would be the last thing to fix.
@ChrisRackauckas

PR #1133, opened 2020-05-13
always color

ChrisRackauckas:
With SparseDiffTools v1.8, the calculus changed. Now in the README example, matrix coloring takes... 

```julia
0.087866 seconds (30.74 k allocations: 3.610 MiB)
```

but it decreases the solve time of even the Brusselator by 7 seconds below. So now that https://github.com/JuliaDiff/SparseDiffTools.jl/pull/107 decreased coloring time by 1000x, it's a no-brainer to just always color if we see a sparse Jacobian.

```julia
using DifferentialEquations
const N = 32
const xyd_brusselator = range(0,stop=1,length=N)
brusselator_f(x, y, t) = (((x-0.3)^2 + (y-0.6)^2) <= 0.1^2) * (t >= 1.1) * 5.
limit(a, N) = a == N+1 ? 1 : a == 0 ? N : a
function brusselator_2d_loop(du, u, p, t)
  A, B, alpha, dx = p
  alpha = alpha/dx^2
  @inbounds for I in CartesianIndices((N, N))
    i, j = Tuple(I)
    x, y = xyd_brusselator[I[1]], xyd_brusselator[I[2]]
    ip1, im1, jp1, jm1 = limit(i+1, N), limit(i-1, N), limit(j+1, N), limit(j-1, N)
    du[i,j,1] = alpha*(u[im1,j,1] + u[ip1,j,1] + u[i,jp1,1] + u[i,jm1,1] - 4u[i,j,1]) +
                B + u[i,j,1]^2*u[i,j,2] - (A + 1)*u[i,j,1] + brusselator_f(x, y, t)
    du[i,j,2] = alpha*(u[im1,j,2] + u[ip1,j,2] + u[i,jp1,2] + u[i,jm1,2] - 4u[i,j,2]) +
                A*u[i,j,1] - u[i,j,1]^2*u[i,j,2]
    end
end
p = (3.4, 1., 10., step(xyd_brusselator))

using SparsityDetection, SparseArrays
input = rand(32,32,2)
output = similar(input)
sparsity_pattern = jacobian_sparsity(brusselator_2d_loop,output,input,p,0.0)
jac_sparsity = Float64.(sparse(sparsity_pattern))
f = ODEFunction(brusselator_2d_loop;jac_prototype=jac_sparsity)

function init_brusselator_2d(xyd)
  N = length(xyd)
  u = zeros(N, N, 2)
  for I in CartesianIndices((N, N))
    x = xyd[I[1]]
    y = xyd[I[2]]
    u[I,1] = 22*(y*(1-y))^(3/2)
    u[I,2] = 27*(x*(1-x))^(3/2)
  end
  u
end
u0 = init_brusselator_2d(xyd_brusselator)
prob_ode_brusselator_2d = ODEProblem(brusselator_2d_loop,
                                     u0,(0.,11.5),p)

prob_ode_brusselator_2d_sparse = ODEProblem(f,
                                     u0,(0.,11.5),p)

@btime solve(prob_ode_brusselator_2d_sparse,save_everystep=false)

using SparseDiffTools, BenchmarkTools
@time colorvec = matrix_colors(jac_sparsity)

f = ODEFunction(brusselator_2d_loop;jac_prototype=jac_sparsity,
                                    colorvec=colorvec)
prob_ode_brusselator_2d_sparse = ODEProblem(f,
                                     init_brusselator_2d(xyd_brusselator),
                                     (0.,11.5),p)
@btime solve(prob_ode_brusselator_2d_sparse,save_everystep=false)
```

```julia
3.230 s (18691 allocations: 881.07 MiB)
10.766 s (19042 allocations: 881.07 MiB)
```


ChrisRackauckas:
Note we should remove color vectors from the tutorial after doing this because (a) it confuses people and (b) since it's fast enough, relying on auto-coloring will make things like supporting fast sparse adjoints easier.

src/derivative_wrappers.jl
```diff
@@ -42,7 +42,8 @@ jacobian_autodiff(f, x, odefun, alg) = (ForwardDiff.derivative(f,x),1, alg)
 function jacobian_autodiff(f, x::AbstractArray, odefun, alg)
   sparsity = odefun.sparsity
   jac_prototype = odefun.jac_prototype
-  colorvec = DiffEqBase.has_colorvec(odefun) ? odefun.colorvec : (isnothing(sparsity) ? 1:length(x) : matrix_colors(sparsity))
+  colorvec = DiffEqBase.has_colorvec(odefun) ? odefun.colorvec :
+                  (isnothing(sparsity) ? (1:length(x)) : matrix_colors(sparsity))
```
devmotion, line 46: Maybe
```suggestion
  colorvec = DiffEqBase.has_colorvec(odefun) ? odefun.colorvec :
                  (sparsity === nothing ? (1:length(x)) : matrix_colors(sparsity))
```
just to make sure that it's easy for the compiler (although I doubt that it makes a difference here)?

devmotion:
I guess one could also think about putting the logic for retrieving `sparsity` (?) and `colorvec` in a separate function, i.e., define
```julia
function sparsity_colorvec(f, x)
  sparsity = f.sparsity
  colorvec = DiffEqBase.has_colorvec(f) ? f.colorvec : (sparsity === nothing ? (1:length(x)) : matrix_colors(sparsity))
  return sparsity, colorvec
```
That could help to reduce the amount of code duplication.

src/derivative_wrappers.jl
ChrisRackauckas: doesn't `isnothing` do `===` ?

devmotion: The result is the same, but it is easier for the compiler to reason about type inference with `===` than with the function call `isnothing`, see, e.g., https://github.com/JuliaLang/julia/issues/35585#issuecomment-619053871. I guess it's much more important though in, e.g., iteration.

test/interface/sparsediff_tests.jl
```diff
@@ -60,39 +60,16 @@ for f in [f_oop, f_ip]
         for (i,prob) in enumerate(map(f->ODEProblem(f,u0,tspan),
                         [ODEFunction(f,colorvec=colors,jac_prototype=jac_sp),
                          ODEFunction(f,jac_prototype=jac_sp),
-                         ODEFunction(f,colorvec=colors)
+                         ODEFunction(f,colorvec=colors,sparsity=jac_sp)
                          ]))
-          # TODO: these broken test-cases need to be investigated.
```
ChrisRackauckas, line 65: all of the sparse diff tests are fixed now.

PR #1138, opened 2020-05-18
a few small optimizations

ChrisRackauckas:


YingboMa:
The test failure looks unrelated.

ChrisRackauckas:
Downstream failure and it was fixed.

PR #1142, opened 2020-05-20
2nd order mass matrix check fix

ChrisRackauckas:


src/solve.jl
```diff
@@ -79,7 +79,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
   elseif !(typeof(prob)<:DiscreteProblem) &&
          !(typeof(prob)<:DiffEqBase.AbstractDAEProblem) &&
          !is_mass_matrix_alg(alg) &&
-         prob.f.mass_matrix != I
+         (prob.f.mass_matrix !== I && prob.f.mass_matrix !== (I,I))
```
YingboMa, line 82: This condition looks like it will always be false.

ChrisRackauckas: oh yeah...

PR #1147, opened 2020-05-27
The end of Wfact: Fully inferred Wfact oop

ChrisRackauckas:
This is just a cheap way to get SOL Wfact timings, and...

```julia
# DAE
using BenchmarkTools, StaticArrays, OrdinaryDiffEq, ModelingToolkit, LinearAlgebra
@parameters t k₁ k₂ k₃
@variables y₁(t) y₂(t) y₃(t)
@derivatives D'~t
eqs = [D(y₁) ~ -k₁*y₁+k₃*y₂*y₃,
       0     ~  y₁ + y₂ + y₃ - 1,
       D(y₂) ~  k₁*y₁-k₂*y₂^2-k₃*y₂*y₃]
sys = ODESystem(eqs)
u0 = SA[y₁ => 1.0,
      y₂ => 0.0,
      y₃ => 0.0]
p  = SA[k₁ => 0.04,
      k₂ => 3e3,
      k₃ => 1e3]
tspan = (0.0,100000.0)
prob1 = ODEProblem{false}(sys,u0,tspan,p)
prob2 = ODEProblem{false}(sys,u0,tspan,p,jac=true)
prob3 = ODEProblem{false}(sys,u0,tspan,p,Wfact=true,Wfact_t=true)

sol1 = @btime solve($prob1, Rodas5()); # 11.537 ms (25593 allocations: 1.02 MiB)
sol2 = @btime solve($prob2, Rodas5()); # 53.101 μs (384 allocations: 38.91 KiB)
sol3 = @btime solve($prob3, Rodas5()); # 61.000 μs (200 allocations: 20.34 KiB)
BLAS.set_num_threads(4)
sol1 = @btime solve($prob1, Rodas5()); # 12.090 ms (25593 allocations: 1.02 MiB)
sol2 = @btime solve($prob2, Rodas5()); # 64.601 μs (384 allocations: 38.91 KiB)
sol3 = @btime solve($prob3, Rodas5()); # 67.799 μs (200 allocations: 20.34 KiB)
```

It turns out that StaticArrays LU is just faster these days (it didn't have an overload when we first did Wfact!). This is even with a pretty big advantage since here Wfact produces a StaticArrays.LU while calc_W normally produces a WOperator and factorizes each time with `\`, so Wfact-reuse is only done with Wfact and it still loses.

In the end, since it's only at StaticArray scales that Wfact is really possible, this is proof for the removal of Wfact.

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -477,30 +477,19 @@ end
   islin, isode = islinearfunction(integrator)
   !isdae && update_coefficients!(mass_matrix,uprev,p,t)
 
-  if islin
-    J = isode ? f.f : f.f1.f # unwrap the Jacobian accordingly
-    W = WOperator{false}(mass_matrix, dtgamma, J, uprev; transform=W_transform)
-  elseif DiffEqBase.has_jac(f)
-    J = f.jac(uprev, p, t)
-    if !isa(J, DiffEqBase.AbstractDiffEqLinearOperator)
-      J = DiffEqArrayOperator(J)
+  # handle Wfact
+  if W_transform && DiffEqBase.has_Wfact_t(f)
+    W = f.Wfact_t(uprev, p, dtgamma, t)
+    is_compos && (integrator.eigen_est = constvalue(opnorm(LowerTriangular(W), Inf)) + inv(dtgamma)) # TODO: better estimate
+    return W
+  elseif !W_transform && DiffEqBase.has_Wfact(f)
+    W = f.Wfact(uprev, p, dtgamma, t)
+    if is_compos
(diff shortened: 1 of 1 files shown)


YingboMa:
Do we want to remove `Wfact` and `Wfact_t` in OrdinaryDiffEq, too?

ChrisRackauckas:
I think yes. If it's not going to be faster anymore (i.e., for systems that are small enough to use it, it's beat out by StaticArray LU), then we might as well remove it. It does kind of serve a second purpose of letting people do approximate matrix factorization and such, but that hasn't really be used (or documented), and is better served by the linear solver refactor. So we should just remove this functionality and then that makes the linear solver refactor easier, and we should just focus on that.

PR #1153, opened 2020-05-30
Single GPU kernel call for low storage methods

kanav99:
Doesn't break anything. Need to profile on an actual GPU now

### DIFF
--- a/src/wrappers.jl
+++ b/src/wrappers.jl
@@ -19,11 +19,12 @@ end
 
 ArrayFuse(visible::AT, hidden::AT, p) where {AT} = ArrayFuse{AT, eltype(visible), typeof(p)}(visible, hidden, p)
 
-@inline function Base.setindex!(af::ArrayFuse, value, index)
-	af.visible[index] = af.p[1] * af.visible[index] + af.p[2] * value
-	af.hidden[index] = af.hidden[index] + af.p[3] * af.visible[index]
+@inline function Base.copyto!(af::OrdinaryDiffEq.ArrayFuse{AT, T, P}, src::Base.Broadcast.Broadcasted) where {AT, T, P}
+	@. af.visible = af.p[1] * af.visible + af.p[2] * src
+	@. af.hidden = af.hidden + af.p[3] * af.visible
 end
 
+# not recommended but good to have
 @inline function Base.getindex(af::ArrayFuse, index)
 	return af.visible[index]
 end
--- a/src/wrappers.jl
+++ b/src/wrappers.jl
@@ -31,7 +31,7 @@ end
 
 @inline function Base.setindex!(af::ArrayFuse, value, index)
 	af.visible[index] = af.p[1] * af.visible[index] + af.p[2] * value
-	af.hidden[index] = af.hidden[index] + af.p[3] * af.visible[index]
+	af.hidden[index] =  muladd(af.p[3], af.visible[index], af.hidden[index])
 end
 
 @inline Base.size(af::ArrayFuse) = length(af.visible)
--- a/src/wrappers.jl
+++ b/src/wrappers.jl
@@ -32,4 +32,4 @@ end
 @inline Base.axes(af::ArrayFuse) = axes(af.visible)
 
 Adapt.adapt_structure(to, af::ArrayFuse{AT, T, P}) where {AT, T, P} = 
-	ArrayFuse(adapt(to, af.visible), adpat(to, af.hidden), af.p)
+	ArrayFuse(adapt(to, af.visible), adapt(to, af.hidden), af.p)
--- a/src/wrappers.jl
+++ b/src/wrappers.jl
@@ -29,6 +29,11 @@ end
 	return af.visible[index]
 end
 
+@inline function Base.setindex!(af::ArrayFuse, value, index)
+	af.visible[index] = af.p[1] * af.visible[index] + af.p[2] * value
+	af.hidden[index] = af.hidden[index] + af.p[3] * af.visible[index]
+end
+
 @inline Base.size(af::ArrayFuse) = length(af.visible)
 @inline Base.axes(af::ArrayFuse) = axes(af.visible)
 
--- a/src/wrappers.jl
+++ b/src/wrappers.jl
@@ -1,0 +1,35 @@
+"""
+	ArrayFuse{AT, T, P} <: AbstractArray{T, 1}
+
+GPU Friendly type to wrap around two arrays - `visible` and `hidden`, for which when we `setindex!` some value `v` at index `i`
+we get
+
+```
+visible[i] = p[1] * visible[i] + p[2] * v
+hidden[i] = hidden[i] + p[3] * visible[i]
+```
+
+where p is a parameter tuple of size 3.
+"""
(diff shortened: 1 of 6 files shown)


src/wrappers.jl
```diff
@@ -0,0 +1,35 @@
...
+
+@inline function Base.getindex(af::ArrayFuse, index)
+	return af.visible[index]
+end
+
+@inline Base.size(af::ArrayFuse) = length(af.visible)
+@inline Base.axes(af::ArrayFuse) = axes(af.visible)
+
+Adapt.adapt_structure(to, af::ArrayFuse{AT, T, P}) where {AT, T, P} = 
+	ArrayFuse(adapt(to, af.visible), adpat(to, af.hidden), af.p)
```
kanav99, line 35:
```suggestion
	ArrayFuse(adapt(to, af.visible), adapt(to, af.hidden), af.p)
```
Would change after tests pass

kanav99:
Lets wait for tests to pass, I have tested 3 methods locally. Also, I would add a GPU test too

ChrisRackauckas:
Yes, please add a GPU test for this. Currently GPU CI is down but we should have it ready.

kanav99:
I am testing on GPU and it fails as of now due to the custom broadcast

ChrisRackauckas:
Open an issue and let's cover that in the next step.

kanav99:
Good results 
```
CPU Times for ORK256(true)
  7.786772 seconds (4.07 k allocations: 23.074 MiB)
GPU Times for ORK256(true)
  3.049972 seconds (1.06 M allocations: 47.637 MiB)
CPU Times for CarpenterKennedy2N54(true)
  7.757759 seconds (4.08 k allocations: 23.075 MiB)
GPU Times for CarpenterKennedy2N54(true)
  3.051859 seconds (1.06 M allocations: 47.637 MiB)
CPU Times for HSLDDRK64(true)
  9.117048 seconds (5.07 k allocations: 23.120 MiB)
GPU Times for HSLDDRK64(true)
  3.570850 seconds (1.20 M allocations: 54.450 MiB)
CPU Times for DGLDDRK73_C(true)
 10.692754 seconds (6.08 k allocations: 23.166 MiB)
GPU Times for DGLDDRK73_C(true)
  4.088580 seconds (1.34 M allocations: 61.261 MiB)```

**Comment by kanav99** [comment]
Oh I just forgot about the original issue, let me fix

**Comment by kanav99** [comment]
This is tough..

**Comment by simonbyrne** [comment]
I'm not sure this is fusing kernels. I just ran
```
using OrdinaryDiffEq, CuArrays

a = cu(ones(10))
b = cu(ones(10))

function f(du,u,p,t)
  @. du = u
end

function foo(a, b, n)
    for i = 1:n
        f(OrdinaryDiffEq.ArrayFuse(a, b, (0.1,0.2,0.3)), b, nothing, 0.01)
    end
end

foo(a,b,1000)
```
under `nvprof` and it said it launched 2000 kernels (so 2 per loop).

**Comment by simonbyrne** [comment]
I should say however that it does seem to work out of the box with KernelAbstractions. The following ran with exactly 1000 kernel launches!

It might be worth adding some KernelAbstractions code as test cases.
```
using OrdinaryDiffEq, CuArrays, KernelAbstractions

a = cu(ones(10))
b = cu(ones(10))

@kernel function kf(du,u)
    i = @index(Global, Linear)
    @inbounds begin
        du[i] = u[i]
    end
end


function f(du,u,p,t)
    groupsize = 256

    event = kf(CUDA(), groupsize)(
        du, u;
        ndrange = length(u),
    )
    wait(event)
end

function foo(a, b, n)
    for i = 1:n
        f(OrdinaryDiffEq.ArrayFuse(a, b, (0.1,0.2,0.3)), b, nothing, 0.01)
    end
end

foo(a,b,1000)
```

**Comment by ChrisRackauckas** [comment]
Interesting! That is worth testing and documenting.

**Comment by kanav99** [comment]
Yes, the PR was not complete, I was working with KernelAbstractions, but
your code snippet looks completely perfect, thanks!

**Comment by kanav99** [comment]
I am actually trying to make that internal, so that you dont have to write GPU code inside the function. So this serves as a great basepoint

**Comment by simonbyrne** [comment]
The downside is that it does assume that you assign to each index in `du` exactly once: unfortunately, this isn't the case in the ClimateMachine code, as we do 2 passes: one to compute the volume integrals:
https://github.com/CliMA/ClimateMachine.jl/blob/766ad0c0354742b08baab5a9d8aef10a511d86c6/src/Numerics/DGMethods/DGModel_kernels.jl#L247
and one for the face integrals:
https://github.com/CliMA/ClimateMachine.jl/blob/766ad0c0354742b08baab5a9d8aef10a511d86c6/src/Numerics/DGMethods/DGModel_kernels.jl#L753
(they are split up to allow for overlapping computation and communication).

So in our case we probably would need a slightly different interface, either:
- evaluates `f` into an additional work array, and then does the update of `hidden` and `visible`, but can fuse them into one additional kernel call, or
- some clever mechanism that lets us do
```
af.visible[index] *= af.p[1]
af.hidden[index] = af.p[3] * af.visible[index]
```
to all elements once, and then overload `setindex!` to
```
af.visible[index] += af.p[2] * value
af.hidden[index] += af.p[3] * af.p[2] * value
```
and combines them into the one kernel. We could easily add extra machinery into the ClimateMachine kernels to do this (e.g. by dispatching on the type of `du`)

**Comment by simonbyrne** [comment]
More specifically, in our case the kernels are applied to every index, and then a subset of indices (the element faces). So we would want two functions:
```
function setindex_once!(af::ArrayFuse, value, indices...) 
	af.visible[indices...] = af.p[1] * af.visible[indices...] + af.p[2] * value
	af.hidden[indices...] = af.hidden[indices...] + af.p[3] * af.visible[indices...]
end
function setindex_additional!(af::ArrayFuse, value, indices...) 
        increment = af.p[2] * value
	af.visible[indices...] += increment
	af.hidden[indices...] += af.p[3] * increment
end
```

**Comment by ChrisRackauckas** [comment]
We can have a DiffEqFunction for the increment=true form.

**Comment by simonbyrne** [comment]
Just so it is clear what is going on, we are doing:
```
du .+= f(u,p,t) 
```
in our "tendency" kernels, and
```
u .+= rkb .* dt .* du
du .+= rka .* du
```
in the time step kernel.

**Comment by simonbyrne** [comment]
Actually, it occurs to me that we could write it as:
```
du .= rka .* du .+ f(u,p,t) 
```
in our tendency kernels, then the timestep kernel would just be
```
u .+= rkb .* dt .* du
```

So if we could expose a BLAS-style scale plus add interface instead of just an increment, we could simplify this considerably.

kanav99:
@simonbyrne In achieving `du .= rka .* du .+ f(u,p,t)`, we would have to make two kernel calls anyway because we would do `du .= rka .* du` first then `du += f(u,p,t)`. How do you think we can fuse these two kernel calls? We can make it like you suggested in setindex_once! but that would make us save a state of whether an index was called first time or not so we would need to have an array of bools, that's not good I guess

simonbyrne:
@kanav99 I've opened another issue for this: https://github.com/SciML/DifferentialEquations.jl/issues/615

PR #1154, opened 2020-05-30
Justify LinearInterpolation and ConstantInterpolation from DiffEqBase

Vaibhavdixit02:


ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/1155 is the right solution

PR #1156, opened 2020-05-30
Adaptive Radau

d-netto:


test/algconvergence/ode_firk_tests.jl
```diff
@@ -56,10 +56,6 @@ for iip in (true, false)
     vanstiff = ODEProblem{false}((u,p,t)->van(u,p,t), [0;sqrt(3)], (0.0,1.0), 1e6)
   end
   sol = solve(vanstiff, RadauIIA3())
-  if iip
-    @test sol.destats.naccept + sol.destats.nreject > sol.destats.njacs # J reuse
-    @test sol.destats.njacs < sol.destats.nw # W reuse
-  end
```
YingboMa: Oh yes, leave the Jacobian and W reuse for now. I have to reform it anyway.

src/perform_step/firk_perform_step.jl
```diff
@@ -305,181 +323,38 @@ end
...
   end
-  cache.ηold = η
-  cache.iter = iter
 
   @. u = uprev + z2
-  f(fsallast, u, p, t+dt)
-  return
-end
-
-@muladd function perform_step!(integrator, cache::RadauIIA5ConstantCache, repeat_step=false)
```
ChrisRackauckas, line 316: why is this deleted?

d-netto: deleted it by accident, just fixed in the the next commits

d-netto:
@YingboMa: the error "LoadError: type RadauIIA3Cache has no field nlsolver" persisted even after adding RadauIIA3 to the controller.jl file. Any idea of what could be causing it?

YingboMa:
Always copy and paste the entire stacktrace. I cannot help if you just tell me the error message.

d-netto:
> Always copy and paste the entire stacktrace. I cannot help if you just tell me the error message.

d-netto:
Ok. The stack trace has a few thousand lines of code, unfortunately.

d-netto:
@ChrisRackauckas, @YingboMa: stack trace above.

ChrisRackauckas:
Set `dt=0.1` for now. It has to do with how it's trying to grab a temporary.

PR #1160, opened 2020-06-02
Fix Trapezoid mass matrix handling

YingboMa:
This PR fixes Trapezoid's Newton iteration per https://github.com/SciML/DiffEqDevMaterials/blob/master/newton/output/main.pdf

I will open separate PRs to fix functional iteration and other multistep methods.

### DIFF
--- a/src/nlsolve/newton.jl
+++ b/src/nlsolve/newton.jl
@@ -49,7 +49,7 @@ Equations II, Springer Series in Computational Mathematics. ISBN
 """
 @muladd function compute_step!(nlsolver::NLSolver{<:NLNewton,false}, integrator)
   @unpack uprev,t,p,dt,opts = integrator
-  @unpack alg,z,tmp,γ,α,cache = nlsolver
+  @unpack z,tmp,γ,α,cache = nlsolver
   @unpack tstep,W,invγdt = cache
 
   f = nlsolve_f(integrator)
--- a/src/nlsolve/newton.jl
+++ b/src/nlsolve/newton.jl
@@ -65,10 +65,10 @@ Equations II, Springer Series in Computational Mathematics. ISBN
       ustep = z
       # tmp = outertmp ./ hγ
       if mass_matrix === I
-        ztmp = tmp .+ f(z, p, tstep) - α * z .* invγdt
+        ztmp = tmp .+ f(z, p, tstep) .- (α * invγdt) .* z
       else
         update_coefficients!(mass_matrix, ustep, p, tstep)
-        ztmp = tmp .+ f(z, p, tstep) - α * mass_matrix * z .* invγdt
+        ztmp = tmp .+ f(z, p, tstep) .- (mass_matrix * z) .* (α * invγdt)
       end
     else
       ustep = @. tmp + γ * z
@@ -127,11 +127,11 @@ end
       ustep = z
       f(k, z, p, tstep)
       if mass_matrix === I
-        @.. ztmp = tmp .+ k - α * z .* invγdt
+        @.. ztmp = tmp + k - (α * invγdt) * z
       else
         update_coefficients!(mass_matrix, ustep, p, tstep)
         mul!(vec(ztmp), mass_matrix, vec(z))
-        @.. ztmp = tmp .+ k - α * ztmp .* invγdt
+        @.. ztmp = tmp + k - (α * invγdt) * ztmp
       end
     else
       @.. ustep = tmp + γ * z
--- a/src/nlsolve/newton.jl
+++ b/src/nlsolve/newton.jl
@@ -31,24 +31,9 @@ end
 Compute next iterate of numerically stable modified Newton iteration
 that is specialized for implicit methods.
 
-It solves
-```math
-z = dt⋅f(tmp + γ⋅z, p, t + c⋅dt)
-```
-by iterating
-```math
-(I + (dt⋅γ)J) Δᵏ = dt*f(tmp + γ⋅zᵏ, p, t + c⋅dt) - zᵏ
-zᵏ⁺¹ = g(zᵏ) = zᵏ - Δᵏ
-```
-or, by utilizing a transformation,
-```math
-W Δᵏ = f(tmp + γ⋅zᵏ, p, t + c⋅dt)/γ - zᵏ/(dt⋅γ)
-zᵏ⁺¹ = g(zᵏ) = zᵏ - Δᵏ/(dt⋅γ)
-```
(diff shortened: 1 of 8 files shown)


src/nlsolve/newton.jl
```diff
@@ -64,27 +49,35 @@ Equations II, Springer Series in Computational Mathematics. ISBN
...
   if isdae
     ustep = @. uprev + z
     dustep = @. (tmp + α * z) * invγdt
     ztmp = f(dustep, ustep, p, t)
   else
-    ustep = @. tmp + γ * z
-    if mass_matrix === I
-      ztmp = (dt .* f(ustep, p, tstep) .- z) .* invγdt
+    mass_matrix = integrator.f.mass_matrix
+    if isodemultistep(unwrap_alg(integrator, true))
```
ChrisRackauckas, line 64: this will cause problems in SDEs.

ChrisRackauckas:
Why does this only fix Trapezoid?

YingboMa:
Other multistep methods need little ticks to get it right, so I have to read original papers to check that. That takes more time, so I think, for the time being, I can fix Trapezoid first and do the rest of mechanical work later.

src/nlsolve/newton.jl
YingboMa: How can I make it StochasticDiffEq friendly?

YingboMa: I think StochasticDiffEq has `unwrap_alg` overload, too.

ChrisRackauckas: Oh nevermind, https://github.com/SciML/OrdinaryDiffEq.jl/pull/1160/files#diff-aac6396f64983b0f4a67297163d3155aR572 that will catch it. But I still think there should be two caches so each algorithm picks which one to use. That would make general linear methods possible.

src/nlsolve/newton.jl
```diff
@@ -70,21 +55,29 @@ Equations II, Springer Series in Computational Mathematics. ISBN
...
+      ustep = z
+      # tmp = outertmp ./ hγ
+      if mass_matrix === I
+        ztmp = tmp .+ f(z, p, tstep) - α * z .* invγdt
```
devmotion, line 68:
```suggestion
        ztmp = tmp .+ f(z, p, tstep) .- (α * invγdt) .* z 
```

src/nlsolve/newton.jl
```diff
@@ -70,21 +55,29 @@ Equations II, Springer Series in Computational Mathematics. ISBN
...
+      else
+        update_coefficients!(mass_matrix, ustep, p, tstep)
+        ztmp = tmp .+ f(z, p, tstep) - α * mass_matrix * z .* invγdt
```
devmotion, line 71:
```suggestion
        ztmp = tmp .+ f(z, p, tstep) .- mass_matrix * z .* (α * invγdt)
```

src/nlsolve/newton.jl
```diff
@@ -119,31 +112,37 @@ end
...
     b = vec(k)
   else
-    if mass_matrix === I
-      @.. ztmp = (dt * k - z) * invγdt
+    mass_matrix = integrator.f.mass_matrix
+    if isodemultistep(unwrap_alg(integrator, true))
+      ustep = z
+      f(k, z, p, tstep)
+      if mass_matrix === I
+        @.. ztmp = tmp .+ k - α * z .* invγdt
```
devmotion, line 130: I guess since we use `@..` the additional broadcasts are not needed?
```suggestion
        @.. ztmp = tmp + k - α * z * invγdt
```

src/nlsolve/newton.jl
```diff
@@ -119,31 +112,37 @@ end
...
+      else
+        update_coefficients!(mass_matrix, ustep, p, tstep)
+        mul!(vec(ztmp), mass_matrix, vec(z))
+        @.. ztmp = tmp .+ k - α * ztmp .* invγdt
```
devmotion, line 134:
```suggestion
        @.. ztmp = tmp + k - α * ztmp * invγdt
```

src/nlsolve/newton.jl
devmotion: Maybe one could precompute `α * invγdt` as in the out-of-place version.

src/nlsolve/newton.jl
devmotion: Same here, maybe precompute `α * invγdt`.

src/perform_step/sdirk_perform_step.jl
```diff
@@ -199,14 +199,16 @@ end
...
+  # initial guess: constant extrapolation
+  nlsolver.z = uprev
 
-  nlsolver.tmp = uprev + γdt*integrator.fsalfirst
-  z = nlsolve!(nlsolver, integrator, cache, repeat_step)
+  if f.mass_matrix === I
+    nlsolver.tmp = uprev * inv(γdt) + integrator.fsalfirst
+  else
+    nlsolver.tmp = f.mass_matrix * uprev * inv(γdt) + integrator.fsalfirst
+  end
```
devmotion, line 209: Maybe use broadcasting?

src/nlsolve/nlsolve.jl
```diff
@@ -3,15 +3,17 @@
...
 """
 function nlsolve!(nlsolver::AbstractNLSolver, integrator, cache=nothing, repeat_step=false)
   @label REDO
   if isnewton(nlsolver)
-    cache isa Nothing && throw(ArgumentError("cache is not passed to `nlsolve!` when using NLNewton"))
-    update_W!(nlsolver, integrator, cache, nlsolver.γ*integrator.dt, repeat_step)
+    cache === nothing && throw(ArgumentError("cache is not passed to `nlsolve!` when using NLNewton"))
+    γW = nlsolver.γ*integrator.dt
+    isodemultistep(unwrap_alg(integrator, true)) && (γW = γW / nlsolver.α)
+    update_W!(nlsolver, integrator, cache, γW, repeat_step)
```
devmotion, line 16: Maybe one could even use a proper `if` - `else` statement to make it more obvious what's happening here. It feels like the difference between both methods is a bit hidden, and maybe an explicit
```suggestion
    if isodemultistep(unwrap_alg(integrator, true))
        update_W!(nlsolver, integrator, cache, γW / nlsolver.α, repeat_step)
    else
        update_W!(nlsolver, integrator, cache, γW, repeat_step)
    end
```
would make it clearer.

YingboMa:
The test failures are unrelated, and the CI timed out for Downstream.

PR #1161, opened 2020-06-03
Update ABDF2 to the new nlsolver

YingboMa:
POLLU WPD:

ABDF2_PR

Note that the `CVODE` here is `CVODE_BDF(max_order=2)` to make it a "fair" fight.

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -568,14 +568,3 @@ isesdirk(alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm}) = false
 is_mass_matrix_alg(alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm}) = false
 is_mass_matrix_alg(alg::RosenbrockAlgorithm) = true
 is_mass_matrix_alg(alg::NewtonAlgorithm) = !isesdirk(alg)
-
-isodemultistep(::Any) = false
-isodemultistep(::Union{
-  # TODO: fix the nlsolve for other multistep methods
-  ABDF2,
-  #AB3,AB4,AB5,ABM32,ABM43,ABM54,
-  #VCAB3,VCAB4,VCAB5,VCABM3,VCABM4,VCABM5,VCABM,
-  #CNAB2,CNLF2,
-  #QNDF1,QNDF2,QNDF,SBDF,
-  Trapezoid,
-}) = true
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -572,7 +572,8 @@ is_mass_matrix_alg(alg::NewtonAlgorithm) = !isesdirk(alg)
 isodemultistep(::Any) = false
 isodemultistep(::Union{
   # TODO: fix the nlsolve for other multistep methods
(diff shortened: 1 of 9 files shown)


src/nlsolve/type.jl
```diff
@@ -17,14 +24,16 @@ mutable struct NLSolver{algType,iip,uType,tType,C<:AbstractNLSolverCache} <: Abs
   maxiters::Int
   status::NLStatus
   cache::C
+  method::MethodType
```
kanav99, line 27: Do we have plans to standardize NLSolver code? We should probably move this to DiffEqBase soon

src/perform_step/bdf_perform_step.jl
```diff
@@ -16,41 +16,48 @@ end
   alg = unwrap_alg(integrator, true)
   dtₙ, uₙ, uₙ₋₁, uₙ₋₂ = integrator.dt, integrator.u, integrator.uprev, integrator.uprev2
 
+  # TODO: this doesn't look right
```
kanav99, line 19: Maybe a later optimization, but we should actually write the Euler step instead of using `perform_step!` for ImplicitEuler. Any change in that function could lead to breakage

src/alg_utils.jl
```diff
@@ -572,7 +572,8 @@ is_mass_matrix_alg(alg::NewtonAlgorithm) = !isesdirk(alg)
 isodemultistep(::Any) = false
 isodemultistep(::Union{
   # TODO: fix the nlsolve for other multistep methods
-  #ABDF2,AB3,AB4,AB5,ABM32,ABM43,ABM54,
+  ABDF2,
+  #AB3,AB4,AB5,ABM32,ABM43,ABM54,
   #VCAB3,VCAB4,VCAB5,VCABM3,VCABM4,VCABM5,VCABM,
   #CNAB2,CNLF2,
   #QNDF1,QNDF2,QNDF,SBDF,
```
devmotion, line 579: Is this still used somewhere now that you introduced `nlsolver.method`?

ChrisRackauckas:
Can't this just be done with two tmps? They can default to `nothing` to skip the extra kernel, and then getting each algorithm right would be just about using the right `tmp`. Also this would open up general linear methods.

src/nlsolve/type.jl
YingboMa: Not soon. We still have quite a bit of refactoring to do.

src/perform_step/bdf_perform_step.jl
YingboMa: `ImplicitEuler` most likely won't change in such a way that breaks `ABDF2`. The comment refers to the discontinuity handling.

src/alg_utils.jl
YingboMa: Yeah, those methods can be removed.

YingboMa:
We still need `MethodType` even with two `tmp`s, because `COEFFICIENT_MULTISTEP` and `NORDSIECK_MULTISTEP` have the same `tmp`, but they should go to different code paths.

YingboMa:
CI failure is due to timeouts.

PR #1165, opened 2020-06-05
Update QNDF to new nlsolver

utkarsh530:


ChrisRackauckas:
How's the latest benchmark? I assume that this would "fix it".

utkarsh530:
Before:
wp_all_prev
After:
wp_all_new

src/perform_step/bdf_perform_step.jl
```diff
@@ -868,28 +879,43 @@ function perform_step!(integrator,cache::QNDFCache,repeat_step=false)
...
-  @.. nlsolver.z = (uprev + tm -nlsolver.tmp)*inv(γ)
+  @.. nlsolver.z = uprev + tm
+
+  mass_matrix = f.mass_matrix
+  
+  if mass_matrix === I
+    @.. tmp = (uprev + tm - ϕ)/ (dt * β₀)
+    
+  else
+    dz = nlsolver.cache.dz
```
kanav99, line 906: I am out of sync due to the MLH things, but we don't handle mass matrix intrinsically in nlsolve? tmp depending on mass matrix sounds spectical

test/interface/mass_matrix_tests.jl
```diff
@@ -78,11 +78,11 @@ dependent_M2 = DiffEqArrayOperator(ones(3,3),update_func=update_func2)
 
     @test _norm_dsol(QBDF1(),prob,prob2) ≈ 0 atol=1e-12
     @test _norm_dsol(QBDF2(),prob,prob2) ≈ 0 atol=1e-12
-    @test _norm_dsol(QBDF(),prob,prob2) ≈ 0 atol=1e-7
+    @test _norm_dsol(QBDF(),prob,prob2) ≈ 0 atol=1e-12
 
     @test _norm_dsol(QNDF1(),prob,prob2) ≈ 0 atol=1e-12
     @test _norm_dsol(QNDF2(),prob,prob2) ≈ 0 atol=1e-12
-    @test _norm_dsol(QNDF(),prob,prob2) ≈ 0 atol=1e-7
+    @test _norm_dsol(QNDF(),prob,prob2) ≈ 0 atol=1e-12
```
kanav99, line 85: this is nice

src/perform_step/bdf_perform_step.jl
YingboMa: No, this is right.

kanav99: Alright 👍 I need to see the new nlsolver changes.

utkarsh530:
I updated the plots slightly for better comparison (due to ode15s).

src/perform_step/bdf_perform_step.jl
```diff
@@ -733,24 +733,35 @@ function perform_step!(integrator,cache::QNDFConstantCache,repeat_step=false)
...
+
+  if mass_matrix == I
+    nlsolver.tmp = @.. (u₀ - ϕ)/ (dt * β₀)
+  else
+    _tmp = mass_matrix * @.. (u₀ - ϕ)
+    nlsolver.tmp = @.. _tmp / (dt * β₀)
+  end
+  nlsolver.γ = β₀
+  nlsolver.α = α₀
+  nlsolver.method = COEFFICIENT_MULTISTEP
```
devmotion, line 760: I'm a bit surprised about the mix of broadcasted and non-broadcasted expressions (not only in these lines, here it would be just `uprev .+ sum(D)`, but also above e.g. in `ϕ += γₖ[i] * D[i]`). Shouldn't we consistently use broadcasting? Or does it still impact performance?

YingboMa:
Why are there two QNDFs in the plots? Is it one QNDF and one QBDF? Which one is better? The WPD curve looks really suspicious since the convergence looks erratic.

utkarsh530:
> Why are there two QNDFs in the plots? Is it one QNDF and one QBDF? Which one is better? The WPD curve looks really suspicious since the convergence looks erratic.

The 'inverted green triangle' legend is QBDF and the other one is QNDF.

YingboMa:
Do you mind to benchmark `QNDF/QBDF` against `TRBDF2`, `QBDF2`, `ABDF2`, `Trapezoid`, and `Rosenbrock23`?

ChrisRackauckas:
Looks like it didn't move... which is odd.

utkarsh530:
> Do you mind to benchmark `QNDF/QBDF` against `TRBDF2`, `QBDF2`, `ABDF2`, `Trapezoid`, and `Rosenbrock23`?

Before:

wp_all_prev
After:
wp_all_new

YingboMa:
What's the benchmark problem?

src/perform_step/bdf_perform_step.jl
utkarsh530: Yeah I agree, actually, there's some dimensional problem like that D is 1x5 matrix and γ is 6-Array, it was originally defined like that. we can change it probably for more consistent operation using broadcast, maybe in another PR.

utkarsh530:
POLLU, with `testsol` with `Rodas5`

utkarsh530:
@YingboMa any further changes to be done? The build is passing [here](https://travis-ci.org/github/SciML/OrdinaryDiffEq.jl/builds/695084963)

PR #1166, opened 2020-06-05
Kernel abstractions for LSRK

kanav99:
After this PR
```
CPU Times
CarpenterKennedy2N54
 14.442381 seconds (92.63 k allocations: 27.064 MiB, 2.79% gc time)
GPU Times
CarpenterKennedy2N54
  4.807376 seconds (899.47 k allocations: 30.019 MiB)
```

Before
```
CPU Times
CarpenterKennedy2N54
  6.195860 seconds (4.09 k allocations: 23.075 MiB)
GPU Times
CarpenterKennedy2N54
  1.473602 seconds (1.08 M allocations: 35.924 MiB)
```

Kernel abstractions do not look like a good solution, or maybe I am doing something wrong somewhere else?


ChrisRackauckas:
I'm not sure we want this as a dependency anyways, since IIRC it's a pretty huge one and brings in a ton of compiler stuff.

kanav99:
Yeah, I didn't realized it had similar big dependencies. Can we have a seperate GPU method in DiffEqGPU?

ChrisRackauckas:
yeah we could

PR #1169, opened 2020-06-07
Migrate to JuliaGPU's CI setup

christopher-dG:
First try at this, might need some fixing up.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,73 +1,33 @@
-stages:
-  - build
-  - test
+include: https://raw.githubusercontent.com/JuliaGPU/gitlab-ci/master/templates/v6.yml
 
 variables:
-  JULIA_DEPOT_PATH: "$CI_PROJECT_DIR/.julia/"
   JULIA_NUM_THREADS: '4'
-  JULIA_VERSION: 1.3.0
 
-.root:
+.base:
+  extends:
+    - .julia:1.4
+    - .test
   tags:
-    - 'p6000'
-  only:
-    - master
-    - tags
-    - external
-    - pushes
-
-build:
-  extends: .root
-  stage: build
-  script:
-    - curl https://julialang-s3.julialang.org/bin/linux/x64/$(echo $JULIA_VERSION | cut -d '.' -f 1,2)/julia-${JULIA_VERSION}-linux-x86_64.tar.gz -o julia.tar.gz
-    - unp julia.tar.gz
-    - export PATH="$(pwd)/julia-${JULIA_VERSION}/bin:$PATH"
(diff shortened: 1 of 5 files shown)


christopher-dG:
Seems like the GPU job is timing out, a lot of time is spent downloading artifacts.

ChrisRackauckas:
Are others having this?

devmotion:
Maybe we should start caching artifacts as described in https://discourse.julialang.org/t/recommendation-cache-julia-artifacts-in-ci-services/35484?u=devmotion (might be useful on Travis as well)?

christopher-dG:
Artifacts are being cached in GitLab, so as soon as one job succeeds it should be stored. I haven't tried any other repos but I'll do that now and see what happens.

christopher-dG:
The `intel` runner seems to have a pretty flaky network connection.

christopher-dG:
Out of memory after 2.5 hours! https://gitlab.com/JuliaGPU/OrdinaryDiffEq.jl/-/jobs/588230472

ChrisRackauckas:
https://gitlab.com/JuliaGPU/OrdinaryDiffEq.jl/-/jobs/588230472#L6614

@maleadt is that statistic correct or is it underreporting how much memory we are using?

ChrisRackauckas:
This repo might need a tag requesting the "not the worst" GPU resources: the GPU cost here isn't all that much (probably not more than 1GB?), but it's not just unit tests: it's running a (small) PDE integration test and that's what's failing.

maleadt:
```
   LoadError: Out of GPU memory trying to allocate 768.000 KiB
  Effective GPU memory usage: 99.98% (31.714 GiB/31.719 GiB)
  CuArrays allocator usage: 8.516 MiB (capped at 512.000 MiB)
  BinnedPool usage: 8.516 MiB (8.516 MiB allocated, 0 bytes cached)
```

So this job was killed because the GPU was almost entirely used, but outside of CuArrays. You should pick a device that the most memory free, e.g., https://github.com/JuliaGPU/CUDA.jl/blob/08a83b5d9072555d469894b6443fced04c9593e9/test/runtests.jl#L106-L127'

Artifacts are being cached by Gitlab indeed, but that caching is fairly slow (it zips up the cache directory). We might have to find a better way, like hosting a caching Pkg server on the runners (which presumably is pretty easy).

ChrisRackauckas:
>I seem to have lost my commit rights to this branch, so please apply this for me -- we should be activating the downstream env in local tests too.

Yeah sorry about that: I removed ownership since I thought we were done with all of the updates. I lost track of this one. I'll merge and we can do a followup PR if needed.

PR #1173, opened 2020-06-09
QNDF Fixes

utkarsh530:
In continuation of comments mentioned in #1158 , here are some fixes made in QNDF.
Results:
```
    function test!(du, u, p, t)
        du[1] = -10 * u[1]
        du[2] = -0.01 * u[2]
    end
    u0 = [1.01;1.01]
    tspan = (0.0, 1.0)
    prob = ODEProblem(test!, u0, tspan)
    @time sol = solve(prob, QNDF()) 
```
Before:
```
julia> @time sol = solve(prob, QNDF())
  0.000780 seconds (2.08 k allocations: 101.531 KiB)
julia> sol.destats
DiffEqBase.DEStats
Number of function 1 evaluations:                  361
Number of W matrix evaluations:                    57
Number of linear solves:                           243
Number of Jacobians created:                       1
Number of nonlinear solver iterations:             243
Number of nonlinear solver convergence failures:   0
Number of accepted steps:                          68
Number of rejected steps:                          46

```
After
```
julia> @time sol = solve(prob, QNDF())
  0.000682 seconds (1.57 k allocations: 81.781 KiB)
julia> sol.destats
DiffEqBase.DEStats
Number of function 1 evaluations:                  247
Number of W matrix evaluations:                    28
Number of linear solves:                           173
Number of Jacobians created:                       1
Number of nonlinear solver iterations:             173
Number of nonlinear solver convergence failures:   0
Number of accepted steps:                          70
Number of rejected steps:                          11
```
@ChrisRackauckas @kanav99 please review.

### DIFF
--- a/src/integrators/controllers.jl
+++ b/src/integrators/controllers.jl
@@ -138,6 +138,7 @@ function stepsize_controller!(integrator, alg::QNDF)
   cnt = integrator.iter
   EEst1 = integrator.cache.EEst1
   EEst2 = integrator.cache.EEst2
+  prev_order = integrator.cache.order
   dttmp = QNDF_stepsize_and_order!(integrator.cache, integrator.EEst, EEst1, EEst2, integrator.dt, integrator.cache.order)
   integrator.cache.nconsteps = min(integrator.cache.nconsteps+1,5+2); #Max-order + 2
   if integrator.EEst > one(integrator.EEst)
@@ -145,11 +146,13 @@ function stepsize_controller!(integrator, alg::QNDF)
     q = integrator.dt/dttmp
     integrator.qold = integrator.dt/q
   else
-    if integrator.cache.nconsteps <= integrator.cache.order + 2
+    if integrator.cache.nconsteps < integrator.cache.order + 2
       q = one(integrator.qold) #quasi-contsant steps
       integrator.qold = integrator.dt/q
     else
-      integrator.cache.nconsteps = 1
+      if (integrator.dt != dttmp || prev_order != integrator.cache.order)
+        integrator.cache.nconsteps = 1
+      end
       q = integrator.dt/dttmp
     end
   end
--- a/src/integrators/controllers.jl
+++ b/src/integrators/controllers.jl
@@ -138,18 +138,20 @@ function stepsize_controller!(integrator, alg::QNDF)
   cnt = integrator.iter
   EEst1 = integrator.cache.EEst1
   EEst2 = integrator.cache.EEst2
-  if integrator.cache.nconsteps < integrator.cache.order + 2
-    integrator.cache.nconsteps += 1
-    q = one(integrator.qold) #quasi-contsant steps
+  dttmp = QNDF_stepsize_and_order!(integrator.cache, integrator.EEst, EEst1, EEst2, integrator.dt, integrator.cache.order)
+  integrator.cache.nconsteps = min(integrator.cache.nconsteps+1,5+2); #Max-order + 2
+  if integrator.EEst > one(integrator.EEst)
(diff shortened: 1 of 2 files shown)


src/integrators/controllers.jl
```diff
@@ -166,7 +171,7 @@ end
 function step_reject_controller!(integrator,alg::QNDF)
   #append no. of consecutive failed steps
   integrator.cache.consfailcnt += 1
-  integrator.dt = integrator.dt/integrator.qold
+  integrator.dt = integrator.qold
```
kanav99, line 174: Why is qold used as dt? I think qold is unitless

src/perform_step/bdf_perform_step.jl
```diff
@@ -936,21 +936,38 @@ function perform_step!(integrator,cache::QNDFCache,repeat_step=false)
...
+      cache.EEst2 = Inf
+    end
     # cnt == 1
   end # integrator.opts.adaptive
 
+  if integrator.EEst > one(integrator.EEst)
+    for i = 1:5
+      fill!(D[i], zero(eltype(u)))
+    end
+    for i = 1:6, j = 1:6
```
kanav99, line 964: there must be a better way to fill it recursively with zeros

src/integrators/controllers.jl
utkarsh530: I have changed some definitions here https://github.com/SciML/OrdinaryDiffEq.jl/pull/1173/files#diff-c90c88b38a0b685e7edd6abc54fa1ef8R147-R151

kanav99: You would probably need to change definitions at many places for q and qold. Also this wont work with units. qold and q are used as ratios, make sure that nothing funny is happening. Also this would make a problem for composite algorithms, but that can be handled later

utkarsh530: https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/integrators/controllers.jl#L33
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/integrators/controllers.jl#L108
I tried to follow what was done in the standard step-size controller and how q and old were used by `step_accept_controller` and `step_reject_controller`. Let me know if it's okay.

kanav99: ohh, that's... strange. I think we have a separate issue then. Surprised this works with units.

utkarsh530: Yup there's a difference of `q` & `qold` as to how algorithms use them.

ChrisRackauckas: yeah... it's kind of kludgy and probably is worth refactoring.

utkarsh530: I would definitely like to look into it!

src/perform_step/bdf_perform_step.jl
ChrisRackauckas: I don't know of one 🥘

ChrisRackauckas:
Looks like this needs to rebase, and one of the regression tests on the interpolation are failing.

src/perform_step/bdf_perform_step.jl
devmotion: Doesn't solve the problem, but one could also just iterate over `D` and `D2` without writing down the indices explicitly. Would still use `fill!` recursively but one doesn't have to know and write down the dimensions explicitly.

YingboMa:
I am trying to fix QNDF in another branch. Could you wait for that?

utkarsh530:
Yeah sure. Let me know if there's something for I could help.

PR #1179, opened 2020-06-13
vec inside of WOperator

ChrisRackauckas:
Since the Jacobian is defined to act like a matrix, things work better if we always vec. It's what's done in all of the direct methods, so this is just matching what's going on there. Fixes downstream DiffEqOperators failures.


src/derivative_utils.jl
```diff
@@ -290,29 +290,30 @@ function Base.:\(W::WOperator, x::Union{AbstractVecOrMat,Number})
 end
 
 function LinearAlgebra.mul!(Y::AbstractVecOrMat, W::WOperator, B::AbstractVecOrMat)
+  vec = Base.vec
```
ChrisRackauckas, line 293: I'm not sure why this is required, but it is.

PR #1185, opened 2020-06-18
Use IncrementingODE in LSRK 2N methods

kanav99:
Test script - 
```julia
       using OrdinaryDiffEq
       function f(du, u, p, t, increment)
           if increment
               @. du += u
           else
               @. du = u
           end
       end
       prob = IncrementingODEProblem(f, [1.0], (0.0, 1.0))
       solve(prob, CarpenterKennedy2N54(;williamson_condition=false); dt=1.0)
```

I don't like the idea of dividing with `dt` in this PR

### DIFF
--- a/src/caches/low_storage_rk_caches.jl
+++ b/src/caches/low_storage_rk_caches.jl
@@ -48,7 +48,7 @@ function alg_cache(alg::ORK256,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUn
     k = zero(rate_prototype)
     williamson_condition = false
   else
-    if williamson_condition
+    if williamson_condition || f isa IncrementingODEFunction
       k = tmp
     else
       k = zero(rate_prototype)
@@ -93,7 +93,7 @@ function alg_cache(alg::CarpenterKennedy2N54,u,rate_prototype,uEltypeNoUnits,uBo
     k = zero(rate_prototype)
     williamson_condition = false
   else
-    if williamson_condition
+    if williamson_condition || f isa IncrementingODEFunction
       k = tmp
     else
       k = zero(rate_prototype)
@@ -142,7 +142,7 @@ function alg_cache(alg::HSLDDRK64,u,rate_prototype,uEltypeNoUnits,uBottomEltypeN
     k = zero(rate_prototype)
     williamson_condition = false
   else
-    if williamson_condition
+    if williamson_condition || f isa IncrementingODEFunction
       k = tmp
     else
       k = zero(rate_prototype)
@@ -193,7 +193,7 @@ function alg_cache(alg::DGLDDRK73_C,u,rate_prototype,uEltypeNoUnits,uBottomEltyp
     k = zero(rate_prototype)
     williamson_condition = false
   else
-    if williamson_condition
+    if williamson_condition || f isa IncrementingODEFunction
       k = tmp
     else
       k = zero(rate_prototype)
@@ -247,7 +247,7 @@ function alg_cache(alg::DGLDDRK84_C,u,rate_prototype,uEltypeNoUnits,uBottomEltyp
     k = zero(rate_prototype)
     williamson_condition = false
   else
-    if williamson_condition
+    if williamson_condition || f isa IncrementingODEFunction
       k = tmp
     else
       k = zero(rate_prototype)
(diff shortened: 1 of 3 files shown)


src/perform_step/low_storage_rk_perform_step.jl
```diff
@@ -52,18 +56,29 @@ end
...
   # other stages
   for i in eachindex(A2end)
-    if williamson_condition
-      f(ArrayFuse(tmp, u, (A2end[i], dt, B2end[i])), u, p, t+c2end[i]*dt)
-    else
-      @.. tmp = A2end[i]*tmp
-      f(k, u, p, t+c2end[i]*dt)
-      @.. tmp += dt * k
+    if f isa IncrementingODEFunction
+      @.. tmp = (A2end[i] / dt)*tmp
```
ChrisRackauckas, line 60: how come it's required here but not in the other?

kanav99: https://github.com/SciML/OrdinaryDiffEq.jl/pull/1185/files#diff-32bcae2e479bdf2c77ea6287240cdcbcR62 we multiply dt again here. In increment form, the function does `du += f(u,p,t)` so to implement `du += k * f(u,p,t)` we need to divide du by k first, then pass it in the function them multiply it back with k

ChrisRackauckas: is that still hitting the minimal number of kernel calls?

kanav99: Nope that's still 2 kernel calls. We can't make it to 1 without kernel abstractions

kanav99:
I changed the usage of tmp from type of `dt*k` to `k`. This helped reduce the kernel calls from 3 to 2 external of `f`

ChrisRackauckas:
So merge and tag the DiffEqBase PR?

kanav99:
Yes. Also should I pull Simon here? I want to make sure if this PR does what https://github.com/SciML/OrdinaryDiffEq.jl/pull/1153#issuecomment-637703922 says

kanav99:
@ChrisRackauckas now we have scaling, we have optimal kernel calls.

kanav99:
Testing script - 
```julia
using OrdinaryDiffEq

function f(du, u, p, t)
    du .= u
end

function f2(du, u, p, t, increment; α=1.0, β=1.0)
    if increment
        @. du = α * u + β * du
    else
        @. du = u
    end
end

algs = [ORK256, CarpenterKennedy2N54, HSLDDRK64, DGLDDRK73_C, DGLDDRK84_C, DGLDDRK84_F, NDBLSRK124, NDBLSRK134, NDBLSRK144]
probIncrementing = IncrementingODEProblem(f2, [1.0], (0.0, 1.0))
probNormal = ODEProblem(f, [1.0], (0.0, 1.0))

for alg in algs
    sol = solve(probIncrementing, alg(); dt=0.1);
    @show sol.u[end]

    sol = solve(probNormal, alg(;williamson_condition=false); dt=0.1);
    @show sol.u[end]

    sol = solve(probNormal, alg(); dt=0.1);
    @show sol.u[end]
end
nothing
```

simonbyrne:
> ```julia
> f2(du, u, p, t, increment; α=1.0, β=1.0)
> ```

I think just using six arguments `f(du, u, p, t, α, β)` is sufficient. Note that the 4-arg form can be derived by using the default values `f(du, u, p, t, α=true, β=false)`

kanav99:
Let's get these merged today, do you have any recommendations @ChrisRackauckas ? We need the DiffEqBase PR merged first, should I do that?

src/caches/low_storage_rk_caches.jl
```diff
@@ -1,3 +1,13 @@
+DiffEqBase.@def lsrkcache begin
+  tmp = zero(rate_prototype)
+  williamson_condition = alg.williamson_condition
+  if williamson_condition || f isa IncrementingODEFunction
+    k = tmp
+  else
+    k = zero(rate_prototype)
+  end
+  LowStorageRK2NCache(u,uprev,k,tmp,tab)
+end
```
devmotion, line 10: I'm wondering, what are the advantages/disadvantages over using a proper function here? Or aybe just use an `@eval` loop that defines the whole `alg_cache` function for a given combination of algorithm type and tableau type? I.e., something like
```julia
for (A, T) in zip(algtypes, tableau)
    @eval begin
        function alg_cache(alg::$A, u, rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Val{true})
            tab = $T(real(uBottomEltypeNoUnits), real(tTypeNoUnits))
            tmp = zero(rate_prototype)
            williamson_condition = alg.williamson_condition
            if williamson_condition || f isa IncrementingODEFunction
                k = tmp
            else
                k = zero(rate_prototype)
            end
            return LowStorageRK2NCache(u,uprev,k,tmp,tab)
        end
    end
end
```

PR #1197, opened 2020-06-30
[WIP] Rice3 Multirate Method

kanav99:
First approach from https://nvlpubs.nist.gov/nistpubs/jres/64B/jresv64Bn3p151_A1b.pdf

- [x] OOP
- [ ] IIP
- [ ] Error Estimation
- [ ] Make it compatible with all 3-order RK methods

Test Script
```julia
using OrdinaryDiffEq

function f(u,v,p,t)
    return u
end

function g(u,v,p,t)
    return v
end

prob = DynamicalODEProblem(f,g,[1.0], [1.0], (0.0, 1.0))
sol = solve(prob, Rice3(BS3(), 5))
@show sol.u
```

### DIFF
--- a/src/perform_step/multirate_perform_step.jl
+++ b/src/perform_step/multirate_perform_step.jl
@@ -1,0 +1,100 @@
+function initialize!(integrator, cache::Rice3ConstantCache)
+  integrator.kshortsize = 2
+  integrator.k = typeof(integrator.k)(undef, integrator.kshortsize)
+  integrator.fsalfirst = integrator.f(integrator.uprev, integrator.p, integrator.t) # Pre-start fsal
+  integrator.destats.nf += 1
+
+  # Avoid undefined entries if k is an array of arrays
+  integrator.fsallast = zero(integrator.fsalfirst)
+  integrator.k[1] = integrator.fsalfirst
+  integrator.k[2] = integrator.fsallast
+end
+
+@muladd function perform_step!(integrator, cache::Rice3ConstantCache, repeat_step=false)
+  @unpack t,dt,uprev,u,p = integrator
+  f, g = integrator.f.f1, integrator.f.f2
+  K = integrator.alg.K
+  xprev, yprev = uprev.x[1], uprev.x[2]
+  @unpack a21,a32,a41,a42,a43,c1,c2,btilde1,btilde2,btilde3,btilde4 = cache.tab
+  # slow time stepping
+  k1, h1 = integrator.fsalfirst.x[1], integrator.fsalfirst.x[2]
+  k2 = f(xprev + dt*a21*k1, yprev + dt*a21*h1, p, t+c1*dt)
+  h2 = g(xprev + dt*a21*k1, yprev + dt*a21*h1, p, t+c1*dt)
+  k3 = f(xprev + dt*a32*k2, yprev + dt*a32*h2, p, t+c2*dt)
+  h3 = g(xprev + dt*a32*k2, yprev + dt*a32*h2, p, t+c2*dt)
+  x = xprev + dt * (a41*k1 + a42*k2 + a43*k3)
+
+  # slow time stepping
+  h = dt/K
+  xj = xprev
+  yj = yprev
+  for j in 1:K
+    # don't delete these comments
+    # tj = t+j*h
+    # d1 = h*g(xj, yj, tj)
+    # d2 = h*g(xj + λ4(j) * k1 + λ5(j) * k2 + λ6(j) * k3, yj + μ1 * d1, tj + μ1*h)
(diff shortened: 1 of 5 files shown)


kanav99:
@ChrisRackauckas do take a look when you are free!

src/perform_step/multirate_perform_step.jl
```diff
@@ -0,0 +1,175 @@
+get_a21(tab::OrdinaryDiffEqTableau{T,T2}) where {T,T2} = hasfield(typeof(tab), :a21) ? tab.a21 : zero(T)
```
ChrisRackauckas, line 1: hmm, interesting. This won't scale... but does it need to?

kanav99: we probably wont need more than 4-5 order methods, base methods are usually small order ones

ChrisRackauckas: why only low order? Why does this not apply to arbitrary methods?

devmotion: Maybe one could define something like
```julia
struct Zero end

function Base.getproperty(tab::OrdinaryDiffEqTableau, sym::Symbol)
    if hasfield(tab, sym)
        getfield(tab, sym)
    else
        Zero()
    end
end

Base.:+(x::Real, ::Zero) = x
Base.:+(::Zero, x::Real) = x
Base.:+(::Zero, ::Zero) = Zero()

Base.:*(x::Real, ::Zero) = Zero()
Base.:*(x::Zero, y::Real) = y * x
Base.:*(::Zero, ::Zero) = Zero()
```
and the compiler would be smart enough to compile away all the unneeded operations if the fields don't exist. 

The annoying thing about just returning `zero(T)` is that all the multiplications and additions will still be performed. Additionally, in this way you could still just use `@unpack a21, .... = cache.tab` instead of defining all these (arguably) ugly `get_...` methods.

kanav99: > why only low order? Why does this not apply to arbitrary methods?  
Not sure but Sandu's paper suggests the same that base methods are generally small order ones

> and the compiler would be smart enough to compile away all the unneeded operations if the fields don't exist.   
This is nice idea, would this be compatible with muladd too?

kanav99:
The original paper explains coefficients for extrapolation too, that should come in a seperate PR I guess?

src/perform_step/multirate_perform_step.jl
devmotion: > This is nice idea, would this be compatible with muladd too?

Probably not with the code snippet above, but you could just define dispatches for `Base.muladd`.

BTW I actually wanted to write "and hopefully the compiler would be smart enough" since I didn't check it carefully, I just assumed it would. I did that now, and it seems to work in the following example:
```julia
struct Tableau{A}
    a::A
end

struct Tableau2{A,B}
    a::A
    b::B
end

function f(t)
    a = t.a
    b = t.b
    return a + 42 * b * (500 + 3.0)
end

@code_typed f(Tableau(1.0))
@code_llvm f(Tableau(1.0))
@code_typed f(Tableau2(1.0, 2.0))
@code_llvm f(Tableau2(1.0, 2.0))
```

devmotion: The following doesn't compile away though (unsurprisingly, I guess):
```julia
function f(t)
    a = t.a
    b = t.b
    return a + 42 * b * rand()
end
```
For broadcasting support one has to add some additional definitions, it seems.

devmotion: BTW `muladd` seems to work for free due to the fallback in https://github.com/JuliaLang/julia/blob/30b09dfd2b3a898237f71d4a2b0812466d017f9e/base/math.jl#L1139.

ChrisRackauckas: >BTW muladd seems to work for free due to the fallback in https://github.com/JuliaLang/julia/blob/30b09dfd2b3a898237f71d4a2b0812466d017f9e/base/math.jl#L1139.

But that means it's not actually doing muladd, which isn't what we want.

devmotion: But isn't that what we want here? We don't want any computations at all since we just add zero or multiply with zero. So the fallback is fine since it allows use to eliminate them at compile time.

ChrisRackauckas: I guess so.

kanav99: I think this approach is good now. But we need to make sure we implement it right, like return `Zero()` only if the requested field is of form `aXY`, `bX`, `cX` etc otherwise debugging stuff won't be easy lol

kanav99: Good for next PR or do it in the same?

ChrisRackauckas: yeah

PR #1200, opened 2020-07-04
DEstats update in Extrapolation Methods & fixes

utkarsh530:
Hi,
I have made two fixes in this PR:
1. `integrator.destats.nf` has been updated where a function call is made.
2. According to the [thesis](https://github.com/AlthausKonstantin/Extrapolation/blob/master/Bachelor%20Theseis.pdf), it uses 2*subsequence for internal discretisation, so I guess it was a fluke that 4 * subsequence was being used. Even the stage_number calculation [here](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/caches/extrapolation_caches.jl#L370-L376) assumes that we are using the latter one. There would be trade-off I guess in the function evaluations and accuracy, so let me know to how judge that. (We can also update to use 4 * subsequence and update the corresponding work). Also the native code [here](https://github.com/AlthausKonstantin/Extrapolation/blob/master/Solver/adaptive-extrapolation-Hairer%26Wanner.jl#L318) and [here](https://github.com/AlthausKonstantin/Extrapolation/blob/master/Solver/adaptive-extrapolation-Hairer%26Wanner.jl#L422-L431) uses the argument that I just mentioned. 

A snippet from the thesis:
image
@ChrisRackauckas please have a look.


utkarsh530:
Here's a work precision diagram on Lodka-Volterra:
(green = 2 * subsequence, orange= 4 * subsequence , blue = 8 * subsequence)
wp_extrplmethod1_comp

ChrisRackauckas:
Look at the linear plot vs the other methods. The very expensive f will be interesting. Looks like making this an option could be useful.

utkarsh530:
Linear Work-Precision
```
 wp = WorkPrecisionSet(prob,abstols,reltols,setups;names=solnames,
                             save_everystep=false,verbose=false,numruns=100)
```
Before:
wp_extrplmethod1_lin_b
After:
wp_extrplmethod1_lin

ChrisRackauckas:
I'd say a good next step would be to make this an option instead of hardcoded to 2, and then play around and see if you could find any setup that is good across different problems. You can see in the benchmarks that I tried a few things but 🤷 . What holds up the profiles for the extrapolation methods? Is multithreading doing well? Can you find a very expensive `f` non-stiff ODE where multithreading improves something? The extrapolating methods will only be good at really high tolerance, so focusing on that zone is the right idea.

If explicit isn't able to find a sweet spot, then just move to trying to get the stiff ones more stable. Are there any benchmarks where they are still diverging? Is there a reason why? Are the lu-factorizations performed in parallel? Is most of the profile spent in the lu-factorization stage? I think the parallelized methods on stiff equations should easily scale on computers with a lot of threads, so it would be good to demonstrate this. However, they can't do W-reuse, so it'll be an interesting game: it'll need a lot of `f` evals to do well, so making it less conservative in `dt` to better exploit parallelism than what Hairer did, possibly modifying the work calculation to take into account the number of threads, would be a good way to go. In theory, having 4 cores all doing `lu` factorizations at the same time should end up better than 1, so I'd try to look at all of those details.

PR #1201, opened 2020-07-05
Remake for Algorithms

kanav99:
Needs SciML/DiffEqBase.jl#545


ChrisRackauckas:
Could we get this for DelayDiffEq and StochasticDiffEq as well?

PR #1216, opened 2020-07-26
Rkmk2 

Biswajitghosh98:
- [x] Add proper test problems

### DIFF
--- a/src/perform_step/linear_perform_step.jl
+++ b/src/perform_step/linear_perform_step.jl
@@ -28,6 +28,39 @@ function perform_step!(integrator, cache::MagnusMidpointCache, repeat_step=false
   integrator.destats.nf += 1
 end
 
+function initialize!(integrator, cache::RKMK2Cache)
+  integrator.kshortsize = 2
+  integrator.fsalfirst = cache.fsalfirst
+  integrator.fsallast = cache.k
+  resize!(integrator.k, integrator.kshortsize)
+  integrator.k[1] = integrator.fsalfirst
+  integrator.k[2] = integrator.fsallast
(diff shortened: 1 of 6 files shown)


Biswajitghosh98:
@ChrisRackauckas Please have a look.

src/perform_step/linear_perform_step.jl
```diff
@@ -28,6 +28,39 @@ function perform_step!(integrator, cache::MagnusMidpointCache, repeat_step=false
...
+
+  L = integrator.f.f
+  update_coefficients!(L,uprev,p,t)
+  A = Matrix(deepcopy(L))
+  k1 = dt*A
+  update_coefficients!(L,exp(k1)*uprev,p,t)
+  B = Matrix(deepcopy(L))
+  k2 = dt*B
+  if integrator.alg.krylov
+    u .= expv(1/2, (k1+k2), uprev; m=min(alg.m, size(L,1)), opnorm=integrator.opts.internalopnorm, iop=alg.iop)
```
ChrisRackauckas, line 55: the dt should probably be handled as a multiplication against the 1/2 so that it gets rid of an O(n^2) operation.

ChrisRackauckas:
Other than that efficiency note, it looks great though. We can loop back around after they are all implemented and tested though to do that kind of clean up.

src/perform_step/linear_perform_step.jl
Biswajitghosh98: Yes. As of now, I'm doing it only to make the steps more understandable for me for debugging at a later point in time. Later I'll loop back through all my PRs, and optimize each of them.

PR #1217, opened 2020-07-26
Update Work Calculation in ImplicitEulerExtrapolation

utkarsh530:
Hi, I have updated work calculation to take account of the Jacobian, using Hairer Wanner Adaptivity, as mentioned in Hairer II. After that, I will update adaptivity coinciding with Hairer Wanner.
Work Calculation:
image

@ChrisRackauckas @kanav99 can you have a look?


ChrisRackauckas:
Let's think about an reconsider the default after it's all optimized and we see some benchmarks. I think we may want to multithread all of them by default just given the region of benchmarkspace where they are good, but that needs to be confirmed.

Issue #1219, opened 2020-07-28
Rodas5 promotes static arrays to arrays when autodiff=false 


andreasnoack:
With OrdinaryDiffEq v5.41.0
```julia
f = (u, p, t) -> -0.7*u
u0  = @SVector([10.0])
tspan = (0.0, 5.0)
prob  = ODEProblem(f, u0, tspan)
sol = solve(prob, Rodas5(autodiff=false))
```
errors with
```julia
ERROR: MethodError: no method matching reshape(::Array{Float64,1}, ::Tuple{SOneTo{1}})
Closest candidates are:
  reshape(::Array{T,M}, ::Tuple{Vararg{Int64,N}}) where {T, N, M} at reshapedarray.jl:40
  reshape(::AbstractArray, ::Int64...) at reshapedarray.jl:116
  reshape(::AbstractArray, ::Union{Int64, AbstractUnitRange}...) at reshapedarray.jl:110
  ...
Stacktrace:
 [1] _reshape(::Array{Float64,1}, ::Tuple{SOneTo{1}}) at /Users/andreasnoack/.julia/packages/DiffEqBase/ytJuW/src/utils.jl:228
[...]

PR #1223, opened 2020-07-29
Extend tests to cover autodiff=false for StaticArrays

andreasnoack:
This triggers https://github.com/SciML/OrdinaryDiffEq.jl/issues/1219. It looks like the root cause might be https://github.com/JuliaDiff/FiniteDiff.jl/blob/310da8c934c249c978a5e359d5bcadeafffe64ce/src/jacobians.jl#L198 since
```julia
julia> reshape(SMatrix{1,1}(1.0), (1,1))
1×1 reshape(::SArray{Tuple{1,1},Float64,2,1}, 1, 1) with eltype Float64:
 1.0

julia> reshape(SMatrix{1,1}(1.0), (1,1))*1.0
1×1 Array{Float64,2}:
 1.0
```
i.e. no longer a StaticArray but I'm not yet completely sure that this is the actual problem.

### DIFF
--- a/test/regression/time_derivative_test.jl
+++ b/test/regression/time_derivative_test.jl
@@ -1,71 +1,97 @@
-using OrdinaryDiffEq, Test
+using OrdinaryDiffEq, StaticArrays, Test
 
 function time_derivative(du,u,p,t)
   du[1] = -t
 end
+function time_derivative_static(u,p,t)
+  SVector(-t)
+end
 function time_derivative_analytic(u0,p,t)
   u0 .- t.^2 ./ 2
 end
-
-ff_time_derivative = ODEFunction(time_derivative,
-                                 analytic=time_derivative_analytic)
-u0 = [1.0]
-tspan = (0.0,1.0)
-prob = ODEProblem(ff_time_derivative,u0,tspan)
-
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Was this supposed to have some `@test_broken`s?

FWIW, the real issue comes down to the fact that FiniteDiff really just needs a separate dispatch to treat StaticArrays fully statically.

andreasnoack:
The idea was to write the tests first and then restart them once the upstream bug has been fixed. I.e. once we have a new FiniteDiff release that includes https://github.com/JuliaDiff/FiniteDiff.jl.

I forgot to mention that I had to adjust the test tolerances when switching to FD. That is expected but for some of the solvers, that adjustment was pretty large. E.g. for `Rodas5` I had to use a factor of a billion
```
@show Rodas5
    sol = solve(prob,Rodas5(autodiff=_autodiff),reltol=1e-9,abstol=1e-9)
    @test sol.errors[:final] < 1e-10*1_000_000_000^(!_autodiff)
```
If the solver is so sensitive to FD, I'm wondering if allowing the option makes sense at all.

ChrisRackauckas:
It's not generally sensitive, it's just an adaptivity worst case scenario because it's a numerical best case scenario of a test case.

PR #1225, opened 2020-07-31
Use AD in initialization when the DAE algorithm uses AD

YingboMa:


src/initialize_dae.jl
```diff
@@ -136,13 +136,14 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
     nlequation! = function (out,u)
       update_coefficients!(M,u,p,t)
       #M * (u-u0)/dt - f(u,p,t)
-      @. tmp = (u - u0)/dt
+      tmp = @. (u - u0)/dt
```
ChrisRackauckas, line 139: why?

ChrisRackauckas: oh, the dual numbers. Can we make a DualCache here?

src/initialize_dae.jl
```diff
@@ -287,12 +288,15 @@ function _initialize_dae!(integrator, prob::ODEProblem,
 	alg_u = @view u[algebraic_vars]
 
 	nlequation = (out, x) -> begin
-		alg_u .= x
-		f(tmp, u, p, t)
-		out .= @view tmp[algebraic_eqs]
+    uu = similar(x, axes(integrator.u)); copyto!(uu, integrator.u)
```
ChrisRackauckas, line 291: whitespace is odd here.

YingboMa: This file is all using tabs for some reason. I used spaces.

src/initialize_dae.jl
YingboMa: I think DualCache is not compatible with arbitrary array length.

PR #1226, opened 2020-08-01
LieRK4

Biswajitghosh98:
image
Bumped up atol to 0.5 for this case, as convergence order was getting close to 4.5 for `dts` in `1 ./2 .^(10:-1:1)`

@ChrisRackauckas Please have a look


ChrisRackauckas:
it's the final dts that's an issue. It gets below 1e-12 and starts hitting floating point errors. Just remove the largest and it should be better behaved.

Biswajitghosh98:
@ChrisRackauckas I'm still getting order close to 4.9 when i used dts in range `1 ./2 .^(7:-1:1)`

image

I think it's probably superconvergence, like you said when there was a similar issue in MagnusAdapt4

ChrisRackauckas:
yes, that seems to be displaying superconvergence to order 5 (it has to be an integer for ODEs). So now just bump the order.

PR #1231, opened 2020-08-04
[WIP] Add duprev, update du and user provided jacobians for DAEAlgorithms

kanav99:
Need to change for DABDF2 too

### DIFF
--- a/src/integrators/integrator_utils.jl
+++ b/src/integrators/integrator_utils.jl
@@ -286,8 +286,14 @@ function apply_step!(integrator)
   end
   if isinplace(integrator.sol.prob)
     recursivecopy!(integrator.uprev,integrator.u)
+    if integrator.alg isa DAEAlgorithm
+      recursivecopy!(integrator.duprev, integrator.du)
+    end
   else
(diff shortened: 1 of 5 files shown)


kanav99:
@ChrisRackauckas this PR adds support for custom jacobians, like we have for IDA, but there's an issue on master which causes dtmin for some reason, I remember last time I worked here it used to work, I'd like to fix it in this PR.

ChrisRackauckas:
v1.5 seems to have caused huge test issues too, so someone really needs to dig into that.

src/integrators/integrator_utils.jl
```diff
@@ -286,8 +286,14 @@ function apply_step!(integrator)
   end
   if isinplace(integrator.sol.prob)
     recursivecopy!(integrator.uprev,integrator.u)
+    if integrator.alg isa DAEAlgorithm
+      recursivecopy!(integrator.duprev, integrator.du)
```
ChrisRackauckas, line 290: will need to figure out the FSAL handling

PR #1235, opened 2020-08-07
Fix tests in ImplicitEulerExtrapolation

utkarsh530:
Hi, I was checking implementation changes in ImplicitEulerExtrapolation, the current master tests were failing https://gitlab.com/JuliaGPU/OrdinaryDiffEq.jl/-/jobs/676218514. I have updated them. I also reflected the changes in out-of-place version too.
@ChrisRackauckas @kanav99 please have a look.

### DIFF
--- a/src/perform_step/extrapolation_perform_step.jl
+++ b/src/perform_step/extrapolation_perform_step.jl
@@ -414,14 +414,27 @@ end
 
 function perform_step!(integrator,cache::ImplicitEulerExtrapolationConstantCache,repeat_step=false)
   @unpack t,dt,uprev,u,f,p = integrator
-  @unpack dtpropose, T, cur_order, work, A, tf, uf = cache
+  @unpack dtpropose, T, n_curr, work, A, tf, uf = cache
   @unpack sequence,stage_number = cache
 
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
That looks reasonable. Are the GPU tests not running?

utkarsh530:
It won't run on PRs created on other forks; some Gitlab issue was there as mentioned by you. So yeah I had to checked afterwards in the master and tests was failing.

ChrisRackauckas:
I see. We might want to get a version of the extrapolation methods running on Travis anyways: it doesn't all have to be multithreaded.

utkarsh530:
Yeah, that would be a good idea for other contributors as well.

utkarsh530:
@ChrisRackauckas I have added ODE Extrapolation tests in Travis. Let me if there are any other changes to be made before merging.

ChrisRackauckas:
https://travis-ci.org/github/SciML/OrdinaryDiffEq.jl/jobs/716290143#L1007 looks like they don't pass.

utkarsh530:
Yeah, a typo was there, my bad.

PR #1236, opened 2020-08-09
similar->zero

ChrisRackauckas:
Fix most common occurrences of uninitialized memory. Fixes https://github.com/SciML/DifferentialEquations.jl/issues/636


src/caches/low_order_rk_caches.jl
```diff
@@ -353,9 +353,9 @@ function alg_cache(alg::Tsit5,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUni
   k5 = zero(rate_prototype)
   k6 = zero(rate_prototype)
   k7 = zero(rate_prototype)
-  utilde = similar(u)
+  utilde = zero(u)
   atmp = similar(u,uEltypeNoUnits)
```
ChrisRackauckas, line 357: Can't do this one because `zero(x,T)` isn't a thing, but `atmp` is always fully initialized since it doesn't hit the user `f` which makes that issue completed.

PR #1237, opened 2020-08-10
Fix Julia 1.5 Tests

kanav99:
cc: @utkarsh530 @ChrisRackauckas


.travis.yml
```diff
@@ -8,7 +8,7 @@ os:
   - linux
 #  - osx
 julia:
-  - 1.4
+  - 1.5
```
devmotion, line 11: I guess we might want to change (back?) to
```suggestion
  - 1
```
such that always the stable Julia version is tested automatically.

kanav99: We can have a separate testing for this, for this PR, I am aiming to see what issues we have in Julia 1.5 tests. See the travis tests, they are all red

devmotion: Travis will use Julia 1.5 when you use `- 1`.

kanav99: Ohh Travis follows semver, then we should do it, you are right

utkarsh530:
Probably splitting of AlgConvergence_I is required too? https://travis-ci.org/github/SciML/OrdinaryDiffEq.jl/jobs/716537488

kanav99:
Regression is real, events tests take 35 minutes on travis.

ChrisRackauckas:
Most of the issue was just because coverage now takes longer to compute.

PR #1240, opened 2020-08-11
CG3 [WIP]

Biswajitghosh98:
TODO : 
- [x] Possibly using the incorrect dexp mapping is what's causing convergence failure. Look for MATLAB implementations or try different maps


ChrisRackauckas:
what's dexp?

Biswajitghosh98:
@ChrisRackauckas Derivative of the exponential

ChrisRackauckas:
That's not just the exponential?

Biswajitghosh98:
@ChrisRackauckas I'm sorry I meant Inverse of the exponential map, which is :
image

And then it's replaced by the appropriate lie polynomials, for example, in RKMK4 : 
image

For CG3, I have the Butcher Tableau, just using the correct ki's would do the job

Biswajitghosh98:
@ChrisRackauckas as of now, this is the convergence plot, I'm constantly getting values close to 2.2 for variable range of dts. I'll try to fix it asap 
image

test/algconvergence/linear_method_tests.jl
```diff
@@ -58,6 +58,15 @@ test_setup = Dict(:alg=>Vern9(),:reltol=>1e-14,:abstol=>1e-14)
 sim = analyticless_test_convergence(dts,prob,CG2(),test_setup)
 @test sim.𝒪est[:l2] ≈ 2 atol=0.2
 
+A = DiffEqArrayOperator(ones(2,2),update_func=update_func)
```
ChrisRackauckas, line 61: since this can in theory handle time and state, I think it should get a new test equation

Issue #1238, opened 2020-08-11
Non-concrete array partition in-place interpolation fails due to element type calculation


ibulu:
this works:
sol_univ = solve(prob_univ, Tsit5(),abstol = 1e-8, reltol = 1e-6);

this leads to an error:
sol_univ = solve(prob_univ, Tsit5(),abstol = 1e-8, reltol = 1e-6;saveat=tsteps);

MethodError: Cannot `convert` an object of type 
  ArrayPartition{Float64{},Tuple{Array{Float64{},1},Array{Float64,1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64,1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1}}} to an object of type 
  ArrayPartition{Float64{},Tuple{Array{Float64{},1},Array{Float32,1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float32,1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1},Array{Float64{},1}}}
Closest candidates are:
  convert(::Type{T}, !Matched::T) where T<:AbstractArray at abstractarray.jl:14
  convert(::Type{T}, !Matched::Factorization) where T<:AbstractArray at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/factorization.jl:55
  convert(::Type{T}, !Matched::T) where T at essentials.jl:171
  ...

ibulu:
a minimal example to reproduce the issue:

```
using Unitful, RecursiveArrayTools, DiffEqBase, OrdinaryDiffEq
using LinearAlgebra

r0 = [1131.340, -2282.343, 6672.423]u"km"
v0 = [-5.64305, 4.30333, 2.42879]u"km/s"
Δt = 86400.0*365u"s"
μ = 398600.4418u"km^3/s^2"
rv0 = ArrayPartition(r0,v0)

function f(dy, y, μ, t)
    r = norm(y.x[1])
    dy.x[1] .= y.x[2]
[...]

PR #1241, opened 2020-08-12
more robust interpolation eltype choice

ChrisRackauckas:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1238

### DIFF
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -365,21 +365,7 @@ function ode_interpolant(Θ,dt,y₀,y₁,k,cache::OrdinaryDiffEqMutableCache,idx
     out = oneunit(Θ) .* y₁
     _ode_interpolant!(out,Θ,dt,y₀,y₁,k,cache,idxs,T)
   else
-    # determine output type
-    # required for calculation of time derivatives with autodifferentiation
-    oneunit_Θ = oneunit(Θ)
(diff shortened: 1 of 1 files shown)


src/dense/generic_dense.jl
```diff
@@ -361,6 +361,9 @@ function ode_interpolant(Θ,dt,y₀,y₁,k,cache::OrdinaryDiffEqMutableCache,idx
   if typeof(idxs) <: Number || typeof(y₀) <: Union{Number,SArray}
     # typeof(y₀) can be these if saveidxs gives a single value
     _ode_interpolant(Θ,dt,y₀,y₁,k,cache,idxs,T)
+  elseif typeof(idxs) <: Nothing
+    out = oneunit(Θ) .* y₁
+    _ode_interpolant!(out,Θ,dt,y₀,y₁,k,cache,idxs,T)
   else
     # determine output type
```
devmotion, line 368: Couldn't we just change the logic in this branch instead of adding a new one? At least it seems that the `if typeof(idxs) <: Nothing` checks in this third branch could be removed.

ChrisRackauckas: I was triggering tests to see if this would work, but yes y1[idxs] would be a simpler way to do the other branch. Looks like it's passing now

PR #1257, opened 2020-08-29
Unnecessary type promotion causes issues with Callbacks  + Tracker

avik-pal:


devmotion:
Downstream test errors should be fixed by https://github.com/SciML/ODEInterfaceDiffEq.jl/pull/38.

ChrisRackauckas:
Can we get a test with Callbacks + Tracker? That is an important one and seems to have previously dodged our tests.

devmotion:
> Downstream test errors should be fixed by SciML/ODEInterfaceDiffEq.jl#38.

Fixed the DataStructures incompatibility of OrdinaryDiffEq but tests still fail since GPU packages are not compatible with DataStructures.

devmotion:
More precisely, the latest releases of GPUCompiler and CUDA are compatible with DataStructures 0.18 but the latest release of CUDA is not compatible with the latest release of GPUCompiler. It's fixed in the master branch, so CUDA just needs a new release.

avik-pal:
> Can we get a test with Callbacks + Tracker? That is an important one and seems to have previously dodged our tests.

Sure. I will add that.

DhairyaLGandhi:
Can you check for the DataStructures.jl compat issue now? Updating to GPUCompiler@0.6.1 should unblock things.

Issue #325, opened 2018-04-16
AutoSwitch algorithms are not thread-safe


ChrisRackauckas:
They use the algorithm itself as a cache:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/composite_algs.jl#L1-L12

Instead, `ComposteCache` should get a slot for `switching_cache` that can be any struct, and then these algorithms can put all of their switching behavior cache in there inside of the algortihm. Then we don't need to rely on mutation of the `alg` and it can be repeated across threads (and make it a struct just to be sure).

PR #1258, opened 2020-09-02
Add reinitialization for AutoSwitch

YingboMa:
Fix https://discourse.julialang.org/t/ode-autotsit5-rodas5-switching-bug-depends-from-tolerance-history/45288

### DIFF
--- a/src/composite_algs.jl
+++ b/src/composite_algs.jl
@@ -1,5 +1,4 @@
-# AutoSwitch needs to be reinitialized in `init`
-mutable struct AutoSwitch{nAlg,sAlg,tolType,T}
+mutable struct AutoSwitchCache{nAlg,sAlg,tolType,T}
   count::Int
   nonstiffalg::nAlg
   stiffalg::sAlg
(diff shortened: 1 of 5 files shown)


src/solve.jl
```diff
@@ -102,7 +102,10 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
                                      prob.f.mass_matrix != I &&
                                      !(typeof(prob.f.mass_matrix)<:Tuple) &&
                                      ArrayInterface.issingular(prob.f.mass_matrix))
-
+  if alg isa AutoSwitch
+    alg.count = 0
```
ChrisRackauckas, line 106: this is still not thread safe since if multiple threads are working then a separate thread can reset the count. We should make the algorithm type be a `struct` instead of a `mutable struct` to enforce correctness here, and have a `switching_cache` in the integrator for this.

YingboMa: Could we just call `copy` here?

ChrisRackauckas: we could copy the alg, but it would be so much nicer if it had immutable semantics like every other one. I think this and the StochasticDiffEq autoswitch are the only two that are mutable?

YingboMa: Yeah

src/composite_algs.jl
```diff
@@ -1,4 +1,4 @@
-mutable struct AutoSwitch{nAlg,sAlg,tolType,T}
+mutable struct AutoSwitchCache{nAlg,sAlg,tolType,T}
```
andreasnoack, line 1: As mentioned in https://github.com/SciML/OrdinaryDiffEq.jl/issues/325, it would make sense to make this immutable.

andreasnoack: Never mind. I see that this became the cache and that the struct below is immutable. I'll grab some coffee.

ChrisRackauckas: It would be better to bring it to full immutability to guarantee safety, but this is the quickest way to working so it's worth it for now. The plan is to probably have a GSoC to follow up and do more stiffness estimates, in which case we'll pick this back up and probably make it immutable, but at least if it's fully thread-safe already that can wait.

ChrisRackauckas:
It would be nice to have a test of thread-safety though... but that might be hard to cook up here.

PR #1264, opened 2020-09-10
Fix some AD issues with KenCarpX and TRBDF2 integrators

andreasnoack:
I'm not sure which tolerances I should use in the tests so please let me know if I should change them. I can also all more solvers to the list. It would probably be good to test all solvers but there might also be some redundancy.

### DIFF
--- a/test/interface/ad_tests.jl
+++ b/test/interface/ad_tests.jl
@@ -211,17 +211,23 @@ end
 
 
 SOLVERS_FOR_AD = (
-    BS3,
-    Tsit5,
-    KenCarp4, KenCarp47, KenCarp5, KenCarp58,
-    TRBDF2,
-    Rodas4, Rodas5,
-    Rosenbrock23, Rosenbrock32,
-    Vern6, Vern7,
-    )
+    (BS3         , 1e-12),
+    (Tsit5       , 1e-12),
+    (KenCarp4    , 1e-11),
+    (KenCarp47   , 1e-11),
+    (KenCarp5    , 1e-11),
+    (KenCarp58   , 1e-12),
+    (TRBDF2      , 1e-07),
+    (Rodas4      , 1e-12),
+    (Rodas5      , 1e-12),
+    (Rosenbrock23, 1e-10),
+    (Rosenbrock32, 1e-11),
+    (Vern6       , 1e-11),
+    (Vern7       , 1e-11),
+)
 
-@testset "$alg can handle ForwardDiff.Dual in u0 when iip=$iip" for
-    alg in SOLVERS_FOR_AD,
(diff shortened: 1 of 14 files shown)


andreasnoack:
So in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1264/commits/5735598516903b035c2061ddd0fed7be2f68f18d I stripped the dual part off `EEst` since caused problems for the Radau integrators and `internalnorm` already strips off the dual part. This mostly looks like it's working but the change breaks this test https://github.com/SciML/DiffEqCallbacks.jl/blob/master/test/saving_tests.jl#L123-L131 because `internalnorm` returns a `TrackedReal`. What's the right solution here? Should the `internalnorm` return `Float32` instead of a `TrackedReal{Float32}` or should the type of `EEst` change?

andreasnoack:
I just found https://github.com/SciML/DiffEqBase.jl/blob/b5ad65e5f6c9900b9feceb33953e5c23552d69f8/src/tracker.jl#L27 so it's unclear to me why https://github.com/SciML/OrdinaryDiffEq.jl/blob/16c4fe3b5df7efced341b22e6c8566e38520821a/src/perform_step/low_order_rk_perform_step.jl#L599 returns a `TrackedReal`. I'll try to figure out what is going on.

test/interface/ad_tests.jl
```diff
@@ -208,3 +208,61 @@ end
...
+    end
+
+    g = u0 -> begin
+        tspan = (0.0, 1.0)
+        prob = ODEProblem(
+            f,
+            u0,
+            tspan
+        )
+        solve(prob, alg())(last(tspan))[1]
```
ChrisRackauckas, line 241: can you set the solver tolerance to `abstol=1e-14,reltol=1e-14`, and lower the test tolerance to something to match? With analytical solution, that test could probably be to like `1e-12`.

andreasnoack: I've not tightened the tolerances in the tests. I've set the test tolerance as tight as possible for each of the algorithms. Notice that I have to use a pretty large relative error for `TRBDF2`.

andreasnoack:
I've split the PR in two. The part the fixed AD for the Radau algorithms will go into a separate PR since it touches more things and it breaks some tests so it might require some more work to figure out what is going on. The changes left in this PR are quite modest: it uses `constvalue` consistently when creating the tableaus and it fixes the initialization of `ηold` in `NLSolver`.

YingboMa (approved):
LGTM

ChrisRackauckas:
That is odd for TRBDF2, but it is very low order so maybe it's not too weird. Good to take note of.

PR #1265, opened 2020-09-11
unroll composite cache perform_step cal to allow inference

ChrisRackauckas:


YingboMa (commented):
You could also unroll
https://github.com/SciML/OrdinaryDiffEq.jl/blob/c05d56244fb8b8fa1834690393d86f7d91a0a1e9/src/alg_utils.jl#L536

PR #1272, opened 2020-09-18
fix get_du for 2N inplace methods

ranocha:
This fixes
```julia
julia> using OrdinaryDiffEq

julia> ode = ODEProblem((du,u,p,t) -> du .= u, [1.0], (0.0, 1.0))
ODEProblem with uType Array{Float64,1} and tType Float64. In-place: true
timespan: (0.0, 1.0)
u0: [1.0]

julia> integrator = init(ode, Tsit5(), dt=0.1)
t: 0.0
u: 1-element Array{Float64,1}:
 1.0

julia> get_du(integrator) # expected behavior
1-element Array{Float64,1}:
 0.0

julia> integrator = init(ode, CarpenterKennedy2N54(williamson_condition=false), dt=0.1)
t: 0.0
u: 1-element Array{Float64,1}:
 1.0

julia> get_du(integrator) # definitely not expected behavior
ERROR: UndefRefError: access to undefined reference
```

### DIFF
--- a/src/perform_step/low_storage_rk_perform_step.jl
+++ b/src/perform_step/low_storage_rk_perform_step.jl
@@ -38,6 +38,8 @@ function initialize!(integrator,cache::LowStorageRK2NCache)
   integrator.kshortsize = 1
   resize!(integrator.k, integrator.kshortsize)
   integrator.k[1] = k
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
great!

PR #1278, opened 2020-09-25
fix resize! for LowStorageRK2NCache; closes #1277

### LINKED ISSUE #1277: resize! is broken for LowStorageRK2NCache
```julia
julia> using Test, OrdinaryDiffEq, DiffEqCallbacks

julia> function f(du, u, p, t)
           du .= u
           nothing
       end
f (generic function with 1 method)

julia> u0 = ones(2)

julia> condition(u, t, integrator) = t - 1
condition (generic function with 1 method)

julia> function affect!(integrator)
         u = integrator.u
         resize!(integrator, length(u) + 1)
         u .= 1
         nothing
       end
affect! (generic function with 1 method)

julia> callback = ContinuousCallback(condition, affect!);

julia> tspan = (0.0, 2.0)
(0.0, 2.0)

julia> prob = ODEProblem(f, u0, tspan)
ODEProblem with uType Array{Float64,2} and tType Float64. In-place: true
timespan: (0.0, 2.0)
u0: [1.0 1.0; 1.0 1.0]

julia> sol = solve(prob, RK4(), dt=0.125, adaptive=false, callback=callback)

julia> @test all(sol[end] .== sol[end][1])
Test Passed

julia> sol = solve(prob, CarpenterKennedy2N54(williamson_condition=false), dt=0.125, adaptive=false, callback=callback)

julia> @test all(sol[end] .== sol[end][1])
Test Failed at REPL[22]:1
  Expression: all(sol[end] .== (sol[end])[1])
```

The `LowStorageRK2NCache` doesn't seem to be updated correctly. In our application in [Trixi.jl](https://github.com/trixi-framework/Trixi.jl), this results in stochastic errors when using adaptive mesh refinement. I will try to look into it but would be glad about any help.

ranocha:
This fixes #1277 for `DiscreteCallback`s. I had to re-eval the FSAL part, since the 2N methods are not recognized as FSAL methods although they evaluate the RHS for the new step. Why?

### DIFF
--- a/src/perform_step/low_storage_rk_perform_step.jl
+++ b/src/perform_step/low_storage_rk_perform_step.jl
@@ -50,6 +50,13 @@ end
   @unpack k,tmp,williamson_condition = cache
   @unpack A2end,B1,B2end,c2end = cache.tab
 
+  if integrator.u_modified
+    if !get_current_isfsal(integrator.alg, integrator.cache)
+      f(k, u, p, t+dt)
+    end
+    @.. tmp = dt*k
+  end
+
   # u1
   @.. u   = u + B1*tmp
   # other stages
--- a/test/integrators/ode_cache_tests.jl
+++ b/test/integrators/ode_cache_tests.jl
@@ -68,7 +68,14 @@ for alg in broken_CACHE_TEST_ALGS
   @test_broken length(solve(prob,alg,callback=callback,dt=1/2)[end]) > 1
 end
 
+sol = solve(prob,Rodas4(chunk_size=1),callback=callback,dt=1/2)
+@test length(sol[end]) > 1
+sol = solve(prob,Rodas5(chunk_size=1),callback=callback,dt=1/2)
+@test length(sol[end]) > 1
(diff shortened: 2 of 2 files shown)


ChrisRackauckas:
Looks like it's hitting a fallback: https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/alg_utils.jl#L5-L29

ranocha:
Well, I'm sure we had `isfsal(alg::CarpenterKennedy2N54) == true` at some time in the past. But now there is https://github.com/SciML/OrdinaryDiffEq.jl/blob/821fc833e0b3c3721d6e91b666c051450ca55833/src/alg_utils.jl#L17
Anyway, we still need an additional check of `u_modified` to update `tmp` for 2N methods.

ChrisRackauckas:
I think @kanav99 did something with that due to Williamson condition and ensuring low storage? I'll merge and we can discuss this more.

ranocha:
Thanks!

PR #1283, opened 2020-10-01
add SSPRK43, cf. #493

ranocha:
This implements the most useful embedded SSPRK method of [Conde, Fekete, Shadid (2018)](https://arxiv.org/abs/1806.08693), cf. #493. Although #493 is "good for beginners", I want to use this method for my research and need it now.


ranocha:
I will include these commits in #1287 to make it easier for me to change all SSPRK methods at once, including this new one.

PR #1287, opened 2020-10-02
SSPRK methods: non-FSAL and p as argument in limiters

### LINKED ISSUE #1285: stage_limiter! and step_limiter! for SSPRK methods should get p as argument
Right now, the `stage_limiter!` and `step_limiter!` of SSPRK methods are called as
```julia
  stage_limiter!(u, f, t+dt)
  step_limiter!(u, f, t+dt)
```
It would be far more convenient if they also received the parameters `p` as additional argument, e.g.
```julia
  stage_limiter!(u, f, p, t+dt)
  step_limiter!(u, f, p, t+dt)
```
If I remember correctly, these limiters were introduced before parameters were handled explicitly in the DiffEq ecosystem. Hence, it seems natural to pass parameters explicitly now.

Technically, this would be a breaking change. However, I doubt many people use this infrastructure right now. However, we want to make use of it in Trixi.jl.

ranocha:
This fixes #1285. Additionally, I made the non-FSAL SSPRK methods really non-FSAL. Thus, they will be more efficient if callbacks modify `u`, as we will do in Trixi.jl using adaptive mesh refinement.


ranocha:
Note to self: Remember to do the same for `SSPRK43`, depending on whether #1283 or this PR is merged first.

src/caches/ssprk_caches.jl
```diff
@@ -321,8 +321,8 @@ end
   u::uType
   uprev::uType
   k::rateType
-  tmp::uType
   fsalfirst::rateType
+  tmp::uType
```
ChrisRackauckas, line 325: You can also override https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_interface.jl#L82 on these methods. `k` might be a suitable cache?

ranocha: But that would work only if `uType === rateType`, wouldn't it? That's definitely okay in our case but won't work if there are units or something similar. However, since the 2N methods already do this and SSP methods are similarly specialized and not really general-purpose methods, that should be okay. I'll do that.

ranocha: Done in 66c83ec.

ChrisRackauckas (approved):

PR #1289, opened 2020-10-12
Import linear and hermite interpolations explicitly

cqql:
DiffEqBase does not export these types and invoking this function without `interp` would produce an unknown identifier error.

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -50,6 +50,8 @@ module OrdinaryDiffEq
 
   import DiffEqBase: calculate_residuals, calculate_residuals!, unwrap_cache, @tight_loop_macros, islinear, timedepentdtmin
 
+  import DiffEqBase: LinearInterpolation, HermiteInterpolation
+
   import SparseDiffTools
   import SparseDiffTools: matrix_colors, forwarddiff_color_jacobian!,
                           forwarddiff_color_jacobian, ForwardColorJacCache,


cqql:
Is this fail really related to this change? Looks unrelated to me.

ChrisRackauckas:
Yes, I think the breaking change might have come from https://github.com/SciML/ArrayInterface.jl/pull/75 so I'll follow up on that. It is indeed unrelated.

That said, I still don't think this PR is needed.

cqql:
They are used in
https://github.com/SciML/OrdinaryDiffEq.jl/blob/f58973b728356e6646feb68191f2ee6e25446126/src/composite_solution.jl#L27
Internally, `interp` is always passed explicitly but I had a setup where some other library from the DifferentialEquations ecosystem was calling `DiffEqBase.build_solution` triggering OrdinaryDiffEq's implementation but without providing `interp`. Then the default value was evaluated and the error occurred.

ChrisRackauckas:
The only call to it passes something into that kwarg (https://github.com/SciML/OrdinaryDiffEq.jl/blob/f58973b728356e6646feb68191f2ee6e25446126/src/solve.jl#L335-L340). We should probably just delete that last portion and make it `interp, ` so it's a required kwarg for safety.

ChrisRackauckas:
>but I had a setup where some other library from the DifferentialEquations ecosystem was calling DiffEqBase.build_solution triggering OrdinaryDiffEq's implementation but without providing interp

Interesting, what's the case for this?

cqql:
I don't remember right now. I fixed it in my local dev'd version a few days ago and since then it has been working. Unfortunately, I forgot where it occurred exactly. I will revert my version to master and report back when I see it again.

ChrisRackauckas:
I'm going to close because I'm very sure this is in the wrong direction, but I'd be happy to follow up with how you got here though. My guess is that you were doing AD and tried to interpolate, but that will set a special interpolation that gives an error message saying that interpolation is incompatible with certain adjoint choices (and there's fundamental reasons for that).

Issue #1292, opened 2020-10-25
Post-stage callbacks


ranocha:
It would be very nice to have (pre- and) post-stage callbacks, similar to [PreStage](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/TS/TSSetPreStage.html) and [PostStage](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/TS/TSSetPostStage.html) in PETSc

### Motivation

I'm rewriting [Trixi.jl](https://github.com/trixi-framework/Trixi.jl), our tree-based numerical simulation framework for hyperbolic PDEs, to be more modular and use the SciML ecosystem for ODE solvers. There, we would like to use positivity-preserving limiters (and multi-physics coupling), which can be implemented conveniently using such post-stage callbacks, see https://github.com/trixi-framework/Trixi.jl/issues/235.

### Existing infrastructure in OrdinaryDiffEq

Right now, we have `stage_limiter!` for SSPRK methods and in-place problems: 
https://github.com/SciML/OrdinaryDiffEq.jl/blob/0c2048a502101647ac35faabd80da8a5645beac7/src/perform_step/ssprk_perform_step.jl#L43
These are basically what I have in mind for post-stage callbacks: Some function that is called after a stage value of an (explicit) Runge-Kutta method has been computed but before the RHS function is evaluated at that stage; the callback is allowed to modify the recently computed stage value. For example, search for `TSPostStage` in https://www.mcs.anl.gov/petsc/petsc-current/src/ts/impls/explicit/rk/rk.c.html.

For OrdinaryDiffEq, I would propose to use the same calling signature as for `stage_limiter!(stage_u, f, p, stage_t)`.

### Questions

Would such a feature be okay in general? How shall we implement it? Similar to `stage_limiters!` in SSPRK methods, i.e. by adding appropriate functions as optional parameters to the algorithms, defaulting to `trivial_limiter!` which does nothing? Then, we could add these additional features algorithm by algorithm, starting by low-storage methods that are most important for us right now.

ChrisRackauckas:
>Would such a feature be okay in general?

Yes, it would be a great feature to have.

>Similar to stage_limiters! in SSPRK methods, i.e. by adding appropriate functions as optional parameters to the algorithms, defaulting to trivial_limiter! which does nothing?

Definitely agree this is the right way to do it.

[...]

PR #1293, opened 2020-10-30
implement post-stage and post-step limiters for 2N methods

ranocha:
As discussed in #1292. I've checked that this works for us in https://github.com/trixi-framework/Trixi.jl/pull/274.


ranocha:
Is this okay or would you like to use a different approach, @ChrisRackauckas?

ChrisRackauckas:
Thanks for the bump. I had just completely missed this PR. It looks great. Exactly what I was thinking.

ranocha:
Thanks! And thanks for creating a new release such that we can use this :+1:

PR #1296, opened 2020-11-06
fix type inference issues

mohamed82008:
This PR fixes some type inference issues in OrdinaryDiffEq when using a composite algorithm. I will add some inference tests tomorrow. With this PR, I have seen a consistent 15-18% speed-up in my use cases.

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -49,12 +49,12 @@ get_current_isfsal(alg::CompositeAlgorithm, cache) = isfsal(alg.algs[cache.curre
 issplit(alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm}) = false
 issplit(alg::SplitAlgorithms) = true
 
-@generated function _composite_beta1_default(algs::T, current, ::Type{QT}) where {T <: Tuple, QT}
+@generated function _composite_beta1_default(algs::T, current, ::Type{QT}, beta2) where {T <: Tuple, QT}
   expr = Expr(:block)
   for i in 1:length(T.types)
     push!(expr.args, quote
       if current == $i
-        return QT(beta1_default(algs[$i], beta2_default(algs[$i])))
+        return QT(beta1_default(algs[$i], beta2))
       end
     end)
   end
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -49,11 +49,35 @@ get_current_isfsal(alg::CompositeAlgorithm, cache) = isfsal(alg.algs[cache.curre
 issplit(alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm}) = false
 issplit(alg::SplitAlgorithms) = true
 
+@generated function _composite_beta1_default(algs::T, current, ::Type{QT}) where {T <: Tuple, QT}
+  expr = Expr(:block)
+  for i in 1:length(T.types)
+    push!(expr.args, quote
+      if current == $i
+        return QT(beta1_default(algs[$i], beta2_default(algs[1])))
+      end
+    end)
+  end
+  return expr
+end
+@generated function _composite_beta2_default(algs::T, current, ::Type{QT}) where {T <: Tuple, QT}
+  expr = Expr(:block)
+  for i in 1:length(T.types)
+    push!(expr.args, quote
+      if current == $i
+        return QT(beta2_default(algs[1]))
+      end
+    end)
+  end
+  return expr
+end
+
+  
 fsal_typeof(alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm},rate_prototype) = typeof(rate_prototype)
 fsal_typeof(alg::ETD2,rate_prototype) = ETD2Fsal{typeof(rate_prototype)}
 function fsal_typeof(alg::CompositeAlgorithm,rate_prototype)
-  fsal = unique(map(x->fsal_typeof(x,rate_prototype), alg.algs))
-  @assert length(fsal) == 1 "`fsal_typeof` must be consistent"
+  fsal = map(x->fsal_typeof(x,rate_prototype), alg.algs)
+  @assert length(unique(fsal)) == 1 "`fsal_typeof` must be consistent"
   return fsal[1]
 end
 
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -54,7 +54,7 @@ issplit(alg::SplitAlgorithms) = true
   for i in 1:length(T.types)
     push!(expr.args, quote
       if current == $i
-        return QT(beta1_default(algs[$i], beta2_default(algs[1])))
+        return QT(beta1_default(algs[$i], beta2_default(algs[$i])))
       end
     end)
   end
@@ -65,7 +65,7 @@ end
   for i in 1:length(T.types)
     push!(expr.args, quote
       if current == $i
-        return QT(beta2_default(algs[1]))
+        return QT(beta2_default(algs[$i]))
       end
     end)
   end
--- a/src/caches/basic_caches.jl
+++ b/src/caches/basic_caches.jl
@@ -10,18 +10,16 @@ mutable struct CompositeCache{T,F} <: OrdinaryDiffEqCache
   current::Int
 end
 
-function alg_cache(alg::CompositeAlgorithm,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Val{false})
-  caches = map(alg.algs) do x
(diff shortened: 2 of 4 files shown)


src/solve.jl
```diff
@@ -7,7 +7,7 @@ function DiffEqBase.__solve(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase
 end
 
 function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.AbstractDAEProblem},
-                           alg::Union{OrdinaryDiffEqAlgorithm,DAEAlgorithm},
```
ChrisRackauckas, line 10: why the name change?

mohamed82008: Because we are defining `alg` below in an if statement. The type of `alg` was not inferred properly previously.

devmotion: Maybe one should rather rename `alg` defined below in the function body? I would assume that this requires fewer renamings and leads to a more "natural" function signature?

src/caches/basic_caches.jl
```diff
@@ -10,18 +10,16 @@ mutable struct CompositeCache{T,F} <: OrdinaryDiffEqCache
...
-  caches = map(alg.algs) do x
-    alg_cache(x,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,Val(false))
-  end
+function alg_cache(alg::CompositeAlgorithm,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Val{V}) where {V}
+  caches = __alg_cache(alg.algs,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,Val(V))
   CompositeCache(caches,alg.choice_function,1)
 end
 
-function alg_cache(alg::CompositeAlgorithm,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Val{true})
-  caches = map(alg.algs) do x
```
devmotion, line 21: Maybe one just has to use the `let`-block trick here?

src/caches/basic_caches.jl
```diff
@@ -10,18 +10,16 @@ mutable struct CompositeCache{T,F} <: OrdinaryDiffEqCache
   current::Int
 end
 
-function alg_cache(alg::CompositeAlgorithm,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol,p,calck,::Val{false})
-  caches = map(alg.algs) do x
```
devmotion, line 14: Same here, is it sufficient to use a `let` block?

src/solve.jl
```diff
@@ -298,8 +298,8 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
 
   if typeof(alg) <: OrdinaryDiffEqCompositeAlgorithm
     id = CompositeInterpolationData(f,timeseries,ts,ks,alg_choice,dense,cache)
-    beta2 === nothing && ( beta2=beta2_default(alg.algs[cache.current]) )
-    beta1 === nothing && ( beta1=beta1_default(alg.algs[cache.current],beta2) )
+    beta2 === nothing && ( beta2=_composite_beta2_default(alg.algs, cache.current, QT) )
+    beta1 === nothing && ( beta1=_composite_beta1_default(alg.algs, cache.current, QT) )
```
devmotion, line 302: Would it be sufficient to use
```suggestion
    beta2 === nothing && ( beta2=QT(beta2_default(alg.algs[cache.current])) )
    beta1 === nothing && ( beta1=QT(beta1_default(alg.algs[cache.current],beta2)) )
```
for type inference of `beta1` and `beta2`?

src/solve.jl
mohamed82008: I don't mind either. What we have here actually does less renaming. The function is huge!

src/caches/basic_caches.jl
mohamed82008: No didn't work. My hypothesis is that there are too many variables closed-over and the Julia compiler gives up.

src/caches/basic_caches.jl
mohamed82008: No it wasn't.

src/solve.jl
mohamed82008: No because the type of `alg.algs[cache.current]` is not known in advance because `alg.algs` is a tuple of different element types. So I did the whole unrolling thing to make sure Julia assigns `alg.algs[1]` to a different variable from `alg.algs[2]`. Your suggestion creates a type instability in the intermediate variable that `alg.algs[cache.current]` is assigned to.

src/solve.jl
devmotion: Renaming `alg` in the `if` block below to `_alg` requires 20 renamings, that's fewer than occurrences of `_alg` in this PR.

mohamed82008: Ok I recounted, and they are very close but your suggestion has a few less renames. I will change it to have a "natural" function signature.

src/solve.jl
devmotion: But that shouldn't matter due to the function barrier, should it? I just saw that you called the QT constructor in the generated function but that could be done without generated function as well.

mohamed82008: it shouldn't matter that much but it's still a dynamic dispatch

src/alg_utils.jl
```diff
@@ -49,11 +49,35 @@ get_current_isfsal(alg::CompositeAlgorithm, cache) = isfsal(alg.algs[cache.curre
...
+    end)
+  end
+  return expr
+end
+@generated function _composite_beta2_default(algs::T, current, ::Type{QT}) where {T <: Tuple, QT}
+  expr = Expr(:block)
+  for i in 1:length(T.types)
+    push!(expr.args, quote
+      if current == $i
+        return QT(beta2_default(algs[1]))
```
devmotion, line 68: This seems incorrect:
```suggestion
        return QT(beta2_default(algs[$i]))
```

src/solve.jl
```diff
@@ -291,19 +291,19 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
...
+    cache = alg_cache(_alg,u,rate_prototype,uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits,uprev,uprev2,f,t,dt,reltol_internal,p,calck,Val(isinplace(prob)))
   end
 
-  if typeof(alg) <: OrdinaryDiffEqCompositeAlgorithm
+  if typeof(_alg) <: OrdinaryDiffEqCompositeAlgorithm
     id = CompositeInterpolationData(f,timeseries,ts,ks,alg_choice,dense,cache)
-    beta2 === nothing && ( beta2=beta2_default(alg.algs[cache.current]) )
-    beta1 === nothing && ( beta1=beta1_default(alg.algs[cache.current],beta2) )
+    beta2 === nothing && ( beta2=_composite_beta2_default(_alg.algs, cache.current, QT) )
+    beta1 === nothing && ( beta1=_composite_beta1_default(_alg.algs, cache.current, QT) )
```
devmotion, line 302: Why not pass `beta2` here?

src/alg_utils.jl
mohamed82008: oh dear :facepalm:

src/solve.jl
mohamed82008: I could but I was worried about the `nothing` case but I think it should be fine.

test/interface/composite_algorithm_test.jl
```diff
@@ -20,3 +20,4 @@ solve!(integrator2)
 @test integrator1.sol.t == integrator2.sol.t
 
 sol = solve(prob,alg_switch)
+@inferred OrdinaryDiffEq.DiffEqBase.__solve(prob,alg_switch)
```
devmotion, line 23: If possible (maybe it's still broken?), it might be better to test
```suggestion
@inferred DiffEqBase.__init(prob, alg_switch)
```
to check the modified function directly and ensure that not only the solution but also the integrator type can be inferred.

mohamed82008:
I think we are good here if there are no further comments and the tests pass.

devmotion:
The tests didn't pass, did they? Has anyone checked locally if they succeed and type inference works?

ChrisRackauckas:
The tests looked like they passed.

devmotion:
But not the final commit it seems? At least the tests https://travis-ci.org/github/SciML/OrdinaryDiffEq.jl/builds/742271865 of the final commit are still pending.

ChrisRackauckas:
Good to know. It passed on master though so we're good.

PR #1297, opened 2020-11-09
Make composite ode_interpolant type stable

mohamed82008:
This PR removes a dynamic dispatch in the composite branch of `ode_interpolant`.


ChrisRackauckas (approved):

devmotion:
Would be good to add some tests 🙂

ChrisRackauckas:
agreed. Let's get some inference tests on the interpolations.

Issue #1298, opened 2020-11-16
Progress bar assumes that initial time is zero


eschnett:
The progress bar assumes that the initial time is zero when it calculates the progress fraction:
```Julia
    progress=integrator.t/integrator.sol.prob.tspan[2]
```
(line 225 of `integrator_utils.jl`). It does not take`integrator.sol.prob.tspan[1]` into account.

PR #1299, opened 2020-11-16
Take initial time into account when calculating progress fraction

eschnett:
Closes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1298.


ChrisRackauckas:
awesome thanks!

PR #1302, opened 2020-11-24
Fix naming ambiguity between algorithms of the same author

grimmmyshini:
Should change the name of `HSLDDRK64` to `SHLDDRK64` to avoid naming ambiguity between the other methods from the same paper and the same author. Moved the S before to match the conventions of having the first author have the first initial. 

**Note**: Not sure how useful this PR is, the naming just confused me a bit and might be misleading so I thought I'd put in a quick PR here.


ChrisRackauckas:
This is a good point. But, we should probably add a deprecation since this is technically breaking.

PR #1307, opened 2020-12-04
Change initialization of heaps and add interface for DelayDiffEq

devmotion:
This PR adds separate functions for the initialization of the `tstops`, `saveat`, and `d_discontinuities` heaps which makes it easier to initialize these heaps with different types in DelayDiffEq. Moreover, it adds an interface for retrieving the next time stop and discontinuity instead of operating with `integrator.opts.tstops` and `integrator.opts.d_discontinuities` directly. This allows to separate user-provided time stops and discontinuities and propagated discontinuities and their corresponding time stops in DelayDiffEq, which allows to fix https://github.com/SciML/DelayDiffEq.jl/issues/188.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.group == 'Downstream' }}
     strategy:
       matrix:
         group:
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
Yeah this is one of those, not breaking but DelayDiffEq uses the internals kinds of things. Github Actions is cancelling all of the CI when it fails though. Can we disable that have to full check? I think it looks good to go, but this is a change I'll want CI to run for.

PR #1309, opened 2020-12-16
Fix setting of rate_prototype in __init for ArrayPartition types

### LINKED ISSUE #1308: Vector of StaticArrays inside ArrayPartition breaks the type of rate_prototype
I have noticed, that if the state vector `u` is an `ArrayPartition` which contains `Vector` of `StaticArray`s, the computed type of the derivative vector `du` comes out wrong. I believe the problem resides in this line: https://github.com/SciML/OrdinaryDiffEq.jl/blob/69484227aa2b069d42f2c269339c84e34bbfa127/src/solve.jl#L184
Consider the following example:
```
using StaticArrays, RecursiveArrayTools

tType = Float64
u = ArrayPartition((zeros(5), zeros(SVector{3,Float64}, 5)))
typeof.(oneunit.(recursive_bottom_eltype.(u.x))./oneunit(tType))
```
gives you
>(Float64, Float64)

As a result, when `rate_prototype` is created, we get `Vector{Float64}` instead of `Vector{SVector{3,Float64}}` for the second part of `ArrayPartition`.

Since the line https://github.com/SciML/OrdinaryDiffEq.jl/blob/69484227aa2b069d42f2c269339c84e34bbfa127/src/solve.jl#L184 is supposed to correspond to the case without units, it actually looks completely out of place. The fix seems to be just to remove this line and calculate `rate_prototype` the same way it is done when `u` is not `ArrayPartition` in line https://github.com/SciML/OrdinaryDiffEq.jl/blob/69484227aa2b069d42f2c269339c84e34bbfa127/src/solve.jl#L182

jdeldre:
I removed the negation on the conditional `typeof(u) <: ArrayPartition` to enable use of `recursivecopy` to set `rate_prototype` for `ArrayPartition` type `u`.

Fixes #1308

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -178,11 +178,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
   # dtmin is all abs => does not care about sign already.
 
   if !isdae && isinplace(prob) && typeof(u) <: AbstractArray && eltype(u) <: Number && uBottomEltypeNoUnits == uBottomEltype && tType == tTypeNoUnits # Could this be more efficient for other arrays?
-    if !(typeof(u) <: ArrayPartition) || isarrays(u)
-      rate_prototype = recursivecopy(u)
-    else # this is for ArrayPartition that is unitless but holds non-arrays
-      rate_prototype = similar(u, typeof.(oneunit.(recursive_bottom_eltype.(u.x))./oneunit(tType))...)
-    end
+    rate_prototype = recursivecopy(u)
   elseif prob isa DAEProblem
     rate_prototype = prob.du0
   else
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -178,9 +178,9 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
   # dtmin is all abs => does not care about sign already.
 
   if !isdae && isinplace(prob) && typeof(u) <: AbstractArray && eltype(u) <: Number && uBottomEltypeNoUnits == uBottomEltype && tType == tTypeNoUnits # Could this be more efficient for other arrays?
-    if typeof(u) <: ArrayPartition
+    if !(typeof(u) <: ArrayPartition) || isarrays(u)
       rate_prototype = recursivecopy(u)
-    else
+    else # this is for ArrayPartition that is unitless but holds non-arrays
       rate_prototype = similar(u, typeof.(oneunit.(recursive_bottom_eltype.(u.x))./oneunit(tType))...)
     end
   elseif prob isa DAEProblem
@@ -470,6 +470,7 @@ function DiffEqBase.solve!(integrator::ODEIntegrator)
 end
 
 # Helpers
+
 
 function handle_dt!(integrator)
   if iszero(integrator.dt) && integrator.opts.adaptive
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Looks like that's a real test failure.

src/solve.jl
```diff
@@ -178,7 +178,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
   # dtmin is all abs => does not care about sign already.
 
   if !isdae && isinplace(prob) && typeof(u) <: AbstractArray && eltype(u) <: Number && uBottomEltypeNoUnits == uBottomEltype && tType == tTypeNoUnits # Could this be more efficient for other arrays?
-    if !(typeof(u) <: ArrayPartition)
+    if typeof(u) <: ArrayPartition
```
devmotion, line 181: This doesn't seem like a good idea. Before, everything apart from ArrayPartition ended up in this branch, and with this change only ArrayPartition would end up in this branch. It seems what you would like is rather to remove the other branch and always use `rate_prototype = recursivecopy(u)`.

src/solve.jl
```diff
@@ -178,7 +178,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
...
       rate_prototype = recursivecopy(u)
     else
       rate_prototype = similar(u, typeof.(oneunit.(recursive_bottom_eltype.(u.x))./oneunit(tType))...)
```
devmotion, line 184: So maybe rather:
```suggestion
    rate_prototype = recursivecopy(u)
```
However, I didn't test it, so maybe it fixes the mentioned issue but breaks some other ArrayPartition cases...

src/solve.jl
```diff
@@ -178,10 +178,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
   # dtmin is all abs => does not care about sign already.
 
   if !isdae && isinplace(prob) && typeof(u) <: AbstractArray && eltype(u) <: Number && uBottomEltypeNoUnits == uBottomEltype && tType == tTypeNoUnits # Could this be more efficient for other arrays?
-    if !(typeof(u) <: ArrayPartition)
-      rate_prototype = recursivecopy(u)
-    else
-      rate_prototype = similar(u, typeof.(oneunit.(recursive_bottom_eltype.(u.x))./oneunit(tType))...)
+    rate_prototype = recursivecopy(u)
     end
```
devmotion, line 182:
```suggestion
```

devmotion:
It would be good to add a small example of the issue that is fixed here to the tests.

devmotion:
It seems this change breaks other `ArrayPartition`s (see, e.g., https://github.com/SciML/OrdinaryDiffEq.jl/pull/1309/checks?check_run_id=1563518704#step:6:255), which is probably what motivates the current implementation.

jdeldre:
Got ahead of myself. This failure is fixed by https://github.com/SciML/DiffEqBase.jl/pull/621. New release is imminent.

devmotion (approved):
LGTM. Thanks for the PR!

jdeldre:
@ChrisRackauckas Do you think you could issue a new release? I've got a few packages that depend on this fix.

ChrisRackauckas:
Done.

PR #1313, opened 2020-12-30
fix symplectic algorithms for arbitrary functions

### LINKED ISSUE #1312: Symplectic algorithms silently calculate wrong results under certain conditions
This is a redirection of an issue, which was found with `DiffEqPhysics`:

https://github.com/SciML/DiffEqPhysics.jl/issues/60#issuecomment-752423827

KlausC:
Once finished this PR shall fix #1312 .

The first commit contains test cases, which reveal the failing algorithms.


test/algconvergence/symplectic_tests.jl
```diff
@@ -0,0 +1,60 @@
+module SymplecticTests
```
devmotion, line 1: You don't have to define a module here, `@safetestset` in `test/runtests.jl` already runs everything inside of a separate module automatically.

KlausC: thanks - will change

KlausC:
Can somebody tell me, what to change in order to get CI start the test job?  https://github.com/SciML/OrdinaryDiffEq.jl/runs/1627221804

KlausC:
I have reverted the change in Project.toml. But something else stops the test run, which is unrelated to my changes.

KlausC:
The results of my local test-run are here. https://github.com/SciML/OrdinaryDiffEq.jl/issues/1312#issuecomment-752702174

ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/1313/checks?check_run_id=1627221804#step:6:235

This looks like it's due to a breaking change in a dependency. 

https://github.com/JuliaLinearAlgebra/IterativeSolvers.jl/pull/280

I'll get that fixed to unblock the tests. Sorry.

ChrisRackauckas:
https://github.com/SciML/DiffEqBase.jl/pull/625 unblocks the test, just waiting on @YingboMa to 👍 it or change it to reltol.

test/runtests.jl
```diff
@@ -14,6 +14,10 @@ end
 #Start Test Script
 
 @time begin
+if GROUP == "All" || GROUP == "Symplectic"
```
YingboMa, line 17: This is not correct. Could you add the tests into a group that already exists?

KlausC: Yes, you are right. That was to see the results of my changes at first. In the current state this PR must not be merged.

src/perform_step/symplectic_perform_step.jl
```diff
@@ -74,10 +75,36 @@ end
...
+
+function verify_f2(::C, f, p, q, pa, t) where C<:CachesNipNewton
+    res = f(p, q, pa, t)
+    res == p ? p : throwex()
+end
+function verify_f2(::C, f, res, p, q, pa, t) where C<:CachesInpNewton
+    f(res, p, q, pa, t)
+    res == p ? res : throwex()
+end
+throwex() = throw(ArgumentError("This algorithm is invalid if f2(p, q, t) != p"))
```
ChrisRackauckas, line 105: which methods throw this argument? That would be good to document. It looks like most are fixed?

src/perform_step/symplectic_perform_step.jl
```diff
@@ -153,8 +176,9 @@ end
   @unpack t,dt,f,p = integrator
   @unpack a1,a2,b1,b2 = cache
   duprev,uprev = integrator.uprev.x
+  kuprev = integrator.k[2].x[2]
```
ChrisRackauckas, line 179: these should use fsallast instead of k directly in case the values are updated by the FSAL due to an event. Though it might work because of aliasing.

KlausC: I can change it to `kuprev = integrator.fsllast.x[2]` instead and verify if it still works.
It looked consistent to use the same variable `k.x`, which is filled at the end of all step methods.
I also recognized, that the same lines are coded multiple times to load and store the state and derivatives in most of the algorithms.
Do you think it is worth while to centralize this code in in-lined functions or macros?
Something like `int_state!(integrator...); uprev, duprev, kuprev = load_state(integrator...); store_state!(integrator)` that would allow to handle this `fsal` thing at one spot.

src/perform_step/symplectic_perform_step.jl
KlausC: There is only one algorithm `VelocityVerletCache`, which I could not generalize to the Hermitian case, because it uses the time derivative of velocity.

KlausC: I could put the name of the algorithm found in `integrator.alg` into the message of the exception.

src/perform_step/symplectic_perform_step.jl
ChrisRackauckas: As a separate PR, yes. For most algorithms the FSAL handling is a trivial single line: partitioned methods get a little tricker though. But the issue isn't where you placed it, it's about the FSAL updates. If someone has an event, say `u` is changed after the last step, then `f` needs to be re-evaluated. The FSAL infrastructure should do this (it does this for most methods, I don't know if I've tested the symplectic methods specifically). It might also update this array due to how things are aliased in the initialization though.

src/perform_step/symplectic_perform_step.jl
ChrisRackauckas: What I meant was that I think that it should get a mention on https://diffeq.sciml.ai/stable/solvers/dynamical_solve/#Symplectic-Integrators

src/perform_step/symplectic_perform_step.jl
ChrisRackauckas: Pushing this off to https://github.com/SciML/OrdinaryDiffEq.jl/issues/1317 since this PR already does a lot and I don't think this is worth blocking it for.

ChrisRackauckas:
Thanks!

PR #1320, opened 2021-01-02
remove deprecated broadcast in test file

KlausC:
this change in a test file remove a test block when StaticArrays@1.0 is used

### DIFF
--- a/test/interface/static_array_tests.jl
+++ b/test/interface/static_array_tests.jl
@@ -9,8 +9,7 @@ ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
 sol = solve(ode, Tsit5())
 
-u0 = fill(zero(SVector{2,Float64}), 2)
-u0[1] = ones(SVector{2,Float64}) .+ 1
+u0 = [fill(2, SVector{2,Float64}), zeros(SVector{2,Float64})]
 ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
 sol = solve(ode, Tsit5())
--- a/test/interface/static_array_tests.jl
+++ b/test/interface/static_array_tests.jl
@@ -9,7 +9,7 @@ ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
 sol = solve(ode, Tsit5())
 
-u0 = fill(zero(SVector{2,Float64}), 2) .+ 1
+u0 = fill(zero(SVector{2,Float64}), 2)
 u0[1] = ones(SVector{2,Float64}) .+ 1
 ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
--- a/test/interface/static_array_tests.jl
+++ b/test/interface/static_array_tests.jl
@@ -2,36 +2,43 @@ using StaticArrays, Test
 using OrdinaryDiffEq
 using RecursiveArrayTools
 
-u0 = fill(zero(MVector{2,Float64}), 2)
-u0[1] = ones(MVector{2,Float64}) .+ 1
+u0 = [fill(2, MVector{2,Float64}), ones(MVector{2,Float64})]
 f = (du,u,p,t) -> du .= u
 ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
 sol = solve(ode, Tsit5())
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
 
-u0 = [fill(2, SVector{2,Float64}), zeros(SVector{2,Float64})]
+u0 = [fill(2, SVector{2,Float64}), ones(SVector{2,Float64})]
 ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
 sol = solve(ode, Tsit5())
-
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
 sol = solve(ode, SSPRK22(), dt=1e-2)
-
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
 
 u0 = ones(MVector{2,Float64})
 ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
 sol = solve(ode, Tsit5(), dt=1e-2)
-
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
 
 u0 = ones(SVector{2,Float64})
 f = (u,p,t) -> u
 ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
 integrator = init(ode, ImplicitEuler())
 @test OrdinaryDiffEq.get_W(integrator.cache.nlsolver) isa StaticArrays.LU
 sol = solve(ode, ImplicitEuler())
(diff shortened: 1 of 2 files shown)


test/interface/static_array_tests.jl
```diff
@@ -9,7 +9,7 @@ ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
 sol = solve(ode, Tsit5())
 
-u0 = fill(zero(SVector{2,Float64}), 2) .+ 1
+u0 = fill(zero(SVector{2,Float64}), 2)
 u0[1] = ones(SVector{2,Float64}) .+ 1
```
devmotion, line 13: This leads to a different initialization than before, it seems? Maybe it would be easier to just use:

```suggestion
u0 = [fill(2, SVector{2,Float64}), ones(SVector{2,Float64})]
```

KlausC: the values may be irrelevant in this test.

devmotion: A solution with a component that is zero constantly did not seem useful but probably it does not matter here and I noticed that the test with `MVector` above also uses a zero component. However, even in this case I'd say the following initialization
```suggestion
u0 = [fill(2, SVector{2,Float64}), zeros(SVector{2,Float64})]
```
is more natural (similarly, the example above could be changed).

KlausC: I think the " .+ 1" it was just a typo, which did not any harm as long as `+(::SArray, ::Number)` was defined.

devmotion: No, it was completely intentional. The tests were added a long time ago, I assume when only `zeros` but not `ones` was defined for StaticArrays. Initially, it was even just `fill(...) + 1` (without the broadcast) since this was valid syntax back then.

ChrisRackauckas:
The values shouldn't matter too much, but making it all zeros makes it a bit weaker. It should have a non-zero initialization and check the final solution isn't pure zero or the same as the initial condition, as that would be a sign of aliasing issues.

ChrisRackauckas:
We should probably strengthen it just a little to add the test for non-zero-ness in the end.

KlausC:
adapted recommendations also to other cases in the surrounding.

ChrisRackauckas (approved):

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
 using OrdinaryDiffEq
 using RecursiveArrayTools
 
-u0 = fill(zero(MVector{2,Float64}), 2)
-u0[1] = ones(MVector{2,Float64}) .+ 1
+u0 = [fill(2, MVector{2,Float64}), ones(MVector{2,Float64})]
 f = (du,u,p,t) -> du .= u
 ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 9: Just a minor suggestion (feel free to ignore it):
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
...
 sol = solve(ode, Tsit5())
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 11: Same here:
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
...
 
-u0 = [fill(2, SVector{2,Float64}), zeros(SVector{2,Float64})]
+u0 = [fill(2, SVector{2,Float64}), ones(SVector{2,Float64})]
 ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 16:
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
...
 sol = solve(ode, Tsit5())
-
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 18:
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
...
 sol = solve(ode, SSPRK22(), dt=1e-2)
-
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 20:
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
...
 
 u0 = ones(MVector{2,Float64})
 ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 25:
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
...
 sol = solve(ode, Tsit5(), dt=1e-2)
-
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 27:
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
...
 
 u0 = ones(SVector{2,Float64})
 f = (u,p,t) -> u
 ode = ODEProblem(f, u0, (0.,1.))
 sol = solve(ode, Euler(), dt=1e-2)
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 33:
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
...
 integrator = init(ode, ImplicitEuler())
 @test OrdinaryDiffEq.get_W(integrator.cache.nlsolver) isa StaticArrays.LU
 sol = solve(ode, ImplicitEuler())
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 37:
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
...
 sol = solve(ode, ImplicitEuler(nlsolve=NLAnderson()))
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 39:
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

test/interface/static_array_tests.jl
```diff
@@ -2,36 +2,43 @@ using StaticArrays, Test
...
 sol = solve(ode, Tsit5(), dt=1e-2)
+@test !any(iszero.(sol(1.0))) && !any(sol(1.0) .== u0)
```
devmotion, line 41:
```suggestion
@test !any(iszero, sol(1.0)) && !any(sol(1.0) .== u0)
```

KlausC:
```
julia> @btime any($iszero, $sol(1.0))
  2.886 μs (34 allocations: 992 bytes)
false

julia> @btime any($iszero.($sol(1.0)))
  2.968 μs (36 allocations: 1.09 KiB)
```

PR #1324, opened 2021-01-08
fix callback double save

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/711 @affans


ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/1324/checks?check_run_id=1669036678#step:6:629

That's exciting.

PR #1333, opened 2021-01-12
Rodas4P2 included

gstein3m:


ChrisRackauckas:
Add a convergence test https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/test/algconvergence/ode_rosenbrock_tests.jl

ChrisRackauckas:
It looks like that tableau as written only gets first order convergence. You will want to play with it locally. Note that you can do `plot(sim)` to plot the convergence test. Most likely one of the values in the tableau were copied incorrectly.

gstein3m:
Could the continuous output be the reason? Maybe I have to check these coefficients.

ChrisRackauckas:
Thanks! 

Add it to the list? https://github.com/SciML/DiffEqDocs.jl/blob/master/docs/src/solvers/ode_solve.md

PR #1337, opened 2021-01-27
Fix ElasticArray tests

ChrisRackauckas:
This is kind of a hack, but it looks like there was a major regression in that library and I'm not quite sure why, but given I found this while working on https://github.com/SciML/SciMLBase.jl/pull/1 I'm putting a fix to ElasticArrays right inline and will upstream that later. I'm kind of confused because I don't see a recent tag from ElasticArrays but it could've been a versioning thing.

### DIFF
--- a/test/integrators/ode_cache_tests.jl
+++ b/test/integrators/ode_cache_tests.jl
@@ -87,6 +87,11 @@ affect_matrix! = function (integrator)
 end
 callback_matrix = ContinuousCallback(condition_matrix, affect_matrix!)
 
+Base.similar(x::ElasticArrays.ElasticArray, ::Type{T}) where T = deepcopy(x)
+Base.copy(x::ElasticArrays.ElasticArray) = deepcopy(x)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Pinging @ranocha since I know you're invested in this (https://github.com/SciML/OrdinaryDiffEq.jl/pull/1275).

ranocha:
After all, we decided to use a different approach in Trixi - instead of using ElasticArrays, we just used plain vectors in OrdinaryDiffEq and wrap them ourselves to have more control. Hence, I didn't follow the development of ElasticArrays etc. further.

oschulz:
This is strange - "integrators/ode_cache_tests.jl" (current master) runs fine for me with the current version of ElasticArrays. Which version of ElasticArrays did the tests fail with?

ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/runs/1773806609?check_suite_focus=true#step:6:96

oschulz:
Ok, new release is ready: https://github.com/JuliaArrays/ElasticArrays.jl/releases/tag/v1.2.5

PR #1342, opened 2021-01-29
Try running GPU tests without code coverage.

maleadt:
It isn't submitted anyway, and this looks to be the cause for a 1000x slowdown.


maleadt:
Seems to be doing the job. I lowered the test timeout to something reasonable to reflect this change. If OK, merge this ASAP and rebase existing PRs on top of this to save on CI resources.

PR #1354, opened 2021-03-04
Remove unnecessary pre-allocation

mohamed82008:
This PR removes an unnecessary pre-allocation that causes some issues when doing higher-order AD and the eltype is incorrect.

### DIFF
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -230,7 +230,7 @@ function ode_interpolation(tvals,id,idxs,deriv,p,continuity::Symbol=:left)
       return ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache,idxs,deriv)
     end
   end
-  vals = _vals[invperm(idx)]
+  invpermute!(vals, idx)
   DiffEqArray(vals, tvals)
 end
 
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -230,7 +230,7 @@ function ode_interpolation(tvals,id,idxs,deriv,p,continuity::Symbol=:left)
       return ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache,idxs,deriv)
     end
   end
-  vals = _vals[sortperm(idx)]
+  vals = _vals[invperm(idx)]
   DiffEqArray(vals, tvals)
 end
 
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -230,7 +230,7 @@ function ode_interpolation(tvals,id,idxs,deriv,p,continuity::Symbol=:left)
       return ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache,idxs,deriv)
     end
   end
-  vals = _vals[sortperm(idx,rev=tdir<0)]
+  vals = _vals[sortperm(idx)]
   DiffEqArray(vals, tvals)
 end
 
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -200,7 +200,7 @@ function ode_interpolation(tvals,id,idxs,deriv,p,continuity::Symbol=:left)
   # start the search thinking it's ts[1]-ts[2]
   i₋ = 1
   i₊ = 2
-  _vals = map(idx) do j
+  vals = map(idx) do j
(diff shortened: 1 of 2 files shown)


src/dense/generic_dense.jl
```diff
@@ -227,18 +219,18 @@ function ode_interpolation(tvals,id,idxs,deriv,p,continuity::Symbol=:left)
...
-      vals[j] = ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache.caches[id.alg_choice[i₊]],idxs,deriv)
+      return ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache.caches[id.alg_choice[i₊]],idxs,deriv)
     else
       DiffEqBase.addsteps!(ks[i₊],ts[i₋],timeseries[i₋],timeseries[i₊],dt,f,p,cache) # update the kcurrent
-      vals[j] = ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache,idxs,deriv)
+      return ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache,idxs,deriv)
     end
   end
-
+  vals = _vals[sortperm(idx)]
```
devmotion, line 233: Is `sortperm` really needed here? Isn't this already handled by https://github.com/mohamed82008/OrdinaryDiffEq.jl/blob/398d5263c93098538ec6e20c799357af4fd47587/src/dense/generic_dense.jl#L198 (which also takes care of the integration direction)?

mohamed82008: The sortperm here is on `idx` to get the "inverse" of `idx` that will put the elements of `_vals` in the original order they should be in. Previously we would just allocated at the proper index `j`. Now I use a map so the order is not right.

devmotion: Ah, I see. This additional allocation and sorting is unfortunate.

devmotion: I guess it might also be more efficient to use
```suggestion
  vals = _vals[invperm(idx)]
```
here.

src/dense/generic_dense.jl
```diff
@@ -197,18 +197,10 @@ function ode_interpolation(tvals,id,idxs,deriv,p,continuity::Symbol=:left)
...
-    vals = Vector{Array{eltype(first(timeseries)),ndims(idxs)}}(undef, length(tvals))
-  else
-    vals = Vector{eltype(timeseries)}(undef, length(tvals))
-  end
-
   # start the search thinking it's ts[1]-ts[2]
   i₋ = 1
   i₊ = 2
-  @inbounds for j in idx
+  _vals = map(idx) do j
```
devmotion, line 203:
```suggestion
  vals = map(idx) do j
```

src/dense/generic_dense.jl
```diff
@@ -227,18 +219,18 @@ function ode_interpolation(tvals,id,idxs,deriv,p,continuity::Symbol=:left)
...
-      vals[j] = ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache.caches[id.alg_choice[i₊]],idxs,deriv)
+      return ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache.caches[id.alg_choice[i₊]],idxs,deriv)
     else
       DiffEqBase.addsteps!(ks[i₊],ts[i₋],timeseries[i₋],timeseries[i₊],dt,f,p,cache) # update the kcurrent
-      vals[j] = ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache,idxs,deriv)
+      return ode_interpolant(Θ,dt,timeseries[i₋],timeseries[i₊],ks[i₊],cache,idxs,deriv)
     end
   end
-
+  vals = _vals[invperm(idx)]
```
devmotion, line 233:
```suggestion
  invpermute!(vals, idx)
```

devmotion (commented):
Just learnt about `invpermute!`, that's even better.

mohamed82008:
>Just learnt about invpermute!, that's even better.

Awesome!

devmotion:
I guess it would be good to check that this change does not break type inference (maybe it's already covered?).

mohamed82008:
We already test `ode_interpolant` but not `ode_interpolation`'s type stability. I added a test on the `sol(..)` call.

ChrisRackauckas (approved):

PR #669, opened 2019-02-23
Implemented KYKDGSSPRK3S2

jprob:
Implemented the SSPRK(3 Stages, Order 2) scheme given in "Optimal Strong-Stability-Preserving Runge–Kutta Time Discretizations for Discontinuous Galerkin Methods" by Kubatko, Yeager, Ketcheson (2014) as requested here #493. 

```
f = (u,p,t)->sin(u)
(::typeof(f))(::Type{Val{:analytic}},u0,p,t) = 2*acot(exp(-t)*cot(0.5))
prob_ode_nonlinear = ODEProblem(f, 1.,(0.,0.5))
dts = 1 .//2 .^(8:-1:4)
testTol = 0.25
sim = test_convergence(dts, prob_ode_nonlinear, KYK2014DGSSPRK_3S2())
plot(sim)
```
image

```
g(u,p,t) = 2*t + u
u0 = 1.0
tspan = (0.0,10.0)
prob = ODEProblem(g,u0,tspan)
sol1 = solve(prob, Tsit5())
sol2 = solve(prob, KYK2014DGSSPRK_3S2(), dt = 1/100)
tf(t) = exp(t)*(2*(-t*exp(-t) - exp(-t)) + 3)
plot(sol2.t, tf, title="Solution for g'=2*t + u",label="True")
plot!(sol1, label="Tsit5")
plot!(sol2, label="KYK2014DGSSPRK_3S2")
```
image

I am not sure if I am doing the `perform_step` and `initialize` version using the cache right, since I barely use it.

PR #1366, opened 2021-03-25
Implemented KYKSSPRK32

ErikQQY:
Hello Chris, I implemented SSPRK32 method in Kubatko's paper, please review it. Then I would implement the other methods in next few days.


ChrisRackauckas:
Looks like there's a test error: https://github.com/SciML/OrdinaryDiffEq.jl/pull/1366/checks?check_run_id=2192281329#step:6:387

ErikQQY:
Oops, I see the error. I will try it again.

ChrisRackauckas:
I would suggest making sure it runs locally before pushing commits. Each of the test files can be run on their own so If you just run through it then most likely all of the tests will pass on CI. That might be a quicker feedback than waiting for CI.

ErikQQY:
Ok, thank you, I am testing it locally.

ChrisRackauckas:
This is named wrong, it should be the KYK one.

Also, isn't this https://github.com/SciML/OrdinaryDiffEq.jl/pull/669 ?

PR #1367, opened 2021-03-30
Implementing KYKSSPRK52

ErikQQY:


src/perform_step/ssprk_perform_step.jl
```diff
@@ -111,6 +111,74 @@ end
...
+  f(k, tmp, p, t+c2*dt)
+  # u3
+  @.. tmp = α30*uprev + α32*tmp + dt*β30*δ + dt*β32*k
+  f(k, tmp, p, t+c3*dt)
+  # u4
+  @.. tmp = α40*uprev + α43*tmp + dt*β40*δ + dt*β43*k
+  f(k, tmp, p, t+c4*dt)
+  # u
+  @.. u = α50*uprev + α51*tmp + dt*β50*δ + dt*β54*k
+  f(k, u, p, t+dt)
```
ChrisRackauckas, line 144: this k isn't used for anything? I think you're missing the last step.

src/perform_step/ssprk_perform_step.jl
```diff
@@ -111,6 +111,74 @@ end
...
+  @unpack t,dt,uprev,u,f,p = integrator
+  @unpack k,tmp, fsalfirst = cache
+  @unpack α20, α21, α30, α32, α40, α41, α43, α50, α51, α52, α54, β10, β21, β30, β32, β40, β41, β43, β50, β51, β52, β54, c1, c2, c3, c4 = cache.tab
+
+  δ = fsalfirst
+  # u1 -> stored as u
+  @.. u = uprev + dt*β10*δ
+  f(k, u, p, t+c1*dt)
+  # u2
+  @.. tmp = α20*uprev + α21*u + dt*β21*k
```
ChrisRackauckas, line 134: uprev should only ever be multiplied by 1?

src/perform_step/ssprk_perform_step.jl
ErikQQY: Am I missing the stage_limiter and step_limiter?

src/perform_step/ssprk_perform_step.jl
ErikQQY: While there are α10 and α20, I think uprev should be multipled by α10？

ChrisRackauckas: clip the relevant part of the paper.

PR #1371, opened 2021-04-15
Linear solver improvements for Rosenbrock solvers

YingboMa:
@aml5600 could you apply similar changes to other Rosenbrock solvers?

### DIFF
--- a/src/perform_step/rosenbrock_perform_step.jl
+++ b/src/perform_step/rosenbrock_perform_step.jl
@@ -24,8 +24,8 @@ function initialize!(integrator, cache::Union{Rosenbrock23ConstantCache,
 end
 
 @muladd function perform_step!(integrator, cache::Rosenbrock23Cache, repeat_step=false)
-  @unpack t,dt,uprev,u,f,p = integrator
-  @unpack k₁,k₂,k₃,du1,du2,f₁,fsalfirst,fsallast,dT,J,W,tmp,uf,tf,linsolve_tmp,jac_config,atmp = cache
+  @unpack t,dt,uprev,u,f,p,opts = integrator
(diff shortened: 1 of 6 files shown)


YingboMa:
Improvements:
1. Only factorize once when W is linear and constant
2. Make Rosenbrock23 work with the recent linear solver interface

src/perform_step/rosenbrock_perform_step.jl
```diff
@@ -37,9 +37,15 @@ end
...
+  new_W = calc_rosenbrock_differentiation!(integrator, cache, γ, γ, repeat_step, false)
+
+  calculate_residuals!(weight, fill!(weight, one(eltype(u))), uprev, uprev,
+                       opts.abstol, opts.reltol, opts.internalnorm, t)
+
+  cache.linsolve(vec(k₁), W, vec(linsolve_tmp), new_W,
+      Pl=DiffEqBase.ScaleVector(weight, true),
+      Pr=DiffEqBase.ScaleVector(weight, false), reltol=opts.reltol)
 
-  cache.linsolve(vec(k₁), W, vec(linsolve_tmp), !repeat_step)
```
ChrisRackauckas, line 42: do all of the Rosenbrocks need this treatment?

ChrisRackauckas:
The failing test is that the IIP and OOP no longer match outputs now. Are the weights changing the solve at all?

But if it seems right, then this should probably be copied down to all other rosenbrocks.

src/perform_step/rosenbrock_perform_step.jl
YingboMa: Yes

YingboMa:
The test failure looks unrelated? @ChrisRackauckas

ChrisRackauckas:
https://github.com/SciML/RecursiveArrayTools.jl/issues/137 is the issue.

PR #1373, opened 2021-04-16
Move isdae to kwargs

aml5600:
* moves the `isdae` value into the `kwargs` so the user can manually set it
* Allows for the user to bypass singularity checks on the mass matrix, which is beneficial when the mass matrix type does not play well with the singular functions or is large
* Default value is to run the checks

MWE of motivation:
```
M = Symmetric(sparse([1,2,3,1],[1,2,3,3],[2.,2.,2.,1.]))
fode = ODEFunction(DiffEqArrayOperator(M'M); mass_matrix=M)
ode_prob = ODEProblem(fode, [1.,1.,1.], (0.,1.))
sol = solve(ode_prob, Rosenbrock23(); mm_singular=false)
```

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -61,7 +61,6 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
                            alias_u0 = false,
                            alias_du0 = false,
                            initializealg = DefaultInit(),
-                           isdae = is_dae(prob,alg),
                            kwargs...) where recompile_flag
 
   if prob isa DiffEqBase.AbstractDAEProblem && alg isa OrdinaryDiffEqAlgorithm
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
kwargs need to match the common arguments. I don't get what case the user would need to set this?

aml5600:
> kwargs need to match the common arguments. I don't get what case the user would need to set this?

I was unaware of that. There are cases where I already know if my mass matrix is nonsingular or singular, it is also a type that does not work with the `issingular` logic. So the thought was this would allow someone to bypass that logic.

aml5600:
obviously open to suggestions of how I can facilitate this and include it here rather than just in my local project!

ChrisRackauckas:
I think it might just be better to improve `issingular`. We can make some better dispatches on that pretty easily, right? What matrix types?

aml5600:
sure, but I don't necessarily know what the types are yet. Currently I use the above type of `Symmetric{Float64, SparseMatrixCSC{Float64, Int64}}` but for a large-large matrix maybe using `LinearMaps` would be beneficial? And in both cases, if I know the matrices are nonsingular by construction, is it possible to leverage that knowledge to skip the check?

aml5600:
I couldn't come up with a solution yesterday to have a solution in `ArrayInterface`. My concern is, I really just don't want to need to check the singularity of a matrix if I know it is singular/nonsingular by construction. 

The "W operator" in the solver methods allows for lazy evaluation of the mass matrix and a rhs `DiffEqArrayOperator`, correct? If I have a matrix big enough that I want to don't want it in memory (or, a sparse-symmetric matrix I don't want to lose the symmetric savings) needing to do any factorization to determine if it is singular could be an issue.

I understand this is a very specific case, so if it only really makes sense to have this on my local copy I can close the PR again. Thanks!

YingboMa:
I feel it's easier to add a `issingular` field in DiffEqOperators. It could default to `nothing` which uses the current behavior, or the user could supply a `Bool` to override the computation.

aml5600:
> I feel it's easier to add a `issingular` field in DiffEqOperators. It could default to `nothing` which uses the current behavior, or the user could supply a `Bool` to override the computation.

@YingboMa this is in regards to the lhs mass matirx, not the rhs array operator.

YingboMa:
We can wrap the mass matrix inside a DiffEqArrayOperator, too.

Issue #1266, opened 2020-09-11
Initial point issue for some interpolants (Rodas4)


moesphere:
The `get_du` function does throw an error for certain customized interpolants, presumably an issue with the initial point and in the case of the Rodas4 family.

MWE:

```julia
using DifferentialEquations

function lorenz!(du,u,p,t)
 du[1] = 10.0*(u[2]-u[1])
 du[2] = u[1]*(28.0-u[3]) - u[2]
 du[3] = u[1]*u[2] - (8/3)*u[3]
end

u0 = [1.0;0.0;0.0]
tspan = (0.0,100.0)
prob = ODEProblem(lorenz!,u0,tspan)

integ1 = init(prob, Tsit5())
get_du(integ1) # works
integ2 = init(prob, Rodas4P())
get_du(integ2) # throws an error
```

See also [this discussion](https://discourse.julialang.org/t/callback-problem-with-differentialequations/46057/8) on julia discourse.

MatFi:
fixed and can be closed.

PR #1382, opened 2021-04-20
fix get_du for Rosenbrock4 methods

MatFi:
I came across https://github.com/SciML/OrdinaryDiffEq.jl/issues/1266 through a failed [terminatesteadystate callback](https://github.com/SciML/DiffEqCallbacks.jl/blob/master/src/terminatesteadystate.jl) and wondered about the special treatment of Rosenbrock4 family methods and the lack of a `Rosenbrock4Cache`. I wasn't sure if it is a good idea to rename `Rodas4Cache` accordingly. Anyway, `get_du(...)` on `Rodas4P()` works now and so does my callback.


src/perform_step/rosenbrock_perform_step.jl
```diff
@@ -612,13 +612,13 @@ end
...
 #### Rodas4 type method
 
-function initialize!(integrator, cache::Rodas4ConstantCache)
-  integrator.kshortsize = 2
-  integrator.k = typeof(integrator.k)(undef, integrator.kshortsize)
-  # Avoid undefined entries if k is an array of arrays
-  integrator.k[1] = zero(integrator.u)
-  integrator.k[2] = zero(integrator.u)
-end
+# function initialize!(integrator, cache::Rodas4ConstantCache)
```
ChrisRackauckas, line 615: ?

src/perform_step/rosenbrock_perform_step.jl
```diff
@@ -736,16 +736,17 @@ end
...
 end
 
 
-function initialize!(integrator, cache::Rodas4Cache)
-  integrator.kshortsize = 2
-  @unpack dense1,dense2 = cache
-  resize!(integrator.k, integrator.kshortsize)
-  integrator.k .= [dense1,dense2]
-end
+# function initialize!(integrator, cache::Rodas4Cache)
```
ChrisRackauckas, line 744: ?

src/perform_step/rosenbrock_perform_step.jl
MatFi: This function is equivalent to the already defined `
function initialize!(integrator, cache::Union{...})` at line 261, except that `integrator.fsallast` and `integrator.fsalfirst` are not set here and remain `undef`. So firstly this dispatch is redundant and secondly it does not initialize compleately

src/perform_step/rosenbrock_perform_step.jl
MatFi: this function is equivalent to the already defined `
function initialize!(integrator, cache::Union{...})` n L276, except that `integrator.fsallast` and `integrator.fsalfirst` are not defined here and remain `undef`. So firstly it is redundant and secondly it does not initialize completely

ChrisRackauckas: these algorithms don't want/have/need FSAL?

MatFi: This is how it looks to me. Anyway it does not hurt much to keep them.

ChrisRackauckas: Yeah I think they are needed since these algorithms aren't FSAL so fsalfirst and fsallast are unused on purpose.

MatFi: I have to admit that I'am not much in to all the details. I just looked up how it was done in the working `Rodas5()` and did some copy-paste instead of waiting for an low priority issue to get a fix.

ChrisRackauckas:
Those aren't the correct values for the `du`. I think you need to have this method generate from the interpolation, since it needs to do it from the `k`. Doing this correctly on this algorithm is much more subtle.

MatFi:
In fact, this is handled in `Rodas5` by an additional call to f, so this can be used here as well, or not?

ChrisRackauckas:
Rodas4 uses a higher order interpolation that is more robust to algebraic variables. There are many advantages to this algorithms, but also some advantages to that of Rodas5. They are doing something entirely different here, so confusing them doesn't make all that much sense. Roadas5's is incorrect for algebraic variables, which is noted in the documentation as a negative on DAEs. 

What you really want to do is just use the interpolation to grab the du, i.e. `get_du(integrator)` on Rodas4 should make use of `integrator(t,Val{1})` for the first derivative of the interpolation, which would then be the correct values for the algebraic variables as well.

devmotion:
> What you really want to do is just use the interpolation to grab the du, i.e. `get_du(integrator)` on Rodas4 should make use of `integrator(t,Val{1})` for the first derivative of the interpolation, which would then be the correct values for the algebraic variables as well.

Maybe one should make this the default, or at least change more generally
https://github.com/SciML/OrdinaryDiffEq.jl/blob/b563fcce03ee83d354c957a8a9c4f71bb85a56c1/src/integrators/integrator_interface.jl#L63
to
```julia
return if isdefined(integrator, :fsallast)
    integrator.fsallast
else
    integrator(integrator.t, Val{1})
end
```
and similarly for `get_du!`?

ChrisRackauckas:
That would be a better default.

MatFi:
>Maybe one should make this the default, or at least change more generally

I gave it a shot in https://github.com/MatFi/OrdinaryDiffEq.jl/commit/222ef916817668003060ac2548fd9549e061085b

but it turns out that this also does not work as expected (MWE):
```julia
using DiffEqBase, OrdinaryDiffEq
function lorenz!(du,u,p,t)
    du[1] = 10.0*(u[2]-u[1])
end

u0 = [1.0;0.0;0.0]
du=similar(u0)
tspan = (0.0,100.0)

prob = ODEProblem(lorenz!,u0,tspan)

integR5 = init(prob, Rodas5())
integR4 = init(prob, Rodas4P())

step!(integR5)
step!(integR4)

lorenz!(du,integR5.u,0,integR5.t)
all(get_du(integR5) .≈ du) #true

lorenz!(du,integR4.u,0,integR4.t)
all(get_du(integR4) .≈ du) # false
(du .- get_du(integR4)).^2 |>sum  # 874.3422007632396
```

ChrisRackauckas:
double check the Rodas4 derivative.

MatFi:
> double check the Rodas4 derivative.

I see, the interpolation only specializes for `Var{0}`,  

https://github.com/SciML/OrdinaryDiffEq.jl/blob/78c6d292bea39b23245f9c30741c5cbe3d049b20/src/dense/rosenbrock_interpolants.jl#L74-L102

ChrisRackauckas:
Interesting that it doesn't error for higher? That's a bug... maybe because of the fallback dispatch?

Anyways, If you write out the derivative w.r.t. theta there it should be good.

MatFi:
> Anyways, If you write out the derivative w.r.t. theta there it should be good. 

Thanks ..I'm on it

src/dense/rosenbrock_interpolants.jl
```diff
@@ -100,3 +100,27 @@ end
   @views @.. out = Θ1*y₀[idxs] + Θ*(y₁[idxs] + Θ1*(k[1][idxs] + Θ*k[2][idxs]))
   out
 end
+
+# First Derivative 
+
+@muladd function _ode_interpolant(Θ,dt,y₀,y₁,k,cache::Rodas4ConstantCache,idxs::Nothing,T::Type{Val{1}})
+  @inbounds (k[1] - y₀ + y₁ - 2*k[1]*Θ + 2*k[2]*Θ-3*k[2]*Θ^2)/dt
```
devmotion, line 107: Is it possible to rearrange this with [Horner's method](https://en.wikipedia.org/wiki/Horner%27s_method), as in the interpolations above?

devmotion: And in the other implementations below, of course.

src/integrators/integrator_interface.jl
```diff
@@ -60,14 +60,22 @@ function set_proposed_dt!(integrator::ODEIntegrator,integrator2::ODEIntegrator)
...
 @inline function DiffEqBase.get_du!(out,integrator::ODEIntegrator)
   if typeof(integrator.cache) <: FunctionMapCache
     out .= integrator.cache.tmp
   else
-    out .= integrator.fsallast
+    return if isdefined(integrator, :fsallast)
+      integrator.fsallast
+    else
+      integrator(integrator.t, Val{1})
+    end
```
devmotion, line 78: This should write the result to `out`:
```suggestion
    return if isdefined(integrator, :fsallast)
      out .= integrator.fsallast
    else
      integrator(out, integrator.t, Val{1})
    end
```

src/dense/rosenbrock_interpolants.jl
MatFi: done.

ChrisRackauckas:
Update https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/test/regression/ode_dense_tests.jl to test its derivative.

MatFi:
I added tests for the Rodas4's but the archived `dreltol = 2e-2` is quite bad. Is this something to worry about?

ChrisRackauckas:
What are the other derivative tests at? That seems a bit high.

MatFi:
Of course, I droped a  minus sign when taking the derivative. Now it's fine.

test.jl
```diff
@@ -0,0 +1,41 @@
...
+(du .- get_du(integR4)).^2 |>sum
+du
+get_du(integR4)
+du
+get_du(integR5)
+integR5.t
+integR4.t
+integR4(integR4.t,Val{1})
+using ForwardDiff
+ForwardDiff.derivative(t->integR4(t,Val{0}),integR4.t).-integR4(integR4.t,Val{1})
```
ChrisRackauckas, line 34: update the real test?

MatFi: sorry for the noise

ChrisRackauckas (approved):

PR #1384, opened 2021-04-22
controllers

### LINKED ISSUE #1379: PID step size controllers (and maybe even more?)
As far as I understand, OrdinaryDiffEq uses PI controllers (with parameters `beta1, beta2`). Some algorithms seem to work best with PID controllers (adding a parameter `beta3` to the mix), as demonstrated in our [recent preprint](http://arxiv.org/abs/2104.06836). It would be great if PID controllers could be added as well. 

Talking about this, could it make sense to restructure the [controllers](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/integrators/controllers.jl) by wrapping controllers in `struct`s and dispatch to distinguish different controllers? For example, the most general controller used in practice I know are the ones based of [Söderlind based on digital signal processing](https://dl.acm.org/doi/10.1145/641876.641877), which come with 5 parameters `alpha2, alpha3, beta1, beta2, beta3` and are [implemented in PETSc](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/TS/TSADAPTDSP.html).

### LINKED ISSUE #1380: Optimized adaptive methods for compressible fluid dynamics
We published a [new preprint](http://arxiv.org/abs/2104.06836) developing optimized low-storage explicit Runge-Kutta methods with automatic error control for compressible fluid dynamics. It plan to add the new methods here. This is to track the progress.

- [x] `3Sstarp35` (https://github.com/SciML/OrdinaryDiffEq.jl/pull/1384)
- [x] `3SstarpFSAL35` (https://github.com/SciML/OrdinaryDiffEq.jl/pull/1384)
- [x] `3Sstarp49` (https://github.com/SciML/OrdinaryDiffEq.jl/pull/1384)
- [x] `3SstarpFSAL49` (https://github.com/SciML/OrdinaryDiffEq.jl/pull/1384)
- [x] `3Sstarp510` (https://github.com/SciML/OrdinaryDiffEq.jl/pull/1384)
- [x] `3SstarpFSAL510` (https://github.com/SciML/OrdinaryDiffEq.jl/pull/1384)

Implementing the optimized controllers for the methods derived in the preprint linked above requires PID step size controllers (#1379).

The label `new-algorithm` would be appropriate but I can't add labels here right now.

ranocha:
I restructured the step size controllers to make them more flexible and extendable. In addition, I will add a new PID controller.

This is still work in progress. Nevertheless, I open this PR to get some CI tests. I will comment here when this is ready for review or I need help.

Closes #1379
Closes #1380


ranocha:
How shall updates of DelayDiffEq be handled? It's another repo, so I can't do it in one PR. On the other hand, tests will fail here unless DelayDiffEq get's updated and the other way round. Do we need to use an approach like
- Set upper compat bounds on OrdinaryDiffEq in DelayDiffEq
- Update OrdinaryDiffEq here
- Make a new release of OrdinaryDiffEq
- Update DelayDiffEq and set new lower compat bounds on OrdinaryDiffEq, remove upper bounds
- Make a new release of DelayDiffEq

src/solve.jl
```diff
@@ -35,11 +35,12 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
                            qmax = qmax_default(alg),
                            qsteady_min = qsteady_min_default(alg),
                            qsteady_max = qsteady_max_default(alg),
+                           beta1 = nothing,
+                           beta2 = nothing,
                            qoldinit = isadaptive(alg) ? 1//10^4 : 0,
+                           controller = nothing,
```
ChrisRackauckas, line 41: we'd need to level this up to a common kwarg if it goes here.

ranocha: What exactly do you mean by that?
- It needs to be documented (sure it does)?
- It needs to be implemented for other solvers/packages (delay, stochastic, quadrature, linear, nonlinear, ...)?
- Something else?

src/alg_utils.jl
```diff
@@ -487,7 +487,7 @@ function default_controller(alg, cache, qoldinit, _beta1=nothing, _beta2=nothing
   if ispredictive(alg)
     return PredictiveController()
   elseif isstandard(alg)
-    return IController()
+    return StandardIController()
```
ChrisRackauckas, line 490: isn't it a P controller? @YingboMa

ranocha: Citing from Hairer-Nørsett-Wanner 2, page 28:

> We saw that it was an I-controler ... and a control-man knows ... that PI is always better than I  (K. Gustafsson, June 1990)

If we order the letters of `PID` in the order of their terms, it should be `IPD` (as far as I know).

ChrisRackauckas: Interesting. But it's using the proportion of the error to the tolerance to do the next step?

YingboMa: Yes, that's an I-controller.

ranocha:
The current version seems to be a (more or less) clean adaptation of the previous infrastructure. What needs to be done:
- [ ] Implement a PID controller
- [ ] Handle downstream packages (DelayDiffEq)
- [ ] Document the new infrastructure
- [ ] Consolidate the interface. Right now, some parts such as `q`, `qold` are stored in the `integrator` while one could argue that they belong to the `controller`. On the other hand, several controllers will use this stuff, so it's probably okay to share this functionality.
- [ ] Maybe implement more general DPS controllers (or leave it for later)

ranocha:
This is now working as expected in some local tests. Documentation is added in the [companion PR to DiffEqDocs](https://github.com/SciML/DiffEqDocs.jl/pull/464). 
- How shall we handle the dependent packages failing right now?
- How shall we "level `controller` up to a common kwarg"?

ChrisRackauckas:
>How shall we handle the dependent packages failing right now?

Get a local PR ready for DelayDiffEq that will work, and we can up the lower bound.

>How shall we "level controller up to a common kwarg"?

I think at least StochasticDiffEq will need to support this too, and then it should be documented in the DiffEq docs common args section.

ranocha:
I opened PRs to SciML/StochasticDiffEq.jl#417 and SciML/DelayDiffEq.jl#214. Locally, all tests of these packages pass for me when using my dev'ed version of OrdinaryDiffEq. I also added documentation in SciML/DiffEqDocs.jl#464.

ChrisRackauckas (approved):

src/solve.jl
```diff
@@ -270,7 +271,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,DiffEqBase.
     # The QT fields are not used for DiscreteProblems
     constvalue(tTypeNoUnits)
   else
-    typeof(internalnorm(u, t))
+    typeof(DiffEqBase.value(internalnorm(u, t)))
```
YingboMa, line 274: This looks incorrect.

Issue #1387, opened 2021-04-25
Velocity Verlet integrator calculates the acceleration twice at each time step


jgreener64:
I think the current `VelocityVerlet` implementation calculates the acceleration twice per time step:

```julia
using DifferentialEquations

accel_counter = 0

function acceleration!(accels, vels, coords, p, t)
    global accel_counter += 1
    accels .= zero(coords)
end

tspan = (0.0, 100.0)
prob = SecondOrderODEProblem(acceleration!, rand(100), rand(100), tspan)
sol = solve(prob, VelocityVerlet(), dt=1.0)

println(accel_counter)
```
```
201
```

There are only 100 steps here. Looking at https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/perform_step/symplectic_perform_step.jl#L192-L196 it seems that the current and next time step accelerations are calculated, but as shown in https://en.wikipedia.org/wiki/Verlet_integration#Velocity_Verlet you can store the last time step acceleration and use it again. This is a good idea, since the acceleration calculation is expensive.

This is on v5.52.2 of OrdinaryDiffEq.jl. I may have misunderstood something, it's my first time looking properly at the SciML code base (it's a fun adventure!).

YingboMa:
Indeed it can reuse the acceleration calculation. I think you can access that via `integrator.fsallast.k[2]`. Do you want to make a PR?

ChrisRackauckas:
Yeah, it just needs to use the FSAL value.

jgreener64:
Great, I can look at making a PR.

PR #1390, opened 2021-04-26
Prevent velocity Verlet integrator calculating the acceleration twice at each time step

jgreener64:
As discussed in https://github.com/SciML/OrdinaryDiffEq.jl/issues/1387, use the FSAL value to avoid calculating the acceleration twice in the `VelocityVerlet` integrator.

Obtains similar results for a small system I tested on. I am running the package tests locally now.

Closes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1387.

### DIFF
--- a/src/perform_step/symplectic_perform_step.jl
+++ b/src/perform_step/symplectic_perform_step.jl
@@ -171,7 +171,7 @@ end
   duprev, uprev = load_symp_state(integrator)
 
   # x(t+Δt) = x(t) + v(t)*Δt + 1/2*a(t)*Δt^2
-  ku = f.f1(duprev,uprev,p,t)
+  ku = integrator.fsallast.x[1]
   dtsq = dt^2
   half = cache.half
   u = uprev + dt*duprev + dtsq*(half*ku)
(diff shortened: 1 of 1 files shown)


src/perform_step/symplectic_perform_step.jl
```diff
@@ -171,7 +171,7 @@ end
   duprev, uprev = load_symp_state(integrator)
 
   # x(t+Δt) = x(t) + v(t)*Δt + 1/2*a(t)*Δt^2
-  ku = f.f1(duprev,uprev,p,t)
+  ku = integrator.fsallast.x[1]
```
YingboMa, line 174: I feel this should be `integrator.fsallast.x[2]`

jgreener64: That leads to immediate divergence. I'm not sure I fully understand the variable naming scheme but in that function it seems `ku` is `a(t)` and `kdu` is `a(t+dt)`. So once you reach the next step, you want to load `kdu` from the last step into the current `ku`.

YingboMa: You are exactly, right.

jgreener64:
Thanks for merge.

Issue #1389, opened 2021-04-26
RadauIIA's  ArgumentError when jac_prototype is of type SparseMatrixCSC


MatFi:
This error should be easy to catch and only happened to me on `Radau` methods so far. However took quite a while to find the reason for it and I want to warn other users.

```julia
using DiffEqBase, OrdinaryDiffEq
using SparseArrays

function f!(du,u,p,t)
    du .= u
end

M= sparse(Diagonal(ones(3)))
jac_prototype = sparse(Diagonal(ones(3))) 
#jac_prototype = Diagonal(ones(3))  # works
f=ODEFunction(f!,mass_matrix=M, jac_prototype=jac_prototype)

u0=zeros(3)
p=ODEProblem(f,u0,(0,1.))
julia> sol =solve(p,RadauIIA3())
ERROR: ArgumentError: output array must not be aliased with input array
Stacktrace:
  [1] solve!(x::Vector{ComplexF64}, lu::SuiteSparse.UMFPACK.UmfpackLU{ComplexF64, Int64}, b::Vector{ComplexF64}, typ::Int64)
    @ SuiteSparse.UMFPACK /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/SuiteSparse/src/umfpack.jl:377
  [2] _AqldivB_kernel!
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/SuiteSparse/src/umfpack.jl:532 [inlined]
  [3] _Aq_ldiv_B!
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/SuiteSparse/src/umfpack.jl:527 [inlined]
  [4] ldiv!(X::Vector{ComplexF64}, lu::SuiteSparse.UMFPACK.UmfpackLU{ComplexF64, Int64}, B::Vector{ComplexF64})
    @ SuiteSparse.UMFPACK /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/SuiteSparse/src/umfpack.jl:510
  [5] (::DefaultLinSolve)(x::Vector{ComplexF64}, A::SparseMatrixCSC{ComplexF64, Int64}, b::Vector{ComplexF64}, update_matrix::Bool; reltol::Nothing, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
[...]

PR #1391, opened 2021-04-26
Add new buffers to prevent aliasing in RadauIIA3 and RadauIIA5

MatFi:
Proposed fix for https://github.com/SciML/OrdinaryDiffEq.jl/issues/1389
Wasn't sure about the naming convention of the new vectors.

### DIFF
--- a/src/perform_step/firk_perform_step.jl
+++ b/src/perform_step/firk_perform_step.jl
@@ -1,4 +1,4 @@
-  function do_newJ(integrator, alg, cache, repeat_step)::Bool # for FIRK
+function do_newJ(integrator, alg, cache, repeat_step)::Bool # for FIRK
   integrator.iter <= 1 && return true
   repeat_step && return false
   first(islinearfunction(integrator)) && return false
@@ -265,7 +265,7 @@ end
     @. dw12 = complex(fw1 - αdt*Mw1 + βdt*Mw2, fw2 - βdt*Mw1 - αdt*Mw2)
     needfactor = iter==1
     linsolve2(vec(cubuff), W1, vec(dw12), needfactor)
-    @.. dw12 = cubuff
+    dw12 = cubuff
     integrator.destats.nsolve += 1
     dw1 = real(dw12)
     dw2 = imag(dw12)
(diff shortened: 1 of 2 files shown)


src/perform_step/firk_perform_step.jl
```diff
@@ -1,4 +1,4 @@
-function do_newJ(integrator, alg, cache, repeat_step)::Bool # for FIRK
+  function do_newJ(integrator, alg, cache, repeat_step)::Bool # for FIRK
```
YingboMa, line 1: mistake?

src/perform_step/firk_perform_step.jl
```diff
@@ -264,7 +264,8 @@ end
 
     @. dw12 = complex(fw1 - αdt*Mw1 + βdt*Mw2, fw2 - βdt*Mw1 - αdt*Mw2)
     needfactor = iter==1
-    linsolve2(vec(dw12), W1, vec(dw12), needfactor)
+    linsolve2(vec(cubuff), W1, vec(dw12), needfactor)
+    @.. dw12 = cubuff
```
YingboMa, line 268: You don't really need to copy the data. Just do a pointer reassignment is enough
```julia
dw12 = cubuff
```
and since `cubuff` and `dw12` will be unpacked again from the cache, no information is lost.

src/perform_step/firk_perform_step.jl
MatFi: For some reason VScode does not highlight diffs in the first line. So this one went trough.

src/perform_step/firk_perform_step.jl
MatFi: I just feared that the reuse of `cubuff`/`ubuff`could cause some trouble, but you are right.

MatFi: I have to take back what I said. 

> and since cubuff and dw12 will be unpacked again from the cache, no information is lost.
this peace of code is running in a loop, so it will brake again if more then one iteration is happens.

MatFi:
Obviously there is a problem because a pointer reassigngment is in a loop https://github.com/SciML/OrdinaryDiffEq.jl/blob/4f09d4a94cb61cd2670eb3bdced1483961d54235/src/perform_step/firk_perform_step.jl#L122-L197

PR #1392, opened 2021-04-27
Fix Radau aliasing during newton iterations

MatFi:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/1391 has not fully solved issue https://github.com/SciML/OrdinaryDiffEq.jl/issues/1389 . The newton iterations still cause aliasing if more the one iteration is needed. This propose should finally do it


YingboMa:
Thanks!

PR #1398, opened 2021-05-02
Remove stiffness detection loop

ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/1358 allowed for skipping the instability check when changing the algorithm. However, this can loop back and forth, as seen in https://github.com/SciML/DiffEqFlux.jl/issues/539. This PR makes it so that way if the error check was off in the last step, it would be on in the next, making it so that way if it's truly unstable this will get caught.

Fixes https://github.com/SciML/DiffEqFlux.jl/issues/539


ChrisRackauckas:
Downstream patch inbound.

PR #1401, opened 2021-05-09
Bump the epsilon used for the tstops checking

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/747 which shows that the restriction is too tight.


ChrisRackauckas:
@devmotion DelayDiffEq might need to have a bigger floating point tolerance somewhere?

https://github.com/SciML/OrdinaryDiffEq.jl/pull/1401/checks?check_run_id=2537760741#step:6:377

ChrisRackauckas:
Seems like it's https://github.com/SciML/DelayDiffEq.jl/pull/212

PR #1411, opened 2021-05-21
Fix QNDF

JunpengGao233:
The QNDF in master branch fails to solve the Lorenz system. This PR fixes various errors in the QNDF implementation. After the fixes, both in-place and out-of-place QNDF implementations are able to solve the Lorenz system within 5000 timesteps.

Here is the plot:
lorenz


ChrisRackauckas:
Can you resolve the conflicts with the controller updates?

src/nlsolve/newton.jl
```diff
@@ -97,6 +98,7 @@ Equations II, Springer Series in Computational Mathematics. ISBN
 
   atmp = calculate_residuals(dz, uprev, ustep, opts.abstol, opts.reltol, opts.internalnorm, t)
   ndz = opts.internalnorm(atmp, t)
+  #@show ndz
```
YingboMa, line 101: Remove this

src/nlsolve/newton.jl
```diff
@@ -52,6 +52,7 @@ Equations II, Springer Series in Computational Mathematics. ISBN
   @unpack z,tmp,γ,α,cache = nlsolver
   @unpack tstep,W,invγdt = cache
 
+
```
YingboMa, line 55: Same here

src/nlsolve/nlsolve.jl
```diff
@@ -54,6 +54,7 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator, cache=nothing, repeat_
     iter > 1 && (η = θ / (1 - θ))
     if (iter == 1 && ndz < 1e-5) || (iter > 1 && (η >= zero(η) && η * ndz < κ))
       nlsolver.status = Convergence
+      #@info "converge"
```
YingboMa, line 57: here

YingboMa (approved):
Looks right to me.

YingboMa:
This is basically a complete rewrite of QNDF. Good work!

src/bdf_utils.jl
```diff
@@ -1,72 +1,23 @@
...
+  @views @.. D[:,k+1] = dd
+  for i in k:-1:1
+    @views @.. D[:,i] = D[:,i] + D[:,i+1]
   end
+  return nothing
 end
 
-global const γₖ = @SVector[sum(1//j for j in 1:k) for k in 1:6]
\ No newline at end of file
+global const γₖ = @SVector[sum(1//j for j in 1:k) for k in 1:6]
```
devmotion, line 23: I don't think `global const` is a thing?
```suggestion
const γₖ = @SVector[sum(1//j for j in 1:k) for k in 1:6]
```
One could also hardcode the values I guess:
```suggestion
# sum(1//j for j in 1:k) for k in 1:6
const γₖ = @SVector[1//1, 3//2, 11//6, 25//12, 137//60, 49//20]
```

src/bdf_utils.jl
```diff
@@ -69,4 +70,25 @@ function reinterpolate_history!(cache::OrdinaryDiffEqConstantCache, D, R, k)
...
+  @inbounds for r = 1:k
+    R[1,r] = -r * ρ
+    for j = 2:k
+      R[j,r] = R[j-1,r] * ((j-1) - r * ρ)/j
+    end
+  end
+  SArray(R)
+end
+
+function update_D!(D, dd, k)
```
YingboMa, line 85:
```suggestion
function update_D!(D, dd, k)
  dd = _vec(dd)
```

YingboMa:
I ran the POLLU benchmark with the new QNDF

pollu_bench

src/bdf_utils.jl
```diff
@@ -69,4 +70,26 @@ function reinterpolate_history!(cache::OrdinaryDiffEqConstantCache, D, R, k)
...
+  dd = _vec(dd)
+  @views @.. D[:,k+2] = dd - D[:,k+1]
+  @views @.. D[:,k+1] = dd
+  for i in k:-1:1
+    @views @.. D[:,i] = D[:,i] + D[:,i+1]
+  end
+  return nothing
+end
+
+const γₖ = @SVector[sum(1//j for j in 1:k) for k in 1:6]
```
devmotion, line 95: Actually, due to https://github.com/JunpengGao233/OrdinaryDiffEq.jl/blob/cc45e743e9505809d268c9c2d4ecd5344eba80be/src/caches/bdf_caches.jl#L290 and a similar line for the mutable cache, I am wondering if this line is needed at all.

YingboMa: This is needed for the old QNDF1 and QNDF2 methods

YingboMa:
The test failures seem unrelated.

YingboMa:
QNDF is actually the best solver for BCR now.

bcr_bench

JunpengGao233:
Thank you for helping make benchmarks. The bdf_utils.jl has some functions and constants that QNDF1 and QNDF2 are dependent so I recover them later. If needed, I could modify these methods later. 

Thanks for the instructions and help of Yingbo.

ChrisRackauckas:
ABDF2 is the other. Yeah, it would be good to get those cleaned up because they are common in PDE applications still.

YingboMa:
I merged this PR since the test failures are unrelated. I will look into these.

ChrisRackauckas:
bcr_bench_2
pollu_bench_2
bcr_step_control

YingboMa:
The above plots are with an optimized QNDF. I have a few more improves locally. I will open a PR soon.

Issue #1412, opened 2021-05-22
Low-storage RK solvers test failure


YingboMa:
![image](https://user-images.githubusercontent.com/17304743/119234919-157e7280-bafe-11eb-8c5d-c74f023c2122.png)

Because of the CI short-circuiting, tests were never run on those newly added solvers. When I checkout the `v5.54.0` tag locally, the tests fail. I think those solvers are merged with tests ever passing.

@ranocha could you fix those test failures?

CC: @ChrisRackauckas

ranocha:
I will have a look

PR #1413, opened 2021-05-22
Mark broken tests broken

YingboMa:
Ref: https://github.com/SciML/OrdinaryDiffEq.jl/issues/1412


ChrisRackauckas (approved):

YingboMa:
@ChrisRackauckas could you take a look at the Tracker failure?

YingboMa:
Ref #1414

Issue #1414, opened 2021-05-22
Tracker downstream test is broken


YingboMa:
Introduced by #1384

We need to respect the return type of `internalnorm`. We want to AD through the controller when `t` is being differentiated.

CC: @ranocha, @ChrisRackauckas

ChrisRackauckas:
>We need to respect the return type of internalnorm. We want to AD through the controller when t is being differentiated.

Normally you don't want to differentiate through the controller even when differentiating through `t`.

YingboMa:
Ah, right. Controller should call `value` to strip the differentiation info.

ranocha:
So what's the conclusion? Is there some need to change something from #1348? Xref https://github.com/SciML/OrdinaryDiffEq.jl/pull/1384#discussion_r637448652

ChrisRackauckas:
Fixed by `ArrayInterface.ismutable(T::Type{<:Tracker.TrackedReal}) = false`. It looks like it was failing earlier with the newer RecursiveArrayTools, just using the wrong path because it thinks a TrackedReal is mutable.

YingboMa:
I don't think this completely fixes it. I had this locally when I was debugging yesterday. Something in the controller is throwing as well.

YingboMa:
Yep https://github.com/SciML/OrdinaryDiffEq.jl/pull/1416/checks?check_run_id=2650244244#step:6:782

ChrisRackauckas:
Fixed by https://github.com/SciML/OrdinaryDiffEq.jl/pull/1417, which is then matched in https://github.com/SciML/StochasticDiffEq.jl/pull/418 and https://github.com/SciML/StochasticDelayDiffEq.jl/pull/43. Interestingly, this must've been noticed in the DelayDiffEq.jl tests, so it already does the upconversion:

https://github.com/SciML/DelayDiffEq.jl/blob/master/src/solve.jl#L227

ChrisRackauckas:
Honestly, given that big of a change, I'm pretty happy this was the only regression we found. I expected more 😉. Let's see what the tests say.

PR #1417, opened 2021-05-23
Do not remove tracking information from time

ChrisRackauckas:
Time has to be differentiated when it is chosen to be differentiated.

Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1414


ChrisRackauckas:
The old downstream test doesn't run master. We should change it sometime soon.

PR #1420, opened 2021-05-24
Optimize QNDF take 1

YingboMa:
This PR optimizes QNDF by:
- reducing the memory footprint (This can be optimized further by removing the `utilde*` cache vectors.)
- computing `scaling_factor * ||vector||` instead of `||scaling_factor * vector||` (This can be optimized further by precomputing the weight vector.)
- minimizing the number of function evaluations for the Hermite interpolant (This can be optimized further by using the NDF/BDF equality.)
- better convergence criterion for the simplified Newton iteration, so that the safety factor kappa can be relaxed. Ref: https://github.com/SciML/DiffEqBase.jl/pull/668

CC: @JunpengGao233


YingboMa:
pollu_qndf_opt

ChrisRackauckas:
Any tests with LinSolveGMRES?

YingboMa:
It has one now :p

ChrisRackauckas:
And performance check, say on some PDE?

JunpengGao233:
I do not even realize many of the points in these optimizations. This teaches me a lot. Thank you! :)

ChrisRackauckas (approved):

src/caches/bdf_caches.jl
```diff
@@ -334,8 +336,10 @@ function alg_cache(alg::QNDF{MO},u,rate_prototype,uEltypeNoUnits,uBottomEltypeNo
   u₀ = zero(u)
   dtprev = one(dt)
   u₀ = zero(u)
-  D = zero(similar(u, length(u), max_order + 2))
-  Dtmp = zero(similar(D))
+  D = similar(u, length(u), max_order + 2)
```
devmotion, line 339: Wouldn't it be better to use
```suggestion
  D = similar(u, uEltypeNoUnits, length(u), max_order + 2)
```
or
```suggestion
  D = Matrix{uEltypeNoUnits}(undef, length(u), max_order + 2)
```
as above since `D` is filled with `zero(uEltypeNoUnits)`?

YingboMa: Yes, I wasn't being consistent because implicit solvers don't support units anyway.

YingboMa:
Here's a performance analysis for the Filament PDE problem:

Filament_t_vs_dt

QNDF has the tendency to vary the step size much more frequently than CVODE_BDF. Therefore, we need to tune the controller heuristic more. This results in more Jacobian evaluations and W factorizations than CVODE_BDF.
```julia
julia> sol.destats # QNDF
DiffEqBase.DEStats
Number of function 1 evaluations:                  2185
Number of W matrix evaluations:                    62
Number of linear solves:                           470
Number of Jacobians created:                       25
Number of nonlinear solver iterations:             328
Number of nonlinear solver convergence failures:   17
Number of accepted steps:                          112
Number of rejected steps:                          6

julia> sol_bdf.destats # CVODE_BDF
DiffEqBase.DEStats
Number of function 1 evaluations:                  170
Number of W matrix evaluations:                    32
Number of linear solves:                           0
Number of Jacobians created:                       23
Number of nonlinear solver iterations:             167
Number of nonlinear solver convergence failures:   1
Number of accepted steps:                          64
Number of rejected steps:                          0
```

YingboMa:
And here is GMRE on the Filament problem
```julia
julia> @time sol = solve(prob, QNDF(autodiff=false, nlsolve=NLNewton(κ=0.8, new_W_dt_cutoff=0.8)), linsolve=LinSolveGMRES());
  1.276866 seconds (27.03 M allocations: 413.025 MiB, 5.18% gc time)

julia> @time sol = solve(prob, QNDF(autodiff=false, nlsolve=NLNewton(κ=0.8, new_W_dt_cutoff=0.8)));
  1.307086 seconds (27.03 M allocations: 413.025 MiB, 7.08% gc time)

julia> @time sol = solve(prob, CVODE_BDF(linear_solver=:GMRES));
  1.067537 seconds (22.68 M allocations: 346.842 MiB, 5.82% gc time)

julia> @time sol = solve(prob, CVODE_BDF());
  0.948050 seconds (20.04 M allocations: 306.333 MiB, 6.53% gc time)
```

ChrisRackauckas:
Oh, could be optimized more but that's really not too bad.

PR #1438, opened 2021-06-28
Add event handling tests for QNDF

JunpengGao233:
Does "not safe for event handling" mean anything other than add QNDF tests to the "alg_events_tests.jl"?


ChrisRackauckas:
That test won't catch it. Try one of the bouncing ball ones. After an event modifies `u`, it will need to drop all history values and act like it was iteration 1 again.

JunpengGao233:
Thank you. I did not realize this previously...

ChrisRackauckas:
Yeah, now add something to the bouncing ball tests that will catch this. I think you can probably just swap out an integrator and test a value at high accuracy or something.

ChrisRackauckas (approved):

JunpengGao233:
Oh no, I just found one typo in the last line. Sorry, I will at least run the whole complete test one time next time.

PR #1440, opened 2021-07-04
Allow resizing with lazy WOperator

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqOperators.jl/issues/419 . Requires https://github.com/SciML/DiffEqBase.jl/pull/680 and https://github.com/SciML/DiffEqOperators.jl/commit/2fd08814b343c15e3d93f7212d1667000d79466a

### DIFF
--- a/src/integrators/integrator_interface.jl
+++ b/src/integrators/integrator_interface.jl
@@ -148,12 +148,15 @@ function resize_J_W!(cache, integrator, i)
     nf = nlsolve_f(f, integrator.alg)
     islin = f isa Union{ODEFunction,SplitFunction} && islinear(nf.f)
     if !islin
-      J = similar(f.jac_prototype, i, i)
-      if !isa(J, DiffEqBase.AbstractDiffEqLinearOperator)
+      if isa(cache.J, DiffEqBase.AbstractDiffEqLinearOperator)
+        resize!(cache.J,i)
+      else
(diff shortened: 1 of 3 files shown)


YingboMa:
Is the code coverage report real? Looks like no branch is covered.

YingboMa (approved):

src/integrators/integrator_interface.jl
```diff
@@ -148,12 +148,15 @@ function resize_J_W!(cache, integrator, i)
     nf = nlsolve_f(f, integrator.alg)
     islin = f isa Union{ODEFunction,SplitFunction} && islinear(nf.f)
     if !islin
-      J = similar(f.jac_prototype, i, i)
-      if !isa(J, DiffEqBase.AbstractDiffEqLinearOperator)
+      if isa(cache.J, DiffEqBase.AbstractDiffEqLinearOperator)
+        resize!(cache.J,i)
```
YingboMa, line 152: Is this right? Should it be `resize!(cache.J,i,i)`?

ChrisRackauckas: It has to be square though, so I'm not sure if that's necessary? It would always just use the first value?

YingboMa: Makes sense.

PR #1445, opened 2021-07-08
update signature of stage/step limiters

### LINKED ISSUE #1444: Change signature of stage limiters
Right now, some methods support stage callback/limiters, e.g. SSP methods Right now, we have `stage_limiter!` for SSPRK methods and in-place problems: 
https://github.com/SciML/OrdinaryDiffEq.jl/blob/0c2048a502101647ac35faabd80da8a5645beac7/src/perform_step/ssprk_perform_step.jl#L43
We would like to use this feature more often in our hyperbolic PDE framework [Trixi.jl](https://github.com/trixi-framework/Trixi.jl). However, we noticed that we need additional information such as the current time step for some applications. Would it be possible to change the signature from
```julia
stage_limiter!(u, f, p, t) 
```
to something like
```julia
stage_limiter!(u, integrator, p, t) 
```
including the `integrator`? Alternatively, we could unify the signature with callbacks, i.e. just use
```julia
stage_limiter!(integrator) 
```
since this will have all of the information.

This would be a breaking change, but I honestly don't know whether anybody except us uses this feature :sweat_smile:  

Xref https://github.com/SciML/OrdinaryDiffEq.jl/issues/1292

ranocha:
Closes #1444

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -47,7 +47,7 @@ struct ExplicitRK{TabType} <: OrdinaryDiffEqAdaptiveAlgorithm
 end
 ExplicitRK(;tableau=ODE_DEFAULT_TABLEAU) = ExplicitRK(tableau)
 
-@inline trivial_limiter!(u, f, p, t) = nothing
+@inline trivial_limiter!(u, integrator, p, t) = nothing
 """
 Euler - The canonical forward Euler method. Fixed timestep only.
 """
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
@chriselrod https://github.com/SciML/OrdinaryDiffEq.jl/pull/1445/checks?check_run_id=3018474439#step:6:369 ?

chriselrod:
Should be fixed by https://github.com/JuliaRegistries/General/pull/40496

ranocha:
Is the failing buildkite check somehow related to this PR?

ChrisRackauckas:
No, something is weird with the GPU test right now. I need to build a new PC to check it out.

Issue #105, opened 2017-08-04
Fully Implicit BDF


ChrisRackauckas:
http://faculty.smu.edu/shampine/cic.pdf

ufechner7:
The link to the paper in the first post is dead. This one works: https://sci-hub.53yu.com/10.1515/jnma.2002.291

PR #1452, opened 2021-07-19
Implementation of paper "Solving 0 = F (t; y(t); y0(t)) in Matlab"

JunpengGao233:
This PR implements the algorithms in this issue https://github.com/SciML/OrdinaryDiffEq.jl/issues/105 .

As the paper points out, for the difficulty due to eigenvalues near the imaginary axis, QNDF could not handle it well. The controller in this algorithm in this PR is similar to the controller of DASSL and could deal with this well.

The example problem (comes from L.F. Shampine, Numerical Solution of Ordinary Differential Equations, Chapman & Hall p387)

```julia
function f_im(du,u,p,t)
    du[1] = -10u[1]+100u[2]
    du[2] = -100u[1]-10u[2]
    du[3] = -4u[3]
    du[4] = -u[4]
    du[5] = -0.5*u[5]
    du[6] = -0.1u[6]
end
tspan = [0,64.0]
u0 = [1.0;1.0;1.0;1.0;1.0;1.0]
prob = ODEProblem(f_im,u0,tspan)
```

Set keywords parameters reltol=1e-4 abstol=1e-6. QNDF takes 7280 steps and gets 48 rejected steps, but the algorithm in this PR uses only 610 steps and gets 12 rejected steps. 

Moreover, in this problem, if we set the maximum order of QNDF to 3, it should be also theoretically quite efficient. However, the current QNDF doesn't behave like this. I will check the reason later.

I have not decided its name so only give a simple name now.  The event handling test is broken. It seems that this implementation now only arrives a little less accurate than expected. Further checks, benchmarks, and optimizations should still be made.


JunpengGao233:
For the dimension of inputs larger than 1, the results of this implementation and QNDF are wrong. I will fix them together in the next PR.

src/perform_step/bdf_perform_step.jl
```diff
@@ -983,3 +983,361 @@ end
...
+  integrator.fsalfirst = cache.fsalfirst
+  integrator.fsallast = du_alias_or_new(cache.nlsolver, integrator.fsalfirst)
+  resize!(integrator.k, integrator.kshortsize)
+  integrator.k[1] = integrator.fsalfirst
+  integrator.k[2] = integrator.fsallast
+  integrator.f(integrator.fsalfirst, integrator.uprev, integrator.p, integrator.t)
+  integrator.destats.nf += 1
+end
+
+function perform_step!(integrator, cache::FBDFCache{max_order}, repeat_step=false) where max_order
```
ChrisRackauckas, line 1196: why is it parameterized on the max order?

YingboMa: The internal coefficients calculation is done on `SArray`s, so `max_order` needs to be a compile time const, so that we know the upper bound of the size of `SArray`s.

JunpengGao233 (commented):
The previous test passed on my Macbook... For the out-of-place version of implementation, it takes around 6100 steps locally but here it shows the length of sol.t equal to 6509 in the previous test.

YingboMa:
The PSOS tests seem to be unrelated.

YingboMa:
Let's handle the event handling in the next PR.

PR #1453, opened 2021-07-22
Fix algevent tests and multi-dimensional input bugs in FBDF

JunpengGao233:
```julia
using OrdinaryDiffEq, DiffEqProblemLibrary.ODEProblemLibrary, Test
ODEProblemLibrary.importodeproblems()
prob = ODEProblemLibrary.prob_ode_2Dlinear
sol = solve(prob, FBDF(), abstol=1e-8,reltol=1e-8)
sol2 = solve(prob, Tsit5(),abstol=1e-8,reltol=1e-8)
@test sol.u[end] ≈ sol2.u[end]

> Test Passed
```

### DIFF
--- a/src/bdf_utils.jl
+++ b/src/bdf_utils.jl
@@ -113,7 +113,6 @@ function compute_weights!(ts,k,weights)
 end
 
 function calc_Lagrange_interp(k,weights,t,ts,u_history,u::Number)
-  #@show t,ts,u_history
   if t in ts
     i = searchsortedfirst(ts,t,rev=true)
     return u_history[i]
@@ -125,22 +124,18 @@ function calc_Lagrange_interp(k,weights,t,ts,u_history,u::Number)
       u *= t-ts[i]
     end
   end
-  #@show weights
   u
 end
 
 function calc_Lagrange_interp(k,weights,t,ts,u_history,u)
+  vc = _vec(u)
   if t in ts
     i = searchsortedfirst(ts,t,rev=true)
-    for j in eachindex(u)
-      u[j] = u_history[j,i]
-    end
+    @.. @views vc = u_history[:,i]
     return u
   else
     for i in 1:k+1
-      for j in eachindex(u)
-        u[j] += weights[i]/(t-ts[i])*u_history[j,i]
-      end
+      @.. @views vc += weights[i]/(t-ts[i])*u_history[:,i]
     end
     for i in 1:k+1
       @.. u *= t-ts[i]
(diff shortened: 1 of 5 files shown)


YingboMa (changes requested):
You should use `@..` whenever possible.

src/bdf_utils.jl
```diff
@@ -132,10 +132,15 @@ end
...
+    for j in eachindex(u)
+      u[j] = u_history[j,i]
+    end
+    return u
   else
     for i in 1:k+1
-      @.. u += weights[i]/(t-ts[i])*u_history[:,i]
+      for j in eachindex(u)
+        u[j] += weights[i]/(t-ts[i])*u_history[j,i]
+      end
```
ChrisRackauckas, line 143: I still see a lot of cases which are de-looped. Use `@view` to make the `@..` non-allocating. Changing to loops will make it not compatible with GPUs.

JunpengGao233: Oh, I do not get this into considerations. I will add @view later. 

Every column of `u_history` saves one history point solution. If `u0` and `u` is an array, 
```julia 
@.. u += weights[i]/(t-ts[i])*u_history[:,i]
```
will not calculate the same results as the loop using linear index. That's why I change to loop here. I will think more about how to deal with this.

src/perform_step/bdf_perform_step.jl
```diff
@@ -1015,21 +1015,21 @@ function perform_step!(integrator, cache::FBDFConstantCache{max_order}, repeat_s
...
     ts[2] = ts[1]
     ts[1] = t
     @.. u_history[:,2] = u_history[:,1]
-    @.. u_history[:,1] = _vec(uprev)
+    @.. u_history[:,1] = $_vec(uprev)
   elseif consfailcnt == 0
     for i in k+2:-1:2
       ts[i] = ts[i-1]
-      u_history[:,i] = u_history[:,i-1]
+      @.. u_history[:,i] = u_history[:,i-1]
```
YingboMa, line 1029: views

src/integrators/controllers.jl
```diff
@@ -580,7 +580,9 @@ function stepsize_controller!(integrator, alg::FBDF{max_order}) where max_order
         terk_tmp *= abs(dt^(k-2))
       else
         for i in 2:k-2
-          @.. @views terk_tmp += fd_weights[i,k-2] * u_history[:,i-1]
+          for j in eachindex(terk_tmp)
+            terk_tmp[j] += fd_weights[i,k-2] * u_history[j,i-1]
```
YingboMa, line 584: Any particular reason to use loops here?

JunpengGao233: No, just forgot it. I will carefully check git diff next time :(

PR #1456, opened 2021-07-26
Add residual to composite solution.

NAThompson:
This *doesn't work*, and is completely inelegant. But I figured I'd just open it to get some feedback-I'm not seeing the residual added to the `sol` object. So will this need to be a multiple repo change?


ChrisRackauckas:
You'd need to add this to SciMLBase.ODESolution. This is only the solution object for a very specific case (multi-algorithm solutions, i.e. algorithms that swap back and forth between different alg_choices. We should actually merge that into the original solution object some time)

NAThompson:
@ChrisRackauckas : Thanks!

PR #1465, opened 2021-08-06
Reduce the first time to solve from 5 seconds to 1 second for Tsit5

ChrisRackauckas:
Ya'll think you write good compilers? Well, I'm the compiler now!

```julia

function lorenz(du,u,p,t)
 du[1] = 10.0(u[2]-u[1])
end

u0 = [1.0;0.0;0.0]
tspan = (0.0,100.0)
alg = Tsit5()
itrig = itrigs[13]
ascend(itrig)
@time solve(prob,alg)

using ProfileView
ProfileView.view(flamegraph(tinf))

v5.60.2
InferenceTimingNode: 1.249748/4.881587 on Core.Compiler.Timings.ROOT() with 2 direct children

Before
InferenceTimingNode: 1.136504/3.852949 on Core.Compiler.Timings.ROOT() with 2 direct children

Without `@turbo`
InferenceTimingNode: 0.956948/3.460591 on Core.Compiler.Timings.ROOT() with 2 direct children

With `@inbounds @simd`
InferenceTimingNode: 0.941427/3.439566 on Core.Compiler.Timings.ROOT() with 2 direct children

With `@turbo`
InferenceTimingNode: 1.174613/11.118534 on Core.Compiler.Timings.ROOT() with 2 direct children

With `@inbounds @simd` everywhere
InferenceTimingNode: 0.760500/1.151602 on Core.Compiler.Timings.ROOT() with 2 direct children
```


ChrisRackauckas:
```julia
Laptop

Before:
InferenceTimingNode: 1.585750/5.363441 on Core.Compiler.Timings.ROOT() with 2 direct children

After:
InferenceTimingNode: 0.885957/1.254411 on Core.Compiler.Timings.ROOT() with 2 direct children
```

ChrisRackauckas:
LOL

```julia
Vern7
Before:
InferenceTimingNode: 5.962703/13.461966 on Core.Compiler.Timings.ROOT() with 1 direct children

After:
InferenceTimingNode: 2.979609/3.563301 on Core.Compiler.Timings.ROOT() with 2 direct children
```

ChrisRackauckas:
```julia
Vern9
Before:
InferenceTimingNode: 17.960255/23.513515 on Core.Compiler.Timings.ROOT() with 2 direct children

After:
InferenceTimingNode: 6.969864/7.531495 on Core.Compiler.Timings.ROOT() with 2 direct children
```

timholy:
Yep, broadcasting is quite expensive for the compiler. Sorry you had to write all this out by hand, but nice outcome!

ChrisRackauckas:
Writing it out by hand is fine. The fact that I cannot do the same for RecursiveFactorization.jl's compile times bug me though...

timholy:
In the next month I'm planning to trying to go through some of https://github.com/JuliaLang/julia/issues?q=is%3Aopen+is%3Aissue+label%3Aprecompile. That might make the inference part go away. Any overhead due to codegen/LLVM won't be helped though (yet).

ChrisRackauckas:
From these studies and https://github.com/SciML/OrdinaryDiffEq.jl/pull/1467, https://github.com/SciML/OrdinaryDiffEq.jl/pull/1467, and https://github.com/JuliaSIMD/TriangularSolve.jl/pull/8, the biggest thing for us would be to figure out why the RecursiveFactorization/TriangularSolve/LoopVectorization stack won't cache the precompiles. In some sense it should be easy: the lowest level is just functions on `Vector{Float64}`? But I wonder if the caches being an `__init__()` time static array is causing issues there. This 20 seconds remaining is almost entirely due to this, and it's the reason why https://discourse.julialang.org/t/catalyst-differentiaequations-startup-time/65946/8 can't be solved with a few cheap tricks like this. I was going to ask @chriselrod if there was another way to try and do the caching.

chriselrod:
I suspect a big part of the problem is that LoopVectorization owns the `_turbo_!` calls / there's nothing to root the inference results into the calling libraries?

timholy:
AFAICT `__init__` is never cached: https://github.com/JuliaGraphics/Gtk.jl/issues/466#issuecomment-891813872. Figuring out whether that's an oversight or deliberate is high on the agenda.

@ChrisRackauckas, you duplicated #1467 up there, feel free to edit and then I will take a look.

ChrisRackauckas:
Edited. Ahh yes, that's the one piece I in the chain I didn't setup to precompile! I'll go add something to DiffEqBase and see if that handles it.

PR #1470, opened 2021-08-11
Define prepare_alg dispatch for OrdinaryDiffEq AD algorithms

ChrisRackauckas:
Requires https://github.com/SciML/DiffEqBase.jl/pull/698


chriselrod:
This may be the primary cause of the inference regression in Pumas.

ChrisRackauckas:
This PR causes the biggest decrease to compile times 😅

chriselrod:
Well, commenting out `prepare_alg` reduced inference of the first `solve` on a single subject in Pumas from 16.8 down to 9.7 seconds.

ChrisRackauckas:
But having this allows for the things to fully precompile, which is the full 22 to 3 seconds precompile 🌊

chriselrod:
Yeah, I think a better fix is to just avoid the type instability in Pumas

PR #1506, opened 2021-10-20
set proposed dt in `auto_reset_dt!`

hexaeder:
Right now, `auto_reset_dt!` does not actually change the next timestep when used inside a callback for example.

I think it would be also nice to think about integrating
```julia
  if integrator.opts.adaptive
    auto_dt_reset!(integrator)
  end
```
next to 

https://github.com/SciML/OrdinaryDiffEq.jl/blob/02f85eae4b281ecdd9db4572aa0439694d5e0c61/src/integrators/integrator_interface.jl#L40-L42

so the adaptive timestepping is reset automaticially after introducing a discontinuity.
I think this is a saner default for callbacks. Well at least in my case, where I often introduce a pertubation to a system in a fixpoint. In the fixpoint the adaptive timestep is to big to capure the behaviour of the system right after the callback.

### DIFF
--- a/src/integrators/integrator_interface.jl
+++ b/src/integrators/integrator_interface.jl
@@ -330,6 +330,7 @@ function DiffEqBase.auto_dt_reset!(integrator::ODEIntegrator)
   integrator.tdir,integrator.opts.dtmax,integrator.opts.abstol,integrator.opts.reltol,
   integrator.opts.internalnorm,integrator.sol.prob,integrator)
   integrator.destats.nf += 2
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
>So maybe instead of using set_proposed_dt!(integrator, dt) one should just set integrator.dtpropose = dt inside the auto_dt_reset!?

That sounds right. `dtcache` is for fixed time integrators to remember the `dt` after changing it around `tstops`.

PR #1515, opened 2021-10-29
Misc type stability improvements.

chriselrod:
Between this PR, 
https://github.com/SciML/DiffEqBase.jl/pull/712
and some local changes I have on the Pumas symblics update, I am one issue away from making the first `simobs` call from making the ad tests type stable.
The remaining issue is this forced dynamic dispatch:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/60f3401909c56a41b9ea647f1c2761f638092e0d/src/alg_utils.jl#L174

So I/someone else needs to make another PR that adds all the plumbing needed to support (in a type stable way) supplying the chunk size from an external repo, and supplying it to `remake` there.

Once that's all said and done, inference and total compile time for that method will go from around 24 -> 12 seconds.
Without the type unstable chunk size, as is currently the case, inference and compile time are both about 19 seconds.

Single threaded minimum (i.e. ignoring G.C.) runtime is mostly unaffected, but memory allocations are much less, so multi-threaded throughput should benefit.

Odds are the `Val` vs `Type` thing here will cause test failures, so I'll fix those later.
I did not actually run tests locally.

Two reasons for the `Val` vs `Type` change:
1) `Val`s force specialization.
2) `Val`s are 0-byte instances, while `Type`s are 80 bytes. If you pass a type to a function that is not inlined, you need to pass the 80 byte object (by reference, at least).

For performance reasons, `::Val{T}` should thus be preferred over `::Type{T}` unless you guarantee inlining and specialization, in which case they're equivalent. Neither was the case here.
(Also, means that `::Type{Val{T}}` should be avoided, too.)

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -36,7 +36,7 @@ function DiffEqBase.remake(thing::Union{OrdinaryDiffEqAdaptiveImplicitAlgorithm{
                         OrdinaryDiffEqImplicitAlgorithm{CS,AD,FDT},
                         DAEAlgorithm{CS,AD,FDT}}; kwargs...) where {CS, AD, FDT}
   T = SciMLBase.remaker_of(thing)
-  T(; chunk_size=CS,autodiff=AD,SciMLBase.struct_as_namedtuple(thing)...,kwargs...)
+  T(; chunk_size=CS,autodiff=Val(AD),SciMLBase.struct_as_namedtuple(thing)...,kwargs...)
 end
 
 ###############################################################################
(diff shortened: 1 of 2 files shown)


src/algorithms.jl
```diff
@@ -36,7 +36,7 @@ function DiffEqBase.remake(thing::Union{OrdinaryDiffEqAdaptiveImplicitAlgorithm{
                         OrdinaryDiffEqImplicitAlgorithm{CS,AD,FDT},
                         DAEAlgorithm{CS,AD,FDT}}; kwargs...) where {CS, AD, FDT}
   T = SciMLBase.remaker_of(thing)
-  T(; chunk_size=CS,autodiff=AD,SciMLBase.struct_as_namedtuple(thing)...,kwargs...)
+  T(; chunk_size=CS,autodiff=Val(AD),SciMLBase.struct_as_namedtuple(thing)...,kwargs...)
 end
```
chriselrod, line 40: This constructor was type unstable, which was the motivation for the `_unwrap_val` change.

Be aware that while `AD` is a compile time constant here, passing it as a runtime value to `T` makes it a runtime value there.
It's possible for it to be inlined and const prop to take care of it, but that isn't guaranteed, and in fact was not happening, making this code needlessly type unstable.

ChrisRackauckas: Awesome. For the tests of https://github.com/SciML/DifferentialEquations.jl/issues/786 it was showing inference working, but it's always better to make it explicit anyways.

chriselrod: Yeah, this should be less brittle to different algorithms / Julia versions.
I was on Julia 1.6 with Rosenbrock23 (as part of a composite algorithm with Tsit5 IIRC).

PR #1518, opened 2021-11-02
Drop AbstractFloat condition when snapping integrator.t to the tstop value

andreasnoack:
Originally, this condition was introduced to avoid calling
eps(::Integer) but this is no longer a risk since eps is only computed
after a division operation that will promote to float.

This change fixes the case the time is a ForwardDiff.Dual and the stepping
is slightly off because of rounding errors.

I haven't added a test as it's not obvious how to construct a simple test case for
this. I've verified that the changes here fixes the more complicated case that
revealed the issue.


ChrisRackauckas:
Real test failure. Probably easiest to just exclude Integer and Rational.

andreasnoack:
I've added a `float` before the `eps` which took care of the error caused by the Rational.

ChrisRackauckas:
Floating point is fun.

PR #1520, opened 2021-11-04
Directly use PreallocationTools.jl

ChrisRackauckas:


src/initialize_dae.jl
```diff
@@ -291,9 +291,9 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::BrownFullBasicInit,
 
   isad = alg_autodiff(integrator.alg)
   if isad
-    chunk = Val{ForwardDiff.pickchunksize(count(algebraic_vars))}
-    _tmp = DiffEqBase.dualcache(tmp, chunk)
-    _du_tmp = DiffEqBase.dualcache(tmp, chunk)
+    chunk = count(algebraic_vars)
```
thomvet, line 294: Is this fine? Doesn't this potentially give very large chunk sizes? Depending on the problem that is supplied? ForwardDiff still caps chunk size at 12 as far as I know.

ChrisRackauckas: oh oops, I only meant to remove the `Val` part.

PR #1532, opened 2021-12-07
(WIP) try immutable WOperator

ChrisRackauckas:
```julia
using OrdinaryDiffEq, StaticArrays, ParameterizedFunctions, BenchmarkTools
rober = @ode_def begin
  dy₁ = -k₁*y₁+k₃*y₂*y₃
  dy₂ =  k₁*y₁-k₂*y₂^2-k₃*y₂*y₃
  dy₃ =  k₂*y₂^2
end k₁ k₂ k₃
prob = ODEProblem{false}(rober,SA[1.0,0.0,0.0],(0.0,1e5),SA[0.04,3e7,1e4])
@btime sol = solve(prob,Rosenbrock23(chunk_size=Val{3}()),save_everystep=false)

mutable struct WOperator:
# 10.000 μs (87 allocations: 16.47 KiB)

struct WOperator:
# 8.900 μs (26 allocations: 4.67 KiB)
```

seems like this could be useful for Pumas cases @chriselrod

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -181,7 +181,7 @@ to be a diffeq operator --- it will automatically be converted to one.
 internal cache (can be specified in the constructor; default to regular `Vector`).
 It supports all of `AbstractDiffEqLinearOperator`'s interface.
 """
-mutable struct WOperator{IIP,T,
+struct WOperator{IIP,T,
   MType,
   GType,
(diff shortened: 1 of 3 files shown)


chriselrod:
>+10% performance gain is nice. Any idea why `mutable struct` -> `struct` made such a large difference here?

PR #1537, opened 2021-12-09
microoptimize by replacing division with multiplication

oscardssmith:
This probably needs a benchmark to confirm this is faster. Any suggestions?
Note that this doesn't work if `dt` is subnormal, but that should never happen.

### DIFF
--- a/src/perform_step/rosenbrock_perform_step.jl
+++ b/src/perform_step/rosenbrock_perform_step.jl
@@ -388,9 +388,10 @@ end
   @unpack a21,a31,a32,C21,C31,C32,b1,b2,b3,btilde1,btilde2,btilde3,gamma,c2,c3,d1,d2,d3 = cache.tab
 
   # Precalculations
-  dtC21 = C21/dt
-  dtC31 = C31/dt
-  dtC32 = C32/dt
+  invdt = inv(dt)
+  dtC21 = C21*invdt
+  dtC31 = C31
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Run at least ROBER https://benchmarks.sciml.ai/html/StiffODE/ROBER.html

ChrisRackauckas:
If you want to micro-optimize a bit, use this example:

```julia
using OrdinaryDiffEq, StaticArrays, BenchmarkTools, Sundials, LSODA
function rober_static(u,p,t)
    y₁,y₂,y₃ = u
    k₁,k₂,k₃ = p
    du1 = -k₁*y₁+k₃*y₂*y₃
    du2 =  k₁*y₁-k₂*y₂^2-k₃*y₂*y₃
    du3 =  k₂*y₂^2
    SA[du1,du2,du3]
end
prob = ODEProblem{false}(rober_static,SA[1.0,0.0,0.0],(0.0,1e5),SA[0.04,3e7,1e4])
@btime solve(prob,Rodas5(),reltol=1.0e-8,abstol=1.0e-8, saveat = 1000) # 42.300 μs (29 allocations: 9.75 KiB)
```

oscardssmith:
That doesn't find any difference.

ChrisRackauckas:
😅 . Then.

ChrisRackauckas:
Good try though.

PR #1538, opened 2021-12-10
Return a bunch of nothings where possible to improve inference

ChrisRackauckas:


oscardssmith:
This isn't needed. Since the return values aren't used, the type stability doesn't matter.

PR #1540, opened 2021-12-11
Make the chunksize determination static when static 

ChrisRackauckas:
A bit more of a sane version of https://github.com/SciML/OrdinaryDiffEq.jl/pull/1536

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -171,8 +171,20 @@ function DiffEqBase.prepare_alg(alg::Union{OrdinaryDiffEqAdaptiveImplicitAlgorit
     if typeof(alg) <: OrdinaryDiffEqImplicitExtrapolationAlgorithm
       return alg # remake fails, should get fixed
     else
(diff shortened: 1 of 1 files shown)


src/alg_utils.jl
```diff
@@ -171,10 +171,22 @@ function DiffEqBase.prepare_alg(alg::Union{OrdinaryDiffEqAdaptiveImplicitAlgorit
...
+      else # statically sized
+        cs = pick_static_chunksize(Val{L}())
+        remake(alg,chunk_size=cs)
+      end
     end
 end
 
+@generated function pick_static_chunksize(::Val{chunksize}) where chunksize
+  x = ForwardDiff.pickchunksize(chunksize)
+  :(Val{$x}())
```
chriselrod, line 187: FWIW, Pumas has been relying on const prop for this, and that seems to be fine so far.

ChrisRackauckas: IIRC it caused a major performance change here 😅

chriselrod: I should double check Pumas, then.

PR #1544, opened 2021-12-11
Chunk size 1 when eltype is already dual

chriselrod:
Would probably be worth actually running a variety of benchmarks here and picking between strategies.

For now, I'm setting the chunk size to `1` if `sizeof(eltype(u0)) > 24`, which would (for example) mean the element type are duals of `Float64` with 3 or more partials.

Pumas hasn't been eager to merge any of the chunk size PRs as they cause a few regressions when applied globally.
But we can apply it more selectively here when choosing the chunk size as a function of the eltype.

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -181,6 +181,12 @@ function DiffEqBase.prepare_alg(alg::Union{OrdinaryDiffEqAdaptiveImplicitAlgorit
       end
     end
 end
+function DiffEqBase.prepare_alg(alg::Union{OrdinaryDiffEqAdaptiveImplicitAlgorithm{0,AD,FDT},
+                        OrdinaryDiffEqImplicitAlgorithm{0,AD,FDT},
+                        DAEAlgorithm{0,AD,FDT}},u0::AbstractArray{<:ForwardDiff.Dual},p,prob) where {AD,FDT}
+# If our element type consists of dual numbers already, we prefer a chunk size of `1`.
+  remake(alg,chunk_size=Val(1))
+end
 
 @generated function pick_static_chunksize(::Val{chunksize}) where chunksize
   x = ForwardDiff.pickchunksize(chunksize)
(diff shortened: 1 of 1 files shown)


src/alg_utils.jl
```diff
@@ -181,12 +182,6 @@ function DiffEqBase.prepare_alg(alg::Union{OrdinaryDiffEqAdaptiveImplicitAlgorit
       end
     end
 end
-function DiffEqBase.prepare_alg(alg::Union{OrdinaryDiffEqAdaptiveImplicitAlgorithm{0,AD,FDT},
```
YingboMa: Accidental?

chriselrod: I decided to change the approach to not set chunk size 1 for small enough duals.

ChrisRackauckas:
Need to exclude or fix the extrapolation methods.

chriselrod:
> Need to exclude or fix the extrapolation methods.

Done.

ChrisRackauckas:
Looks like a real test failure: https://github.com/SciML/OrdinaryDiffEq.jl/runs/4498861629?check_suite_focus=true#step:6:499

chriselrod:
Only remaining failure
```julia
 Test Summary:          | Pass  Total
Non-diagonal SDE Tests |    2      2
└ @ SciMLBase ~/.julia/packages/SciMLBase/OpCVa/src/integrator_interface.jl:345
 61.386006 seconds (108.90 M allocations: 6.584 GiB, 3.04% gc time, 97.67% compilation time)
┌ Warning: Instability detected. Aborting
└ @ SciMLBase ~/.julia/packages/SciMLBase/OpCVa/src/integrator_interface.jl:351
No Index Tests: Test Failed at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.7/Test/src/Test.jl:806
  Expression: isempty(stderr_content)
   Evaluated: isempty("┌ Warning: Instability detected. Aborting\n└ @ SciMLBase ~/.julia/packages/SciMLBase/OpCVa/src/integrator_interface.jl:351\n")
Stacktrace:
 [1] macro expansion
   @ /opt/hostedtoolcache/julia/1.7.0/x64/share/julia/stdlib/v1.7/Test/src/Test.jl:445 [inlined]
 [2] top-level scope
   @ ~/work/OrdinaryDiffEq.jl/OrdinaryDiffEq.jl/downstream/test/noindex_tests.jl:48
Test Summary:  | Pass  Fail  Total
No Index Tests |    5     1      6
```

ChrisRackauckas:
Ignore that. It's part of https://github.com/SciML/StochasticDiffEq.jl/pull/445

PR #1545, opened 2021-12-12
Add a standardtag mechanism so that more precompilation occurs

ChrisRackauckas:
Requires https://github.com/JuliaDiff/SparseDiffTools.jl/pull/170


ChrisRackauckas:
@chriselrod

chriselrod (approved):
LGTM, do you have any timings?

ChrisRackauckas:
Not yet, first I'm going to get tests to pass. I think I want to do this anyways just because it greatly improves the stacktraces and it's not unsafe because the eltype is in the tag still so nesting OrdinaryDiffEq on itself still works. Literally the only way to get perturbation confusion is for someone to use ForwardDiff inside of the `f` function and purposefully use `OrdinaryDiffEq.OrdinaryDiffEqTag()` (since it needs to use the same module too), which is so insane that it would only happen adversarially. So, shorter tags it is.

chriselrod:
Did you define tag ordering somewhere?

ChrisRackauckas:
No, if you do the tags correctly you never have to do that. You have to do it like `SMALLTAG = typeof(ForwardDiff.Tag(DiffEqTag(),Float64))`.

ChrisRackauckas:
```julia
function lorenz(du,u,p,t)
 du[1] = 10.0(u[2]-u[1])
end

u0 = [1.0;0.0;0.0]
tspan = (0.0,100.0)

alg = Rodas5(chunk_size = Val{3}())

# Before:
InferenceTimingNode: 1.043785/1.836156 on Core.Compiler.Timings.ROOT() with 4 direct children

# After:
InferenceTimingNode: 1.081242/1.899589 on Core.Compiler.Timings.ROOT() with 6 direct children
```

Compile times are actually a little bit slower because of the inference trigger on `t = similar(x,Dual{T,eltype(x),length(first(first(p)))})` in SparseDiffTools, but the change is well worth it because the stacktraces are absolutely beautiful now.

PR #1547, opened 2021-12-14
Remove a bunch of inline annotations

ChrisRackauckas:
Julia's inline cost model keeps getting better and better, so it's generally not a good idea to mess with it these days.


ChrisRackauckas:
This actually makes compile times slower? Ehh.

Issue #1551, opened 2021-12-27
Finishing the linear solve and Krylov interfaces


ChrisRackauckas:
I plan to finally finish the linear solver Krylov preconditioner etc. interfaces this week. The big breaking LinearSolve.jl change https://github.com/SciML/OrdinaryDiffEq.jl/pull/1543 of course was a prerequisite to doing this correctly, and so now we're set to solve https://github.com/SciML/DifferentialEquations.jl/issues/453 and https://github.com/SciML/DifferentialEquations.jl/issues/521 at the same time. The test cases are:

- https://github.com/SciML/DiffEqBase.jl/issues/589
- https://github.com/SciML/OrdinaryDiffEq.jl/issues/722
- https://github.com/SciML/OrdinaryDiffEq.jl/issues/1430

The plan for this last little part is (mostly documenting for myself to remember the whole plan a few days from now 😅)

- Add and document algorithm traits for LinearSolve.jl algorithms for `needsconcreteA(alg)` for whether a concrete `A` is required or whether only the `A*x` is used. 
- Internally to OrdinaryDiffEq, https://github.com/SciML/OrdinaryDiffEq.jl/blob/v6.0.3/src/derivative_utils.jl#L680 `elseif jac_prototype === nothing && needsconcreteA(alg.linsolve)`, automatically choose `J` a JacVecOperator matching the autodiff setting.  This would make it the same as the user choosing `jac_prototype = JacVecOperator()`, but now if they choose GMRES it will automatically make this swap and set autodiff appropriately. This is then inverted for a good reason: if the user wants to specify their own matrix-free operator or approximate Jacobian, then they just as before set `jac_prototype` and `jac` 
- Algorithms then get a `precs` argument for a function `Pl,Pr = alg.precs(W,du,u,p,t,newW,solverdata)` for building preconditioners which are dependent on the current `W` matrix. To handle the type-groundedness for the solver's init phase, `newW` will be passed as `nothing` to allow a separate dispatch for handling that phase. `solverdata` is kind of an escape hatch so future additions can be non-breaking. Currently it's just set to include `gamma` for the `W = I - gamma*J` since that will be needed for @tjdiamandis's randomized preconditioners, but that list of "things from the solver someone needs" can easily grow, and so it's non-breaking to add more little scalars to the struct in the future. `dt`, etc. who knows.
[...]

PR #1552, opened 2021-12-28
Automate the Jacobian-free part of JFNK

ChrisRackauckas:


ChrisRackauckas:
This fixes a good chunk of https://github.com/SciML/OrdinaryDiffEq.jl/issues/1551

PR #1555, opened 2021-12-31
New dense output for rodas5

gstein3m:


ChrisRackauckas:
The interpolation, when checked, should get an interpolation regression test:

https://github.com/gstein3m/OrdinaryDiffEq.jl/blob/master/test/regression/ode_dense_tests.jl#L439

A special interps test to compare the oop and iip versions

https://github.com/gstein3m/OrdinaryDiffEq.jl/blob/master/test/regression/special_interps.jl

and the dense convergence L2 should be double checked.

https://github.com/gstein3m/OrdinaryDiffEq.jl/blob/master/test/algconvergence/ode_rosenbrock_tests.jl#L399-L408

ChrisRackauckas:
It will also need an `add_steps!` test: https://github.com/gstein3m/OrdinaryDiffEq.jl/blob/master/test/integrators/ode_add_steps_tests.jl

and an `add_steps!` implementation

https://github.com/gstein3m/OrdinaryDiffEq.jl/blob/master/src/dense/stiff_addsteps.jl

add_steps! is kind of a kludge because it's a post-event handling rebuild of the interpolation which is only required for the algorithms with special interpolations, and there's not a good way to reuse the normal stepping code.

gstein3m:
In one addsteps function of Rodas5 I had to include a call to f, because the fsalfirst seemed to be false:

    dtgamma = dt*gamma
#!!!!!
    f( fsalfirst,  uprev, p, t)
    @inbounds @simd ivdep for i in eachindex(u)
      linsolve_tmp[i] = @muladd fsalfirst[i] + dtgamma*dT[i]
    end

ChrisRackauckas:
Yes, that's expected because it's the cache reset call, so the left-hand `f` is no longer correct when that is called.

test/integrators/ode_add_steps_tests.jl
```diff
@@ -32,7 +32,9 @@ for inplace in [false,true], alg in algs
     global cur_itr
     prob = ODEProblem{inplace}(test_ode, [0.], tspan, [1.])
     sol = solve(prob, alg(); callback=cb,dt=0.0013)
+    println(sol.destats);
```
ChrisRackauckas, line 35:
```suggestion
```

test/integrators/ode_add_steps_tests.jl
```diff
@@ -32,7 +32,9 @@ for inplace in [false,true], alg in algs
...
     pass = all(isapprox(sol(t)[1], test_solution(t); atol=0.05) for t in testtimes)
+    println(alg," ",pass," ",sol(9.0)[1]," ",test_solution(9.0))
```
ChrisRackauckas, line 37:
```suggestion
```

ChrisRackauckas:
Checking in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1557 if the downstream etc. issues are on master or induced by this PR.

ChrisRackauckas:
https://buildkite.com/julialang/ordinarydiffeq-dot-jl/builds/743#d1cd7ebb-5b5f-4437-a303-daa310f40fca/353-806

That looks like a real test failure and not from master, but I'll look into that.

ChrisRackauckas:
https://buildkite.com/julialang/ordinarydiffeq-dot-jl/builds/745#18b57e4c-aecf-4892-aea1-75374125af1f/371-965

Looks like the add_steps! test fails. 

https://github.com/SciML/OrdinaryDiffEq.jl/runs/4715772426?check_suite_focus=true

Looks like that just needs a tolerance adjustment. Can fix after merging.

https://github.com/SciML/OrdinaryDiffEq.jl/runs/4715772299?check_suite_focus=true#step:6:620

Looks like a win, and a tolerance thing.

So yeah, don't worry about the downstream tests, it looks like this was a fix for a few things 🎉, but there's still a failure.

gstein3m:
The tests ode_dense_test, special_interps, ode_rosenbrock_tests passed well.

I failed to adopt stiff_add_steps, because  type Rosenbrock5ConstantCache has no field autodiff in contrast ro rodas4 ?

gstein3m:
In my environment the stiff_add_steps test passed well: 
Rosenbrock32 true
Rodas4 true
Rodas5 true
Rosenbrock32 true
Rodas4 true
Rodas5 true
Test Passed

ChrisRackauckas:
Are you running the rebased version?

ChrisRackauckas:
oh shit... in the future please do not PR from master... 😅

PR #1573, opened 2022-01-19
Move default alg choice sooner to improve inner loop stability

ChrisRackauckas:


ChrisRackauckas:
```julia
using OrdinaryDiffEq, ParameterizedFunctions, BenchmarkTools
using LinearSolve

f = @ode_def Orego begin
  dy1 = p1*(y2+y1*(1-p2*y1-y2))
  dy2 = (y3-(1+y1)*y2)/p1
  dy3 = p3*(y1-y3)
end p1 p2 p3

p = [77.27,8.375e-6,0.161]
prob = ODEProblem(f,[1.0,2.0,3.0],(0.0,30.0),p)
@btime solve(prob,Rosenbrock23(),abstol=1e-6,reltol=1e-6)
# Before:
# 1.405 ms (23476 allocations: 1.95 MiB)
# After:
# 1.423 ms (23478 allocations: 1.95 MiB)

@btime solve(prob,Rosenbrock23(linsolve=RFLUFactorization()),abstol=1e-6,reltol=1e-6)
# Before:
# 1.262 ms (10521 allocations: 952.25 KiB)
# After:
# 1.252 ms (10521 allocations: 952.25 KiB)
```

It's surprising this isn't helpful while making the choice is.

ChrisRackauckas:
@chriselrod @YingboMa that seems to be something in RecursiveFactorization or JuliaSIMD stuff. Don't know if there's a workaround https://github.com/JuliaLang/julia/issues/35800

chriselrod:
> @chriselrod @YingboMa that seems to be something in RecursiveFactorization or JuliaSIMD stuff. Don't know if there's a workaround [JuliaLang/julia#35800](https://github.com/JuliaLang/julia/issues/35800)

Yeah, that's why I removed the precompilation from TriangularSolve.jl.
With precompilation, performance was bad until you forced it to recompile with Revise.jl.
Without the precompilation, it was fast the first time. =/

I don't know inference or the compiler well enough to know what causes.
Jameson suggested it probably means the code is bad.
I define code quality mostly based on outcomes, so that seems true by definition and not helpful for finding the problem and fixing it.

YingboMa:
The failure looks real.

src/alg_utils.jl
```diff
@@ -161,22 +161,42 @@ function DiffEqBase.prepare_alg(alg::Union{OrdinaryDiffEqAdaptiveImplicitAlgorit
...
-    x = if prob.f.colorvec === nothing
-      length(u0)
+
+    if alg.linsolve === nothing
+      if (prob.f isa ODEFunction && prob.f.f isa SciMLBase.AbstractDiffEqOperator) ||
+         (prob.f isa SplitFunction && prob.f.f1 isa SciMLBase.AbstractDiffEqOperator)
+
+        linsolve = LinearSolve.defaultalg(prob.f.f,u0)
+      elseif prob.f.mass_matrix === nothing || (prob.f.mass_matrix !== nothing &&
+                              typeof(prob.f.mass_matrix) == prob.f.jac_prototype)
```
YingboMa, line 171:
```suggestion
      elseif (!@isdefined(prob.f, :mass_matrix)) || (prob.f.mass_matrix === nothing || (prob.f.mass_matrix !== nothing &&
                              typeof(prob.f.mass_matrix) == prob.f.jac_prototype))
```

PR #1574, opened 2022-01-19
Use NonlinearSolve.jl for DAE initialization

ChrisRackauckas:
This fixes master tests because NLsolve.jl does not seem to directly support static arrays, and is moving the code in the right direction anyways. For the full version, we need https://github.com/SciML/SciMLNLSolve.jl to complete registration.


src/initialize_dae.jl
```diff
@@ -250,12 +250,14 @@ function _initialize_dae!(integrator, prob::DAEProblem,
     f((u-u0)/dt,u,p,t)
   end
 
-  nlequation! = @closure (out,u) -> out .= nlequation_oop(u)
+  nlequation = @closure (u,_) -> nlequation_oop(u)
```
YingboMa, line 253: This doesn't seems to create a closure.

src/initialize_dae.jl
```diff
@@ -250,12 +250,14 @@ function _initialize_dae!(integrator, prob::DAEProblem,
...
 
   resid = f(integrator.du,u0,p,t)
   integrator.opts.internalnorm(resid,t) <= integrator.opts.abstol && return
 
-  integrator.u = nlsolve(nlequation!, u0).zero
+  nlprob = NonlinearProblem(nlequation,u0)
+  sol = solve(nlprob,NewtonRaphson(autodiff=false))
```
YingboMa, line 259: Can we get autodiff from `unwrap_alg(integrate)` or sth?

ChrisRackauckas: No, because NonlinearSolve will need an update for its autodiff to be compatible on the same functions do to how it under-types the duals. It needs standard tags, etc. That's a big update but I'll do it next.

PR #892, opened 2019-08-23
Disable threading in ImplicitEulerExtrapolation

saurabhkgp21:

PR #1579, opened 2022-01-21
Enable multi-threading in ImplictEulerExtrapolation

utkarsh530:


src/algorithms.jl
```diff
@@ -94,7 +94,7 @@ end
 
 function ImplicitEulerExtrapolation(;chunk_size=Val{0}(),autodiff=true, standardtag = Val{true}(), concrete_jac = nothing,
     diff_type=Val{:forward},linsolve=nothing,precs = DEFAULT_PRECS,
-    max_order=12,min_order=3,init_order=5,threading=true,sequence = :bulirsch)
+    max_order=12,min_order=3,init_order=5,threading=false,sequence = :harmonic)
```
ChrisRackauckas, line 97: why default to false while the others default to true?

src/algorithms.jl
```diff
@@ -103,18 +103,23 @@ function ImplicitEulerExtrapolation(;chunk_size=Val{0}(),autodiff=true, standard
     n_min = max(3,min_order)
     n_init = max(n_min + 1,init_order)
     n_max = max(n_init + 1, max_order)
-    if threading
-      @warn "Threading in `ImplicitEulerExtrapolation` is currently disabled. Thus `threading` has been changed from `true` to `false`."
-      threading = false
```
ChrisRackauckas, line 108: Nothing needs to change in the implementation?

src/algorithms.jl
utkarsh530: Extrapolation methods have default threading to false. https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/algorithms.jl#L280

ChrisRackauckas: https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/algorithms.jl#L79 missed one then.

src/algorithms.jl
utkarsh530: Yep. Everything is there. Ran tests as well as benchmarks. This was on bruss problem:
download

I really don't know why it was turned off: 
https://github.com/SciML/OrdinaryDiffEq.jl/pull/892

PR #1583, opened 2022-01-22
Convert tstops and saveat to time types

ChrisRackauckas:
Found as a potential issue in https://github.com/SciML/DiffEqFlux.jl/issues/675


ChrisRackauckas:
This shouldn't be needed because it happens implicitly in the push! due to the type of the heaps.

Issue #1587, opened 2022-01-27
Rosenbrock23 performs 3 rhs evaluations per step


simonbyrne:
```
using OrdinaryDiffEq

function f!(du, u, p, t)
    println("f!: t=$t")
    du .= u
end

prob = ODEProblem(
    ODEFunction(
        f!;
        jac = (J, u, p, t) -> (J[1,1] = 1),
        jac_prototype = zeros(1,1),
        tgrad = (dT, u, p, t) -> fill!(dT, 0),
    ),
    [1.0],
    (0.0, 5.0),
)

sol = solve(
    prob,
    Rosenbrock23(),
    dt = 1.0,
    adaptive = false,
)
```

gives

```
f!: t=0.0
f!: t=0.0
f!: t=0.5
f!: t=1.0
f!: t=1.0
f!: t=1.5
f!: t=2.0
f!: t=2.0
f!: t=2.5
f!: t=3.0
f!: t=3.0
f!: t=3.5
f!: t=4.0
f!: t=4.0
f!: t=4.5
```

The 3 calls appear to be:

1. once in reset_fsal!  at time t https://github.com/SciML/OrdinaryDiffEq.jl/blob/774b202f20d63bd1011cfc0dc04d7e8035730b45/src/integrators/integrator_utils.jl#L391
2. once in calc_tderivative! also at time t https://github.com/SciML/OrdinaryDiffEq.jl/blob/0f74871e199148deec548d78feeff0b0ed9ad621/src/derivative_utils.jl#L37
3. once in perform_step! at time t+dt/2 https://github.com/SciML/OrdinaryDiffEq.jl/blob/e82fa88a79252357340e797f0ac18ca5dac55d38/src/perform_step/rosenbrock_perform_step.jl#L61

ChrisRackauckas:
The `calc_tderivative!` one looks like a mistake that hid for 4 years, mostly because some tests would fail because they were assuming FSAL when they shouldn't have. A bit of clean up can remove that.

simonbyrne:
Thanks, much appreciated!

PR #1588, opened 2022-01-27
remove extra FSAL evaluation in calc_tderivative!

ChrisRackauckas:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1587

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -34,7 +34,6 @@ function calc_tderivative!(integrator, cache, dtd1, repeat_step)
       end
     end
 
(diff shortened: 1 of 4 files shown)


src/derivative_utils.jl
```diff
@@ -34,7 +34,6 @@ function calc_tderivative!(integrator, cache, dtd1, repeat_step)
       end
     end
 
-    f(fsalfirst, uprev, p, t)
     integrator.destats.nf += 1
```
ranocha, line 37:
```suggestion
```
Shouldn't the counter update be removed, too, if `f` isn'tcalled anymore?

PR #1593, opened 2022-02-03
fix problem with zero EEst and PIDController

ranocha:
I noticed some trouble with `PIDController`s when the error estimates vanish. This fixes the problem.

### DIFF
--- a/test/interface/controllers.jl
+++ b/test/interface/controllers.jl
@@ -4,6 +4,9 @@ f(u,p,t) = zero(u)
 prob = ODEProblem(f, 1.0, (0.0, 10.0))
 
 # Shouldn't error
+sol = solve(prob, Tsit5(), controller=IController())
+@test sol.retcode === :Success
+
 sol = solve(prob, Tsit5(), controller=PIController(7//50, 2//25))
 @test sol.retcode === :Success
 


test/runtests.jl
```diff
@@ -38,6 +38,7 @@ if GROUP == "All" || GROUP == "InterfaceI" || GROUP == "Interface"
   @time @safetestset "Composite Interpolation Tests" begin include("interface/composite_interpolation.jl") end
   @time @safetestset "Export tests" begin include("interface/export_tests.jl") end
   @time @safetestset "Type Handling Tests" begin include("interface/type_handling.jl") end
+  @time @safetestset "Controller Tests" begin include("interface/controllers.jl") end
```
ChrisRackauckas, line 41: oh no, was that missing before? Does this have a test case for this?

ranocha: I added this test case here (but noticed that I forgot to push it in the first commit - fixed in e886266e). This file contains a new test case for this specific case.

ChrisRackauckas: Ahh okay. Slap a few other controllers in there and see if they have a similar issue with 0 EEst?

ChrisRackauckas (approved):

test/runtests.jl
ranocha: I did that - everything looks good so far

Issue #1596, opened 2022-02-09
QNDF and QBDF solve fails where Rodas4 has success 


jd-lara:
When trying to solve stiff systems with M-matrix, the fast methods fail after a callback into the Algebraic part where Rodas is successful. From the multistep methods only FBDF return a successful result. 

Tested methods: 
QNDF()
QNDF2()
QBDF()
QBDF2()

MWE 

```julia
using OrdinaryDiffEq
using LinearAlgebra
using NLsolve

p_inv = [500.0
0.084
4.69
2.0
400.0
20.0
0.2
1000.0
0.59
736.0
0.0
0.0
0.2
1.27
14.3
0.0
50.0
0.2
0.08
0.003
0.074
0.2
0.01 # Matt
1.0  # V_ref
1.0  # ω ref
0.7  # Pref
0    # Q_ref
0.5  #Xtrans
0.0  # Rtrans
1.01 #Vm
0.0] # Vθ

function vsm(dx, x, p, t)
    #PARAMETERS
    ω_lp = p[1]
    kp_pll = p[2]
    ki_pll = p[3]
    Ta = p[4]
    kd = p[5]
    kω = p[6]
    kq = p[7]
    ωf = p[8]
    kpv = p[9]
    kiv = p[10]
    kffv = p[11]
    rv = p[12]
    lv = p[13]
    kpc = p[14]
    kic = p[15]
    kffi = p[16]
    ωad = p[17]
    kad = p[18]
    lf = p[19]
    rf = p[20]
    cf = p[21]
    lg = p[22]
    rg = p[23]
    Vref = p[24]
    ωref = p[25]
    Pref = p[26]
    Qref = p[27]
    Xtrans = p[28]
    Rtrans = p[29]
    Vm = p[30]
    Vθ = p[31]

    #STATE INDEX AND STATES
    i__vi_filter, vi_filter = 1, x[1]
    i__γd_ic, γd_ic = 2, x[2]
    i__vq_pll, vq_pll = 3, x[3]
    i__γq_ic, γq_ic = 4, x[4]
    i__ir_filter, ir_filter = 5, x[5]
    i__ξd_ic, ξd_ic = 6, x[6]
    i__ϕd_ic, ϕd_ic = 7, x[7]
    i__ε_pll, ε_pll = 8, x[8]
    i__ir_cnv, ir_cnv = 9, x[9]
    i__vr_filter, vr_filter = 10, x[10]
[...]

PR #1604, opened 2022-02-23
New error estimate heuristic for QNDF on DAEs

YingboMa:
Fix https://github.com/SciML/OrdinaryDiffEq.jl/issues/1596

### DIFF
--- a/src/perform_step/bdf_perform_step.jl
+++ b/src/perform_step/bdf_perform_step.jl
@@ -761,7 +761,7 @@ function perform_step!(integrator,cache::QNDFConstantCache{max_order},repeat_ste
 
   if integrator.opts.adaptive
     @unpack abstol, reltol, internalnorm = integrator.opts
-    if cache.consfailcnt >= 1 && mass_matrix !== I
+    if cache.consfailcnt > 1 && mass_matrix !== I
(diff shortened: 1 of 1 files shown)


YingboMa:
The solution now looks like:
1596_plot

src/perform_step/bdf_perform_step.jl
```diff
@@ -863,7 +873,13 @@ function perform_step!(integrator, cache::QNDFCache{max_order}, repeat_step=fals
   update_D!(D, dd, k)
   if integrator.opts.adaptive
     @unpack abstol, reltol, internalnorm = integrator.opts
-    calculate_residuals!(atmp, dd, uprev, u, abstol, reltol, internalnorm, t)
+    if cache.consfailcnt > 1 && mass_matrix !== I
+      # if we get repeated failure and mass_matrix !== I it's likely that
+      # there's a discontinuity on the algebraic equations
+      calculate_residuals!(atmp, mul!(nlsolver.tmp, mass_matrix, dd), uprev, u, abstol, reltol, internalnorm, t)
```
ChrisRackauckas, line 879: makes sense.

jd-lara: FWIW @YingboMa this is exactly the case we are testing in #1606

PR #1605, opened 2022-02-23
Update uprev* when iter == 0 if there's an event already

YingboMa:
Fix https://github.com/SciML/DifferentialEquations.jl/issues/848


ChrisRackauckas:
Shouldn't that be handled by https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L595 ?

ChrisRackauckas (approved):

PR #1607, opened 2022-02-25
New method Rodas5P and new fourth order dense output Rodas5 .

gstein3m:
New coefficients were constructed for Rodas5. The construction is along the lines of Rodas4P/Rodas4P2. Therefore the method is named Rodas5P.
Additionally a new fourth order dense output is defined for Rodas5 and Rodas5P.
Benchmarks_Rodas.jl and Benchmarkas_WP_Rodas.jl show some comparisons for constant stepsizes and adaptiv stepsizes.
A paper about the construction of Rodas5P is in progress.


benchmark/Benchmarks_Rodas.jl
```diff
@@ -0,0 +1,137 @@
+using DifferentialEquations, Plots, LaTeXStrings, LinearAlgebra
```
ChrisRackauckas, line 1: This should go into SciMLBenchmarks.jl so it can run on the clusters and get automatically updated.

src/derivative_utils.jl
```diff
@@ -639,7 +639,11 @@ function calc_rosenbrock_differentiation!(integrator, cache, dtd1, dtgamma, repe
   # we need to skip calculating `W` when a step is repeated
   new_W = false
   if !repeat_step
+#      println("a) Cache:",cache.linsolve_tmp)
```
ChrisRackauckas, line 642: ?

ChrisRackauckas (approved):

ChrisRackauckas:
That test failure is interesting. It's using the Rodas5 interpolation and doesn't seem to be able to keep positivity anymore. Could you look into that?

https://github.com/SciML/DiffEqCallbacks.jl/blob/master/test/domain_tests.jl

gstein3m:
I see two problems in domain.jl (see DiffEqCallbacks):

1) The function affect! is called after the computation of stage k5 in Rodas4 (this is ok) and after k7 in Rodas5. This is not ok, since the fourth order interpolation needs the last stage value k8 too. But I don't find, where to change that.
2) When the stpesize is reduced in domain.jl, a new interpolation is computed via function 
@muladd function _ode_interpolant!(out,Θ,dt,y₀,y₁,k,cache::Rosenbrock5Cache{<:Array},idxs::Nothing,T::Type{Val{0}})
Here the new dt is applied, but it needs the old one, since the stages were computed with this old value. Thus this interplation gives wrong values, also for Rodas4.

ChrisRackauckas:
> The function affect! is called after the computation of stage k5 in Rodas4 (this is ok) and after k7 in Rodas5. This is not ok, since the fourth order interpolation needs the last stage value k8 too. But I don't find, where to change that.

If `calck` is true (which it should be with any callbacks?) then it should always compute k8?

> When the stpesize is reduced in domain.jl, a new interpolation is computed via function
@muladd function _ode_interpolant!(out,Θ,dt,y₀,y₁,k,cache::Rosenbrock5Cache{<:Array},idxs::Nothing,T::Type{Val{0}})
Here the new dt is applied, but it needs the old one, since the stages were computed with this old value. Thus this interplation gives wrong values, also for Rodas4.

Ahh, https://github.com/SciML/DiffEqCallbacks.jl/pull/110 should fix that.

gstein3m:
Yes, whenever interpolation is needed (which seems the case for any callback) k8 must be computed.

ChrisRackauckas:
Thanks!

PR #1611, opened 2022-03-04
Improve Rosenbrock finite difference by swapping dt with J

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/773


src/derivative_utils.jl
```diff
@@ -645,12 +645,13 @@ end
 
 function calc_rosenbrock_differentiation!(integrator, cache, dtd1, dtgamma, repeat_step, W_transform)
   nlsolver = nothing
+  calc_tderivative!(integrator, cache, dtd1, repeat_step)
```
YingboMa, line 648: I am not sure if I understand this. What different does it make if we call it earlier?

src/nlsolve/nlsolve.jl
```diff
@@ -18,6 +18,10 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator, cache=nothing, repeat_
     else
       γW = nlsolver.γ * integrator.dt / nlsolver.α
     end
+
+    # This is for numerical differentiation cache correctness
+    # Requires Newton methods are FSAL
+    nlsolver.cache.du1 .= integrator.fsalfirst
```
YingboMa, line 24: Right, all the methods are programmatically FSAL for dense outputs, we should use it well. BTW, can we simply set `integrator.fsalfirst === nlsolver.cache.du1`? Will that lead to incorrectness?

YingboMa (approved):

src/nlsolve/nlsolve.jl
ChrisRackauckas: `du1` can get overwritten/changed, for example in the central differencing method, so it's not always correct.

PR #1612, opened 2022-03-04
Delete special ODECompositeSolution overloads

ChrisRackauckas:
Since we are now >=1.6 we are good.

### DIFF
--- a/src/composite_solution.jl
+++ b/src/composite_solution.jl
@@ -13,8 +13,6 @@ struct ODECompositeSolution{T,N,uType,uType2,EType,tType,rateType,P,A,IType,DE} 
   destats::DE
   retcode::Symbol
 end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
@devmotion looks like the DiffEqCallbacks change caused a downstream issue? https://buildkite.com/julialang/ordinarydiffeq-dot-jl/builds/935#ab4b1844-b6d7-47ed-a33a-98974f2abbb0/345-775

ChrisRackauckas:
oh it just needs parameters.

devmotion:
Oh yeah I guess you made a non-breaking release? Maybe we should keep supporting the previous non-parameter version, possibly with a deprecation warning. I could make a PR later today.

Issue #1617, opened 2022-03-15
2-norm hardcoded in DP5 stiffness detection


danielwe:
Thanks for the quick fix to #1614 in #1615. However, the fix inherited a hardcoded 2-norm, while all other stiffness detecting code use `integrator.opts.internalnorm`. Compare

https://github.com/SciML/OrdinaryDiffEq.jl/blob/79ba28b6ff19e8424fb5926485bb90ab631b0ebd/src/perform_step/low_order_rk_perform_step.jl#L818-L824

to

https://github.com/SciML/OrdinaryDiffEq.jl/blob/79ba28b6ff19e8424fb5926485bb90ab631b0ebd/src/perform_step/low_order_rk_perform_step.jl#L767-L768

PR #1620, opened 2022-03-15
Use internal norm in the stiffness detection

ChrisRackauckas:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1617

### DIFF
--- a/src/perform_step/low_order_rk_perform_step.jl
+++ b/src/perform_step/low_order_rk_perform_step.jl
@@ -819,7 +819,7 @@ end
     ϱu, ϱd = zero(eltype(k7))^2, zero(eltype(g7))^2
     @. g6 = g7 - g6
     ϱd = integrator.opts.internalnorm(g6,t)
-    @. tmp = (k7 - k6)^2
+    @. tmp = k7 - k6
     ϱu = integrator.opts.internalnorm(tmp,t)
     integrator.eigen_est = sqrt(ϱu/ϱd)*oneunit(t)
   end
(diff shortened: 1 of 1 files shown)


src/perform_step/low_order_rk_perform_step.jl
```diff
@@ -818,9 +818,9 @@ end
     # Hairer II, page 22
     ϱu, ϱd = zero(eltype(k7))^2, zero(eltype(g7))^2
     @. g6 = (g7 - g6)^2
```
devmotion, line 820: This should be changed I guess:
```suggestion
    @. g6 = g7 - g6
```

src/perform_step/low_order_rk_perform_step.jl
```diff
@@ -818,9 +818,9 @@ end
...
-    ϱd = sum(g6)
+    ϱd = integrator.opts.internalnorm(g6,t)
     @. tmp = (k7 - k6)^2
```
devmotion, line 822:
```suggestion
    @. tmp = k7 - k6
```

src/perform_step/low_order_rk_perform_step.jl
```diff
@@ -817,10 +817,10 @@ end
...
     ϱu, ϱd = zero(eltype(k7))^2, zero(eltype(g7))^2
-    @. g6 = (g7 - g6)^2
-    ϱd = sum(g6)
-    @. tmp = (k7 - k6)^2
-    ϱu = sum(tmp)	
+    @. g6 = g7 - g6
+    ϱd = integrator.opts.internalnorm(g6,t)
+    @. tmp = k7 - k6
+    ϱu = integrator.opts.internalnorm(tmp,t)
     integrator.eigen_est = sqrt(ϱu/ϱd)*oneunit(t)
```
danielwe, line 824: I think you need to drop the sqrt here

danielwe: Also noticing that `oneunit(t)` is only used here and not in similar code for other algorithms. Any reason for that?

PR #1622, opened 2022-03-20
Add fast broadcast and a proper test for #1615

danielwe:


ChrisRackauckas (approved):

test/gpu/autoswitch.jl
```diff
@@ -0,0 +1,14 @@
+using OrdinaryDiffEq, CUDA, Test
```
ChrisRackauckas, line 1: If you add a new test file then you need to add it to runtests.jl

danielwe: Ah right, thanks!

ChrisRackauckas:
I swapped the test order so we can ignore https://github.com/SciML/OrdinaryDiffEq.jl/issues/1626 for now and see if we can get a fix there in the next day. It seems like something changed in the GPU compiler that doesn't like our broadcast hack there 😅

Issue #1618, opened 2022-03-15
ROCK algorithms fail to initialize when part of a composite algorithm


danielwe:
AutoDP5(ROCK4()) errors out at the following line when it switches to  ROCK4:

https://github.com/SciML/OrdinaryDiffEq.jl/blob/79ba28b6ff19e8424fb5926485bb90ab631b0ebd/src/perform_step/rkc_perform_step.jl#L267

Here, `integrator.alg` points to the CompositeAlgorithm instance, not the ROCK4 instance, so it doesn't have a `max_stages` field. Glancing through the rest of the file, the issue also affects ROCK2, and more generally I guess any algorithm that references `integrator.alg` may fail to work as part of a CompositeAlgorithm.

danielwe:
I guess the least invasive solution would be to do something like

```julia
alg = (integrator.alg isa CompositeAlgorithm) ? integrator.alg.algs[current] : integrator.alg
```

in each of the affected functions, and then proceed using `alg`. The problem is that the `current` index is stored on the CompositeCache, which is not passed to these functions, so a slight redesign is needed one way or the other.

ChrisRackauckas:
there's an internal function `_unwrap_alg(alg)` for this. We just need to use it here. BTW, what's the use case for this? Since ROCK is already an explicit algorithm, I would assume that this wouldn't do much?

danielwe:
Aha, should have noticed that. (Please excuse my earlier musings then.)

And yeah, I guess I had that question myself. ROCK4 and the other stabilized explicit methods are listed among the methods for stiff equations in the docs, so I figured I'd try this in my experimentation/benchmarking even though I think I was somewhat aware of the association between stiff equations and implicit methods. Is this a documentation bug?

danielwe:
Speaking of documentation bugs, there is no SERK2v2, only SERK2. Separate issue perhaps.

ChrisRackauckas:
[...]

PR #1625, opened 2022-03-20
Fix missing and incorrect uses of `unwrap_alg` in `perform_step!`

danielwe:
I took a comprehensive approach to fixing #1618. I'm not equally confident in all of the changes so feel free to pick and choose commits. There are also potential issues that I wasn't certain enough to do something about at all, see below. Feedback appreciated.

**Summary**

* Added missing `unwrap_alg` calls to make all RKC methods, including ROCK4, work in composite algorithms (`is_stiff = true`)
* Same for `QNDF` (`is_stiff = true`)
* Same for all extrapolation methods (`is_stiff = false` for explicit methods, `is_stiff = true` for implicit methods including the undocumented `ImplicitEulerBarycentricExtrapolation`)
* Same for Anas5, KuttaPRK2p5, and PDIRK44 (`is_stiff = false` for Anas5 and KuttaPRK2p5, `is_stiff = true` for PDIRK44)
* Changed `is_stiff` from `true` to `false` for FRK65 and PFRK87 (both are listed as non-stiff in the docs)
* Removed a bunch of redundant unpackings of `alg`, mostly from Lie group and DAE methods but also a couple from RKC methods

Here are the remaining methods that still reference `integrator.alg` directly and are thus incompatible with any form of `CompositeAlgorithm`:

* FunctionMap
* ExplicitRK
* McAte42 (will fail silently when part of a CompositeAlgorithm---it's just a branch adding extra steps `if typeof(integrator.alg) <: McAte42`)
* CNAB2, CNLF2, and SBDF
* Every Krylov-enabled Lie group solver _except_ LinearExponential
  * Full list: MagnusMidpoint, LieRK4, RKMK4, RKMK2, MagnusNC8, MagnusGL4, MagnusGL8, MagnusNC6, MagnusGL6, MagnusGauss4, LieEuler, MagnusLeapfrog 

Things that kinda seem wrong to me but I wasn't sure:

* LinearExponential calls `unwrap_alg(integrator, true)`, i.e., it claims to be a stiff algorithm. This is the exact solver for LTI systems. Does this make sense at all? 

**Questions/TL;DR**

* Was I correct assuming that it's a bug for FRK65 and PFRK87 to identify as stiff?
* Do any of the following have good choices for `is_stiff` or should they remain uncomposable?
  * FunctionMap (wouldn't ever make sense to compose would it?)
  * ExplicitRK (general, takes tableau, so probably can't make such an assumption?)
  * McAte42 (symplectic, perhaps that's too specialized for composition anyway, but also I'm guessing `false`?)
  * CNAB2, CNLF2, SBDF (split solvers, perhaps don't make much sense to compose but also probably `true`?)
  * Lie group solvers
* Should LinearExponential really identify as stiff?
* Would it make sense to do `unwrap_alg(integrator, nothing)` for algorithms that aren't unambiguously stiff or non-stiff (ExplicitRK, perhaps Lie group solvers?)? That way they could still participate in custom CompositeAlgorithms, just not the built-in stiffness-based autoswitching, and one could make `unwrap_alg` throw a friendly, descriptive error about the latter.

Apologies for handing out lots of work with all the recent PRs---please take it as a sign of appreciation of this package.

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -747,6 +747,9 @@ function unwrap_alg(alg::SciMLBase.DEAlgorithm, is_stiff)
   if !iscomp
     return alg
   elseif typeof(alg.choice_function) <: AutoSwitchCache
+    if is_stiff === nothing
+      throwautoswitch(alg)
+    end
     num = is_stiff ? 2 : 1
     if num == 1
       return alg.algs[1]
@@ -764,6 +767,9 @@ function unwrap_alg(integrator, is_stiff)
   if !iscomp
     return alg
   elseif typeof(alg.choice_function) <: AutoSwitchCache
+    if is_stiff === nothing
+      throwautoswitch(alg)
+    end
     num = is_stiff ? 2 : 1
     if num == 1
       return alg.algs[1]
@@ -779,6 +785,10 @@ function unwrap_alg(integrator, is_stiff)
       return alg.algs[integrator.cache.current]
     end
   end
+end
+
+function throwautoswitch(alg)
+  throw(ArgumentError("one of $(alg.algs) is not compatible with stiffness-based autoswitching"))
 end
 
 # Whether `uprev` is used in the algorithm directly.
--- a/src/perform_step/adams_bashforth_moulton_perform_step.jl
+++ b/src/perform_step/adams_bashforth_moulton_perform_step.jl
@@ -1516,7 +1516,7 @@ function initialize!(integrator,cache::CNAB2ConstantCache)
 end
 
 function perform_step!(integrator,cache::CNAB2ConstantCache,repeat_step=false)
-  @unpack t,dt,uprev,u,f,p,alg = integrator
+  @unpack t,dt,uprev,u,f,p = integrator
   @unpack k2,nlsolver = cache
   cnt = integrator.iter
   f1 = integrator.f.f1
@@ -1567,7 +1567,7 @@ function initialize!(integrator, cache::CNAB2Cache)
 end
 
 function perform_step!(integrator, cache::CNAB2Cache, repeat_step=false)
-  @unpack t,dt,uprev,u,f,p,alg = integrator
+  @unpack t,dt,uprev,u,f,p = integrator
   @unpack k1,k2,du₁,nlsolver = cache
   @unpack z,tmp = nlsolver
   @unpack f1 = f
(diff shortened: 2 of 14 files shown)


ChrisRackauckas:
> including the undocumented ImplicitEulerBarycentricExtrapolation

@utkarsh530 can you add documentation for this one? Nice catch @danielwe

ChrisRackauckas:
> ExplicitRK

Most of that exclusion list looks about right, but this one seems like an exception. You should be able to make the generic explicit RK work with this though. At least let's get an issue open for it.

ChrisRackauckas:
> LinearExponential calls unwrap_alg(integrator, true), i.e., it claims to be a stiff algorithm. This is the exact solver for LTI systems. Does this make sense at all?

Ehh, this really is quite an exception since it fails except in the specific case where it's an analytical solution, so yeah... I don't think there's ever a reason to autoswitch, but we might as well label it like the other exponential integrators for continuity.

ChrisRackauckas:
> Was I correct assuming that it's a bug for FRK65 and PFRK87 to identify as stiff?

They are non-stiff.

> Would it make sense to do unwrap_alg(integrator, nothing) for algorithms that aren't unambiguously stiff or non-stiff (ExplicitRK, perhaps Lie group solvers?)? That way they could still participate in custom CompositeAlgorithms, just not the built-in stiffness-based autoswitching, and one could make unwrap_alg throw a friendly, descriptive error about the latter.

That would probably be nice. Better errors are always helpful!

danielwe:
Thanks for the feedback. Changes made:

* `unwrap_alg` throws descriptive error if `integrator.alg` is an AutoSwitch and `is_stiff = nothing`
* `unwrap_alg(integrator, nothing)` now used for the remaining algorithms that need access to `alg`: FunctionMap, ExplicitRK, McAte42, SBDF, and all exponential solvers including LinearExponential
* Never mind CNAB2 and CNLF2, they didn't need `alg` and the redundant unpacking has been removed

Requesting final confirmation that `is_stiff = nothing` is the right call for symplectic integrators (McAte42) and SBDF. Wondering if one could also argue in favor of `false` and `true`, respectively.

ChrisRackauckas:
symplectic integrators are only non-stiff. Ever using autoswitch would make it non-symplectic though since it's a global not a local property. SBDF is for stiff equations (or semi-stiff because of IMEX)

danielwe:
Yes, exactly. There! (I know it makes no sense to autoswitch so this is entirely inconsequential, but might as well make it as consistent as it can get.)

ChrisRackauckas:
Yeah, might as well let people explore. Who knows what they will find!

PR #1628, opened 2022-03-22
use integrator tolerance in init

jd-lara:
This PR addresses an inconsistency in the tolerances used in `BrownFullBasicInit`. The current code uses a default tolerance of `1e-10` even if the solver tolerances are higher. Also, in cases where the nlsolve call is triggered, it uses Nlsolve default tolerances `1e-8` which is less than the default `1e-10`

### DIFF
--- a/src/initialize_dae.jl
+++ b/src/initialize_dae.jl
@@ -38,13 +38,13 @@ end
 function _initialize_dae!(integrator, prob::ODEProblem,
              alg::DefaultInit, x::Val{true})
   _initialize_dae!(integrator, prob,
-          BrownFullBasicInit(), x)
+          BrownFullBasicInit(integrator.opts.abstol), x)
 end
 
 function _initialize_dae!(integrator, prob::ODEProblem,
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

src/initialize_dae.jl
```diff
@@ -65,7 +65,7 @@ function _initialize_dae!(integrator, prob::DAEProblem,
             ShampineCollocationInit(), x)
   else
     _initialize_dae!(integrator, prob,
-            BrownFullBasicInit(), x)
+            BrownFullBasicInit(integrator.opts.abstol ), x)
```
ChrisRackauckas, line 68: and reltol?

jd-lara: should we pick the minimum? The function implementation for the check is 
```
integrator.opts.internalnorm(tmp,t) <= alg.abstol && return
```

ChrisRackauckas: ahh yeah, that check only makes sense for abstol.

PR #1633, opened 2022-03-26
Remove a broadcast for slightly faster compilation

ChrisRackauckas:


ChrisRackauckas:
Supersceded by https://github.com/SciML/OrdinaryDiffEq.jl/pull/1634

PR #1634, opened 2022-03-26
@.. -> @.. broadcast=false

YingboMa:


src/dense/generic_dense.jl
```diff
@@ -474,13 +474,13 @@ Hairer Norsett Wanner Solving Ordinary Differential Euations I - Nonstiff Proble
 Herimte Interpolation, chosen if no other dispatch for ode_interpolant
 """
 @muladd function hermite_interpolant(Θ,dt,y₀,y₁,k,::Type{Val{false}},idxs::Nothing,T::Type{Val{0}}) # Default interpolant is Hermite
-  #@.. (1-Θ)*y₀+Θ*y₁+Θ*(Θ-1)*((1-2Θ)*(y₁-y₀)+(Θ-1)*dt*k[1] + Θ*dt*k[2])
+  #@.. broadcast=false broadcast=false (1-Θ)*y₀+Θ*y₁+Θ*(Θ-1)*((1-2Θ)*(y₁-y₀)+(Θ-1)*dt*k[1] + Θ*dt*k[2])
```
ChrisRackauckas, line 477: double?

YingboMa:
With
```julia

function f(du, u, p, t)
    du[1] = 0.2u[1]
    du[2] = 0.4u[2]
end
u0 = ones(2)
tspan = (0.0, 1.0)
prob = ODEProblem(f, u0, tspan)
tinf = @snoopi_deep sol = solve(prob, Rosenbrock23(chunk_size=Val{1}()))

```
Before:
```julia
julia> tinf = @snoopi_deep sol = solve(prob, Rosenbrock23(chunk_size=Val{1}()))
InferenceTimingNode: 1.231902/2.255570 on Core.Compiler.Timings.ROOT() with 6 direct children
```
After:
```julia
julia> tinf = @snoopi_deep sol = solve(prob, Rosenbrock23(chunk_size=Val{1}()))
InferenceTimingNode: 1.172056/2.128723 on Core.Compiler.Timings.ROOT() with 6 direct children
```

YingboMa:
The test failure seems unrelated, but since it's a large change, let's wait the downstream fix and run the CI fully before we merge it.

ChrisRackauckas:
If this works, we should delete the Array specialized code.

YingboMa:
OK to merge?

PR #1645, opened 2022-04-06
Add formatter?

ChrisRackauckas:


devmotion (commented):
In JuliaGaussianProcesses, TuringLang, ChainRules, ... we run a format action only in PRs, install and run JuliaFormatter manually, and provide suggestions to PR authors/reviewers. E.g., https://github.com/JuliaDiff/ChainRulesCore.jl/blob/main/.github/workflows/format.yml

Issue #1601, opened 2022-02-18
New 5(4) RK coefficients


ChrisRackauckas:
https://arxiv.org/abs/2202.08443

dynamic-queries:
Found a similar 7th order method https://arxiv.org/pdf/2108.12590.pdf

dynamic-queries:
- [ ] Implement 6th order method from table 4 - MSRK6
- [ ] Adaptive method MSRK54 and MSRK64

pri1311:
Could give implementing the 6th order method a try.

ChrisRackauckas:
It would be good to finish the 5h order method first. Adaptivity and interpolation. https://docs.sciml.ai/DiffEqDevDocs/stable/contributing/adding_algorithms/ is a good overview. Then see:

https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/tableaus/low_order_rk_tableaus.jl#L501-L609
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/perform_step/low_order_rk_perform_step.jl#L903-L910

For how the adaptive estimates are computed using the btilde values. Let me know if you need any help there. One that and the interpolation are in, we can add it to our benchmarking suite: the author is curious about what the results will be.

PR #1646, opened 2022-04-06
New RK Tableau - MSRK5

dynamic-queries:
Ref : https://github.com/SciML/OrdinaryDiffEq.jl/issues/1601

The method is of fifth order.
My implementation has a couple of issues that are dubious. 

1. On testing this locally with the exponential growth function from the dev-docs,  convergence factor (infinity norm) of 4.82 is obtained. 
```julia
f = ODEFunction((u,p,t)->1.01u, 
            analytic = (u0,p,t) -> u0*exp(1.01t))
prob = ODEProblem(f,1.01,(0.0,1.0))

N = 7
begin
    dts = 0.5.^(N:-1:1)
    sim = test_convergence(dts,prob,MSRK5())
    @show sim.𝒪est[:l∞]
    fig = plot(sim)
    display(fig)
    savefig("MSRK5.png")
end
```
MSRK5

At `N=8` , the errors are of the order of machine precision to conclude anything. Nonetheless the convergence_test function in the toolchain returns 4.46. 

2. In addition, the ode_convergence_test fails despite ~5 order convergence for the exponential growth problem. 
```julia 
  sim_ms5 = test_convergence(dts,prob,MSRK5())
  @test sim_ms5.𝒪est[:l∞] ≈ 5 atol=testTol
```

I would like to understand, if this issue is a consequence of my implementation or the method itself. Further, I would like your feedback to help close this issue.


src/OrdinaryDiffEq.jl
```diff
@@ -229,7 +229,7 @@ using DocStringExtensions
 
   export FunctionMap, Euler, Heun, Ralston, Midpoint, RK4, ExplicitRK, OwrenZen3, OwrenZen4, OwrenZen5,
          BS3, BS5, RK46NL, DP5, Tsit5, DP8, Vern6, Vern7, Vern8, TanYam7, TsitPap8,
-         Vern9, Feagin10, Feagin12, Feagin14, CompositeAlgorithm, Anas5, RKO65, FRK65, PFRK87, RKM
+         Vern9, Feagin10, Feagin12, Feagin14, CompositeAlgorithm, Anas5, RKO65, FRK65, PFRK87, RKM , MSRK5, MSRK6
```
ChrisRackauckas, line 232:
```suggestion
         Vern9, Feagin10, Feagin12, Feagin14, CompositeAlgorithm, Anas5, RKO65, FRK65, PFRK87, RKM , MSRK5
```

src/algorithms.jl
```diff
@@ -416,6 +416,13 @@ RK4: Explicit Runge-Kutta Method
 """
 struct RK4 <: OrdinaryDiffEqAdaptiveAlgorithm end
 struct RKM <: OrdinaryDiffEqAlgorithm end
+
+"""
+MSRK6 : 5th order Explicit RK method.
```
ChrisRackauckas, line 421:
```suggestion
MSRK5 : 5th order Explicit RK method.
```

ChrisRackauckas:
Notice that in the convergence test file that the `dt`s are adjusted on a per-method basis. This is because a convergence test can only occur in the range where `dt` is small enough, but also above floating point epsilon (`1e-16`). So it's completely expected to level off at that point! You need to make sure that the convergence test file is using values of dt in the appropriate range, and then it should work out.

src/OrdinaryDiffEq.jl
dynamic-queries: FYI, 
I also implemented a 6 order Tableau that the paper discusses. Hence the residual exports

PR #1655, opened 2022-05-02
Reduce number of calls in typed code

chriselrod:
Mostly this just makes Cthulhu look prettier by cutting down the number of branches.
Doesn't seem to help compile or runtime.

Co-authored-by: Yingbo Ma <mayingbo5@gmail.com>

### DIFF
--- a/src/perform_step/composite_perform_step.jl
+++ b/src/perform_step/composite_perform_step.jl
@@ -33,26 +33,48 @@ end
 
 =#
 
+function _initialize!(integrator, caches::Tuple{A,B}, current) where {A,B}
+  if current == 1
+    initialize!(integrator, @inbounds(caches[1]))
+  else
+    initialize!(integrator, @inbounds(caches[2]))
+  end
+end
+function _initialize!(integrator, caches::Tuple{A,B,C,Vararg}, current) where {A,B,C}
+  if current == 1
+    initialize!(integrator, @inbounds(caches[1]))
+  elseif current == 2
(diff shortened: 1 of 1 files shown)


YingboMa (approved):

ChrisRackauckas:
Update the failing regression test by increasing the tolerance.

Though we need to fix the buildkite to not give us ARM. I don't know who added that to the ecosystem queue but that's the most breaking thing I've seen and it probably breaks every CI using that queue, so we should get that reversed.

devmotion:
> Though we need to fix the buildkite to not give us ARM. I don't know who added that to the ecosystem queue but that's the most breaking thing I've seen and it probably breaks every CI using that queue, so we should get that reversed.

I noticed this as well, it broke eg also TuringTutorials (and hence I assume SciMLTutorials, SciMLBenchmarks etc are broken as well). We explicitly specified the desired architecture to fix it: https://github.com/TuringLang/TuringTutorials/pull/298 Probably a good thing to do anyway, but it would have been less inconvenient if aarch64 would have been opt-in at least initially.

ChrisRackauckas:
The error here is that ARM doesn't allow closures. We shouldn't consider that a standard architecture for CI if that's true.

devmotion:
The tutorials didn't even start running since one can't use bash (and probably other programs) on these machines, so I don't know if there are additional Julia-specific problems.

ChrisRackauckas:
Rebase?

PR #1657, opened 2022-05-03
Force to use amd64

YingboMa:


ChrisRackauckas:
@YingboMa can you help look into that GPU failure? @frankschae you're looking at the DiffEqSensitivity stuff? All of the GalacticOptim.jl and DiffEqFlux.jl updates are in, along with all of the Flux v0.13 stuff and the recent Zygote bump, so in theory AD versioning should be fine though I know there's some stuff in https://github.com/SciML/DiffEqSensitivity.jl/pull/570

frankschae:
yep, I was/am looking into the test failure in DiffEqSensitivity. This week, the preparation for our move took quite some time. I wasn't able to reproduce the remaining DiffEqFlux downstream failure locally so far.

YingboMa:
The CUDA failure is https://github.com/JuliaGPU/CUDA.jl/issues/1476

PR #1661, opened 2022-05-09
Lee1

Leebre:
Development to add a new exponential integrator, ETD2RK4 from Krogstad's 2005 paper. Krylov/non-krylov OOP versions are implemented and seem to be working (although tests are throwing an error). IIP versions still need to be added. First step is calculated using DP5 (which is also 4th-order).

### DIFF
--- a/src/perform_step/exponential_rk_perform_step.jl
+++ b/src/perform_step/exponential_rk_perform_step.jl
@@ -650,59 +650,61 @@ function perform_step!(integrator, cache::ETD2RK4ConstantCache, repeat_step=fals
 
   #Draft code for new ETD2RK4 integrator
   #Variable names follow those in Krogstad, 2005
-  if integrator.iter == 1
-    print("Using Krogstad's ETD2RK4 algorithm!\n")
-  end
 
   @unpack t,dt,uprev,uprev2,f,p = integrator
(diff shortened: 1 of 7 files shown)


ranjanan:
Rename PR to reflect contribution?

Leebre:
> Rename PR to reflect contribution?

Done! Thanks for the suggestion :-)

src/perform_step/exponential_rk_perform_step.jl
```diff
@@ -646,6 +646,94 @@ function perform_step!(integrator, cache::HochOst4Cache, repeat_step=false)
...
+
+  @unpack t,dt,uprev,uprev2,f,p = integrator
+  A = isa(f, SplitFunction) ? f.f1.f : calc_J(integrator, cache) # get linear operator
+  alg = unwrap_alg(integrator, true)
+  #F1 = integrator.fsalfirst #not sure if FSAL, so ignore for now
+  halfdt = dt/2
+
+  if integrator.iter == 1
+    # Initialize the first step using Dorm-Prince-5 (also 4th-order)
+    aij = [
```
YingboMa, line 662: We normally write the tableau in https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/tableaus/low_order_rk_tableaus.jl so it is only initialized once.

PR #1697, opened 2022-07-14
fix check for polyesterthreads

chriselrod:


ChrisRackauckas:
More general with https://github.com/SciML/OrdinaryDiffEq.jl/pull/1700

PR #1698, opened 2022-07-14
Add always_new and check_div options for NLsolve

YingboMa:
`always_new` forces Newton iterations to always evaluate new J and W
matrices. `check_div = false` disables divergence checking.

This PR requires https://github.com/SciML/DiffEqBase.jl/pull/794 .


YingboMa:
I'll add a test for this.

YingboMa:
Ah, I need to update StochasticDiffEq as well.

PR #1702, opened 2022-07-17
Keep a consistent sparsity pattern

ChrisRackauckas:
See the comment. This fixes the MTK downstream failure that was introduced when symbolic factorizations started getting cached


ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/runs/7382501873?check_suite_focus=true#step:6:449
https://github.com/SciML/OrdinaryDiffEq.jl/runs/7382501932?check_suite_focus=true#step:6:756

Things are pointing at KLU. @wkimmerer is the current overload with the symbolic factorization storing incorrect?

PR #1709, opened 2022-08-01
Symbolic `set_u!` method

AayushSabharwal:
Requires https://github.com/SciML/SciMLBase.jl/pull/209
Closes https://github.com/SciML/SciMLBase.jl/issues/206


ChrisRackauckas:
This should be done at the interface level (SciMLBase), not OrdinaryDiffEq.

AayushSabharwal:
Oh. Okay. I assumed since the other `set_u!` methods are implemented here, I'd add this here too. I'll move the changes over to SciMLBase.jl

ChrisRackauckas:
Only things which require specific internals of the integrator should be done at the solver level.

PR #1713, opened 2022-08-03
change to scimlops

vpuri3:


vpuri3:
ref https://github.com/SciML/SciMLOperators.jl/issues/55

Issue #1710, opened 2022-08-01
SingularException with stiff integrators and Float32 data


IlyaOrson:
The ODE system is not the nicest but maybe the error could be improved. It does not error with Float64 and non-stiff integrators.
```julia
using OrdinaryDiffEq

r(x1, x2) = (x1 * (-2.1474936f0 * (x2 + x1)))
function f(u, p, t)
    return [-r(u...), r(u...)]
end

# original problem
# tspan = (0.0f0, 2f0)
# initial_condition = [3.6f0, 0.4f0]
# problem around t = 1.6078221f0 and state [2.1349438f6, -2.1349438f6]

tspan=(1.6078221f0, 2f0)
initial_condition = [2.1349438f6, -2.1349438f6]
prob = ODEProblem(f, initial_condition, tspan)
sol = solve(prob, Tsit5())  # works
sol = solve(prob, Rodas5())  # fails

ERROR: LinearAlgebra.SingularException(2)
Stacktrace:
  [1] checknonsingular
    @ C:\Users\ilyao\.julia\juliaup\julia-1.8.0-rc3+0.x64\share\julia\stdlib\v1.8\LinearAlgebra\src\factorization.jl:19 [inlined]
  [2] checknonsingular
    @ C:\Users\ilyao\.julia\juliaup\julia-1.8.0-rc3+0.x64\share\julia\stdlib\v1.8\LinearAlgebra\src\factorization.jl:21 [inlined]
  [3] #lu!#172
    @ C:\Users\ilyao\.julia\juliaup\julia-1.8.0-rc3+0.x64\share\julia\stdlib\v1.8\LinearAlgebra\src\lu.jl:82 [inlined]
  [4] #lu#179
    @ C:\Users\ilyao\.julia\juliaup\julia-1.8.0-rc3+0.x64\share\julia\stdlib\v1.8\LinearAlgebra\src\lu.jl:279 [inlined]
  [5] lu (repeats 2 times)
    @ C:\Users\ilyao\.julia\juliaup\julia-1.8.0-rc3+0.x64\share\julia\stdlib\v1.8\LinearAlgebra\src\lu.jl:278 [inlined]
  [6] default_factorize
    @ C:\Users\ilyao\.julia\packages\DiffEqBase\HDcso\src\DiffEqBase.jl:121 [inlined]
[...]

PR #1715, opened 2022-08-05
Early terminate perform step when factorization fails

YingboMa:
Using the example in https://github.com/SciML/OrdinaryDiffEq.jl/issues/1710, we have
```julia
julia> DiffEqBase.default_factorize(A) = lu(A, check=false)

julia> sol = solve(prob, Rodas5()); # before
┌ Warning: Instability detected. Aborting
└ @ SciMLBase ~/.julia/packages/SciMLBase/YasGG/src/integrator_interface.jl:512

julia> sol = solve(prob, Rodas5()); # after

julia> sol.retcode
:Success
```

### DIFF
--- a/src/misc_utils.jl
+++ b/src/misc_utils.jl
@@ -134,3 +134,7 @@ function wrapprecs(_Pl,_Pr,weight)
   end
   Pl, Pr
 end
+
+issuccess_W(W::LinearAlgebra.Factorization) = LinearAlgebra.issuccess(W)
+issuccess_W(W::Number) = !iszero(W)
+issuccess_W(::Any) = true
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
This'll need to happen to all Rosenbrocks. I assume Newton methods already have a safety catch for this?

YingboMa:
Yeah, Newton is fine because of https://github.com/SciML/OrdinaryDiffEq.jl/blob/cd406fc03da416b6e2a8c1e296afbf072d93d74d/src/nlsolve/nlsolve.jl#L40

By in-place methods will error because `dolinsolve` doesn't handle this.

YingboMa:
The test depends on https://github.com/SciML/DiffEqBase.jl/pull/799 . We need to rerun the CI after that's released.

ChrisRackauckas:
Yes we can handle in-place in a following PR. That needs retcodes to be added to LinearSolve.jl.

PR #1723, opened 2022-08-19
Fix the `always_new` option in NLNewton

YingboMa:


src/derivative_utils.jl
```diff
@@ -129,28 +129,30 @@ If `integrator.f` has a custom Jacobian update function, then it will be called.
 either automatic or finite differencing will be used depending on the `cache`.
 """
 function calc_J!(J, integrator, cache)
-  @unpack t,uprev,f,p,alg = integrator
+  @unpack dt,t,uprev,f,p,alg = integrator
 
+  tdt = t + dt
```
YingboMa, line 134: This makes sense for always_new where we want the Jacobian at `t + dt`. Do we want this in general? @ChrisRackauckas

src/nlsolve/nlsolve.jl
```diff
@@ -31,7 +31,12 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator, cache=nothing, repeat_
 
   local ndz
   for iter in 1:maxiters
-    always_new && update_W!(nlsolver, integrator, cache, γW, repeat_step, (true, true))
+    if always_new
+        @. nlsolver.cache.ustep = integrator.uprev + nlsolver.γ * nlsolver.z
```
YingboMa, line 35: This is a bit wasteful because the same `nlsolver.cache.ustep` will be evaluated later. I'll handle this later.

src/derivative_utils.jl
YingboMa: I guess Rosenbrock methods might not like this.

ChrisRackauckas: But it's supposed to be evaluated at `t` if it's at `uprev`?

YingboMa: Yes, I guess we should only do that when it's always_new.

PR #1725, opened 2022-08-22
Add downstream tests for ProbNumDiffEq.jl

nathanaelbosch:
[ProbNumDiffEq.jl](https://github.com/nathanaelbosch/ProbNumDiffEq.jl) currently builds on quite a few non-public parts of OrdinaryDiffEq.jl to implement its solvers. Therefore, [as discussed on discourse](https://discourse.julialang.org/t/building-on-ordinarydiffeq-jl-vs-diffeqbase-jl/85620/4) it would be great to add downstream tests that ensure that ProbNumDiffEq.jl does not break due to changes to OrdinaryDiffEq.jl - or at least, to provide some very early notifications on possible upcoming problems.

I added a very simple test to `./test/downstream/probnumdiffeq.jl`, which is called as part of the `Downstream` group in `./test/runtests.jl`. Is there more that should be added here in this PR? E.g. should there be a corresponding entry in `./.github/workflows/Downstream.yml`? Let me know what's missing.

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -26,6 +26,7 @@ jobs:
           - {user: SciML, repo: StochasticDiffEq.jl, group: AlgConvergence}
           - {user: SciML, repo: DiffEqCallbacks.jl, group: All}
           - {user: SciML, repo: DiffEqSensitivity.jl, group: Downstream}
+          - {user: nathanaelbosch, repo: ProbNumDiffEq.jl, group: Downstream}
 
     steps:
       - uses: actions/checkout@v2


devmotion:
I wonder if it would be sufficient to only add it to Downstream.yml?

ChrisRackauckas:
That would probably be better. We might as well run your whole test suite.

devmotion:
As an explanation, activating project environments during tests can cause difficult to debug issues as it can cause downgrades of dependencies that are not picked up in the tests. Also running your test suite seems a bit more complete and more likely to catch issues.

nathanaelbosch:
Oh okay, I didn't quite understand what Downstream.yml actually does, but now I guess it just runs the tests in the specified packages but with the new version of OrdinaryDiffEq.jl? If that is the case then that's probably better yes. Then I should maybe first create a smaller/faster/more stable subset of the tests that should be run, similar to the groups that you use here and in the downstream packages.

PR #1728, opened 2022-08-23
simplify controller setup logic

ranocha:
I looked over some of the code while writing down a pseudocode of the time integration loop for my students. While doing that, I noticed that this part can be simplified - at least from my point of view. The original code I wrote in #1384 needed an additional check, the modification from https://github.com/SciML/OrdinaryDiffEq.jl/commit/d30cb473c2e395c5e7612fbb11e916997e0ceac6 does not need this check anymore.


ChrisRackauckas:
Thanks, good catch.

Issue #1732, opened 2022-08-24
`jacobian!` does not work with the new wrapped functions


nathanaelbosch:
Minimal example:
```julia
using OrdinaryDiffEq
function lorenz(du, u, p, t)
    du[1] = 10.0(u[2] - u[1])
    du[2] = u[1] * (28.0 - u[3]) - u[2]
    du[3] = u[1] * u[2] - (8 / 3) * u[3]
    return nothing
end
prob = ODEProblem{true,false}(lorenz, [1.0; 0.0; 0.0], (0.0, 1.0))
solve(prob, EPIRK4s3A(), dt=1e-1)
```

Similarly, `ProbNumDiffEq.EK1` now fails since it relies on `jacobian!` to compute Jacobians. What needs to be done to fix this?

ChrisRackauckas:
Fixed.

PR #1733, opened 2022-08-24
Fix interface breaks with Exponential integrators

ChrisRackauckas:
They had a few non-conformities that were causing issues once no recompile mode was enabled.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -3885,8 +3885,6 @@ for Alg in [:Exp4, :EPIRK4s3A, :EPIRK4s3B, :EPIRK5s3, :EXPRB53s3, :EPIRK5P1, :EP
     adaptive_krylov::Bool
     m::Int
     iop::Int
-    autodiff::Bool
-    chunksize::Int
   end
   @eval $Alg(;adaptive_krylov=true, m=30, iop=0, autodiff=true, standardtag = Val{true}(), concrete_jac = nothing,
               chunk_size=Val{0}(), diff_type = Val{:forward}) =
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1732

src/algorithms.jl
```diff
@@ -3856,50 +3856,49 @@ RosenbrockW6S4OS(;chunk_size=Val{0}(),autodiff=true, standardtag = Val{true}(),
...
     iop::Int
     autodiff::Bool
     chunksize::Int
   end
   @eval $Alg(;adaptive_krylov=true, m=30, iop=0, autodiff=true, standardtag = Val{true}(), concrete_jac = nothing,
-              chunksize=0, diff_type = Val{:forward}) =
-              $Alg{diff_type,_unwrap_val(standardtag),_unwrap_val(concrete_jac)}(adaptive_krylov, m, iop, _unwrap_val(autodiff), chunksize)
+              chunk_size=Val{0}(), diff_type = Val{:forward}) =
+              $Alg{_unwrap_val(chunk_size),_unwrap_val(autodiff),diff_type,
+              _unwrap_val(standardtag),_unwrap_val(concrete_jac)}(adaptive_krylov, m, iop)
```
YingboMa, line 3894: Remember to pass `autodiff` and `chunksize` as well here.

ChrisRackauckas: Those should all be removed: all of the other algorithms only have it type parameters 😰

ChrisRackauckas:
I considered this non-breaking because the algorithms were defined in the docs like this already 😅. I guess those are just not options that are used much (they are only used for the exponential Rosenbrock form)

PR #1736, opened 2022-08-25
slightly more robust way of handling tiny EEst for PID controllers

ranocha:


ranocha:
The CI failure
https://github.com/SciML/OrdinaryDiffEq.jl/runs/8013328762?check_suite_focus=true#step:6:688
in IntegrationTest / DiffEqSensitivity.jl/Downstream/1 (pull_request) looks weird enough to me that I think it should not be related to this PR :sweat_smile:

ChrisRackauckas:
Yeah that's not you.

PR #1745, opened 2022-08-30
Fix incorrect FSAL logic causing instability in KenCarp algorithms

### LINKED ISSUE #1730: `SplitFunction` causes instability in split ode solvers
Using a `SplitFunction` seems to cause instabilities in split ode solvers such as `KenCarp4`. A test case is given below where a fourth-order PDE is solved by finite differences. The standard `ODEFunction` formulation works fine. However, using a `SplitFunction` with a dummy `f2` causes immediate instablity while I would expect it to be functionally equivalent.

https://gist.github.com/superckl/3d03324c1aee64a40a1cf3b5acf97c7f.

superckl:
Fixes some incorrect FSAL logic in the `KenCarp` algorithms. The same logic is also present in the `CFNLIRK3` algorithm, but I am unable to test it efficiently because it requires a constant timestep so I have left that algorithm untouched. 

The provided tests use the stiff problem of axisymmetric droplet spreading which would cause instability (dtmin too small) if a `SplitFunction` is used (due to the incorrect FSAL logic). A broken test is also included for `CFNLIRK3` since I can't test it properly.

Fixes #1730

### DIFF
--- a/src/perform_step/kencarp_kvaerno_perform_step.jl
+++ b/src/perform_step/kencarp_kvaerno_perform_step.jl
@@ -569,13 +569,9 @@ end
 
     markfirststage!(nlsolver)
 
-    if typeof(integrator.f) <: SplitFunction
-        # Explicit tableau is not FSAL
-        # Make this not compute on repeat
-        if !repeat_step && !integrator.last_stepfail
-            f(z₁, integrator.uprev, p, integrator.t)
-            z₁ .*= dt
-        end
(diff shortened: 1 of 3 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
I went ahead and did the same to `CFNLIRK3` since we know that it would need it. Since it's an unreleased method it's not a big deal though (and it'll probably never be released).

ChrisRackauckas:
Thanks!

PR #1752, opened 2022-09-05
[DO NOT MERGE] precompile all routes

ChrisRackauckas:
This is purely for testing compilation


ChrisRackauckas:
Handled by the preferences system.

PR #1754, opened 2022-09-08
Fixing Rosenbrock_tableaus cant handle float32 datatype

TheZombie1999:
While Working on my Baechlors thesis i found out that when i tried to run a rosenbrock solver on the gpu with float32 numbers that one one the rosenbrock tableaus made an casting error.

The fix is pretty simple.
I just casted some of the constant values to the correct data type.

Any kind of feed back would be welcome.

This is my very first contribution to an Opensource Project.

### DIFF
--- a/src/tableaus/rosenbrock_tableaus.jl
+++ b/src/tableaus/rosenbrock_tableaus.jl
@@ -40,7 +40,7 @@ struct ROS3PTableau{T, T2}
     d2::T
     d3::T
 end
-
+ 
 function ROS3PTableau(T, T2)
     gamma = convert(T, 1 / 2 + sqrt(3) / 6)
     igamma = inv(gamma)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
Thanks! This looks correct.

Though I would be curious where this Rosenbrock method would be useful for GPUs. If something is GPU-sized, other methods like QNDF are surely better?

TheZombie1999:
It's probably not very useful but I found the error while working on it so I fixed it.

YingboMa:
The format checker fails. You can run formatting after each commit.

TheZombie1999:
Can you show me how to run the formatting? I can't figure out how to do that.

ChrisRackauckas:
No worries, I'll just merge and run the formatter. Thanks!

TheZombie1999:
Thanks

PR #1757, opened 2022-09-14
Preferencially enable or disable "snoop-precompilation" with `set_snoop_precompile!`

ven-k:
- Enable or disable the snoop-precompilation by passing the appropriate boolean value. By default it is enabled aka `true`.
- These preferences are stored in a LocalPrefernces.toml.
- Instead, to store it in (OrdinaryDiffEq)Project.toml, set `export_prefs = true`


src/preferences.jl
```diff
@@ -0,0 +1,19 @@
...
+Instead, to store it in (OrdinaryDiffEq)Project.toml, set `export_prefs = true`
+"""
+function set_snoop_precompile!(
+    include_snoop_precompile::Bool = true;
+    export_prefs::Bool = false,
+    )
+    set_preferences!(@__MODULE__,
+        "include_snoop_precompile" => include_snoop_precompile;
+        export_prefs,
+        force = true)
```
devmotion, line 16: I wonder what's the advantage of using a separate function instead of `set_preferences!` directly, as eg ForwardDiff recommends in its docs? You would have to load OrdinaryDiffEq first before you can call the function anyway, so if you want to disable precompilation you have to wait until the (by default enabled) precompilation is done. And in the other case you would have to restart Julia and reload OrdinaryDiffEq to trigger precompilation.

I think the better approach might be to tell users to use `set_preferences!`, but with the UUID instead of the module. Then you could change the setting before loading OrdinaryDiffEq, it seems.

ven-k: I wanted to assert the types of value `include_snoop_precompile` can have; so wrapped it in a function.

There isn't much difference in either of methods. We will have to `import` or be `using` a package while setting a preference (while passing either UUID or Module)

```julia
julia> using Preferences, UUIDs

julia> ordinary_diff_eq_uuid = UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed")
UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed")

julia> set_preferences!(ordinary_diff_eq_uuid, "enable_snoop_precompilation" => true)
ERROR: Cannot set preferences of an unknown package that is not loaded!
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] set_preferences!(u::UUID, prefs::Pair{String, Bool}; export_prefs::Bool, active_project_only::Bool, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Preferences ~/.julia/packages/Preferences/VmJXL/src/Preferences.jl:256
 [3] set_preferences!(u::UUID, prefs::Pair{String, Bool})
   @ Preferences ~/.julia/packages/Preferences/VmJXL/src/Preferences.jl:210
 [4] top-level scope
   @ REPL[5]:1
```

ven-k: I've fixed the formatting and haven't dropped the function yet. 
LMK

ven-k:
Closing in favor of #1759

src/preferences.jl
devmotion: Strange, it worked for me with eg ForwardDiff without loading it: https://github.com/JuliaDiff/ForwardDiff.jl/pull/598 I only had to add it to my environment.

devmotion: I rechecked and it definitely works for me to set the preference without loading the package.

PR #1759, opened 2022-09-15
Allow for precompilation choices via Preferences

ChrisRackauckas:


ChrisRackauckas:
@devmotion @ven-k @chriselrod any reason for it to be more complicated than this?

chriselrod:
> @devmotion @ven-k @chriselrod any reason for it to be more complicated than this?

This looks fine to me.

ven-k:
This granular control helps out. LGTM

src/OrdinaryDiffEq.jl
```diff
@@ -265,23 +271,59 @@ SnoopPrecompile.@precompile_all_calls begin
...
-                 #ODEProblem{true, SciMLBase.FunctionWrapperSpecialize}(lorenz, [1.0; 0.0; 0.0], (0.0, 1.0));
-                 #ODEProblem{true, SciMLBase.FunctionWrapperSpecialize}(lorenz, [1.0; 0.0; 0.0], (0.0, 1.0), Float64[])
-                 #ODEProblem{true, SciMLBase.NoSpecialize}(lorenz, [1.0; 0.0; 0.0], (0.0, 1.0));
-                 #ODEProblem{true, SciMLBase.NoSpecialize}(lorenz, [1.0; 0.0; 0.0], (0.0, 1.0), Float64[])
-                 #ODEProblem{false,false}(lorenz_oop,[1.0;0.0;0.0],(0.0,1.0))
-                 #ODEProblem{false,false}(lorenz_oop,[1.0;0.0;0.0],(0.0,1.0),Float64[])
-                 ]
+    solver_list = []
+
+    if Preferences.@load_preference("PrecompileNonStiff", true)
```
devmotion, line 276: One could enforce booleans here with e.g.
```suggestion
    if Preferences.@load_preference("PrecompileNonStiff", true) === true
```
But I don't know if that's necessary.

devmotion (commented):
It would be good to document that one might want to set the preference via the UUID (see https://github.com/JuliaDiff/ForwardDiff.jl/pull/598). Otherwise, if one wants to disable precompilation and loads OrdinaryDiffEq first, then everything is already precompiled and changes are only picked up after restarting Julia and reloading OrdinaryDiffEq. And the other way around, if you want to re-enable it, you have to restart Julia and reload OrdinaryDiffEq if it's already loaded.

One approach to make this a bit more convenient and avoid that people have to figure out the UUID could be to add helper functions in e.g. DiffEqBase such as `set_ordinarydiffeq_preference!(args::Pairs{String}...; kwargs...) = Preferences.set_preference!(UUID("...."), args...; kwargs...)`.

ChrisRackauckas:
> It would be good to document that one might want to set the preference via the UUID (see https://github.com/JuliaDiff/ForwardDiff.jl/pull/598). Otherwise, if one wants to disable precompilation and loads OrdinaryDiffEq first, then everything is already precompiled and changes are only picked up after restarting Julia and reloading OrdinaryDiffEq. And the other way around, if you want to re-enable it, you have to restart Julia and reload OrdinaryDiffEq if it's already loaded.

Agreed. I still haven't figured out how/where to document all of the new specialization controls, I'll keep this in mind.

src/OrdinaryDiffEq.jl
ChrisRackauckas: Doesn't using a value as an expression for an if statement enforce that it should be a boolean?

devmotion: No, Julia will just throw an error. As stated in https://docs.julialang.org/en/v1/manual/control-flow/#man-conditional-evaluation,

> Unlike C, MATLAB, Perl, Python, and Ruby – but like Java, and a few other stricter, typed languages – it is an error if the value of a conditional expression is anything but true or false:

ChrisRackauckas: yes, it will error saying it needs to be a bool, just like the `=== true`?

devmotion: No, `=== true` will never error as it is guaranteed to always return a boolean. In case `@load_preference` returns `true`, it will be `true`, and otherwise it will be `false`.

ChrisRackauckas: I see, but it would be good to through an error instead of making it silent when it's incorrectly defined?

devmotion: I guess it is a design choice but I think it would be reasonable to throw an informative error. Probably the only question is if the default error message is clear enough:
```julia
julia> function f(x)
           if x
               return 42
           else
               return -1
           end
       end
f (generic function with 1 method)

julia> f(true)
42

julia> f(1)
ERROR: TypeError: non-boolean (Int64) used in boolean context
Stacktrace:
 [1] f(x::Int64)
   @ Main ./REPL[7]:2
 [2] top-level scope
   @ REPL[9]:1

julia> f("hello")
ERROR: TypeError: non-boolean (String) used in boolean context
Stacktrace:
 [1] f(x::String)
   @ Main ./REPL[7]:2
 [2] top-level scope
   @ REPL[10]:1
```

PR #1762, opened 2022-09-16
Use both sides of `tspan` when determining `dtmin`

oscardssmith:
This prevents spurious errors when solving with small `tspan` (e.g. circuits).

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -260,7 +260,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
             steps = length(tstops)
         else
             # For fixed dt, the only time dtmin makes sense is if it's smaller than eps().
-            # Thereore user specified dtmin doesn't matter, but we need to ensure dt>=eps()
+            # Therefore user specified dtmin doesn't matter, but we need to ensure dt>=eps()
             # to prevent infinite loops.
             dtmin = DiffEqBase.prob2dtmin(prob)
             abs(dt) < dtmin && throw(ArgumentError("Supplied dt is smaller than dtmin"))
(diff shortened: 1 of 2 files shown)


oscardssmith:
Is the gpu test known to have problems? It looks very unrelated.

ChrisRackauckas:
It does but ignore it for now since I don't have a GPU I can use to debug it right now. It's been like that for about a month.

oscardssmith:
I didn't cancel them, so I don't know.

ChrisRackauckas:
It looks like the Interface I tests have a real issue with this.

oscardssmith:
any idea why? this pr really seems like it shouldn't change much.

ChrisRackauckas:
Maybe it pushed dtmin below some eps where it gets stuck in an infinite loop? You'll need to look at that example and figure out it.

src/solve.jl
```diff
@@ -259,8 +259,10 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
         if dt == 0
             steps = length(tstops)
         else
-            # if not adaptive, user set `dt` should be respected.
-            dtmin = something(dtmin, zero(tspan[1]))
+            # For fixed dt, the only time dtmin makes sense is if it's smaller than eps().
+            # Thereore user specified dtmin doesn't matter, but we need to ensure dt>=eps()
```
ChrisRackauckas, line 263:
```suggestion
            # Therefore user specified dtmin doesn't matter, but we need to ensure dt>=eps()
```

ChrisRackauckas:
Can this get a test case?

oscardssmith:
I believe this is ready to merge.

PR #1763, opened 2022-09-16
make initial `dt` guess for DAEs. based on `tspan`

oscardssmith:
This is much better when simulating over small or large time ranges.


oscardssmith:
@ChrisRackauckas merge?

ChrisRackauckas (approved):

PR #1767, opened 2022-09-22
Allow for choosing only autodiff or finitediff in stiff precompile

ChrisRackauckas:
This may cut down the precompilation time.

Thoughts:

* Autoswitch is generally only used by the default algorithm, and the default algorithm only uses finite difference, so only precompile the finite difference one.
* Maybe move the autoswitch precompilation to DifferentialEquations.jl itself if it's a big piece. We should test the difference here
* Stiff is generally used with autodiff, so don't precompile the finite diff there.

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -221,60 +221,41 @@ SnoopPrecompile.@precompile_all_calls begin
         BS3(), Tsit5(), Vern7(), Vern9(),
     ]
 
-    stiff = [Rosenbrock23(), Rosenbrock23(autodiff = false),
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
```julia
@time using OrdinaryDiffEq

# Before
171.884657 seconds (21.87 M allocations: 1.608 GiB, 0.50% gc time, 1.01% compilation time: 56% of

# After
146.106109 seconds (21.62 M allocations: 1.562 GiB, 0.64% gc time, 1.36% compilation time: 56% of which was recompilation)

# No Switching
143.139529 seconds (21.32 M allocations: 1.528 GiB, 1.10% gc time, 2.11% compilation time: 44% of which was recompilation)
```

it's not a substantial difference.

PR #1768, opened 2022-09-22
fix remake for extrapolation methods

ChrisRackauckas:


utkarsh530:
Thanks for the PR! Apologies for how this mistake crept on.

PR #1248, opened 2020-08-17
add OrdinaryDiffEqLinearExponentialAlgorithm abstract type

kleinhenz:
* this allows get_tmp_cache to be properly specialized
* closes #1247

I'm not sure if this is the right approach but this seems to fix my problem.

PR #1777, opened 2022-10-08
Fixed error for LinearExponential algorithm when using Callbacks

albertomercurio:
Hello,

similarly to #1248 I created a separate definition of `get_tmp_cache` for the LinearExponential algorithm. Indeed, prior to this change the following code failed

```julia
using OrdinaryDiffEq
using LinearAlgebra
using SparseArrays
A = sprand(ComplexF64, 100, 100, 0.5)
A += A'

t_l = LinRange(0, 1, 100)

saved_values = SavedValues(Float64, Float64)
function save_func(u, t, integrator)
    real(u' * A * u)
end
cb = SavingCallback(save_func, saved_values, saveat = t_l)

u0 = normalize(rand(ComplexF64, 100))
A = DiffEqArrayOperator(-1im * A)
prob = ODEProblem(A, u0, (0, 1.0))
solve(prob, LinearExponential(), dt = t_l[2] - t_l[1], callback = cb);




ERROR: LoadError: type LinearExponential has no field dz
```

### DIFF
--- a/src/integrators/integrator_interface.jl
+++ b/src/integrators/integrator_interface.jl
@@ -165,6 +165,11 @@ end
                                           cache::OrdinaryDiffEqMutableCache)
     (cache.tmp,)
 end
+@inline function DiffEqBase.get_tmp_cache(integrator,
+    alg::LinearExponential,
+    cache::OrdinaryDiffEqMutableCache)
+(cache.tmp,)
+end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Thanks this looks correct. Can you add your example as a test?

albertomercurio:
Ok, I added the test.

ChrisRackauckas:
Thanks!

PR #1778, opened 2022-10-13
Use `undefmatrix` instead of `zeromatrix`

oscardssmith:
`undefmatrix` is roughly 10x faster which yields a roughly 2x improvement for autostiff solvers on large non-stiff equations.


oscardssmith:
Specifically, I'm now seeing
```
f(u, p, t) = -u
prob = ODEProblem(f, rand(1000), (0,1));
julia> @btime sol = solve(prob, Tsit5());
  188.681 μs (358 allocations: 2.50 MiB)

julia> @btime sol = solve(prob, AutoTsit5(Rodas5P()));
  213.582 μs (389 allocations: 10.19 MiB)
```
While before it was
```
julia> @btime sol = solve(prob, Tsit5());
  187.081 μs (358 allocations: 2.50 MiB)

julia> @btime sol = solve(prob, AutoTsit5(Rodas5P()));
  674.642 μs (389 allocations: 10.19 MiB)
```

oscardssmith:
CI failure appears to be that `zeromatrix` works for numbers, but `undefmatrix` doesn't. This therefore needs a followup PR to `ArrayInterface` :(

PR #1779, opened 2022-10-13
Added modification and test to allow for AD of LinearExponential

aelmokadem:
Updated line 735 in `src/perform_step/linear_perform_step.jl` to swap `exp` with `exponential!` which would allow for AD of LinearExponential. Also added an accompanying test to the end of `test/interface/ad_tests.jl`. Related to the issue here https://discourse.julialang.org/t/can-we-use-linearexponential-with-turing-jl-model/88479


ChrisRackauckas (approved):

ChrisRackauckas:
I'll handle the formatting issue. Thanks.

aelmokadem:
@ChrisRackauckas No problem. Glad I was able to help.

Issue #1773, opened 2022-09-29
Saving Jacobians created


gstein3m:
using DifferentialEquations

function ode_foo(du, u, p, t)
 du[1] = u[1] + t^10
end

prob = ODEProblem(ode_foo, [0.0], (0.0, 1.0))
sol = solve(prob, Rodas4())
println(sol.destats)

gives the statistics: 

Number of function 1 evaluations:                  210
Number of function 2 evaluations:                  0
Number of W matrix evaluations:                    26
Number of linear solves:                           156
Number of Jacobians created:                       26
Number of nonlinear solver iterations:             0
Number of nonlinear solver convergence failures:   0
Number of rootfind condition calls:                0
Number of accepted steps:                          21
Number of rejected steps:                          5

Rosenbrock methods need only a new W matrix but not a new Jacobian when a step is rejected. Thus only 21 Jocobians would be required?

ChrisRackauckas:
Thanks! Yes, that does look like it could save some calculations. @YingboMa is investigating it in #1780

PR #1780, opened 2022-10-13
Don't recompute Jacobian when it's fresh

YingboMa:
Fix https://github.com/SciML/OrdinaryDiffEq.jl/issues/1773

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -16,10 +16,24 @@ steps:
     # Don't run Buildkite if the commit message includes the text [skip tests]
     if: build.message !~ /\[skip tests\]/
 
-  - label: "Regression_I"
+  - label: "{{matrix.group}} v{{matrix.version}}"
+    matrix:
+      setup:
+        version:
+          - "1.6"
+          - "1"
+        group:
+          - "Regression_I"
+          - "Regression_II"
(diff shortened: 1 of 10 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
@YingboMa you're keeping on this one?

gstein3m:
In order to save one more function evaluation when the Jacobian is fresh, we could call         
   calc_tderivative!(integrator, cache, dtd1, true)
instead of        
  calc_tderivative!(integrator, cache, dtd1, repeat_step)

ChrisRackauckas:
yes, that's a good point. @YingboMa were you finishing this one?

YingboMa:
Somehow I cannot reproduce the test failure.

YingboMa:
Let's wait the new DelayDiffEq.jl's release and run the CI again.

ChrisRackauckas:
> In order to save one more function evaluation when the Jacobian is fresh, we could call

I believe that is already done?
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/derivative_utils.jl#L807-L808

YingboMa:
That's from this PR.

ChrisRackauckas:
oh oops haha brain dead

PR #1782, opened 2022-10-15
ci: use build matrix

thazhemadam:
Use buildkite's build matrix instead of having duplicated CI steps.


ChrisRackauckas:
Can you help do the same for SciMLSensitivity and StochasticDiffEq?

thazhemadam:
Yep, on it.

PR #1783, opened 2022-10-17
Check error after each `loop_header!` in `step!`

### LINKED ISSUE #1769: non-adaptive `step!` can get stuck in infinite loop
If using iterator_interface.jl `step!(integrator::ODEIntegrator)` and we get `integrator.accept_step=false` then we enter the `while` loop.  If then `perform_step!`  hits `nlsolvefail(nlsolver) = true` (in sdirk_perform_step.jl) then we will never exit the while loop.  When entering the `loopfooter` I can see that `integrator.last_stepfail = true`, therefore the `loopfooter` simply returns and the `while` loop continues.

YingboMa:
Fix #1769

### DIFF
--- a/src/iterator_interface.jl
+++ b/src/iterator_interface.jl
@@ -13,6 +13,7 @@
         @inbounds loopfooter!(integrator)
         @inbounds while !integrator.accept_step
             loopheader!(integrator)
+            check_error!(integrator) != :Success && return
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

YingboMa:
Downstream failures seem unrelated.

PR #1786, opened 2022-10-31
allow precompiling low-storage methods via preferences

ranocha:
This would be nice for us in Trixi.jl.


ranocha:
The downstream test failure of SciMLSensitivity.jl seems unrelated to me.

ChrisRackauckas:
And it's just hitting a Github Actions limit.

PR #1788, opened 2022-11-02
Update Benchmarks_WP_Rodas.jl

gstein3m:
There was a sign error.


ChrisRackauckas:
These benchmarks aren't used anymore and can be removed, but I'll merge this anyways.

PR #1790, opened 2022-11-04
switch `_unwrap_val` to `SciMLBase`

oscardssmith:


src/algorithms.jl
```diff
@@ -4406,8 +4406,7 @@ Scientific Computing, 18 (1), pp. 1-22.
 différentiels-algébriques. MSc mathematics thesis, Faculty of Science,
 University of Geneva, Switzerland.
 =#
-_unwrap_val(::Val{B}) where {B} = B
-_unwrap_val(B) = B
+import SciMLBase._unwrap_val
```
devmotion, line 4409: Could this be moved to the other imports in src/OrdinaryDiffEq.jl? And maybe don't use `import` since we don't want to extend it?

YingboMa: Yeah, I would prefer that as well.

oscardssmith: done.

PR #1798, opened 2022-11-15
Optimize QNDF's step size control

YingboMa:
QNDF_t_dt


oscardssmith:
what are these changes doing morally?

YingboMa:
QNDF_t_dt

YingboMa:
Two changes:
1. The old logic doesn't change the step size at all when the step is accepted and order=1. This is likely a bug. It now applies the normal step size selection heuristic in this case as well.
2. Since QNDF requires O+2 constant sized steps to be taken to increase the order, the old logic just holds the step size to be constant when the step is accepted and the number of constant sized steps isn't reached. The new logic only prefers to not change the step size by clamping `q`, so it doesn't totally disengages the controller most of the time.

PR #1806, opened 2022-11-30
Remove any dependence of the nlsolve pieces from DiffEqBase

ChrisRackauckas:
Just a little clean up.


src/nlsolve/type.jl
```diff
@@ -6,6 +9,62 @@
     GLM
 end
 
+@enum NLStatus::Int8 begin
```
oscardssmith, line 12: Do these really belong here?

ChrisRackauckas: Yes, it's specific to these solvers and internal to the package.

ChrisRackauckas:
I will release, do the DiffEqBase deletion, then a release which bumps the bound, then a yank.

PR #1815, opened 2022-12-09
Initialize DAE using NonlinearSolve

MarcBerliner:
+ Adds an `nlsolve` kwarg to `ShampineCollocationInit` and `BrownFullBasicInit` to select custom `NonlinearSolve` solvers
+ Supports `jac_prototype` and `jac`

cc @oscardssmith


test/interface/dae_initialize_integration.jl
```diff
@@ -35,7 +35,7 @@ p0 = [
 ]
 
 prob = ODEProblem(connected, u0, tspan, p0)
-sol = solve(prob, Rodas5(), initializealg = BrownFullBasicInit())
```
oscardssmith, line 38: why are you changing the initial test?

src/initialize_dae.jl
```diff
@@ -1,14 +1,18 @@
 struct DefaultInit <: DiffEqBase.DAEInitializationAlgorithm end
 
-struct ShampineCollocationInit{T} <: DiffEqBase.DAEInitializationAlgorithm
+struct ShampineCollocationInit{T,F} <: DiffEqBase.DAEInitializationAlgorithm
     initdt::T
+    nlsolve::F
 end
-ShampineCollocationInit() = ShampineCollocationInit(nothing)
+ShampineCollocationInit(;initdt=nothing,nlsolve=nothing) = ShampineCollocationInit(initdt,nlsolve)
+ShampineCollocationInit(initdt) = ShampineCollocationInit(;initdt=initdt,nlsolve=nlsolve)
```
oscardssmith, line 8: where is `nlsolve` coming from here?

test/interface/dae_initialize_integration.jl
MarcBerliner: This is an ODE. If you run it with `initializealg` it tells you the kwarg is unsupported.

src/initialize_dae.jl
```diff
@@ -36,36 +40,52 @@ end
...
 end
 
 function _initialize_dae!(integrator, prob::ODEProblem,
                           alg::DefaultInit, x::Val{false})
+    isad = SciMLBase.forwarddiffs_model(integrator.alg)
+    if integrator.u isa Number || integrator.u isa StaticArray
+        nlsolve = SimpleNewtonRaphson(autodiff=isad)
+    else
+        nlsolve = NewtonRaphson(autodiff=isad)
+    end
```
oscardssmith, line 56: this functionality is duplicated. Should it be a method?

src/initialize_dae.jl
```diff
@@ -36,36 +40,52 @@ end
...
     _initialize_dae!(integrator, prob,
-                     BrownFullBasicInit(integrator.opts.abstol), x)
+                     BrownFullBasicInit(integrator.opts.abstol, nlsolve), x)
 end
 
 function _initialize_dae!(integrator, prob::DAEProblem,
                           alg::DefaultInit, x::Val{false})
+    isad = SciMLBase.forwarddiffs_model(integrator.alg)
+    if (integrator.u isa Number && integrator.du isa Number) ||
+       (integrator.u isa StaticArray && integrator.du isa StaticArray)
```
oscardssmith, line 66: will u and `du` ever be different types?

src/initialize_dae.jl
MarcBerliner: Should be `nothing`, will fix

src/initialize_dae.jl
```diff
@@ -150,15 +171,25 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
             out .-= tmp
             nothing
         end
-        r = nlsolve(nlequation!, u0, autodiff = isad ? :forward : :central,
-                    method = :newton)
-        integrator.u .= r.zero
+
+        if alg.nlsolve !== nothing
+            nlsolve = alg.nlsolve
+        else
```
oscardssmith, line 177: why doesn't this have the simplenewtonrhapson case?

src/initialize_dae.jl
```diff
@@ -227,22 +269,39 @@ function _initialize_dae!(integrator, prob::DAEProblem,
 
     dt = t != 0 ? min(t / 1000, dtmax) : dtmax # Haven't implemented norm reduction
 
-    nlequation! = @closure (out, u) -> begin
+    f(resid, integrator.du, u0, p, t)
+    integrator.opts.internalnorm(resid, t) <= integrator.opts.abstol && return
+
+    isad = SciMLBase.forwarddiffs_model(integrator.alg)
+    _tmp = isad ?
```
oscardssmith, line 276: IMO if you need 3 lines, don't use a terinary.

src/initialize_dae.jl
```diff
@@ -261,13 +320,26 @@ function _initialize_dae!(integrator, prob::DAEProblem,
...
-    nlprob = NonlinearProblem(nlequation, u0)
-    sol = solve(nlprob, SimpleNewtonRaphson(autodiff = false))
-    integrator.u = sol.u
+    if alg.nlsolve !== nothing
+        nlsolve = alg.nlsolve
+    elseif u0 isa Number || u0 isa StaticArray
+        nlsolve = SimpleNewtonRaphson()
+    else
+        nlsolve = NewtonRaphson()
+    end
```
oscardssmith, line 329: this code is duplicated in multiple places.

src/initialize_dae.jl
```diff
@@ -317,12 +390,24 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::BrownFullBasicInit,
...
-    alg_u .= r.zero
+    if f.jac_prototype === nothing
+        J = nothing
+    else
+        J = f.jac_prototype[algebraic_eqs, algebraic_vars]
+    end
+
+    if alg.nlsolve !== nothing
+        nlsolve = alg.nlsolve
+        @show nlsolve
```
oscardssmith, line 401: delete the `@show`

src/initialize_dae.jl
MarcBerliner: Chris said "for the oop ones, we want to defualt to SimpleNonlinearSolve when scalar or staticarray so it's fully non-allocating, then the standard NonlinearSolve otherwise"

src/initialize_dae.jl
```diff
@@ -399,16 +509,47 @@ function _initialize_dae!(integrator, prob::DAEProblem,
...
-        f(out, du, u, p, t)
+    isad = SciMLBase.forwarddiffs_model(integrator.alg)
+    if isad
+        chunk = ForwardDiff.pickchunksize(length(tmp))
+        _tmp = PreallocationTools.dualcache(tmp, chunk)
+        _du_tmp = PreallocationTools.dualcache(du_tmp, chunk)
+    else
+        _tmp, _du_tmp = tmp, similar(tmp)
+    end
+    # nlequation = @closure (out, x) -> begin
```
oscardssmith, line 520: delete obsolete code

test/integrators/dae_initialization_tests.jl
```diff
@@ -164,7 +164,8 @@ du₀ = [0.0, 0.0, 0.0]
 tspan = (0.0, 1.0)
 differential_vars = [true, true, false]
 prob = DAEProblem(f, du₀, u₀, tspan, differential_vars = differential_vars)
-integrator = init(prob, DABDF2())
+# Initializing the problem fails with SingularException(3)
+@test_broken integrator = init(prob, DABDF2())
```
oscardssmith, line 168: this needs to be solved before we're ready to merge.

src/initialize_dae.jl
MarcBerliner: Unsure. I copied the format from https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/initialize_dae.jl#L432

src/initialize_dae.jl
MarcBerliner: Simple is only for oop problems

src/initialize_dae.jl
```diff
@@ -298,15 +370,16 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::BrownFullBasicInit,
     integrator.opts.internalnorm(tmp, t) <= alg.abstol && return
     alg_u = @view u[algebraic_vars]
 
-    isad = alg_autodiff(integrator.alg)
+    isad = SciMLBase.forwarddiffs_model(integrator.alg)
```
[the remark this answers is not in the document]
MarcBerliner: `alg_autodiff` errors if the algorithm doesn't have AD defined. `SciMLBase.forwarddiffs_model` returns a Boolean

src/initialize_dae.jl
oscardssmith: agreed. I just mean that these 5 lines or so are duplicated in about 3 places so I was wondering if it should be pulled out into a function

src/initialize_dae.jl
```diff
@@ -1,14 +1,36 @@
...
+struct BrownFullBasicInit{T, F} <: DiffEqBase.DAEInitializationAlgorithm
     abstol::T
+    nlsolve::F
+end
+function BrownFullBasicInit(; abstol = 1e-10, nlsolve = nothing)
+    BrownFullBasicInit(abstol, nlsolve)
+end
+BrownFullBasicInit(abstol) = BrownFullBasicInit(; abstol = abstol, nlsolve = nothing)
+
+default_nlsolve(alg::Any, isinplace, u, autodiff = false) = alg
```
ChrisRackauckas, line 23:
```suggestion
default_nlsolve(alg, isinplace, u, autodiff = false) = alg
```

src/initialize_dae.jl
```diff
@@ -1,14 +1,36 @@
...
+function default_nlsolve(::Nothing, ::Val{false}, u, autodiff = false)
+    SimpleNewtonRaphson(autodiff = autodiff)
+end
+function default_nlsolve(::Nothing, ::Val{true}, u::T,
+                         autodiff = false) where {T <: Union{<:Number, StaticArray}}
```
ChrisRackauckas, line 28: if it's in place, then those can't be `T`

src/initialize_dae.jl
```diff
@@ -49,23 +71,21 @@ end
...
     if prob.differential_vars === nothing
-        _initialize_dae!(integrator, prob,
-                         ShampineCollocationInit(), x)
+        initializealg = ShampineCollocationInit()
     else
-        _initialize_dae!(integrator, prob,
-                         BrownFullBasicInit(integrator.opts.abstol), x)
+        initializealg = BrownFullBasicInit(integrator.opts.abstol)
     end
+    _initialize_dae!(integrator, prob, initializealg, x)
```
ChrisRackauckas, line 78: This change gives a dynamic dispatch. Revert it so it uses branching instead.

src/initialize_dae.jl
```diff
@@ -49,23 +71,21 @@ end
...
```
ChrisRackauckas, line 88: Same.

src/initialize_dae.jl
```diff
@@ -150,19 +174,26 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
             out .-= tmp
             nothing
         end
-        r = nlsolve(nlequation!, u0, autodiff = isad ? :forward : :central,
-                    method = :newton)
-        integrator.u .= r.zero
+
+        nlsolve = default_nlsolve(alg.nlsolve, isinplace, u0, isad)
+
+        nlfunc = NonlinearFunction(nlequation!; jac = f.jac,
```
ChrisRackauckas, line 180: The Jacobian of this function is not the same as the Jacobian of the ODE. `f.jac` would have to be modified to be useful here. Instead it's `M/dt - J` for J being `f.jac` (I think, someone would need to double check. Shampine collocation init is simply an implicit Euler step that is used to converge the algebraic variables, so it should be that, one would just have to double check the place of the constant dt), so one would need to `if has_jac(f); new_jac(du,u,p,) = ...` and define that equation. Let's leave that for a follow up PR though.

src/initialize_dae.jl
```diff
@@ -150,19 +174,26 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
...
+                                   jac_prototype = f.jac_prototype)
```
ChrisRackauckas, line 181: though it is true that this should be the same sparsity pattern, because by definition it's M/gamma - J

src/initialize_dae.jl
```diff
@@ -205,19 +236,25 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
             M * (u - u0) / dt - f(u, p, t)
         end
 
-        nlprob = NonlinearProblem(nlequation_oop, u0)
-        nlsol = solve(nlprob, SimpleNewtonRaphson(autodiff = false))
+        nlsolve = default_nlsolve(alg.nlsolve, isinplace, u0)
+
+        nlfunc = NonlinearFunction(nlequation_oop; jac = f.jac,
```
ChrisRackauckas, line 241: same jac issue here

src/initialize_dae.jl
```diff
@@ -227,27 +264,43 @@ function _initialize_dae!(integrator, prob::DAEProblem,
 
     dt = t != 0 ? min(t / 1000, dtmax) : dtmax # Haven't implemented norm reduction
 
-    nlequation! = @closure (out, u) -> begin
+    f(resid, integrator.du, u0, p, t)
+    integrator.opts.internalnorm(resid, t) <= integrator.opts.abstol && return
+
+    isad = SciMLBase.forwarddiffs_model(integrator.alg)
```
ChrisRackauckas, line 270: same here

src/initialize_dae.jl
```diff
@@ -227,27 +264,43 @@ function _initialize_dae!(integrator, prob::DAEProblem,
...
         f(out, tmp, u, p, t)
         nothing
     end
 
-    nlequation!(tmp, u0)
-    f(resid, integrator.du, u0, p, t)
-    integrator.opts.internalnorm(resid, t) <= integrator.opts.abstol && return
+    nlsolve = default_nlsolve(alg.nlsolve, isinplace, u0, isad)
+
+    nlfunc = NonlinearFunction(nlequation!; jac = f.jac, jac_prototype = f.jac_prototype)
```
ChrisRackauckas, line 288: same here

src/initialize_dae.jl
```diff
@@ -261,13 +314,20 @@ function _initialize_dae!(integrator, prob::DAEProblem,
     resid = f(integrator.du, u0, p, t)
     integrator.opts.internalnorm(resid, t) <= integrator.opts.abstol && return
 
-    nlprob = NonlinearProblem(nlequation, u0)
-    sol = solve(nlprob, SimpleNewtonRaphson(autodiff = false))
-    integrator.u = sol.u
+    nlsolve = default_nlsolve(alg.nlsolve, isinplace, u0)
+
+    nlfunc = NonlinearFunction(nlequation; jac = f.jac, jac_prototype = f.jac_prototype)
```
ChrisRackauckas, line 319: same here

src/initialize_dae.jl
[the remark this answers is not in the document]
ChrisRackauckas: What algorithm can hit here without having autodiff defined? That seems like something that should be caught earlier.

src/initialize_dae.jl
```diff
@@ -348,25 +418,39 @@ function _initialize_dae!(integrator, prob::ODEProblem,
 
     integrator.opts.internalnorm(resid, t) <= alg.abstol && return
 
+    isad = SciMLBase.forwarddiffs_model(integrator.alg)
```
ChrisRackauckas, line 421: same here

src/initialize_dae.jl
```diff
@@ -399,16 +484,37 @@ function _initialize_dae!(integrator, prob::DAEProblem,
         error("differential_vars must be set for DAE initialization to occur. Either set consistent initial conditions, differential_vars, or use a different initialization algorithm.")
     end
 
-    nlequation = @closure (out, x) -> begin
-        @. du = ifelse(differential_vars, x, du)
-        @. u = ifelse(differential_vars, u, x)
-        f(out, du, u, p, t)
+    isad = SciMLBase.forwarddiffs_model(integrator.alg)
```
ChrisRackauckas, line 487: same here

src/initialize_dae.jl
```diff
@@ -444,11 +550,15 @@ function _initialize_dae!(integrator, prob::DAEProblem,
         f(du, u, p, t)
     end
 
-    nlprob = NonlinearProblem(nlequation, ifelse.(differential_vars, du, u))
-    sol = solve(nlprob, SimpleNewtonRaphson(autodiff = false))
+    nlsolve = default_nlsolve(alg.nlsolve, isinplace, integrator.u)
+
+    nlfunc = NonlinearFunction(nlequation; jac = f.jac, jac_prototype = f.jac_prototype)
```
ChrisRackauckas, line 555: same here

test/interface/dae_initialize_integration.jl
ChrisRackauckas: It should be supported. That's a bug that should get fixed then.

test/interface/dae_initialize_integration.jl
```diff
@@ -35,7 +35,7 @@ p0 = [
...
+sol = solve(prob, Rodas5())
 @test prob.u0 == sol[1]
-sol = solve(prob, Rodas5(), initiailizealg = ShampineCollocationInit())
```
ChrisRackauckas, line 40: same here

test/interface/dae_initialize_integration.jl
ChrisRackauckas: It's in the supported list: https://github.com/SciML/DiffEqBase.jl/blob/master/src/solve.jl#L61 is this just a spelling error?

oscardssmith:
I've made https://github.com/SciML/OrdinaryDiffEq.jl/pull/1816 because it seems like a good idea (and will give you a fun rebase :smile: )

ChrisRackauckas:
Superseded by https://github.com/SciML/OrdinaryDiffEq.jl/pull/1817

PR #1816, opened 2022-12-13
improve `ShampineCollocationInit` initial dt choice

oscardssmith:
this makes the in place and OOP more aligned for `DAEProblem` and makes the `ODEProblem` `dt` timespan aware.

### DIFF
--- a/src/initialize_dae.jl
+++ b/src/initialize_dae.jl
@@ -105,7 +105,8 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
     u0 = integrator.u
 
     dt = if alg.initdt === nothing
-        integrator.dt != 0 ? min(integrator.dt / 5, dtmax) : (prob.tspan[end] - prob.tspan[begin]) // 1000 # Haven't implemented norm reduction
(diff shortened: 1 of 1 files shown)


src/initialize_dae.jl
```diff
@@ -225,7 +227,7 @@ function _initialize_dae!(integrator, prob::DAEProblem,
     tmp = get_tmp_cache(integrator)[1]
     resid = get_tmp_cache(integrator)[2]
 
-    dt = t != 0 ? min(t / 1000, dtmax) : dtmax # Haven't implemented norm reduction
+    dt = t != 0 ? min(t / 1000, dtmax / 10) : dtmax / 10 # Haven't implemented norm reduction
```
oscardssmith, line 230: @ChrisRackauckas is there a reason this isn't checking for `alg.initdt`?

oscardssmith:
why does CI think things are failing? the logs don't show an error

ChrisRackauckas:
Valentin turned off the MIT computers.

src/initialize_dae.jl
ChrisRackauckas: maybe it should

PR #1818, opened 2022-12-16
allow init and solve! on DEProblems with u=nothing

oscardssmith:
@ChrisRackauckas can you confirm if this is a reasonable design for this? It still needs tests, but I'm unsure if this is the right way to do this so I want feedback before I deal with all the edge cases.


oscardssmith:
@ChrisRackauckas thoughts?

ChrisRackauckas:
Why would it not occur at the DiffEqBase level? How would IDA be supported?

oscardssmith:
@ChrisRackauckas thoughts?

PR #1820, opened 2022-12-24
Handle sparsity patterns which are specified as missing some diagonal

ChrisRackauckas:
Since it's M - gamma*J where gamma changes around, it always needs to have the non-zeros of M be nonzeros of the sparsity pattern. We just slap those in there.

### DIFF
--- a/test/interface/nonfulldiagonal_sparse.jl
+++ b/test/interface/nonfulldiagonal_sparse.jl
@@ -1,0 +1,115 @@
+using OrdinaryDiffEq, DiffEqOperators, SparseArrays
+using UnPack
+using ComponentArrays
+using Symbolics
+
+function enclosethetimedifferential(parameters::NamedTuple)::Function
+    @info "Enclosing the time differential"
+
+    @unpack Δr, r_space, countorderapprox = parameters.compute
+    countdiscretizationsteps = length(r_space)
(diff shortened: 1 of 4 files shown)


test/interface/nonfulldiagonal_sparse.jl
```diff
@@ -0,0 +1,102 @@
+using OrdinaryDiffEq, DiffEqOperators, SparseArrays
```
ChrisRackauckas, line 1:
```suggestion
using OrdinaryDiffEq, DiffEqOperators, SparseArrays, LinearSolve
```

test/interface/nonfulldiagonal_sparse.jl
```diff
@@ -0,0 +1,102 @@
...
+                       V_c = 18,
+                       Q_l = 20,
+                       Q_r = 3.6,
+                       V_b = 1490,
+                       S = 52,
+                       Lm = 0.05,
+                       Dm = 0.046,
+                       V_v = 18.0)
+
+r_space = collect(range(start = 0.0, stop = 2.0, length = 15))
```
ChrisRackauckas, line 75:
```suggestion
r_space = collect(range(0.0, 2.0, length = 15))
```

PR #1822, opened 2022-12-25
Throw customized error messages with automatic differentiation

ChrisRackauckas:
```julia
using OrdinaryDiffEq, Test

const a = Float64[1.0]

function lorenz(u,p,t)
    du1 = 10.0(u[2]-u[1])
    a[1] = u[2]
    du2 = u[1]*(28.0-u[3]) - u[2]
    du3 = u[1]*u[2] - (8/3)*u[3]
    [du1,du2,du3]
end
u0 = [1.0;0.0;0.0]
tspan = (0.0,1.0)
solve(prob,Rosenbrock23())
```

```
ERROR: First call to automatic differentiation for time gradient
failed. This means that the user `f` function is not compatible
with automatic differentiation. Methods to fix this include:

1. Turn off automatic differentiation (e.g. Rosenbrock23() becomes
   Rosenbrock23(autodiff=false)). More details can befound at
   https://docs.sciml.ai/DiffEqDocs/stable/features/performance_overloads/
2. Improving the compatibility of `f` with ForwardDiff.jl automatic
   differentiation (using tools like PreallocationTools.jl). More details
   can be found at https://docs.sciml.ai/DiffEqDocs/stable/basics/faq/#Autodifferentiation-and-Dual-Numbers
3. Defining analytical Jacobians and time gradients. More details can be
   found at https://docs.sciml.ai/DiffEqDocs/stable/types/ode_types/#SciMLBase.ODEFunction

Note 1: this failure occured inside of the time gradient function. These
time gradients are only required by Rosenbrock methods (`Rosenbrock23`,
`Rodas4`, etc.) are are done by automatic differentiation w.r.t. the
argument `t`. If your function is compatible with automatic differentiation
w.r.t. `u`, i.e. for Jacobian generation, another way to work around this
issue is to switch to a non-Rosenbrock method.

Note 2: turning off automatic differentiation tends to have a very minimal
performance impact (for this use case, because it's forward mode for a
square Jacobian. This is different from optimization gradient scenarios).
However, one should be careful as some methods are more sensitive to
accurate gradients than others. Specifically, Rodas methods like `Rodas4`
and `Rodas5P` require accurate Jacobians in order to have good convergence,
while many other methods like BDF (`QNDF`, `FBDF`), SDIRK (`KenCarp4`),
and Rosenbrock-W (`Rosenbrock23`) do not. Thus if using an algorithm which
is sensitive to autodiff and solving at a low tolerance, please change the
algorithm as well.
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEq"
 uuid = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
-version = "6.36.0"
+version = "6.36.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/derivative_wrappers.jl
+++ b/src/derivative_wrappers.jl
@@ -2,7 +2,7 @@ const FIRST_AUTODIFF_TGRAD_MESSAGE = """
                                First call to automatic differentiation for time gradient
                                failed. This means that the user `f` function is not compatible
                                with automatic differentiation. Methods to fix this include:
-                               
+
                                1. Turn off automatic differentiation (e.g. Rosenbrock23() becomes
                                   Rosenbrock23(autodiff=false)). More details can befound at
                                   https://docs.sciml.ai/DiffEqDocs/stable/features/performance_overloads/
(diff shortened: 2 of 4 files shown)


ChrisRackauckas:
```julia
# Before
# 47.900 μs (443 allocations: 38.97 KiB)
# 48.400 μs (443 allocations: 38.97 KiB)
# 48.200 μs (443 allocations: 38.97 KiB)
# After
# 48.300 μs (443 allocations: 38.97 KiB)
# 48.400 μs (443 allocations: 38.97 KiB)
# 49.000 μs (443 allocations: 38.97 KiB)
```

Impact is tiny.

PR #1824, opened 2022-12-28
Downstream test and fix MTK by re-enabling autodiff in SciMLNLsolve

ChrisRackauckas:
Fixes https://github.com/SciML/ModelingToolkit.jl/issues/2012

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -30,6 +30,7 @@ jobs:
           - {user: SciML, repo: SciMLSensitivity.jl, group: Core3}
           - {user: SciML, repo: SciMLSensitivity.jl, group: Core4}
           - {user: SciML, repo: SciMLSensitivity.jl, group: Core5}
+          - {user: SciML, repo: ModelingToolkit.jl, group: All}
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
@oscardssmith make note.

oscardssmith:
Good catch.

ChrisRackauckas:
@chriselrod looks like there was a regression in PolyesterWeave.jl where it no longer takes Integer but needs Int in some spot

chriselrod:
That's what I get for not running downstream tests. @Krastanov, think you can fix this? https://buildkite.com/julialang/ordinarydiffeq-dot-jl/builds/1790#0185596c-f7d5-4364-988c-410cf30e74d8

chriselrod:
@Krastanov, I think it's useful to still have some concept of per-machine maximum number of threads (i.e., just like the num cores) that does not get invalidated.

Polyester has a thread mask indicating which threads are occupied. For systems with less than 64 threads, that mask can be a single `UInt64`.
But for systems with more threads than this, it needs to use multiple `UInt64`s.

Issue #1511, opened 2021-10-25
More stage/step limiters and threaded broadcasting (in explicit RK methods)


ranocha:
This project will continue the work discussed in https://github.com/SciML/OrdinaryDiffEq.jl/issues/1423 and started in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1508. The basic idea is to add fields

- `stage_limiter!`
- `step_limiter!`
- `thread`

to more/all (explicit RK) algorithms and extend the corresponding `create_cache` and `perform_step!` methods, adding 

- calls to the `stage_limiter!` after computing an RK stage and before evaluating the RHS at that stage
- calls to the `step_limiter!` after computing the new value after a step and before evaluating the RHS there
- the keyword argument `thread=thread` to all calls of `@..` from FastBroadcast.jl and the argument `thread` as last argument of `calculate_residuals!` for adaptive methods

Citing https://github.com/SciML/OrdinaryDiffEq.jl/pull/1508#issuecomment-949714235
> I enabled threaded broadcasting and stage/step limiters for
> 
> * 3Sp and 3SpFSAL low-storage methods
> * 2N low-storage methods
> * `Tsit5`
> * `BS3`
> * SSPRK methods (`SSPRK22`, `SSPRK33`, `SSPRK53`, `SSPRK53_2N1`, `SSPRK53_2N2`, `SSPRK53_H`, `SSPRK63`, `SSPRK73`, `SSPRK83`, `SSPRK43`, `SSPRK432`, `SSPRK932`, `SSPRK54`, `SSPRK104`)
> 
> Left for another PR:
> 
> * SSP RK MS methods (`SSPRKMSVS43`, `SSPRKMSVS32`): they already have limiters
> * All other methods (without limiters so far)

PR #1833, opened 2023-01-11
Add more optional threading

ErikQQY:
This PR continues the work by @ranocha in PR #1508 and adds the optional threading to all of the explicit RK methods.

Fix: #1511 

I still have some questions:

1. As for the algorithm ```Vern7``` and ```Vern9```, in the second perform step function concerning ```Vern7Cache``` and ```Vern9Cache```:

```julia
perform_step!(integrator, cache::Vern7Cache{<:Array}, repeat_step = false)
```

```julia
perform_step!(integrator, cache::Vern9Cache{<:Array}, repeat_step = false)
```
  it uses ```@simd``` rather than ```@..```, do we need to add optional threading here too?

2. As for the algorithms ```Feagin```, I didn't see any ```@..``` used in these algorithms, do we need to add optional threading for these algorithms?


ErikQQY:
I think maybe the buildkite test failure is not caused by this PR?

ChrisRackauckas:
That's a stochastic failure. Don't worry about that one.

ChrisRackauckas:
> it uses @simd rather than @.., do we need to add optional threading here too?

Yes. Those branches are for compile time reduction, though I'm not sure that they are necessary anymore since when you do `broadcast=false` on `@..` the compile time should be sane (without it, `@..` does have compilation over head). Could you take the time to profile whether those dispatches are still needed? If not, the ones for Verner and Tsit5 methods can just be removed.

ChrisRackauckas:
That or Polyester. Will you be following up with that? Otherwise we should make sure that we document that threading choices don't apply to those methods when Float64, but I think a simple deletion is all that's necessary (and testing first solve times)

ErikQQY:
Sure, I will follow up with that. I just benchmarked the ```Vern7``` scheme:

Before I removed that ```@muladd function perform_step!(integrator, cache::Vern7Cache{<:Array}, repeat_step = false)``` dispatch:

didntremove

However, after I removed that dispatch, the performance got a little declined🤔:

afterremove

ChrisRackauckas:
That's because it wasn't actually using multithreading. If you solve with Array{Float64}, it's to hit the Array{Float64} branches. My guess is that this case is too small for multithreading to help.

But `@benchmark` removes compile times. We need to also bench compile times like https://github.com/SciML/DifferentialEquations.jl/issues/786#issuecomment-1000596500

ErikQQY:
I used SnoopCompile.jl to test the compile time:

### Before I removed that branch:
```julia
julia> tinf = @snoopi_deep solve(prob,Vern7(OrdinaryDiffEq.True()))
InferenceTimingNode: 2.508216/3.608172 on Core.Compiler.Timings.ROOT() with 3 direct children
```

And the corresponding flamegraph:

before
 
### After I removed that branch:
```julia
julia> tinf = @snoopi_deep solve(prob,Vern7(OrdinaryDiffEq.True()))
InferenceTimingNode: 2.427201/3.539494 on Core.Compiler.Timings.ROOT() with 3 direct children
```

The corresponding flamegraph:

after


Is this the correct way to test the compile time of the ```Vern7``` scheme? Given the result is approximately equal, I think we can remove that branch now?

ChrisRackauckas:
Yeah, looks like it can be safely removed.

ErikQQY:
OK, I will send another PR to remove these branches.

PR #1834, opened 2023-01-14
fix `ode_determine_initdt` for zero-length arrays

ranocha:
We need to be able to handle zero-length arrays with Trixi.jl - aggressive adaptive mesh refinement for massively parallel MPI simulations can have empty arrays on some ranks initially.

I don't know whether this is the best way to deal with it since it may introduce type instabilities for non-homogeneous arrays. On the other hand, the working version was introduced in #1810 and reverted in #1813 because (citing @oscardssmith)

> the PR caused test suite errors that were not caught before merging

I don't know which kind of errors appeared and whether it would be better to fix these errors somewhere else. But we definitely need to be able to use empty arrays initially.

CC @JoshuaLampert

### DIFF
--- a/src/initdt.jl
+++ b/src/initdt.jl
@@ -57,7 +57,12 @@
     #tmp = cache[2]
 
     if u0 isa Array
-        tmp = similar(u0, typeof(u0[1] / sk[1]))
+        if length(u0) > 0
+            T = typeof(u0[1] / sk[1])
+        else
+            T = promote_type(eltype(u0), eltype(sk))
+        end
+        tmp = similar(u0, T)
         @inbounds @simd ivdep for i in eachindex(u0)
             tmp[i] = u0[i] / sk[i]
         end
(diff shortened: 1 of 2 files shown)


ranocha:
`format-check` fails with a lot of stuff unrelated to this PR. Shall I fix it nevertheless here? It appears to be more sane to do this somewhere else.

oscardssmith:
would https://github.com/SciML/DiffEqBase.jl/pull/857 work for you? in my opinion is a bunch more reliable approach

ChrisRackauckas:
I would think https://github.com/SciML/DiffEqBase.jl/pull/857 in general, avoiding the solvers when it's empty, is the way to go.

ranocha:
I don't think so. We will use adaptive mesh refinement with load balancing, so every MPI process needs to run and enter the callbacks, even if its own share is empty initially - over time, we will increase the workload so that every process needs to be in the main loop, not just returning an empty solution at the very beginning.

PR #1837, opened 2023-01-19
Add docs for optional threading

ErikQQY:
This PR finishes the work in #1833


docs/src/nonstiff/explicitrk.md
```diff
@@ -1,5 +1,7 @@
 # Explicit Runge-Kutta Methods
 
+With the help of [FastBroadcast.jl](https://github.com/YingboMa/FastBroadcast.jl), we can use threaded parallelism to reduce compilation time for all of the explicit Runge-Kutta methods!  The ```thread``` option determines whether internal broadcasting on appropriate CPU arrays should be serial (`thread = OrdinaryDiffEq.False()`, default) or use multiple threads (`thread = OrdinaryDiffEq.True()`) when Julia is started with multiple threads. When we call ```solve(prob, alg(thread=OrdinaryDiffEq.True()))```, we can turn on the multithreading option so that the compiler can easily optimize our code and achieve acceleration.
```
ranocha, line 3:
```suggestion
With the help of [FastBroadcast.jl](https://github.com/YingboMa/FastBroadcast.jl), 
we can use threaded parallelism to reduce compute time for all of the explicit Runge-Kutta methods!
The `thread` option determines whether internal broadcasting on appropriate CPU arrays should be serial
(`thread = OrdinaryDiffEq.False()`, default) or use multiple threads 
(`thread = OrdinaryDiffEq.True()`) when Julia is started with multiple threads. 
When we call `solve(prob, alg(thread=OrdinaryDiffEq.True()))`,
we can turn on the multithreading option to achieve acceleration
(for sufficiently large problems).

```

ranocha: Multithreading is not about compilation time

ErikQQY: Thanks!

Issue #1738, opened 2022-08-26
Error in IRKN4 `initialize!`


baggepinnen:
```julia
using OrdinaryDiffEq
dyn = (u,p,t)->-u
u0 = randn(1)
prob = ODEProblem(dyn, u0, (0.0, Inf))
int  = OrdinaryDiffEq.init(prob, IRKN4(), dt=0.1, adaptive=false) 
```
```julia
julia> using OrdinaryDiffEq

julia> dyn = (u,p,t)->-u
#15 (generic function with 1 method)

julia> u0 = randn(1)
1-element Vector{Float64}:
 -0.18213108995807548

julia> prob = ODEProblem(dyn, u0, (0.0, Inf))
ODEProblem with uType Vector{Float64} and tType Float64. In-place: false
timespan: (0.0, Inf)
u0: 1-element Vector{Float64}:
 -0.18213108995807548

julia> int  = OrdinaryDiffEq.init(prob, IRKN4(), dt=0.1, adaptive=false)
ERROR: type Array has no field x
Stacktrace:
 [1] getproperty
   @ ./Base.jl:38 [inlined]
[...]

PR #1843, opened 2023-01-23
Throw error when using partitioned ODE methods with ODEProblem

ErikQQY:
This PR is a simple fix of the error message for the case when we use the methods for partitioned ODE with ODEProblem.

Fix: #1738 

error

I think maybe we can also define a specific type for these partitioned ODE i.e. `OrdinaryDiffEqPartitionedAlgorithms`, it would be easier for the judgement.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -78,16 +78,12 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
         error("You cannot use an DAE Algorithm with a ODEProblem")
     end
 
-    partitioned_ode_solvers = [Nystrom4(), IRKN3(), IRKN4(), ERKN4(), ERKN5(), ERKN7(), Nystrom4VelocityIndependent(),
-                               Nystrom5VelocityIndependent(), DPRKN4(), DPRKN5(), DPRKN6(), DPRKN6FM(), DPRKN8(), DPRKN12(),
-                               SymplecticEuler(), VelocityVerlet(), VerletLeapfrog(), PseudoVerletLeapfrog(), McAte2(), Ruth3(),
-                               McAte3(), CandyRoz4(), McAte4(), CalvoSanz4(), McAte42(), McAte5(), Yoshida6(), KahanLi6(),
-                               McAte8(), KahanLi8(), SofSpa10()]
+    if prob.problem_type isa DiffEqBase.AbstractDynamicalODEProblem && !(alg isa PartitionedAlgorithm)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

src/solve.jl
```diff
@@ -78,6 +78,18 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
         error("You cannot use an DAE Algorithm with a ODEProblem")
     end
 
+    partitioned_ode_solvers = [Nystrom4(), IRKN3(), IRKN4(), ERKN4(), ERKN5(), ERKN7(), Nystrom4VelocityIndependent(),
+                               Nystrom5VelocityIndependent(), DPRKN4(), DPRKN5(), DPRKN6(), DPRKN6FM(), DPRKN8(), DPRKN12(),
+                               SymplecticEuler(), VelocityVerlet(), VerletLeapfrog(), PseudoVerletLeapfrog(), McAte2(), Ruth3(),
+                               McAte3(), CandyRoz4(), McAte4(), CalvoSanz4(), McAte42(), McAte5(), Yoshida6(), KahanLi6(),
+                               McAte8(), KahanLi8(), SofSpa10()]
```
ChrisRackauckas, line 85: Use a subtype instead

src/solve.jl
```diff
@@ -78,6 +78,18 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
...
+
+    if prob isa DiffEqBase.AbstractODEProblem
+        if in(alg, partitioned_ode_solvers)
+            error("You cannot use an algorithm designed for partitioned ODE with an ODEProblem, please use DynamicalODEProblem/SecondOrderODEProblem/HamiltonianProblem instead.")
```
ChrisRackauckas, line 89: Those generate an ODEProblem though. It should instead check the `prob.problem_type`.

src/solve.jl
ErikQQY: I am not sure if I am getting it right, so we need to specify a type of algorithms like ```abstract OrdinaryDiffEqPartitionedAlgorithms <: OrdinaryDiffEqAlgorithms end```, and then redefine all of the partitioned ODE methods like ```struct Nystrom4 <: OrdinaryDiffEqPartitionedAlgorithms end```?

ChrisRackauckas: Yes, otherwise that type check allocates and is not inferred.

ErikQQY:
odewithrkn
odewithtsit5

ChrisRackauckas (approved):

ChrisRackauckas:
Looks like tests are failing.

PR #1850, opened 2023-01-27
-0.0 to optimize away additions

chriselrod:
Adding `0.0` cannot be optimized away, but adding `-0.0` can. Thus, we replace all `0` with `Float16(-0f0)`, to allow them to be compiled away.
This will be most beneficial to CPUs without FMA support. For CPUs with FMA support, it'll also help them by not requiring a zero-register.

### DIFF
--- a/src/dense/interpolants.jl
+++ b/src/dense/interpolants.jl
@@ -2846,7 +2846,7 @@ end end
     b5Θ = Θ * @evalpoly(Θ, r51, r52, r53, r54)
     b6Θ = Θ * @evalpoly(Θ, r61, r62, r63, r64)
 
-    bp1Θ = Θ * @evalpoly(Θ, rp10, rp11, rp12, rp13, rp14)
+    bp1Θ = @evalpoly(Θ, rp10, rp11, rp12, rp13, rp14)
     bp3Θ = Θ * @evalpoly(Θ, rp31, rp32, rp33, rp34)
     bp4Θ = Θ * @evalpoly(Θ, rp41, rp42, rp43, rp44)
     bp5Θ = Θ * @evalpoly(Θ, rp51, rp52, rp53, rp54)
--- a/src/dense/interpolants.jl
+++ b/src/dense/interpolants.jl
@@ -1,3 +1,5 @@
+const NEGZERO = Float16(-0.0f0)
+
 @muladd function _ode_interpolant(Θ, dt, y₀, y₁, k,
                                   cache::Union{FunctionMapConstantCache, FunctionMapCache},
                                   idxs, T::Type{Val{0}})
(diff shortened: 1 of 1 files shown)


oscardssmith:
why evalpoly these coefficients at all? dropping them and multiplying by a power instead will be faster

chriselrod:
> why evalpoly these coefficients at all? dropping them and multiplying by a power instead will be faster (and more accurate)

Sorry, I don't quite follow?
Mind making a ```suggestion?

ChrisRackauckas (approved):

src/dense/interpolants.jl
```diff
@@ -14,10 +16,10 @@ end
...
 """
 @def dp5pre0 begin
-    b10 = @evalpoly(Θ, 0, 1)
-    b20 = @evalpoly(Θ, 0, 1, -1)
-    b30 = @evalpoly(Θ, 0, 0, 1, -1)
-    b40 = @evalpoly(Θ, 0, 0, 1, -2, 1)
+    b10 = @evalpoly(Θ, NEGZERO, 1)
+    b20 = @evalpoly(Θ, NEGZERO, 1, -1)
+    b30 = @evalpoly(Θ, NEGZERO, NEGZERO, 1, -1)
+    b40 = @evalpoly(Θ, NEGZERO, NEGZERO, 1, -2, 1)
```
devmotion, line 22: IIUC @oscardssmith suggested to use e.g.
```suggestion
    b40 = Θ^2 * @evalpoly(Θ, 1, -2, 1)
```

chriselrod: Ah, sorry, of course. That makes sense.

chriselrod: That should be faster.

src/dense/interpolants.jl
```diff
@@ -14,10 +14,11 @@ end
 Hairer Norsett Wanner Solving Ordinary Differential Euations I - Nonstiff Problems Page 192
 """
 @def dp5pre0 begin
-    b10 = @evalpoly(Θ, 0, 1)
-    b20 = @evalpoly(Θ, 0, 1, -1)
-    b30 = @evalpoly(Θ, 0, 0, 1, -1)
-    b40 = @evalpoly(Θ, 0, 0, 1, -2, 1)
+    Θ² = Θ * Θ
+    b10 = Θ * @evalpoly(Θ, 1)
```
devmotion, line 18: That's just
```suggestion
    b10 = Θ
```
isn't it?

ChrisRackauckas:
I think we just need to bump that tolerance downstream but other than that this looks all correct now. Thanks @devmotion for catching that one

PR #1854, opened 2023-01-29
Add MSRK54 method

ErikQQY:
Part of the #1601 

Corresponding paper: https://arxiv.org/pdf/2108.12590.pdf

MSRK54

PR #1855, opened 2023-01-29
MSRK54 -> Stepanov5 and correct doc string

YingboMa:
Ref: https://github.com/SciML/OrdinaryDiffEq.jl/pull/1854#issuecomment-1407672737

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -369,7 +369,7 @@ export FunctionMap, Euler, Heun, Ralston, Midpoint, RK4, ExplicitRK, OwrenZen3, 
        OwrenZen5,
        BS3, BS5, RK46NL, DP5, Tsit5, DP8, Vern6, Vern7, Vern8, TanYam7, TsitPap8,
        Vern9, Feagin10, Feagin12, Feagin14, CompositeAlgorithm, Anas5, RKO65, FRK65, PFRK87,
-       RKM, MSRK5, MSRK54
+       RKM, MSRK5, Stepanov5
 
 export SSPRK22, SSPRK33, KYKSSPRK42, SSPRK53, SSPRK53_2N1, SSPRK53_2N2, SSPRK53_H, SSPRK63,
        SSPRK73, SSPRK83, SSPRK43, SSPRK432,
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -571,7 +571,7 @@ alg_order(alg::RK4) = 4
 alg_order(alg::RKM) = 4
 alg_order(alg::ExplicitRK) = alg.tableau.order
 alg_order(alg::MSRK5) = 5
-alg_order(alg::MSRK54) = 5
+alg_order(alg::Stepanov5) = 5
 
 alg_order(alg::BS3) = 3
 alg_order(alg::BS5) = 5
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -793,17 +793,17 @@ struct Stepanov5{StageLimiter, StepLimiter, Thread} <: OrdinaryDiffEqAdaptiveAlg
 end
 
 function Stepanov5(; stage_limiter! = trivial_limiter!, step_limiter! = trivial_limiter!,
-               thread = False())
+                   thread = False())
     Stepanov5{typeof(stage_limiter!), typeof(step_limiter!), typeof(thread)}(stage_limiter!,
-                                                                         step_limiter!,
-                                                                         thread)
+                                                                             step_limiter!,
+                                                                             thread)
 end
 
 # for backwards compatibility
 function Stepanov5(stage_limiter!, step_limiter! = trivial_limiter!)
     Stepanov5{typeof(stage_limiter!), typeof(step_limiter!), False}(stage_limiter!,
-                                                                step_limiter!,
-                                                                False())
+                                                                    step_limiter!,
+                                                                    False())
 end
 
 function Base.show(io::IO, alg::Stepanov5)
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -762,11 +762,11 @@ function Base.show(io::IO, alg::MSRK5)
 end
 
 """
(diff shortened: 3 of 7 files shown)


YingboMa:
Pleiades_WPD

misha-stepanov:
Dear OrdinaryDiffEq.jl team,

Thank you for sharing this performance data.

The Butcher tableau in the code seems to be the correct.

A question: are the performance curves in invariant to the scaling of vector btilde? I.e., if I multuply the vector btilde by 2, will the performance curve maybe just shift to the left or right, but overall go over the same curve in the (error, time) plane?

In https://arxiv.org/abs/2108.12590v2 (or Calcolo 59 (4) 41 (2022), DOI: 10.1007/s10092-022-00486-1 ) I was not considering the question of the magnitude of the vector d = btilde, and the performance curves in the article are invariant to its scaling.

Just in case, I attach the Julia code for PLEI problem (it is a bit shortened version of the code I used to generate data in Figure 3, lower panel, test PLEI) of https://arxiv.org/abs/2108.12590v2 ) (and the data it produces if executed as
julia PLEI.jl DP > PLEI_DP
julia PLEI.jl Ts5 > PLEI_Ts5
julia PLEI.jl tAp > PLEI_tAp
the 1st column is the number of r.h.s. evaluations (rejected evaluations are in the 2nd column and are included in the 1st column) (number of r.h.s. evaluations is not execution time, but they should be related), the 3rd column is the L2-norm of the error in coordinates only (1st 14 components of the position vector), and the 4th column is "tolerance", the way how the curves look like is not dependent on the scaling of vector d). In my tests with PLEI problem the performance of all 3 methods is fairly similar, with my method being slightly better (it is problem specific, see data for all 4 problems in Figure 3).  The attached picture graphically shows the data from PLEI_DP, PLEI_Ts5, and PLEI_tAp.

In order to attach the files to this post, I had to rename some of them to have *.txt extension (e.g., not *.jl).

In the graph above the error is in the range [1e-5 : 0.1]. How is the error in defined? In my plots I measure the error as L2-norm of the difference between the numerical and "exact" solutions at the end of the integration interval (t = 3), taking into account only the first 14 components of the 28-dimensional x(t) vector (these 14 components correspond to the coordinates of the stars, while the other 14 are velocities).

Notice that the range of error in my plot is about [1e-11 : 1e-7]. Smaller errors would be already in the regime where round-off errors would dominate (and the error would saturate with the tolerance level). Larger errors would correpond to large steps, and the performance notably depends on what particular and how many steps were rejected, etc. (it is an interesting region still if one is not picky about the accuracy of solution, just wants to get it fast).

I do not expect my method to perform significantly better than Tsit5, as it was constructed to be an equivalent of Tsit5, that satisfies the simplifying assumption (A c)_i = c_i^2 / 2 for all i =/= 2. It has only slightly smaller leading error coefficient T_6, and slightly smaller maximal coefficient max_ij |a_ij| and minimal weight min_j b_j (see Table 2, page 11 of https://arxiv.org/abs/2108.12590v2 ). (One can achieve even smaller value of T_6 by making magnitudes of a_ij and b_j larger.)

Thank you and best regards,
Misha Stepanov

[PLEI.txt](https://github.com/SciML/OrdinaryDiffEq.jl/files/10778393/PLEI.txt)
[PLEI_DP.txt](https://github.com/SciML/OrdinaryDiffEq.jl/files/10778394/PLEI_DP.txt)
[PLEI_Ts5.txt](https://github.com/SciML/OrdinaryDiffEq.jl/files/10778395/PLEI_Ts5.txt)
[PLEI_tAp.txt](https://github.com/SciML/OrdinaryDiffEq.jl/files/10778396/PLEI_tAp.txt)
gnuplot

YingboMa:
Hey, thanks for the feedback.

> In the graph above the error is in the range [1e-5 : 0.1].

The rationale for high tolerance is that higher order methods like Verner methods would dominate anyway.

> How is the error in defined?

By default we only use the final result. https://github.com/SciML/DiffEqDevTools.jl/blob/0b22c286dbb2f88e8caeaa4cd218d1e58b4c00c5/src/benchmark.jl#L181
And it computes the mean of the absolute value of the difference https://github.com/SciML/DiffEqDevTools.jl/blob/0b22c286dbb2f88e8caeaa4cd218d1e58b4c00c5/src/test_solution.jl#L84

I rerun the benchmark with a higher tolerance range and got
bench

Here's the full benchmark code:
```julia
using OrdinaryDiffEq, DiffEqDevTools, Plots

f = (du,u,p,t) -> begin
  @inbounds begin
  x = view(u,1:7)   # x
  y = view(u,8:14)  # y
  v = view(u,15:21) # x′
  w = view(u,22:28) # y′
  du[1:7] .= v
  du[8:14].= w
  for i in 15:28
    du[i] = zero(u[1])
  end
  for i=1:7,j=1:7
    if i != j
      r = ((x[i]-x[j])^2 + (y[i] - y[j])^2)^(3/2)
      du[14+i] += j*(x[j] - x[i])/r
      du[21+i] += j*(y[j] - y[i])/r
    end
  end
  end
end

prob = ODEProblem{true, SciMLBase.FullSpecialize}(f,[3.0,3.0,-1.0,-3.0,2.0,-2.0,2.0,3.0,-3.0,2.0,0,0,-4.0,4.0,0,0,0,0,0,1.75,-1.5,0,0,0,-1.25,1,0,0],(0.0,3.0))
sol = solve(prob,Vern9(),abstol=1e-14,reltol=1e-14,maxiters=1000000)
test_sol = TestSolution(sol)

abstols = 1.0 ./ 10.0 .^ (3:12)
reltols = abstols

setups = [Dict(:alg=>DP5())
          Dict(:alg=>Tsit5())
          Dict(:alg=>Stepanov5())
          Dict(:alg=>Vern6())
          Dict(:alg=>Vern7())
          Dict(:alg=>Vern9())]
wp = WorkPrecisionSet(prob,abstols,reltols,setups;appxsol=test_sol,save_everystep=false,numruns=100,maxiters=10000)
plot(wp)
```

PR #1858, opened 2023-01-31
Unbox interpolant

chriselrod:


ChrisRackauckas:
Why would putting it in a ref help? It's a stack allocated int?

chriselrod:
> Why would putting it in a ref help? It's a stack allocated int?

```julia
     @ /home/chriselrod/.julia/packages/OrdinaryDiffEq/5B7d7/src/dense/generic_dense.jl:218 within `ode_interpolation`
    ┌ @ boot.jl:381 within `Box`
1 ──│ %1  = %new(Core.Box)::Core.Box
│   │ %2  = %new(Core.Box)::Core.Box
```

chriselrod:
If a variable is assigned to inside both a scope and a closure inside that same scope, it will be boxed.

chriselrod:
If you'd prefer, I could use `Base.promote_op` to get the return type of `ode_interpolant` to replace the `map` + closure with a loop.

ChrisRackauckas:
That's probably better. I'd try to avoid ref here if we can. The code should be simple?

chriselrod:
> That's probably better. I'd try to avoid ref here if we can. The code should be simple?

Well, there is also a `cache.caches[id.alg_choice[iplus]]` which should be expanded out.

Is there some obvious return type from `ode_interpolant`? There are a lot of dispatches, broadcasting over arrays or operating on scalars.

chriselrod:
Can I put an `@assert Base.isconcretetype(T)` there (where `T` is the `promote_op`)?

chriselrod:
Replace the `map` with a loop breaks `Zygote`.
Updating the implementation to calculate both the loop and the map, but only use 1 of them:
```julia
# use the map
julia> dp2 = Zygote.gradient((p) -> inner_loop(prob, p, loss_fun), p)[1]
vals == vals2 = true
1-element Vector{Float64}:
 12.156525271268848

# use the loop
julia> dp2 = Zygote.gradient((p) -> inner_loop(prob, p, loss_fun), p)[1]
vals == vals2 = true
```
If we need to support `Zygote`, the easiest way would probably be to just go back to the map + `Ref`.

YingboMa:
Yeah, I agree that Ref is better than `promote_op`. Since the Ref never escapes so it won't allocate in the heap. It's a very common patter in `ccall`s for example, and they don't allocate.

PR #1875, opened 2023-02-16
Replace staticarrays with staticarrayscore

ChrisRackauckas:
Make things tuples that should just be tuples


src/bdf_utils.jl
```diff
@@ -90,7 +90,7 @@ function update_D!(D, dd, k)
     return nothing
 end
 
-const γₖ = @SVector[sum(1 // j for j in 1:k) for k in 1:6]
+const γₖ = ntuple(k->sum(1 // j for j in 1:k), Val(6))
```
oscardssmith, line 93: is the `Val` necessary?

ChrisRackauckas: nope

ChrisRackauckas, line 93:
```suggestion
const γₖ = ntuple(k->sum(1 // j for j in 1:k), 6)
```

Issue #1448, opened 2021-07-14
New DIRK methods of Kennedy and Carpenter


ranocha:
A came across the article ["Diagonally implicit Runge–Kutta methods for stiff ODEs" of Christopher Kennedy and Mark Carpenter (2019)](https://doi.org/10.1016/j.apnum.2019.07.008) and couldn't find an implementation in OrdinaryDiffEq.jl. (Sorry for the noise if I missed it, but that might indicate that there is some room for better documentation :wink:).

ChrisRackauckas:
Shouldn't that be KenCarp47?

ranocha:
Citing the abstract of the paper linked above:

> As the present methods must make no concessions to an explicit method as part of an additive Runge–Kutta method, the present ESDIRKs are generally better than their IMEX analogs.

So I don't think these SDIRK methods are the implicit parts of their IMEX methods published in the same year. Additionally, there are several methods in that paper, up to sixth order

ChrisRackauckas:
Oh interesting. Yes, KenCarp47 and KenCarp58 are https://www.sciencedirect.com/science/article/pii/S0168927418302332 . I didn't know they had another recent paper. Nice.

ranocha:
Should get the label "new algorithm", I guess

PR #1877, opened 2023-02-18
Add ESDIRK437L2SA method

ErikQQY:
Corresponding paper: https://www.sciencedirect.com/science/article/abs/pii/S0168927419301801

Part of https://github.com/SciML/OrdinaryDiffEq.jl/issues/1448

ESDIRK437L2SA


ChrisRackauckas:
I'm not so happy with the naming, but let's go with it for now. Add it to the stiff ODE benchmarks and see how it does? Sounds promising in principle.

ChrisRackauckas:
Add it to SciMLBenchmarks? I think some of the larger ones like Pollu will be hte most interesting test for it.

PR #1889, opened 2023-03-01
Replace destats with stats

avik-pal:


ranocha (commented):
Since this field is documented and part of the public API (see https://docs.sciml.ai/DiffEqDocs/stable/basics/solution/#Special-Fields), this requires a breaking change - or at least a fallback `getproperty` overload, maybe with a deprecation.

ChrisRackauckas:
> this requires a breaking change - or at least a fallback getproperty overload, maybe with a deprecation.

That already exists in the PR making the change.

https://github.com/SciML/SciMLBase.jl/pull/391

Project.toml
```diff
@@ -64,7 +64,7 @@ PreallocationTools = "0.2, 0.3, 0.4"
 Preferences = "1.3"
 RecursiveArrayTools = "2.36"
 Reexport = "0.2, 1.0"
-SciMLBase = "1.83"
+SciMLBase = "1.89"
```
ChrisRackauckas, line 67:
```suggestion
SciMLBase = "1.90"
```

PR #1891, opened 2023-03-03
More fixes on newly introduced show(...) methods

thomvet:
Follow up from https://github.com/SciML/OrdinaryDiffEq.jl/pull/1890

1890 fixed my immediate issue, but then I went back and checked whether the same typo is present in other caches as well and found a few. I hope that I caught them all. 

This pull request should fix the same issue on the other algorithms as well. 

Sorry for not thinking immediately about it and creating double work with merging two pull requests.


ChrisRackauckas:
Thanks!

ChrisRackauckas:
No worries, it's not much extra work. These are hard to test since they are just the show methods, but hopefully everyone is happier with the stacktraces now.

thomvet:
I do find them more useful now for quick debugging and since one can get all the information back by setting the verbose flag it's the best of both worlds. So, great change from my perspective!

PR #1892, opened 2023-03-04
Type-stable operator assumptions about being square

ChrisRackauckas:
I went on a date with Cthulhu and compile times lost.

PR #1893, opened 2023-03-04
Deprecate @unpack to reduce specializations (but keep v1.6 compat)

ChrisRackauckas:


ranocha:
Did you measure the impact on precompile/load times? This would be quite interesting for a lot of my packages, too.

ChrisRackauckas:
image

It's at least a piece of the puzzle

I'll figure out where we're at after https://github.com/SciML/OrdinaryDiffEq.jl/pull/1892 + https://github.com/SciML/SciMLOperators.jl/pull/155

ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/1892 had a bigger effect so far, but we're almost <100mb now when it's all together.

src/adams_utils.jl
```diff
@@ -3,7 +3,11 @@
 # III.5 Variable Step Size Multistep Methods: Formulae 5.9
 function ϕ_and_ϕstar!(cache, du, k)
     @inbounds begin
-        @unpack dts, ϕstar_nm1, ϕ_n, ϕstar_n, β = cache
+        @static if VERSION >= v"1.8"
```
devmotion, line 6: I wonder if this could be made into a custom `@unpack` macro to avoid the code duplications? Or even just changed in UnPack.jl?

ChrisRackauckas: Unpack.jl handles cases other than types, while this does not. And this will be easy to delete once LTS changes.

PR #1895, opened 2023-03-06
don't use the ODE solver for the dae_init.

oscardssmith:
Not sure how this slipped through.

### DIFF
--- a/src/initialize_dae.jl
+++ b/src/initialize_dae.jl
@@ -211,37 +211,19 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
 
     integrator.opts.internalnorm(resid, t) <= integrator.opts.abstol && return
 
-    if isdefined(integrator.cache, :nlsolver)
-        # backward Euler
-        nlsolver = integrator.cache.nlsolver
-        oldγ, oldc, oldmethod, olddt = nlsolver.γ, nlsolver.c, nlsolver.method,
-                                       integrator.dt
-        nlsolver.tmp .= integrator.uprev
-        nlsolver.γ, nlsolver.c = 1, 1
-        nlsolver.method = DIRK
-        integrator.dt = dt
-        z = nlsolve!(nlsolver, integrator, integrator.cache)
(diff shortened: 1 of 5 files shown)


YingboMa (changes requested):
format

ChrisRackauckas:
And we should find a way to test this.

oscardssmith:
any testing suggestions? the only thing I can think of is making a very broken nonlinear solver in the tests.

oscardssmith:
test added.

src/initialize_dae.jl
```diff
@@ -211,37 +211,19 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
...
-        @.. broadcast=false integrator.u=integrator.uprev + z
-    else
-        nlequation_oop = @closure (u, _) -> begin
-            update_coefficients!(M, u, p, t)
-            M * (u - u0) / dt - f(u, p, t)
-        end
+    nlequation_oop = @closure (u, _) -> begin
+        update_coefficients!(M, u, p, t)
+        M * (u - u0) / dt - f(u, p, t)
+    end
```
ChrisRackauckas, line 217: This isn't right, it should default to using the nonlinear solver.

oscardssmith: isn't that what `nlsolve = default_nlsolve(alg.nlsolve, isinplace, u0)` does?

ChrisRackauckas: No, my point is that you shouldn't use NonlinearSolve.jl unless you have to. It should just default to using the internal one.

oscardssmith: ok, so the order you want is `initializealg.nlsolve > integrator.cache.nlsolver > default`. Should this change also be made to the rest of the dae initialization methods?

oscardssmith: wait, even if the user passes a nlsolve to their initialization, you want us to ignore that if the integrator has a solver?

ChrisRackauckas: yes, that's the point of passing something that overrides the default right?

ChrisRackauckas (approved):

PR #1896, opened 2023-03-08
Revert #1893 and switch to SimpleUnPack

devmotion:
Reverts SciML/OrdinaryDiffEq.jl#1893 and switches to SimpleUnPack as an alternative solution for fixing the `@unpack`-induced number of specializations. In contrast to the current approach on master, this PR should also reduce specializations for Julia < 1.8 as `SimpleUnPack.@unpack` is just rewritten to a sequence of `getproperty` calls without any `Val` on older Julia versions and to the Base destructuring syntax on Julia >= 1.7.


devmotion:
Ah, OrdinaryDiffEq uses `@pack!` as well?! Well, that needs some additional functionality in SimpleUnPack then (or has to be replaced) :smile:

ChrisRackauckas:
yeah...

devmotion:
Since `@pack!` is only used twice, it seemed faster to just replace them with hand-written assignments. Could be a nice addition to SimpleUnPack nevertheless (based on `setproperty!` instead of `pack!(..., Val(..), ...)`, of course).

ChrisRackauckas (approved):

PR #1888, opened 2023-03-01
Add SIR54 method

ErikQQY:
Fix: #1408 
Corresponding paper: [Runge–Kutta pairs suited for SIR-type epidemic models](https://onlinelibrary.wiley.com/doi/abs/10.1002/mma.7104)

Convergence test:

SIR54

PR #1898, opened 2023-03-10
Adding Alshina2, Alshina3, Alshin6 from Alshina, Zaks, and Kalitkin (2008)

ayushinav:
Adding the 2nd order from [Alshina, Zaks, and Kalitkin (2008), Optimal first- to sixth-order accurate Runge-Kutta schemes](https://link.springer.com/article/10.1134/S0965542508030068). 
This is not a new algo per se (trying to get familiar with the code structure), but if it works fine, I'll add the other orders soon.


ErikQQY (commented):
Here are some suggestions:
1. Please make sure the new scheme passes the convergence tests to ensure the implementation is correct, for example, you can put this new scheme in a [self-contained example](https://docs.sciml.ai/DiffEqDevDocs/stable/contributing/adding_algorithms/#Self-Contained-Example) to easily test the implementation. Also, you can learn from existing schemes or merged PRs to get familiar with the structure.
2. Add convergence plots so that it would be convenient for committers to be sure this new scheme "behaves" well, see [devdocs](https://docs.sciml.ai/DiffEqDevDocs/stable/contributing/adding_algorithms/) about the convergence plots.
3. Since all of the SciML projects follow the [SciML style](https://github.com/SciML/SciMLStyle), it would be nice if you can format your commit before pushing, see the [CONTRIBUTING.md](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/CONTRIBUTING.md) to know how.

ayushinav:
Alshina2
Alshina3
Alshina4
Alshina6

Thanks for the suggestions, @ErikQQY. Updated the codes and added the other methods.

src/algorithms.jl
```diff
@@ -4086,6 +4086,240 @@ function Base.show(io::IO, alg::OwrenZen5)
...
+  title = {Optimal first- to sixth-order accurate Runge-Kutta schemes},
+  journal = {Computational Mathematics and Mathematical Physics}
+}
+"""
+struct Alshina2{StageLimiter, StepLimiter, Thread} <: OrdinaryDiffEqAlgorithm
+  stage_limiter!::StageLimiter
+  step_limiter!::StepLimiter
+  thread::Thread
+end
+export Alshina2
```
ErikQQY, line 4126: Since we already export this algorithm in the OrdinaryDiffEq.jl file, this export is redundant.

src/algorithms.jl
```diff
@@ -4086,6 +4086,240 @@ function Base.show(io::IO, alg::OwrenZen5)
...
+  title = {Optimal first- to sixth-order accurate Runge-Kutta schemes},
+  journal = {Computational Mathematics and Mathematical Physics}
+}
+"""
+struct Alshina3{StageLimiter, StepLimiter, Thread} <: OrdinaryDiffEqAlgorithm
+  stage_limiter!::StageLimiter
+  step_limiter!::StepLimiter
+  thread::Thread
+end
+export Alshina3
```
ErikQQY, line 4184: Same here.

src/algorithms.jl
```diff
@@ -4086,6 +4086,240 @@ function Base.show(io::IO, alg::OwrenZen5)
...
+  title = {Optimal first- to sixth-order accurate Runge-Kutta schemes},
+  journal = {Computational Mathematics and Mathematical Physics}
+}
+"""
+struct Alshina4{StageLimiter, StepLimiter, Thread} <: OrdinaryDiffEqAlgorithm
+  stage_limiter!::StageLimiter
+  step_limiter!::StepLimiter
+  thread::Thread
+end
+export Alshina4
```
ErikQQY, line 4242: Here

src/algorithms.jl
```diff
@@ -4086,6 +4086,240 @@ function Base.show(io::IO, alg::OwrenZen5)
...
+  title = {Optimal first- to sixth-order accurate Runge-Kutta schemes},
+  journal = {Computational Mathematics and Mathematical Physics}
+}
+"""
+struct Alshina6{StageLimiter, StepLimiter, Thread} <: OrdinaryDiffEqAlgorithm
+  stage_limiter!::StageLimiter
+  step_limiter!::StepLimiter
+  thread::Thread
+end
+export Alshina6
```
ErikQQY, line 4301: And here

test/algconvergence/ode_convergence_tests.jl
```diff
@@ -61,6 +61,18 @@ testTol = 0.2
...
+
+    sim3 = test_convergence(dts3, prob, Alshina3())
+    @test sim3.𝒪est[:l2]≈1 atol=testTol
+
+    sim4 = test_convergence(dts3, prob, Alshina4())
+    @test sim4.𝒪est[:l2]≈3 atol=testTol
+
+    sim6 = test_convergence(dts1, prob, Alshina6())
+    @test sim6.𝒪est[:l2]≈4 atol=testTol
+
```
ErikQQY, line 75: The convergence tests here look weird. What we are doing here is testing whether the order of the convergence estimate is equal to the order of this method. If it doesn't equal, you can adjust ```dts``` so that it doesn't saturate, if it still doesn't converge, then there are bugs in your implementation, you need to carefully compare your implementation with this paper, probably some coefficients or the ```initialize!``` and ```perform_step!``` are wrong.

ayushinav: Right, my implementation had bugs. I was so confused by FSAL being used inspite of the algorithm not being FSAL'd. 

The convergence tests make so much sense now. Thanks!

src/tableaus/low_order_rk_tableaus.jl
```diff
@@ -1884,3 +1884,175 @@ function SIR54ConstantCache(T::Type{<:CompiledFloats}, T2::Type{<:CompiledFloats
...
+    b2 = convert(T, 0.3333333333333333)
+    b3 = convert(T, 0.3333333333333333)
+    b4 = convert(T, 0.16666666666666666)
+
+    c2 = convert(T2, 0.5)
+    c3 = convert(T2, 0.5)
+    c4 = convert(T2, 1.0)
+    
+    Alshina4ConstantCache(a21, a31, a32, a41, a42, a43, b1, b2, b3, b4, c2, c3, c4)
+end
```
ChrisRackauckas, line 1976: This is just RK4?

src/tableaus/low_order_rk_tableaus.jl
```diff
@@ -1884,3 +1884,175 @@ function SIR54ConstantCache(T::Type{<:CompiledFloats}, T2::Type{<:CompiledFloats
...
+
+function Alshina2ConstantCache(T::Type{<:Float64}, T2::Type{<:Float64})
+    a21 = convert(T, 0.6666666666666666)
+
+    b1 = convert(T, 0.25)
+    b2 = convert(T, 0.75)
+
+    c2 = convert(T2, 0.6666666666666666)
+    
+    Alshina2ConstantCache(a21, b1, b2, c2)
```
ChrisRackauckas, line 1906: the non float64 dispatches are missing

src/tableaus/low_order_rk_tableaus.jl
```diff
@@ -1884,3 +1884,175 @@ function SIR54ConstantCache(T::Type{<:CompiledFloats}, T2::Type{<:CompiledFloats
...
+    b3::T
+
+    c2::T2
+    c3::T2
+    
+end
+
+function Alshina3ConstantCache(T::Type{<:Float64}, T2::Type{<:Float64})
+    a21 = convert(T, 0.5)
+    a31 = convert(T, 0.0)
```
ChrisRackauckas, line 1925: remove the zeros from the calculation

src/tableaus/low_order_rk_tableaus.jl
```diff
@@ -1884,3 +1884,175 @@ function SIR54ConstantCache(T::Type{<:CompiledFloats}, T2::Type{<:CompiledFloats
...
+    a32 = convert(T, 0.5)
+    a41 = convert(T, 0.0)
+    a42 = convert(T, 0.0)
+    a43 = convert(T, 1.0)
+
+    b1 = convert(T, 0.16666666666666666)
+    b2 = convert(T, 0.3333333333333333)
+    b3 = convert(T, 0.3333333333333333)
+    b4 = convert(T, 0.16666666666666666)
+
```
ChrisRackauckas, line 1970: Embedded schemes are missing?

ayushinav: Ahh, I guess I did not include them because the paper had only for RK2 and RK3. My bad, will include them shortly.

src/tableaus/low_order_rk_tableaus.jl
ayushinav: Do I need to resolve anything here? This is indeed just RK4.

ChrisRackauckas: Delete it because it's not useful.

ayushinav:
Updated

ChrisRackauckas:
This is missing adaptive regression tests.

ayushinav:
> This is missing adaptive regression tests.

I felt this might come up, but I sort of developed based on [this](https://github.com/SciML/OrdinaryDiffEq.jl/pull/1888/files#), I guessed it might have been taken care of. I'll look into it.

ayushinav:
Hi, I am a bit confused about what to add in the adaptivity tests. My implementation passes the tests for most parts but I guess it still has some bugs.
```
function lorenz(du, u, p, t)
    [10.0(u[2] - u[1]) - du[1]
     u[1] * (28.0 - u[3]) - u[2] - du[2]
     u[1] * u[2] - (8 / 3) * u[3] - du[3]]
end
u0 = [1.0; 0.0; 0.0]
tspan = (0.0, 100.0)

ode = ODEProblem((du, u, semi, t) -> du .= u, Float64[], (0.0, 1.0))

integrator = init(ode, Alshina2())
@test integrator.dt ≈ 1e-6

# Adaptivity regression tests for Alshina2

prob_linear = ODEProblemLibrary.prob_ode_linear

prob_lorenz = ODEProblem{true}(lorenz, u0, tspan)

sol_linear = solve(prob_linear, Alshina2())
@show length(sol_linear.u)

sol_lorenz = solve(prob_lorenz, Alshina2())
@show length(sol_lorenz.u)
```
The above (not an MWE exactly) returns 
```
length(sol_linear.u) = 2040
length(sol_lorenz.u) = 8
```
`Tsit5()`, `DP5()` return lower values for the first case but then they are higher orders. Just not sure if I am making some obvious error somewhere.

ChrisRackauckas:
Those values don't seem reasonable. I'd double check that the error estimator is giving reasonable values.

ayushinav:
> Those values don't seem reasonable. I'd double check that the error estimator is giving reasonable values.

I think I found what's the issue. In the above code, `lorenz` is not updating `du`, which is why the lorenz solution was absurdly small compared to the linear one. Instead if we use
```
function lorenz!(du,u,p,t)
 du[1] = 10.0(u[2]-u[1])
end
u0 = [1.0;0.0;0.0]
tspan = (0.0,1.0)
prob_lorenz = ODEProblem(lorenz!,u0,tspan)
```
I get 
```
length(sol_linear.u) = 2040
length(sol_lorenz.u) = 64110
```
which seems reasonable.

ayushinav:
Also, I guess there's a small bug in [here](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/perform_step/low_order_rk_perform_step.jl#L1787), `utilde` is not brought in. I get the `utilde not defined` when I run the in place `lorenz` fn using `SIR54()`.

ChrisRackauckas:
Test failure looks real.

ayushinav:
I was not clear what the tests meant. Made the changes. Hope it looks better now.

test/regression/ode_adaptive_tests.jl
```diff
@@ -125,3 +125,22 @@ sol_linear = solve(prob_linear, ESDIRK547L2SA2())
...
+    du[1] = 10.0(u[2]-u[1])
+    du[2] = u[1]*(28.0-u[3]) - u[2]
+    du[3] = u[1]*u[2] - (8/3)*u[3]
+end
+u0 = [1.0;0.0;0.0]
+tspan = (0.0,1.0)
+prob_lorenz = ODEProblem(lorenz!,u0,tspan)
+
+sol_linear = solve(prob_linear, Alshina2())
+sol_lorenz = solve(prob_lorenz, Alshina2())
```
ChrisRackauckas, line 141: and 6?

test/regression/ode_adaptive_tests.jl
```diff
@@ -125,3 +125,22 @@ sol_linear = solve(prob_linear, ESDIRK547L2SA2())
...
+@test length(sol_linear.u) < length(sol_lorenz.u)
```
ChrisRackauckas, line 142: I'm not sure this relative test makes sense.

test/regression/ode_adaptive_tests.jl
```diff
@@ -125,3 +125,22 @@ sol_linear = solve(prob_linear, ESDIRK547L2SA2())
...
+# Adaptivity tests for Alshina2, 3
+
+function lorenz!(du,u,p,t)
+    du[1] = 10.0(u[2]-u[1])
+    du[2] = u[1]*(28.0-u[3]) - u[2]
+    du[3] = u[1]*u[2] - (8/3)*u[3]
+end
+u0 = [1.0;0.0;0.0]
+tspan = (0.0,1.0)
+prob_lorenz = ODEProblem(lorenz!,u0,tspan)
```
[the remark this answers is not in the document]
ayushinav: True. The same function is defined on line 33. While I was trying to understand the tests, I realized that `ESDIRK` tests use the `lorenz` defined on line 59, which does not update the `du` and will always give 8 as the length of the soln, irrespective of the adaptive method used. Though `ESDIRK` pass the tests on right `lorenz` too, I did it to explicitly show what's happening.

test/regression/ode_adaptive_tests.jl
ayushinav: The embedded schemes were not defined for ERK6. [Reference](https://link.springer.com/article/10.1134/S0965542508030068)

test/regression/ode_adaptive_tests.jl
ayushinav: I am not entirely surely sure about what to put in. Could you please point to what exactly I might be putting in? Thanks

ChrisRackauckas:
Rebase onto master.

test/regression/ode_adaptive_tests.jl
ChrisRackauckas: Make it a regression test like the ones on linear above.

ayushinav: Would this be fine?
```
for prob in [prob_ode_2Dlinear, prob_ode_linear]
    sol = solve(prob, Alshina2())
    val= maximum(abs.(sol.u[end] - sol.u_analytic[end]))
    @test val<1e-6
    
    sol = solve(prob, Alshina3())
    val= maximum(abs.(sol.u[end] - sol.u_analytic[end]))
    @test val<1e-6
end
```

ayushinav:
Rebased again. Please let me know if I have to make any changes in here.

ChrisRackauckas (approved):

ErikQQY:
After adding these algorithms, it would be nice if you can record these new methods in solvers docs: [DiffEqDocs.jl](https://docs.sciml.ai/DiffEqDocs/stable/)

PR #1899, opened 2023-03-11
Minor code cleanup

gaurav-arya:


gaurav-arya:
@ChrisRackauckas there was a typo here (tests totally failed), fixed in #1900

PR #1908, opened 2023-03-16
TrucatedStacktraces: use macro not `show`

Vaibhavdixit02:


src/algorithms.jl
```diff
@@ -6927,7 +6799,7 @@ for Alg in [
                            t::Type{$Alg{CS, AD, F, P, FDT, ST, CJ}}) where {CS, AD, F, P,
                                                                             FDT, ST, CJ}
             if TruncatedStacktraces.VERBOSE[]
```
Vaibhavdixit02, line 6801: @ChrisRackauckas do you know if there is some other work around for this? The only problem I see with this current approach is that it might miss any updates that happen in TruncatedStacktraces

ChrisRackauckas: Why not just use the macro here?

Vaibhavdixit02: Because of `eval`'s scoping the `__module__` here https://github.com/SciML/TruncatedStacktraces.jl/blob/master/src/TruncatedStacktraces.jl#L59 doesn't contain the algorithms

ChrisRackauckas: @avik-pal will come back to handle this, but let's merge what we have for now.

Issue #1064, opened 2020-03-12
Lie Group Integrators


ChrisRackauckas:
For `u' = A(u,p,t)*u`. In theory this could apply to arbitrary `f`, but we can just show in documentation how to use the Jacobian for this.

From https://arxiv.org/pdf/1207.0069.pdf : 

- [ ] Lie–Euler 
- [ ]  Runge–Kutta–Munthe-Kaas 2
- [ ]  Crouch-Grossman 2
- [ ] Runge–Kutta–Munthe-Kaas 4 (RKMK4)
- [ ] LieRK4 (extension to 16)

http://papers.rgrossman.com/journal-011.pdf

- [ ] CG3
- [ ] CGMS3

https://link.springer.com/content/pdf/10.1023/A:1016645730465.pdf

- [ ] CG4a
- [ ] CG5a

https://link.springer.com/content/pdf/10.1023/A:1022325426017.pdf

- [ ] CG3 
- [ ] CG4
- [ ] CGSDIRK3
- [ ] CGESDIRK3

Biswajitghosh98:
Lie-Euler implemented in #1090

mateuszbaran:
Out of curiosity, how is one supposed to supply the function for the group exponential?

ChrisRackauckas:
You define your `f` as an operator with and the operator specification has https://docs.juliadiffeq.org/latest/features/diffeq_operator/#AbstractDiffEqLinearOpeartor-Interface-Description-1 `exp` and `expv`

mateuszbaran:
Nice. I'll try integrating it with the Lie group functionality of Manifolds.jl.

ChrisRackauckas:
oh interesting. Yes, let us know what you need. If we're missing anything in the interface to do this effectively, we can make modifications.

mateuszbaran:
Sure, I'll let you know if I encounter any problems.

mateuszbaran:
Could you always use calls like `expv(dt, L, u)` when performing a step for Lie group integrators?Converting the operator to matrix for all non-Krylov methods seems restrictive. Similarly I'd like to be able to separately get `dt`, `L` and `u` for other methods (in case they are required by other integrators then `LieEuler`) like the Lie group action, logarithmic map or the differential of the group action.

By the way, wouldn't your `LieEuler` work for any Riemannian manifold given the right exponential map? Is there a reason for restricting it to Lie groups?

[...]

PR #1921, opened 2023-03-30
Add CG4a method

ErikQQY:
Part of #1064 

Corresponding paper: https://link.springer.com/content/pdf/10.1023/A:1016645730465.pdf

Convergence test:

CG4a_convergence


mateuszbaran:
For reference, ManifoldDiffEq.jl has this method (https://github.com/JuliaManifolds/ManifoldDiffEq.jl/blob/main/src/frozen_solvers.jl) with additional functionalities like possibility to use optimized or approximate exponential maps, or exponential maps not defined by matrix exponential, as well as support for arbitrary definitions of frozen coefficient vector fields.

ChrisRackauckas:
Yeah and using these with defined manifolds is much better. This is fine for completeness though.

PR #1923, opened 2023-03-30
remove "Exiting" from @warn message

bennn:


ChrisRackauckas:
It does force exit though?

ranocha:
No, warnings do not interrupt the execution of code:

```julia
julia> function foo(x)
           if x > 0
               @warn "x is positive"
           end
           
           return x^2
       end

julia> foo(1)
┌ Warning: x is positive
└ @ Main REPL[1]:3
1
```

By the way, `@error` also does not throw an exception, it just prints stuff in red to the terminal:

```julia
julia> function foo(x)
           if x > 0
               @error "x is positive"
           end
           
           return x^2
       end

julia> foo(1)
┌ Error: x is positive
└ @ Main REPL[4]:3
1
```

You need to `throw` of you want to stop the execution:

```julia
julia> function foo(x)
           if x > 0
               @error "x is positive"
               throw(ArgumentError("x is really positive"))
           end
           
           return x^2
       end

julia> foo(1)
┌ Error: x is positive
└ @ Main REPL[6]:3
ERROR: ArgumentError: x is really positive
Stacktrace:
 [1] foo(x::Int64)
   @ Main ./REPL[6]:4
 [2] top-level scope
   @ REPL[7]:1
```

ChrisRackauckas:
That exact line doesn't do it, but the ODE solver directly exits right after that.

bennn:
Oh! Didn't realize it's supposed to exit.

It wasn't exiting for us, though. We'll take a closer look & report back.
cc @ashton314

ashton314:
I'm having a hard time seeing where this is supposed to exit.

Moreover, I've been able to reproduce the bug where this goes into an infinite loop. What's supposed to be the catch to exit after this error is tripped?

ChrisRackauckas:
> Is the program supposed to stop before it gets to the solve! function?

Yes, initdt will detect a NaN https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/initdt.jl#L130-L133 and it'll exit via the same path.

> If tdir is set to NaN on solve.jl:106

How does that happen? Who's setting it? That doesn't make sense. It's defined using the `sign` function https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L106 , so that could only happen if `tspan` has NaNs, which would hit the NaN exit I just mentioned above.

> What's supposed to be the catch to exit after this error is tripped?

The default unstable check will fail.

https://github.com/SciML/DiffEqBase.jl/blob/master/src/common_defaults.jl#L57-L67

Which would then make the standard error check fail (so this should be common to all solver packages, not even just this one)

https://github.com/SciML/SciMLBase.jl/blob/master/src/integrator_interface.jl#L555-L607

Which will cause an exit in the solve loop since the return code has changed https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L517-L518

bennn:
> > If tdir is set to NaN on solve.jl:106
> 
> How does that happen? Who's setting it? That doesn't make sense. It's defined using the sign function https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L106 , so that could only happen if tspan has NaNs, which would hit the NaN exit I just mentioned above.

We did that.

We have a tool that injects NaNs to (hopefully) discover issues and it replaced l.106 with `tdir = NaN`. 

If that NaN could never happen, then this is a false alarm and we need to refine the tool!

ChrisRackauckas:
Do you have an MWE of a case where a NaN doesn't exit?

ashton314:
You can check out this repo: https://github.com/utahplt/FloatTrackerExamples

Install the deps per usual. If you've already got OrdinaryDiffEq installed, I don't think you'll have a problem running the example.

Then run the example from the root of the Examples repo with:

```
julia --project=. examples/nbody_replay.jl
```

There's a replay file included in the repo that injects a NaN at just the right point.

ChrisRackauckas:
The first step to debugging here would be to take that an make an MWE. Have you created this without NBodySimulator involved and just an ODE definition?

ashton314:
No we have not. That's a good idea though. I'll see what I can do in the next little bit.

ashton314:
Got it:

```julia
println("Loading ODE...")
using OrdinaryDiffEq
println("Loading ODE...done")
f(u,p,t) = 1.01*u
u0=1/2
tspan = (0,NaN)
println("Setting up problem...")
prob = ODEProblem(f,u0,tspan)
println("Setting up problem...done")
println("Solving...")
sol = solve(prob,Tsit5(),reltol=1e-8,abstol=1e-8)
println("Solving...done")
```

Running this produces a warning message:

```
Loading ODE...
Loading ODE...done
Setting up problem...
Setting up problem...done
Solving...
┌ Warning: Automatic dt set the starting dt as NaN, causing instability. Exiting.
└ @ OrdinaryDiffEq ~/Research/ode_debug/dev/OrdinaryDiffEq/src/solve.jl:554
```

but it never finishes.

Yes, we just stuffed a `NaN` into the `tspan`—if we were computing `tspan` dynamically and accidentally got a `NaN` in there, this might be a bit tricky to debug.

bennn:
@ChrisRackauckas ok to open an issue & close this PR?
(removing "Exiting" is definitely not the right thing!)

PR #1927, opened 2023-04-11
Add ESDIRK659L2SA method

sar1kumar:
Corresponding paper: [https://www.sciencedirect.com/science/article/abs/pii/S0168927419301801](https://www.sciencedirect.com/science/article/abs/pii/S0168927419301801)
Part of https://github.com/SciML/OrdinaryDiffEq.jl/issues/1448


test/regression/ode_adaptive_tests.jl
```diff
@@ -126,6 +126,14 @@ sol_linear = solve(prob_linear, ESDIRK547L2SA2())
 sol_lorenz = solve(prob_lorenz, ESDIRK547L2SA2())
 @test length(sol_lorenz.u) < 1000
 
+# ESDIRK659L2SA
+
+sol_linear = solve(prob_linear, ESDIRK659L2SA())
+@test length(sol_linear.u) < 10
```
ErikQQY, line 132: I tested this method before, and it converges as expected, but the regression tests always failed(Yeah, I checked the adaptivity parameters and I can sure they are correct), really confusing🤔

see

image

ChrisRackauckas:
Getting first order convergence.

sar1kumar:
> Getting first order convergence.

Yeah trying to figure out the reason for the error in convergence and regression tests.

ChrisRackauckas:
Add `@test_broken` and mention this issue in the method's docstring.

PR #1929, opened 2023-04-14
use SnoopPrecompile.@precompile_setup

aviatesk:
This allows us to turn off precompilation of OrdinaryDiffEq via Preferences.jl

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -212,148 +212,150 @@ include("composite_algs.jl")
 
 import SnoopPrecompile
 
-SnoopPrecompile.@precompile_all_calls begin
-    function lorenz(du, u, p, t)
-        du[1] = 10.0(u[2] - u[1])
-        du[2] = u[1] * (28.0 - u[3]) - u[2]
-        du[3] = u[1] * u[2] - (8 / 3) * u[3]
+SnoopPrecompile.@precompile_setup let
+    SnoopPrecompile.@precompile_all_calls let
+        function lorenz(du, u, p, t)
+            du[1] = 10.0(u[2] - u[1])
+            du[2] = u[1] * (28.0 - u[3]) - u[2]
+            du[3] = u[1] * u[2] - (8 / 3) * u[3]
+        end
+
+        function lorenz_oop(u, p, t)
+            [10.0(u[2] - u[1]), u[1] * (28.0 - u[3]) - u[2], u[1] * u[2] - (8 / 3) * u[3]]
+        end
+
+        nonstiff = [
+            BS3(), Tsit5(), Vern7(), Vern9(),
+        ]
+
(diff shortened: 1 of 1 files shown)


aviatesk:
By wrapping the precompilation workload by `SnoopPrecompile.@precompile_setup`, we can turn off the precompilation via Preferences.jl.
For example, you can have `LocalPreferences.toml` at the root of the project with the following contents
```toml
[SnoopPrecompile]
skip_precompile = ["OrdinaryDiffEq"]
```
and then the precompilation will be turned off. This setting can be useful when we develop OrdinaryDiffEq or a parent package using it since precompiling OrdinaryDiffEq takes quite some time.

ChrisRackauckas:
But you could've already done that before with the more refined precompilation controls? I guess having another way to do it is fine. Can you add this to the documentation page on precompilation controls in DiffEqDocs?

aviatesk:
> But you could've already done that before with the more refined precompilation controls?

Yes, we can always edit the package file directly. The Preferences.jl approach is more useful when we develop a package that has OrdinaryDiffEq.jl as its dependency, since it allows us to turn of the compilation of this package just by placing `LocalPreferences.toml` on the parent package directly.

> Can you add this to the documentation page on precompilation controls in DiffEqDocs?

Yeah, but I guess OrdinaryDiffEq/docs is the better place for that?

ChrisRackauckas:
> The Preferences.jl approach is more useful when we develop a package that has OrdinaryDiffEq.jl as its dependency, since it allows us to turn of the compilation of this package just by placing LocalPreferences.toml on the parent package directly.

That doesn't answer the question though, since it's still Preferences.

> Yeah, but I guess OrdinaryDiffEq/docs is the better place for that?

But the old docs also need to be updated.

aviatesk:
> That doesn't answer the question though, since it's still Preferences.

I'm not sure I understand the question. SnoopPrecompile automatically loads Preferences.jl and looks for configurations specified in LocalPreferences.toml when a package that uses SnoopPrecompile is precompiled.

> But the old docs also need to be updated.

Okay, I will submit a PR to DiffEqDocs.jl too. I will push a commit that update this package's `docs` quickly.

aviatesk:
Wow, it looks like SciML's documentation is even more sophisticatedly structured than I anticipated. It seems like we just need to update DiffEqDocs. I believe that this pull request is good to go as it stands.

aviatesk:
@ChrisRackauckas bump?

ChrisRackauckas:
I don't recall seeing the DiffEqDocs PR?

ChrisRackauckas:
I don't recall seeing the DiffEqDocs PR?

ChrisRackauckas:
@oscardssmith is there something to update from this?

oscardssmith:
I think this can just be closed, but ideally @aviatesk would confirm.

PR #1934, opened 2023-04-19
always use `integrator.opts.abstol`

oscardssmith:
BrownFullBasic was using `alg.abstol` which seems like a bug.


staticfloat:
Trying this out further, I think this was actually correct; we want to be able to set initialization `abstol` separately from solve `abstol`, and setting it in the `daeinit` algorithm object allows us to do that.

PR #1935, opened 2023-04-20
Added documentation for the Runge-Kutta-Nyström methods.

HenryLangner:
Added a few lines of documentation for the Runge-Kutta-Nyström methods.  Emphasize is on whether the second-order ODE, which is being solved, is allowed to depend on the first derivative or not.

CC @ranocha

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -5247,7 +5247,7 @@ struct Nystrom4 <: OrdinaryDiffEqPartitionedAlgorithm end
 """
 A 4th order explicit Runkge-Kutta-Nyström method. Used directly on second order ODEs, where the acceleration is independent from velocity (ODE Problem is not dependent on the first derivative).
 
-More efficient then Nystrom4 on velocity independent Problems, since less evaluations are needed.
+More efficient then [`Nystrom4`](@ref) on velocity independent problems, since less evaluations are needed.
 
 Fixed time steps only.
 
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -5232,6 +5232,8 @@ Second order ODE should not depend on the first derivative.
 struct IRKN3 <: OrdinaryDiffEqPartitionedAlgorithm end
 
 """
+    Nystrom4
+
 A 4th order explicit Runge-Kutta-Nyström method which can be applied directly on second order ODEs. Can only be used with fixed time steps.
 
 In case the ODE Problem is not dependent on the first derivative consider using 
@@ -5245,6 +5247,8 @@ E. Hairer, S.P. Norsett, G. Wanner, (1993) Solving Ordinary Differential Equatio
 struct Nystrom4 <: OrdinaryDiffEqPartitionedAlgorithm end
 
 """
+    Nystrom4VelocityIdependent
+
 A 4th order explicit Runkge-Kutta-Nyström method. Used directly on second order ODEs, where the acceleration is independent from velocity (ODE Problem is not dependent on the first derivative).
 
 More efficient then [`Nystrom4`](@ref) on velocity independent problems, since less evaluations are needed.
@@ -5259,6 +5263,8 @@ E. Hairer, S.P. Norsett, G. Wanner, (1993) Solving Ordinary Differential Equatio
 struct Nystrom4VelocityIndependent <: OrdinaryDiffEqPartitionedAlgorithm end
 
 """
+    IRKN4
+
 Improves Runge-Kutta-Nyström method of order four, which minimizes the amount of evaluated functions in each step. Fixed time steps only.
 
 Second order ODE should not be dependent on the first derivative.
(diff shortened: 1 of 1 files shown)


src/algorithms.jl
```diff
@@ -5216,6 +5216,11 @@ struct SofSpa10 <: OrdinaryDiffEqPartitionedAlgorithm end
 # Nyström methods
 
 """
+Improved Runge-Kutta-Nyström method of order three, which minimizes the amount of evaluated functions in each step. Fixed time steps only..
```
ranocha, line 5219:
```suggestion
"""
    IRKN3

Improved Runge-Kutta-Nyström method of order three, which minimizes the amount of evaluated functions in each step. Fixed time steps only.
```
Could you please add the name of the struct you describe in each docstring as in the suggestion above? Indenting it by 4 spaces lets it be rendered as code.

src/algorithms.jl
```diff
@@ -5225,20 +5230,39 @@ struct SofSpa10 <: OrdinaryDiffEqPartitionedAlgorithm end
 struct IRKN3 <: OrdinaryDiffEqPartitionedAlgorithm end
 
 """
+A 4th order explicit Runge-Kutta-Nyström method which can be applied directly on second order ODEs. Can only be used with fixed time steps.
+
+In case the ODE Problem is not dependent on the first derivative consider using Nystrom4VelocityIndependent to increase performance.
```
ranocha, line 5235:
```suggestion
In case the ODE Problem is not dependent on the first derivative consider using 
[`Nystrom4VelocityIndependent`](@ref) to increase performance.
```
This let's Documenter add a reference/link.

src/algorithms.jl
```diff
@@ -5225,20 +5230,39 @@ struct SofSpa10 <: OrdinaryDiffEqPartitionedAlgorithm end
...
 E. Hairer, S.P. Norsett, G. Wanner, (1993) Solving Ordinary Differential Equations I.
   Nonstiff Problems. 2nd Edition. Springer Series in Computational Mathematics,
   Springer-Verlag.
 """
 struct Nystrom4 <: OrdinaryDiffEqPartitionedAlgorithm end
 
 """
+A 4th order explicit Runkge-Kutta-Nyström method. Used directly on second order ODEs, where the acceleration is independent from velocity (ODE Problem is not dependent on the first derivative).
+
+More efficient then Nystrom4 on velocity independent Problems, since less evaluations are needed.
```
ranocha, line 5247:
```suggestion
More efficient then [`Nystrom4`](@ref) on velocity independent problems, since less evaluations are needed.
```

src/algorithms.jl
```diff
@@ -5294,6 +5331,11 @@ struct DPRKN5 <: OrdinaryDiffEqAdaptivePartitionedAlgorithm end
 struct DPRKN6 <: OrdinaryDiffEqAdaptivePartitionedAlgorithm end
 
 """
+6th order explicit Runge-Kutta-Nyström method. The second order ODE should not depend on the first derivative.
+
+Compared to DPRKN6, this method has smaller truncation error coefficients which leads to performance gain when only the main solution points are considered.
```
ranocha, line 5336:
```suggestion
Compared to [`DPRKN6`](@ref), this method has smaller truncation error coefficients which leads to performance gain
when only the main solution points are considered.
```

src/algorithms.jl
```diff
@@ -5306,6 +5348,11 @@ struct DPRKN6 <: OrdinaryDiffEqAdaptivePartitionedAlgorithm end
 struct DPRKN6FM <: OrdinaryDiffEqAdaptivePartitionedAlgorithm end
 
 """
+8th order explicit Runge-Kutta-Nyström method. The second order ODE should not depend on the first derivative.
+
+Not as efficient as DPRKN12 when high accuracy is needed, however this solver is competitive with DPRKN6 at lax tolerances and, depending on the problem, might be a good option between performance and accuracy.
```
ranocha, line 5353:
```suggestion
Not as efficient as [`DPRKN12`](@ref) when high accuracy is needed, however this solver is competitive with
[`DPRKN6`](@ref) at lax tolerances and, depending on the problem, might be a good option between performance and accuracy.
```

ranocha (changes requested):
Thanks! Please find some suggestions/comments below.

src/algorithms.jl
```diff
@@ -5286,6 +5294,9 @@ E. Hairer, S.P. Norsett, G. Wanner, (1993) Solving Ordinary Differential Equatio
 struct Nystrom5VelocityIndependent <: OrdinaryDiffEqPartitionedAlgorithm end
 
 """
+    DPRKN4
+
+
```
ranocha, line 5299:
```suggestion
```

ranocha (approved):
Thanks!

ranocha:
@ChrisRackauckas This should be ready to merge as far as I can tell

ChrisRackauckas:
Beautiful. I really want to get these docs completed so we can then simplify the DiffEq ones a bit.

PR #1944, opened 2023-05-12
use jac for the ShampineCollocationInit nlsolve

oscardssmith:
This should improve speed and numerical reliability of the nonlinear solve.

### DIFF
--- a/src/initialize_dae.jl
+++ b/src/initialize_dae.jl
@@ -286,7 +286,7 @@ function _initialize_dae!(integrator, prob::DAEProblem,
     f(resid, integrator.du, u0, p, t)
     integrator.opts.internalnorm(resid, t) <= integrator.opts.abstol && return
 
-    isAD = alg_autodiff(integrator.alg)
+    isAD = alg_autodiff(integrator.alg) isa AutoForwardDiff
     if isAD
         chunk = ForwardDiff.pickchunksize(length(tmp))
         _tmp = PreallocationTools.dualcache(tmp, chunk)
--- a/src/initialize_dae.jl
+++ b/src/initialize_dae.jl
@@ -286,8 +286,8 @@ function _initialize_dae!(integrator, prob::DAEProblem,
     f(resid, integrator.du, u0, p, t)
     integrator.opts.internalnorm(resid, t) <= integrator.opts.abstol && return
 
-    isad = alg_autodiff(integrator.alg)
-    if isad
+    isAD = alg_autodiff(integrator.alg)
+    if isAD
         chunk = ForwardDiff.pickchunksize(length(tmp))
         _tmp = PreallocationTools.dualcache(tmp, chunk)
     else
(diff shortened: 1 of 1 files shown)


src/initialize_dae.jl
```diff
@@ -87,12 +87,6 @@ function _initialize_dae!(integrator, prob::DAEProblem,
     end
 end
 
-## NoInit
-
-function _initialize_dae!(integrator, prob::Union{ODEProblem, DAEProblem},
-                          alg::NoInit, x::Union{Val{true}, Val{false}})
```
ChrisRackauckas, line 93: This requires a bump of the minimum.

src/initialize_dae.jl
```diff
@@ -152,20 +146,32 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
 
         nlequation! = @closure (out, u, p) -> begin
             update_coefficients!(M, u, p, t)
-            #M * (u-u0)/dt - f(u,p,t)
+            # f(u,p,t) + M * (u0 - u)/dt
```
ChrisRackauckas, line 149: why the sign flip?

ChrisRackauckas:
Add some tests?

src/initialize_dae.jl
oscardssmith: the sign flip made the Jacobian a little cleaner IMO. I could revert if you prefer.

src/initialize_dae.jl
```diff
@@ -152,20 +146,32 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
...
         end
 
-        nlsolve = default_nlsolve(alg.nlsolve, isinplace, u0, isad)
+        jac = if isnothing(f.jac)
+            f.jac
+        else
+            @closure (J, u, p) -> begin
+                # f(u,p,t) + M * (u0 - u)/dt
+                # df(u,p,t)/du - M/dt
+                f.jac(J, u, p, t)
```
oxinabox, line 164: for a DAE, `f.jac` takes a gamma argument.
I suspect this should be:
```suggestion
                f.jac(J, u, p, 0.0, t)
```

ChrisRackauckas:
format and add tests

ChrisRackauckas (approved):

ChrisRackauckas:
https://buildkite.com/julialang/ordinarydiffeq-dot-jl/builds/2597#0188b521-4359-4210-9420-0c870f644166/835-1252 the test failures look real

oscardssmith:
@ChrisRackauckas good call. Silly typo.

ChrisRackauckas:
Test tolerance may need an update there. Is that within chosen tolerance?

oscardssmith:
The thing that makes no sense to me is that the failing test doesn't have a jacobian, so I'm very confused as to why this is causing a difference. The initialization is failing, but unless passing `jac=nothing` into the nonlinear function is changing something from not passing a `jac`, it really should be doing exactly the same thing as before.

oscardssmith:
Turns out the tests are actually pretty smart. I had a sign error :)

PR #1952, opened 2023-06-01
New formatting

YingboMa:


src/nordsieck_utils.jl
```diff
@@ -302,41 +302,41 @@ function nordsieck_adjust_order!(cache::T, dorder) where {T}
...
-
-        for j in 2:(order - 1)
-            cache.l[j + 1] = order * cache.l[j] / j
-        end
-        for j in 3:order
+        # Adams order increase
+        if dorder == 1
             if isconstcache
-                cache.z[j] = muladd.(-cache.l[j], cache.z[order + 1], cache.z[j])
+                cache.z[order + 2] = zero(cache.z[order + 2])
```
ChrisRackauckas, line 308: reading this diff is a bit odd.

YingboMa: This refers to the
```julia
            cache.z[order + 2] = zero(cache.z[order + 2])
```
above.

PR #1969, opened 2023-06-17
Set u_modified in DAE initialization if the u was changed

Keno:
Some integrators (such as Sundials) keep a shadow copy of `u` and need to reflect this copy back into the solver internals to use them. This happens e.g. during callbacks, but the same situation applies to initialization, so use the same mechanism (setting u_modified to `true`) to let the solver know that it needs to update its internal datastructures.

### DIFF
--- a/src/initialize_dae.jl
+++ b/src/initialize_dae.jl
@@ -183,6 +183,7 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
         integrator.u .= nlsol.u
         failed = nlsol.retcode != ReturnCode.Success
     end
+    u_modified!(integrator, true)
     recursivecopy!(integrator.uprev, integrator.u)
     if alg_extrapolates(integrator.alg)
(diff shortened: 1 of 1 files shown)


src/initialize_dae.jl
```diff
@@ -183,6 +183,7 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
         integrator.u .= nlsol.u
         failed = nlsol.retcode != ReturnCode.Success
     end
+    u_modified!(integrator, true)
```
ChrisRackauckas, line 186: This is the default assumption. What's a case where this makes a difference?

ChrisRackauckas:
Needs a test case

oscardssmith:
I don't believe this is necessary. Specifically, setting u_modified in inits is only needed for Sundials solvers, and sundials solvers don't use ODE/DAEIntegrator so these initialization methods will never be used with them,

PR #1970, opened 2023-06-17
Re-evaluate u_initial after DAE initialization

Keno:
Found by inspection - I think this acidentally had the same bug as https://github.com/SciML/Sundials.jl/pull/407 if save_idxs is not `nothing`.


ChrisRackauckas:
Needs a test

ChrisRackauckas:
@oscardssmith do you know a test for this?

oscardssmith:
yeah. The test here is the same (morally) as the one I added for IDA where you use an initialization that changes `u0` dramatically.

ChrisRackauckas:
can you add it?

PR #1975, opened 2023-06-19
Added Algorithm ```FineRKN4```

HenryLangner:
Added the 4th order Nyström-Method presented by J. M. Fine in the article "Low Order Practical Runge-Kutta-Nyström Methods" (Also linked here: https://github.com/SciML/OrdinaryDiffEq.jl/issues/677). Works with ```adaptive step-size```. The Method passes the implemented convergence tests (harm. Oscillator):

```
using OrdinaryDiffEq, Test, RecursiveArrayTools, DiffEqDevTools, Statistics

u0 = fill(0.0, 2);
v0 = ones(2);
function f1_harmonic(dv, v, u, p, t)
    dv .= -u
end;
function f2_harmonic(du, v, u, p, t)
    du .= v
end;
function harmonic_analytic(y0, p, x)
    v0, u0 = y0.x
    ArrayPartition(-u0 * sin(x) + v0 * cos(x), u0 * cos(x) + v0 * sin(x))
end;
ff_harmonic = DynamicalODEFunction(f1_harmonic, f2_harmonic; analytic = harmonic_analytic);
prob = DynamicalODEProblem(ff_harmonic, v0, u0, (0.0, 5.0));

dts = 1.0 ./ 2.0 .^ (5:-1:0);

println("##### In Place #####")
sim = test_convergence(dts, prob, FineRKN4(), dense_errors = true);
println(@test sim.𝒪est[:l2]≈5 rtol=1e-1)
println(@test sim.𝒪est[:L2]≈4 rtol=1e-1)
println("with:")
println(sim.𝒪est)
println("##### Out of Place ##### ")

u0 = 0.0;
v0 = 1.0;
function f1_harmonic_nip(v, u, p, t)
    -u
end;
function f2_harmonic_nip(v, u, p, t)
    v
end;

ff_harmonic_nip = DynamicalODEFunction(f1_harmonic_nip, f2_harmonic_nip;
                                       analytic = harmonic_analytic);
prob = DynamicalODEProblem(ff_harmonic_nip, v0, u0, (0.0, 5.0));

csim = test_convergence(dts, prob, FineRKN4(), dense_errors = true);
println(@test sim.𝒪est[:l2]≈5 rtol=1e-1)
println(@test sim.𝒪est[:L2]≈4 rtol=1e-1)
println("with:")
println(sim.𝒪est)

```
Yields:
grafik

There is one more PR on the way which adds ```adaptive step-size``` to ```FineRKN5```.
[...]

PR #1976, opened 2023-06-19
Adaptive step-size for ```FineRKN5```

HenryLangner:
The changes enable the use of adaptive step sizes for ```FineRKN5()```. 
The already implemented convergence tests are all passed. I will add convergence tests for the damped oscillator introduced in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1973, once this PR and https://github.com/SciML/OrdinaryDiffEq.jl/pull/1975 are merged. 

CC @ranocha

PR #1973, opened 2023-06-19
fix out-of-place version of FineRKN5

ranocha:
I found some typos in out-of-place version of `FineRKN5()` and fixed them. I also added some tests checking that the in-place and out-of-place versions of some Runge-Kutta-Nyström methods. Some of these are currently broken, in particular with adaptive time stepping. I propose to open an issue and check them later.
I also fixed the reported number of function evaluations of some methods and added some checks for them.

CC @HenryLangner

PR #1977, opened 2023-06-22
New algorithm ```FineRKN4()```

HenryLangner:
The last PR https://github.com/SciML/OrdinaryDiffEq.jl/pull/1975 got a little messed up with all the changes across the three PRs (https://github.com/SciML/OrdinaryDiffEq.jl/pull/1975, https://github.com/SciML/OrdinaryDiffEq.jl/pull/1973 and https://github.com/SciML/OrdinaryDiffEq.jl/pull/1976). 

Just to be safe I added all the features introduced in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1975 to this up to date branch. With this PR I will close https://github.com/SciML/OrdinaryDiffEq.jl/pull/1975.

The new algorithm ```FineRKN4()``` is added. In addition I have added the necessary tests (convergence, regression and the comparison tests between fixed step size and adaptive step size). 

Furthermore there is now a convergence test for the damped oscillator. Since ```FineRKN4()``` and ```FineRKN5()``` are generalized RK-Nyström methods, they can be directly applied on velocity depended second order ODEs. To test their convergence on these type of ODEs the damped oscillator is added.  

CC @ranocha

### DIFF
--- a/src/caches/rkn_caches.jl
+++ b/src/caches/rkn_caches.jl
@@ -36,6 +36,48 @@ end
 
 # alg_cache(alg::Nystrom4,u,rate_prototype,::Type{uEltypeNoUnits},::Type{uBottomEltypeNoUnits},::Type{tTypeNoUnits},uprev,uprev2,f,t,dt,reltol,p,calck,::Val{false}) where {uEltypeNoUnits,uBottomEltypeNoUnits,tTypeNoUnits} = Nystrom4ConstantCache(constvalue(uBottomEltypeNoUnits),constvalue(tTypeNoUnits))
 
+@cache struct FineRKN4Cache{uType, rateType, reducedRateType, uNoUnitsType, TabType} <:
+              OrdinaryDiffEqMutableCache
+    u::uType
+    uprev::uType
+    fsalfirst::rateType
+    k2::reducedRateType
+    k3::reducedRateType
+    k4::reducedRateType
+    k5::reducedRateType
+    k::rateType
+    utilde::uType
+    tmp::uType
+    atmp::uNoUnitsType
+    tab::TabType
+end
+
+function alg_cache(alg::FineRKN4, u, rate_prototype, ::Type{uEltypeNoUnits},
+    ::Type{uBottomEltypeNoUnits}, ::Type{tTypeNoUnits}, uprev, uprev2, f, t,
+    dt, reltol, p, calck,
+    ::Val{true}) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
+    reduced_rate_prototype = rate_prototype.x[2]
+    tab = FineRKN4ConstantCache(constvalue(uBottomEltypeNoUnits), constvalue(tTypeNoUnits))
+    k1 = zero(rate_prototype)
+    k2 = zero(reduced_rate_prototype)
+    k3 = zero(reduced_rate_prototype)
+    k4 = zero(reduced_rate_prototype)
+    k5 = zero(reduced_rate_prototype)
+    k = zero(rate_prototype)
+    utilde = zero(u)
+    atmp = similar(u, uEltypeNoUnits)
+    recursivefill!(atmp, false)
+    tmp = zero(u)
+    FineRKN4Cache(u, uprev, k1, k2, k3, k4, k5, k, utilde, tmp, atmp, tab)
+end
+
+function alg_cache(alg::FineRKN4, u, rate_prototype, ::Type{uEltypeNoUnits},
+    ::Type{uBottomEltypeNoUnits}, ::Type{tTypeNoUnits}, uprev, uprev2, f, t,
+    dt, reltol, p, calck,
+    ::Val{false}) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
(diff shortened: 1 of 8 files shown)


src/caches/rkn_caches.jl
```diff
@@ -36,6 +36,48 @@ end
...
+    tmp::uType
+    atmp::uNoUnitsType
+    tab::TabType
+end
+
+function alg_cache(alg::FineRKN4, u, rate_prototype, ::Type{uEltypeNoUnits},
+    ::Type{uBottomEltypeNoUnits}, ::Type{tTypeNoUnits}, uprev, uprev2, f, t,
+    dt, reltol, p, calck,
+    ::Val{true}) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
+    reduced_rate_prototype = rate_prototype.x[2]
```
ranocha, line 59: @ChrisRackauckas I really do not like this new style of formatting since it makes it quite hard to see where the function body begins... But it's imposed by JuliaFormatter and used everywhere else in OrdinaryDiffEq.jl, so nothing to change in this PR

test/algconvergence/partitioned_methods_tests.jl
```diff
@@ -368,6 +378,76 @@ sol = solve(prob, ERKN5(), reltol = 1e-8)
...
+                 sqrt(15) * (du0_u0[1] + 4 * du0_u0[2]) * sin(sqrt(15) * t / 4)),
+            ], # du
+            [
+                exp(-t / 4) / 15 * (15 * du0_u0[2] * cos(sqrt(15) * t / 4) +
+                 sqrt(15) * (4 * du0_u0[1] + du0_u0[2]) * sin(sqrt(15) * t / 4)),
+            ])),
+    OrdinaryDiffEq.SciMLBase.ArrayPartition([0.0], [1.0]), # du0, u0
+    (0.0, 10.0), # tspan
+    DiffEqBase.NullParameters(), # p
+    SecondOrderODEProblem{false}())
```
ranocha, line 398: @ChrisRackauckas: Is there a less ugly looking way to create this - ideally using the `SecondOrderODEProblem` constructor? This is just a working version we came up with.

test/algconvergence/partitioned_methods_tests.jl
```diff
@@ -368,6 +378,76 @@ sol = solve(prob, ERKN5(), reltol = 1e-8)
...
+dts = 1.0 ./ 2.0 .^ (5:-1:0)
+sim = test_convergence(dts, prob, Nystrom4(), dense_errors = true)
+@test sim.𝒪est[:l2]≈4 rtol=1e-1
+@test sim.𝒪est[:L2]≈4 rtol=1e-1
+sim = test_convergence(dts, prob, FineRKN4(), dense_errors = true)
+@test sim.𝒪est[:l2]≈4 rtol=1e-1
+@test sim.𝒪est[:L2]≈4 rtol=1e-1
+sim = test_convergence(dts, prob, FineRKN5(), dense_errors = true)
+@test sim.𝒪est[:l2]≈5 rtol=1e-1
+@test sim.𝒪est[:L2]≈4 rtol=1e-1
```
ranocha, line 409: Interesting! So the methods seem to have some superconvergence properties for the harmonic oscillator without damping but not for the version with damping used here.
(nothing blocking this PR from being merged, just an observation)

ranocha (approved):
Thanks!

src/caches/rkn_caches.jl
ChrisRackauckas: Man everyone is mad at it both ways. We should do a thread in the JuliaFormatter and argue about all of these glorious details

test/algconvergence/partitioned_methods_tests.jl
ChrisRackauckas: Why would this not use `SecondOrderODEProblem`? I'm a bit confused as to how this would be different.

src/caches/rkn_caches.jl
ranocha: https://github.com/domluna/JuliaFormatter.jl/issues/741

test/algconvergence/partitioned_methods_tests.jl
ranocha: I couldn't find a constructor of a `SecondOrderODEProblem` that accepts an analytic solution. The only way I found to make it work was to write the conversion to an `ODEProblem` with `DynamicalODEFunction` and `ArrayPartition` by hand. That's okay and works well for the tests - I was just wondering whether there is a less manual way to achieve this.

ranocha: I think this shouldn't block this PR from being merged. I see some reasons why it would be difficult to get another nice interface with the current structure right now. Thus, we may file an issue to remember this but should go ahead with this PR.

ranocha:
@ChrisRackauckas This is ready to be merged from my point of view. I don't see how the SciMLSensitivity failures could be related to this.

ChrisRackauckas:
Oh yes, I just missed this one sorry.

ranocha:
No problem, this can easily happen given the amount of packages you maintain. Thanks

PR #1984, opened 2023-07-06
Don't set autodiff chunksize when autodiff=false

oscardssmith:
when autodiff is false we don't want to set a chunksize since we aren't going to use it anyway. Ideally this would also turn off when we have a jacobian (and tgrad if necessary) but I don't yet know the correct conditions there

### DIFF
--- a/src/alg_utils.jl
+++ b/src/alg_utils.jl
@@ -259,11 +259,11 @@ function DiffEqBase.prepare_alg(alg::Union{
         linsolve = alg.linsolve
     end
 
-    # If norecompile mode or very large bitsize, like a dual number u0 already, then
+    # If not using autodiff or norecompile mode or very large bitsize (like a dual number u0 already)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
I think the test failures look real on SciMLSensitivity.

ChrisRackauckas:
@vpuri3 is this interface 3 test failure known? It looks like a SciMLOperators thing

vpuri3:
looks like an issue with the ADTypes wrappers. lemme take a look

vpuri3:
weird. that shouldn't be happening because default for function operator is batch = false. lemme take a look

vpuri3:
Looks like Krylov.jl is `vec`'ing the array somewhere in there. default algorithm (GenericLU) is working just fine for this problem.

vpuri3:
The culprit is here

https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/derivative_utils.jl#L388-L393

PR #1996, opened 2023-07-22
Support user-specified `W_prototype <: AbstractSciMLOperator` 

gaurav-arya:
- Updates the interface of the existing `WOperator` to match that of a SciMLOperator.
- Detects and uses a user-specified `W_prototype <: AbstractSciMLOperator`.

Dependent on https://github.com/SciML/SciMLBase.jl/pull/473 and https://github.com/SciML/StochasticDiffEq.jl/pull/540. The intent is for this feature to be undocumented until it is more battle tested. Needs tests 

(I have a much more involved set of changes that replaces the existing `WOperator` with a lazy SciMLOperator. But I've come to the conclusion that this is not something to be done lightly, e.g. there are `muladd` optimizations that are not easy to replace with the lazy purely general approach. Instead, I'm just adding `AbstractSciMLOperator` branches to first allow the W-operator to be manually swapped out for an `AbstractSciMLOperator`, while keeping the original `WOperator` around.)


gaurav-arya:
@ChrisRackauckas, @utkarsh530, @vpuri3 this is finally ready:)

vpuri3:
why do we need a type? why is it not just

```julia
integrator.gamma = ScalarOp(0.0; update_func)
 integrator.W = integrator.M - integrator.gamma * integrator.J
```

gaurav-arya:
You mean why not remove the `WOperator`? That's because it has optimizations, e.g. related to Newton methods and the `_concrete_form`, that make it a bit dangerous from a performance standpoint to replace by default. This PR just aims to give the user a choice to specify their own WOperator

utkarsh530:
LGTM. Yes, I believe `WOperator` should remain as is currently. The purpose mainly for `W_prototype` is to add support for approx. matrix factorization techniques. 
If this is ready, maybe we should add an example to test and update docs on using AMF in ODE solvers.

vpuri3:
@gaurav-arya can we resolve this here?

https://github.com/SciML/SciMLOperators.jl/issues/191#issuecomment-1616687458

test/interface/wprototype_tests.jl
```diff
@@ -0,0 +1,47 @@
...
+
+    # Test that the custom W operator is used
+    integrator = init(prob_W, Rosenbrock23())
+    @test integrator.cache.W isa typeof(W_op)
+
+    # Run solves. Ensure all solutions use the same linear solve for fair comparison.
+    # TODO: in future, ensure and test that polyalg chooses the best linear solve for all. 
+    sol = solve(prob, Rosenbrock23(linsolve = KrylovJL_GMRES()))
+    sol_J = solve(prob_J, Rosenbrock23(linsolve = KrylovJL_GMRES())) # note: direct linsolve in this case is broken, see #1998
+    sol_W = solve(prob_W, Rosenbrock23(linsolve = KrylovJL_GMRES()))
```
ChrisRackauckas, line 41: Test a newton solver as well, like FBDF. Also test Radau5

gaurav-arya: Added an FBDF test. RadauIIA5 does not seem to support sparse Jacobians though: https://github.com/SciML/OrdinaryDiffEq.jl/blob/b807375e4f728e61dff5ff3934c370affc0d3d6b/src/caches/firk_caches.jl#L238

ChrisRackauckas: Yeah Radau5 will need more work.

PR #2017, opened 2023-08-16
start reworking docstrings of SSP Optimized RK Methods

ArnoStrouwen:


ArnoStrouwen:
Is codecov broken? It fails after 1s, so it never got any files to analyze?

ChrisRackauckas (approved):

PR #2018, opened 2023-08-16
Change exp to exponential! in linear methods

### LINKED ISSUE #2009: Use `ExponentialUtilities.exponential!` instead of `exp` and add `method` as an option in solvers that currently use `exp` and `expv`
As per title. My main motive for this is that `exp` doesn't support AD (https://github.com/JuliaDiff/ForwardDiff.jl/issues/174). This shouldn't affect existing code if we set the default method to be [ExpMethodNative](https://docs.sciml.ai/ExponentialUtilities/stable/matrix_exponentials/#ExponentialUtilities.ExpMethodNative).

ErikQQY:
Fix #2009 

Before:

old_generic

After:

new_generic

It is weird that I didn't see any acceleration here. Maybe I did something wrong here. But I think now it is supporting AD now?


oscardssmith:
Shouldn't pretty much all of these be using `expv` Matrix exponential times vector will almost always be a lot slower.

src/perform_step/linear_perform_step.jl
```diff
@@ -23,7 +24,8 @@ function perform_step!(integrator, cache::MagnusMidpointCache, repeat_step = fal
             opnorm = integrator.opts.internalopnorm, iop = alg.iop)
     else
         A = Matrix(L) #size(L) == () ? convert(Number, L) : convert(AbstractMatrix, L)
-        u .= exp(dt * L) * u
+        cacheA = ExponentialUtilities.alloc_mem(A, exp_method)
```
ChrisRackauckas, line 27: This cache should be constructed during the cache construction time.

ChrisRackauckas:
> Shouldn't pretty much all of these be using expv Matrix exponential times vector will almost always be a lot slower.

No, the krylov methods are only faster when the system is sufficiently big. The exponentiated matrix can be faster than a krylov technique for cases not large enough. Additionally, some of the methods can cache the matrix exponentiation and reuse it, which can save performance.

Share a profile, that will tell us where the hotspots are.

oscardssmith:
shouldn't it be the job of `expv` to make the size dependent determination of which way to do the matrix exponential * vector?

ChrisRackauckas:
the two methods cache differently. In theory it could adapt itself but in practice it does not.

src/perform_step/linear_perform_step.jl
ErikQQY: Sorry, I am a little confused, while the cache construction is located in ```linear_cache.jl```, you mean we should construct the exponential matrix cache during the algorithm cache construction process right? My understanding is that we can allocate the matrix once in algorithms cache construction like ```exp_cache = ExponentialUtilities.alloc_mem(f, exp_method)``` and pass the allocated matrix to perform_step and use copies of that allocated matrix to complete ```exponential!``` instead of allocating many times in perform_step.

ChrisRackauckas: > My understanding is that we can allocate the matrix once in algorithms cache construction like exp_cache = ExponentialUtilities.alloc_mem(f, exp_method) and pass the allocated matrix to perform_step and use copies of that allocated matrix to complete exponential! instead of allocating many times in perform_step.

Yes exactly. The `alloc_mem` call is intended to be called just once, and its cache can be reused in every `exponential!` call from that point on. That would greatly reduce the allocations of these calls.

ErikQQY:
Old:
old

New:
new

The linear solvers are so fast now!!!

ErikQQY:
bump

ChrisRackauckas (approved):

PR #2019, opened 2023-08-16
Simplify precompiled solvers

ChrisRackauckas:
Goes with the DiffEq default changes


oscardssmith:
How much faster is the precompile?

PR #2022, opened 2023-08-25
Faster linear_perform_step by reducing deepcopies and using sparse matrix when possible.

howsiyu:


src/perform_step/linear_perform_step.jl
```diff
@@ -486,20 +472,17 @@ function perform_step!(integrator, cache::MagnusGL8Cache, repeat_step = false)
...
-    update_coefficients!(L2, uprev, p, t - dt * v2 + dt / 2)
-    A2 = Matrix(L2)
-    update_coefficients!(L3, uprev, p, t + dt * v2 + dt / 2)
-    A3 = Matrix(L3)
+    update_coefficients!(L, uprev, p, t - dt * v1 + dt / 2)
+    A1 = deepcopy(L.A)
+    update_coefficients!(L, uprev, p, t + dt * v1 + dt / 2)
+    A4 = deepcopy(L.A)
+    update_coefficients!(L, uprev, p, t - dt * v2 + dt / 2)
+    A2 = deepcopy(L.A)
```
ChrisRackauckas, line 483: This isn't correct. L.A doesn't necessarily exist. This is only specialized to a single SciMLOperator type, and instead needs to work for the general interface. Also, `Matrix(L)` on a `MatrixOperator` is a no-op to grab `L.A`, so is there a reason that wouldn't be sufficient?

ChrisRackauckas:
The correctness issue is pointed out above. The general thing here is that the copying should only exist in the cache construction, not in the perform step. That would be the proper way to handle this. Also, for these methods there should be some checks to see if the operator is allowed to concretize, otherwise it needs to be handled through the operations or error.

You might want to see the operator interface: https://docs.sciml.ai/SciMLOperators/stable/interface/

src/perform_step/linear_perform_step.jl
ChrisRackauckas: Yes, `convert(AbstractMatrix, L)` would work.

ChrisRackauckas (approved):

ChrisRackauckas:
Hmm https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/6149342613/job/16688343305?pr=2022#step:6:516 looks real. It looks like Julia v1.6 is missing certain dispatches to make this work. I'd be willing to just disable that test for pre-v1.9 and move on though, since LTS is going to bump in the near future.

ChrisRackauckas:
Thanks!

PR #2032, opened 2023-09-29
log progress with configurable ID

pepijndevos:
Needed for https://github.com/SciML/SciMLBase.jl/pull/514 to allow logging progress with a unique ID

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -63,6 +63,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
     progress_steps = 1000,
     progress_name = "ODE",
     progress_message = ODE_DEFAULT_PROG_MESSAGE,
+    progress_id = gensym("OrdinaryDiffEq"),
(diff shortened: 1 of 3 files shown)


src/solve.jl
```diff
@@ -63,6 +63,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
     progress_steps = 1000,
     progress_name = "ODE",
     progress_message = ODE_DEFAULT_PROG_MESSAGE,
+    progress_id = gensym("OrdinaryDiffEq"),
```
ChrisRackauckas, line 66: This needs to be added as a common interface arg in DiffEqBase and the docs

pepijndevos: I've added it in https://github.com/SciML/DiffEqBase.jl/pull/939 but not to the docs

pepijndevos: Ok docs updated as well

ChrisRackauckas (approved):

PR #2040, opened 2023-10-19
Switch to robust multi-newton to improve initialization robustness

ChrisRackauckas:


oscardssmith:
how do the benchmarks compare?

avik-pal:
NonlinearSolve `2` drops `1.6`

ChrisRackauckas:
> how do the benchmarks compare?

https://docs.sciml.ai/SciMLBenchmarksOutput/dev/NonlinearProblem/nonlinear_solver_23_tests/

It's a lot more robust. As a default we went for pure robustness here, since it's only one nonlinear solve of many so hyper-optimizing its performance is probably not worth it. Optimizing its robustness is though, since the initialization is by far the hardest of the nonlinear solves that occurs in the process. The polyalgorithm used here is a mixture of all of the "winners" of these 23 model benchmarks. 4 methods which solve all 23 equations are now tried in succession, as opposed to one method which only solved 21/23 equations.

ChrisRackauckas:
>  NonlinearSolve 2 drops 1.6

I created a v1.6 backport branch that we can backport patches to. Let's keep it moving.

src/initialize_dae.jl
```diff
@@ -23,7 +23,7 @@ BrownFullBasicInit(abstol) = BrownFullBasicInit(; abstol = abstol, nlsolve = not
 using SciMLNLSolve
 default_nlsolve(alg, isinplace, u, autodiff = false) = alg
 function default_nlsolve(::Nothing, isinplace, u, autodiff = false)
-    NLSolveJL(autodiff = autodiff ? :forward : :central)
+    RobustMultiNewton(autodiff = autodiff)
```
ChrisRackauckas, line 26:
```suggestion
    TrustRegion(autodiff = autodiff)
```

PR #2044, opened 2023-10-30
chore: bump RecursiveArrayTools compat

AayushSabharwal:
All tests that don't end up pulling in MTK pass


Project.toml
```diff
@@ -69,7 +69,7 @@ Polyester = "0.3, 0.4, 0.5, 0.6, 0.7"
 PreallocationTools = "0.2, 0.3, 0.4"
 PrecompileTools = "1"
 Preferences = "1.3"
-RecursiveArrayTools = "2.36"
+RecursiveArrayTools = "2, 3"
```
oscardssmith, line 72: this needs to be `"2.36, 3"` since versions prior are missing features.

PR #2046, opened 2023-11-01
Remove inv of singular matrix in tests

ChrisRackauckas:
Some BLAS implementations do find it to be singular 😅


ChrisRackauckas:
Unblocks https://github.com/SciML/LinearSolve.jl/pull/408

PR #2048, opened 2023-11-02
Linear interpolation for algebraic variables

pepijndevos:
Except for those solvers with specialized stiffness aware* interpolation, hermite interpolation is used which uses the derivative. For algebraic variables this ends up incorrectly using the residual. This PR aims to make it fall back to linear interpolation for algebraic variables.

*what does that actually mean?

At the moment I just pushed whatever I was doing with @oscardssmith last time


pepijndevos:
I made a very first attempt to get this to work, and am now at a stage again where I could use some help from someone more familiar with the code. There are several questions

* Which paths into the code need to be handled to cover all uses
* Which types of states and indexes need to be handled
* Something about QR factorization and cacheing the computation

I'm currently testing with the rober example from https://docs.sciml.ai/DiffEqDocs/stable/tutorials/dae_example/#Mass-Matrix-Differential-Algebraic-Equations-(DAEs)

To elaborate on the first point, there are two methods of ode_interpolation that currently don't pass differential_vars anywhere, and I'm not sure when they are used. Are there other code paths that lead here that we need to handle?

The second point is a bit annoying. It seems like `u` can be an array or a scalar, and that `idxs` can be an array, nothing, or a single value, but it's not clear to me what assumptions I can make. Like, are they always numbers? Are there combinations that can't happen? It seems almost impossible to write a generic method that handles all cases, so I've started to implement special cases, but am uncertain of the total set of special cases  I need to handle.

pepijndevos:
I did the implementation as best as I can, from here on it's a tone of clean up and review and handling edge cases I suppose.

I'm currently throwing an error in the base case to make sure we handle all the cases that we should, and I see that is getting hit by a `NoIndexArray` whatever that may be https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/6853617794/job/18634794102?pr=2048#step:6:690

src/dense/interpolants.jl
```diff
@@ -1,12 +1,12 @@
 @muladd function _ode_interpolant(Θ, dt, y₀, y₁, k,
     cache::Union{FunctionMapConstantCache, FunctionMapCache},
-    idxs, T::Type{Val{0}})
+    idxs, T::Type{Val{0}}, differential_vars=trues(size(y₀)))
```
ChrisRackauckas, line 3: No need to allocate the vector if it's not used, just default to `nothing`

pepijndevos:
Summary of the current failures and my ongoing understanding of them

- [ ] This test appears to just have slightly worse error now, could this be expected? https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/6862847015/job/18661406104?pr=2048#step:6:1007
- [ ] idk what's going on here, my code doesn't seem to be in the stacktrace but maybe my matrix handling is slightly wrong? https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/6862847015/job/18661405628?pr=2048#step:6:595
- [ ] this is probably an easy fix in my dispatch where a scalar ODE with `idxs::Nothing` gets in the wrong place https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/6862847015/job/18661404322?pr=2048#step:6:597
- [ ] seems like I introduced allocation somewhere, but in the in-place one so will require some digging https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/6862847015/job/18661403836?pr=2048#step:6:616

src/dense/generic_dense.jl
```diff
@@ -324,6 +340,7 @@ function ode_interpolation(tvals, id::I, idxs, deriv::D, p,
     @unpack ts, timeseries, ks, f, cache = id
     @inbounds tdir = sign(ts[end] - ts[1])
     idx = sortperm(tvals, rev = tdir < 0)
+    differential_vars = get_differential_vars(f, size(timeseries[begin]))
```
pepijndevos, line 343: This should maybe be passed somewhere, but @oscardssmith had some argument why not all cache types need it that I forgot so idk if deep down in the evaluate_interpolant machinery there is somewhere this needs to go

src/dense/generic_dense.jl
```diff
@@ -580,23 +602,108 @@ function ode_interpolant(Θ, dt, y₀, y₁, k, cache::OrdinaryDiffEqMutableCach
...
+    if only(differential_vars)
+        hermite_interpolant!(out, Θ, dt, y₀, y₁, k, idxs, T)
+    else
+        linear_interpolant!(out, Θ, dt, y₀, y₁, idxs, T)
+    end
 end
 
-function _ode_interpolant!(out, Θ, dt, y₀, y₁, k, cache, idxs, T::Type{Val{TI}}) where {TI}
+function partial_hermite_interpolant!(out, Θ, dt, y₀, y₁, k, cache, idxs, T::Type{Val{TI}}, differential_vars) where {TI}
+    throw("how did we get here")
```
pepijndevos, line 704: Once I'm confident I've handled all the cases I guess either completely remove this method, or remove the error and leave it as a fallback if it turns out there is some insane edge case somewhere in the ecosystem.

src/dense/generic_dense.jl
oscardssmith: we do want to cache this eventually, but at first it's not necessary.

pepijndevos:
Memory profile:
image

It seems like pretty much everything is getting inlined except `idxs[.!sel]` which allocates a new BitVector. Not sure what I can do about that.

Would be funny if the broadcasted version is faster because it doesn't allocate. Math is fast, memory is slow.

oscardssmith:
You should probably have a fast path for no algebraic variables though. Lots of people only solve ODEs.

pepijndevos:
Ok so basically a complete rewrite to avoid allocations and use broadcasting and optimize for ODE.

I think if we default `differential_vars::Nothing` we can dispatch on that and do the optimal thing. Is it sufficient to have `nothing` in case of no mass matrix or uniform scaling, or is it common to have some other type of diagonal matrix with all ones that we need to test explicitly?

For broadcasting we need to change `differential_vars` to be output/index shaped.

And I fear that for completely avoiding allocations in the in-place version I need to write a fully integrated and broadcasted interpolation.

For the out of place version you can just do
```julia
    h = hermite_interpolant(Θ, dt, y₀, y₁, k, Val{cache isa OrdinaryDiffEqMutableCache}, idxs, T)
    l = linear_interpolant(Θ, dt, y₀, y₁, idxs, T)
    @.. broadcast=false h*differential_vars + l*!differential_vars
```
but for the in-place version you can't have both write to `out` without indexing or allocating:
```julia
    hermite_interpolant!(out, Θ, dt, y₀, y₁, k, idxs, T)
    l = linear_interpolant(Θ, dt, y₀, y₁, idxs, T)
    @.. broadcast=false out=out*differential_vars + l*!differential_vars
```

So I'm thinking to change 
```julia
    return (1 - Θ) * y₀[idxs] + Θ * y₁[idxs] +
           Θ * (Θ - 1) *
           ((1 - 2Θ) * (y₁[idxs] - y₀[idxs]) + (Θ - 1) * dt * k[1][idxs] +
            Θ * dt * k[2][idxs])
```
to
```julia
    return (1 - Θ) * y₀[idxs] + Θ * y₁[idxs] +
           isdiff * Θ * (Θ - 1) *
           ((1 - 2Θ) * (y₁[idxs] - y₀[idxs]) + (Θ - 1) * dt * k[1][idxs] +
            Θ * dt * k[2][idxs])
```
That adds like one multiplication. Maybe that's insignificant enough that we don't need a complete fast path?

Would it make sense to pass it as a type-level boolean so in the fast path you can compile away the multiplication, or would that cause more trouble than it's worth when it can't be statically inferred?

I'd really like to avoid having two complete copies of hermite interpolation for a single multiplication.

pepijndevos:
That wasn't so bad for the primal. For derivatives it's a bit more tricky but if it passes the tests I'll do the rest.

src/dense/generic_dense.jl
```diff
@@ -580,24 +606,31 @@ function ode_interpolant(Θ, dt, y₀, y₁, k, cache::OrdinaryDiffEqMutableCach
...
 end
 
-function _ode_interpolant!(out, Θ, dt, y₀, y₁, k, cache, idxs, T::Type{Val{TI}}) where {TI}
-    hermite_interpolant!(out, Θ, dt, y₀, y₁, k, idxs, T)
+function _ode_interpolant!(out, Θ, dt, y₀, y₁, k, cache, idxs, T::Type{Val{TI}}, differential_vars::DV=nothing) where {TI, DV}
+    if DV === Nothing
+        hermite_interpolant!(out, Θ, dt, y₀, y₁, k, idxs, T)
+    else
+        hermite_interpolant!(out, Θ, dt, y₀, y₁, k, idxs, T, differential_vars)
+    end
```
ChrisRackauckas, line 633: Why not just use dispatch?

src/dense/generic_dense.jl
```diff
@@ -313,6 +313,26 @@ function evaluate_interpolant(f, Θ, dt, timeseries, i₋, i₊, cache, idxs,
...
+            differential_vars = reshape(Diagonal(mm).diag .!= 0, size)
+        else
+            # QR factorization
+            # @show typeof(mm)
+        end
+    end
+    if idxs === nothing || differential_vars === nothing
+        return differential_vars
+    else
+        return differential_vars[idxs]
```
ChrisRackauckas, line 332: not a view?

src/dense/generic_dense.jl
pepijndevos: Ambiguous method error

src/dense/generic_dense.jl
pepijndevos: Good idea I suppose

pepijndevos:
I'm honestly a bit puzzled by some of these CI errors. Like the `foop` one.

oscardssmith:
The foop one isn't your fault.

pepijndevos:
Is this my fault? https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/6940604309/job/18879835886?pr=2048#step:6:637

It seems to be a regression test for  #2055 and I don't really see the connection or the bug.

ChrisRackauckas:
I just merged a PR that was all green except for the format check (which the formatter is still having issues, I'm going to put a bounty on that) https://github.com/SciML/OrdinaryDiffEq.jl/pull/2069, so if interpolation things are failing I'd venture to guess there was a merge issue.

src/dense/generic_dense.jl
```diff
@@ -313,6 +313,26 @@ function evaluate_interpolant(f, Θ, dt, timeseries, i₋, i₊, cache, idxs,
...
 end
 
+function get_differential_vars(f, idxs, size)
+    differential_vars = nothing
+    if hasproperty(f, :mass_matrix)
+        mm = f.mass_matrix
+        if mm isa UniformScaling
+            # already correct
+        elseif isdiag(mm)
+            differential_vars = reshape(Diagonal(mm).diag .!= 0, size)
```
oxinabox, line 323: Isn't this the same as:
```suggestion
            differential_vars = reshape(diag(mm)  .!= 0, size)
```

pepijndevos: Seems like it

pepijndevos:
I'm hoping `UndefVarError: linearizing_save_finalize not defined` isn't my fault?

Meanwhile I've started adding all the derivative implementations.

src/dense/generic_dense.jl
```diff
@@ -738,62 +778,57 @@ end
...
                                      12 * y₀[idxs] - 12 * y₁[idxs]) + 6 * y₁[idxs]) /
                                    (dt * dt)
 end
 
-@muladd function hermite_interpolant!(out::Array, Θ, dt, y₀, y₁, k, idxs, T::Type{Val{2}}) # Default interpolant is Hermite
-    @views @.. broadcast=false out=(-4 * dt * k[1][idxs] - 2 * dt * k[2][idxs] -
-                                    6 * y₀[idxs] +
-                                    Θ * (6 * dt * k[1][idxs] + 6 * dt * k[2][idxs] +
-                                     12 * y₀[idxs] - 12 * y₁[idxs]) + 6 * y₁[idxs]) /
-                                   (dt * dt)
```
pepijndevos, line 794: this one had two complete copies of the interpolation??? I just kept the last one which was effectively being used I guess.

topolarity:
Seems to throw an error for `sol(0.0, Val{1}, idxs=3)`

pepijndevos:
For what problem/solver? Anyway, I'll add a bunch of tests to gain some confidence that all of those dozens of methods I changed actually work.

pepijndevos:
I'm pretty sure I've seen matrix-shaped idxs and outputs, but when I just pass a matrix to sol, it errors. Is there some other code path that results in a matrix?

pepijndevos:
It seems like master fails the same checks as this branch.

If the tests I've just added pass on CI and didn't break anything else, this should be good for another round of reviews and hopefully merge.

topolarity:
This is definitely a step in the right direction, but it's worth mentioning that the derivatives will still need some work after this PR:
image

In general, the derivative of an interpolant doesn't have anything to do with the derivative of the interpolated function, unless that property was guaranteed by construction (e.g. the first-derivative of a hermite interpolation), so I think we need to use a different trick here.

src/dense/rosenbrock_interpolants.jl
```diff
@@ -282,7 +282,7 @@ end
 
 @muladd function _ode_interpolant!(out, Θ, dt, y₀, y₁, k,
     cache::Union{Rosenbrock5ConstantCache, Rosenbrock5Cache},
-    idxs, T::Type{Val{1}})
+    idxs, T::Type{Val{1}}, dv=nothing)
```
ChrisRackauckas, line 285: why would dv show up here?

test/interface/inplace_interpolation.jl
```diff
@@ -25,7 +25,7 @@ out_VMF = vecarrzero(ntt, size(prob_ode_2Dlinear.u0))   # Vector{Matrix{Float64}
         @test_throws MethodError sol_ODE(out_VF, tt; idxs = 1:1)
         @test sol_ODE(out_VF, tt) isa Vector{Float64}
         @test sol_ODE(out_VVF_1, tt) isa Vector{Vector{Float64}}
-        @test sol_ODE_interp.u == out_VF
+        @test sol_ODE_interp.u ≈ out_VF
```
ChrisRackauckas, line 28: why would this only end up approximate?

src/dense/generic_dense.jl
```diff
@@ -309,7 +309,27 @@ function evaluate_interpolant(f, Θ, dt, timeseries, i₋, i₊, cache, idxs,
...
+            # QR factorization
+            # @show typeof(mm)
+        end
+    end
+    if idxs === nothing || differential_vars === nothing
+        return differential_vars
+    else
+        return @view differential_vars[idxs]
     end
 end
```
ChrisRackauckas, line 334:
```suggestion
function get_differential_vars(f, idxs, size)
    differential_vars = nothing
    if hasproperty(f, :mass_matrix)
        mm = f.mass_matrix
        if mm isa UniformScaling
            return nothing
        elseif isdiag(mm)
            differential_vars = reshape(diag(mm)  .!= 0, size)
        else
            return missing # interpret missing downstream as not implemented
        end
    end
    if idxs === nothing
        return differential_vars
    else
        return @view differential_vars[idxs]
    end
end
```

src/dense/generic_dense.jl
```diff
@@ -324,6 +343,7 @@ function ode_interpolation(tvals, id::I, idxs, deriv::D, p,
     @unpack ts, timeseries, ks, f, cache = id
     @inbounds tdir = sign(ts[end] - ts[1])
     idx = sortperm(tvals, rev = tdir < 0)
+    differential_vars = get_differential_vars(f, idxs, size(timeseries[begin]))
```
ChrisRackauckas, line 346: You can't just assume the sizes are all the same...

ChrisRackauckas:
> This is definitely a step in the right direction, but it's worth mentioning that the derivatives will still need some work after this

In what way? I think the key is really just that the free interpolation cannot give you any more insight and you have to compute more (like a higher order Hermite)

ChrisRackauckas (approved):

ChrisRackauckas:
```julia
using StaticArrays, LinearAlgebra
mm = SMatrix{2,2}(rand(2,2))
u = SA[1.0,2.0]
reshape(diag(mm)  .!= 0, size(u)) isa SVector # true
```

src/dense/rosenbrock_interpolants.jl
pepijndevos: I think we pass it as nothing to all interpolants

test/interface/inplace_interpolation.jl
pepijndevos: That may actually have been because of the missing `dv` in `rosenbrock_interpolants` which caused hermite to be used when it shouldn't.

ChrisRackauckas: So it was just a temporary thing that should be fixed?

ChrisRackauckas:
```suggestion
        @test sol_ODE_interp.u == out_VF
```

ChrisRackauckas:
I'll need to do DelayDiffEq.jl because it reaches into internals that are changed on this.

topolarity:
> In what way? I think the key is really just that the free interpolation cannot give you any more insight and you have to compute more (like a higher order Hermite)

A higher-order analytic derivative would be lovely, but this is effectively a finite-differencing scheme anyway, so choosing an improved scheme would already provide a significant improvement. For example, the purple line here is just a linear interpolation of the "staircase" this PR provides for `sol(t, Val{1})` and pushes down the error of the derivative by more than an order of magnitude:
image (6)

Similarly, I don't see why we'd give up for `sol(t, Val{2})` and report an incorrect zero derivative, instead of continuing to do a finite-differencing scheme of some kind.

topolarity:
BTW by finite-differencing, I mean that the coefficients of the polynomial interpolants and their derivatives for algebraic variables are (usually linear) functions of `u(tᵢ)`. They don't analytically measure u'(t) anywhere, so they can at best perform like a decent finite-differencing scheme rather than a proper analytic derivative

ChrisRackauckas:
That's not a local scheme though, as you're using more than just the current step information? That can only be done in the post solution interpolation.

ChrisRackauckas:
> That just means you have to restrict to one-sided finite differencing techniques

But you only have u_n and u_{n+1} and the k's, which means if you finite difference the u's in a step then it must be a linear interpolation.

devmotion:
> I'll need to do DelayDiffEq.jl because it reaches into internals that are changed on this.

Someone apparently already ran into this problem 😅 https://github.com/SciML/DelayDiffEq.jl/issues/274

ChrisRackauckas:
Yeah it was a fundamental break and just needs a downstream fix. I'll put that in ASAP. Just been doing grant writing all week and was using this as a sidepiece while avoiding the real work.

jebej:
I suspect this PR caused a regression: #2086

pepijndevos:
What makes you say so?

If that's the case the thing to look for is a broadcast in Hermite interpolant that returns an array rather than a scalar.

jebej:
There are only two changes between 6.60 and 6.61, and the other doesn't appear related, but I might be wrong. https://github.com/SciML/OrdinaryDiffEq.jl/compare/v6.60.0...v6.61.0

Issue #2045, opened 2023-11-01
Voltage divider gives wrong answer with Rosenbrock23


oxinabox:
I haven't yet got a MWE without ModellingToolkit standard library.
but I am pretty sure this isn't MSL, 
as this is a reproduction of the same failure on the same circuit when implemented in Cedar, which doesn't use MSL

This is a classic circuit, with the well known solution that the voltage across each of 2 equal resistors in series is half the voltage across both of them.

```julia
using ModelingToolkitStandardLibrary.Electrical, ModelingToolkit, OrdinaryDiffEq, Test
using ModelingToolkitStandardLibrary.Blocks: Step,
    Constant, Sine, Cosine, ExpSine, Ramp,
    Square, Triangular
using ModelingToolkitStandardLibrary.Blocks: square, triangular
using OrdinaryDiffEq: ReturnCode.Success

@parameters t
@testset "voltage divider" begin
    @named source = Sine(offset = 0, amplitude = 1.0, frequency = 1e3, start_time = 0.5,phase = 0)
    @named voltage = Voltage()
    @named R1 = Resistor(R = 1e3)
    @named R2 = Resistor(R = 1e3)
    @named ground = Ground()

    connections = [connect(source.output, voltage.V)
        connect(voltage.p, R1.p)
        connect(R1.n, R2.p)
        connect(R2.n, voltage.n, ground.g)]

    @named model = ODESystem(connections, t,
        systems = [R1, R2, source, voltage, ground])
    sys = structural_simplify(model)
    prob = ODEProblem(sys, Pair[R2.i => 0.0], (0, 2.0))
    sol = solve(prob,  Rosenbrock23())
    @test sol.retcode == Success
    @test sol[voltage.p.v] ≈ 2*sol[R2.v]
end
```
output is
```
voltage divider: Test Failed at /home/oxinabox/JuliaEnvs/ModelingToolkitStandardLibrary.jl/test/Electrical/analog.jl:29
  Expression: sol[voltage.p.v] ≈ 2 * sol[R2.v]
   Evaluated: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.6422526531740537, -5.09502586896226e-13] ≈ [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -0.9082824125929643, 3029.774147437094]
```
WE can see at the end the voltage goes completely wild
[...]

PR #2050, opened 2023-11-02
warn for Rosenbrock methods with no differential variables

oscardssmith:
fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2045

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -96,6 +96,15 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
         error("This solver is not able to use mass matrices.")
     end
 
+    if alg isa OrdinaryDiffEqRosenbrockAdaptiveAlgorithm &&
+            !(prob.f.mass_matrix isa SciMLOperators.AbstractSciMLOperator) &&
+            all(isequal(0), prob.f.mass_matrix)
+        alg isa Union{Rosenbrock23, Rosenbrock32} && error("Rosenbrock23 and Rosenbrock32 require at least one differential variable to produce valid solutions")
+        if dense && verbose
+            @warn("Rosenbrock methods have questionable interpolations when applied to equations without differential states.")
+        end
+    end
+
     if !isempty(saveat) && dense
         @warn("Dense output is incompatible with saveat. Please use the SavingCallback from the Callback Library to mix the two behaviors.")
     end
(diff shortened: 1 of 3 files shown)


src/solve.jl
```diff
@@ -96,6 +96,16 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
         error("This solver is not able to use mass matrices.")
     end
 
+    if alg isa OrdinaryDiffEqRosenbrockAdaptiveAlgorithm &&
+            prob.f.mass_matrix isa AbstractMatrix &&
+            all(isequal(0), prob.f.mass_matrix)
+        # technically this should also warn for zero operators but those are hard to check for
+        alg isa Union{Rosenbrock23, Rosenbrock32} && error("Rosenbrock23 and Rosenbrock32 require at least one differential variable to produce valid solutions")
+        if dense && verbose
+            @warn("Rosenbrock methods have questionable interpolations when applied to equations without differential states.")
```
ChrisRackauckas, line 105:
```suggestion
            @warn("Rosenbrock methods on equations without differential states do not bound the error on interpolations.")
```

src/solve.jl
```diff
@@ -96,6 +96,16 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
         error("This solver is not able to use mass matrices.")
     end
 
+    if alg isa OrdinaryDiffEqRosenbrockAdaptiveAlgorithm &&
+            prob.f.mass_matrix isa AbstractMatrix &&
+            all(isequal(0), prob.f.mass_matrix)
+        # technically this should also warn for zero operators but those are hard to check for
+        alg isa Union{Rosenbrock23, Rosenbrock32} && error("Rosenbrock23 and Rosenbrock32 require at least one differential variable to produce valid solutions")
+        if dense && verbose
```
ChrisRackauckas, line 104: Should also warn if `saveat` is not empty.

oscardssmith: why is `saveat` related?

ChrisRackauckas: `saveat` uses interpolation

oscardssmith: oh, so you want `if (dense || !isempty(saveat)) && verbose?

ChrisRackauckas: yes

ChrisRackauckas:
will bump to require v1.9 and merge.

PR #2051, opened 2023-11-03
fix Rosenbrock methods when `eltype(u)!=eltype(tspan)`

oscardssmith:
MWE
```
using OrdinaryDiffEq
ode_f(du, u, p, t) = du[1] = -u[1]
prob = ODEProblem(ode_f, [1f0], (0.0,1.0))
solve(prob, Rodas5P())
```


ChrisRackauckas:
Test failures look real.

oscardssmith:
@ChrisRackauckas the main real test failure I'm seeing is from https://github.com/oscardssmith/OrdinaryDiffEq.jl/blob/5ab14bf88d72f2b74ba1253aa6e04874e8a5e06c/test/interface/mass_matrix_tests.jl#L302 where we are testing solving an ODE with rosenbrock methods that has a zero mass matrix (but not looking at the results). Can I just change the mass matrix here, or is the test testing something reasonable?

ChrisRackauckas:
It's from https://github.com/SciML/DifferentialEquations.jl/issues/757 as the comment says, so yes as long as it keeps the key property of testing non vector u0 it's fine.

ChrisRackauckas:
https://buildkite.com/julialang/ordinarydiffeq-dot-jl/builds/2906 looks like it's just throwing the wrong error now in that case, but is otherwise fine? Or maybe that test was just relying on the different types to be broken 😅 .

oscardssmith:
no. this is still slightly wrong. the remaining problem is that we are still making a minor mistake calculating the rate_prototype in the unitful case that is hard to fix without depending on unitful

ChrisRackauckas:
Is this done? Rebase?

PR #2053, opened 2023-11-08
bump lower to v1.9

ChrisRackauckas:


.buildkite/pipeline.yml
```diff
@@ -20,7 +20,6 @@ steps:
     matrix:
       setup:
         version:
-          - "1.6"
           - "1"
```
ranocha, line 23:
```suggestion
          - "1.9"
          - "1"
```
If we set the lower version to v1.9, shouldn't we run tests specifically on Julia v1.9 (given that v1 may soon be v1.10)?

.github/workflows/CI.yml
```diff
@@ -26,7 +26,6 @@ jobs:
           - Multithreading
         version:
           - '1'
```
ranocha, line 28:
```suggestion
          - '1.9'
          - '1'
```
See above

.github/workflows/Downstream.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        julia-version: [1,1.6]
+        julia-version: [1]
```
ranocha, line 17:
```suggestion
        julia-version: [1.9, 1]
```
See above

ChrisRackauckas:
For now this is just doubling CI, but why not. In the near future v1.10 should just be the new LTS and we'll drop this.

Issue #677, opened 2019-02-28
More Runge-Kutta-Nyström Methods


ranocha:
Here are some additional papers describing/developing/analysing Runge-Kutta-Nyström methods:
- [ ] [Dormand, J. R., and PJi Prince. "New Runge-Kutta algorithms for numerical simulation in dynamical astronomy." Celestial Mechanics 18.3 (1978): 223-232.](https://link.springer.com/article/10.1007/BF01230162)
- [ ] [Filippi, S., and J. Gräf. "New Runge–Kutta–Nyström formula-pairs of order 8 (7), 9 (8), 10 (9) and 11 (10) for differential equations of the form y ″= f (x, y)." Journal of computational and applied mathematics 14.3 (1986): 361-370.](https://www.sciencedirect.com/science/article/pii/0377042786900737)
- [ ] [Fehlberg, E., S. Filippi, and J. Gräf. "Ein Runge‐Kutta‐Nyström‐Formelpaar der Ordnung 10 (11) für Differentialgleichungen der Form y′= f (x, y)." ZAMM‐Journal of Applied Mathematics and Mechanics/Zeitschrift für Angewandte Mathematik und Mechanik 66.7 (1986): 265-270.](https://onlinelibrary.wiley.com/doi/abs/10.1002/zamm.19860660703)
- [ ] [Marthinsen, Arne. "Continuous extensions to Nyström methods for second order initial value problems." BIT Numerical Mathematics 36.2 (1996): 309-332.](https://link.springer.com/article/10.1007%2FBF01731986)
- [ ] [Fine, Jerry Michael. "Low order practical Runge-Kutta-Nyström methods." Computing 38.4 (1987): 281-297.](https://link.springer.com/article/10.1007/BF02278707)
- [ ] [Fine, Jerry Michael. "Interpolants for Runge-Kutta-Nyström methods." Computing 39.1 (1987): 27-42.](https://link.springer.com/article/10.1007/BF02307711)
- [ ] [Sharp, P. W., and J. M. Fine. "Some Nyström pairs for the general second-order initial-value problem." Journal of Computational and applied mathematics 42.3 (1992): 279-291.](https://www.sciencedirect.com/science/article/pii/0377042792900818)
- [ ] [Bettis, D. G. "A Runge-Kutta Nyström algorithm." Celestial mechanics 8.2 (1973): 229-233.](https://link.springer.com/article/10.1007/BF01231421)
[...]

PR #2056, opened 2023-11-10
Dense output for FineRKN5

HenryLangner:
Implementation of a specialized `dense_output` (6th order Hermite-Interpolation) for `FineRKN5` derived in https://doi.org/10.1007/BF02307711 (for ref. view https://github.com/SciML/OrdinaryDiffEq.jl/issues/677). 

As shown in the convergence plots below, the Interpolation works fine (for `In-Place` ODEs, more on that below) and the `dense_error` is reduced significantly compared to the standard 3rd order Hermite-Interpolation:

In-Place (i) and Out-of-Place (o) convergence tests with and without dense error for the harmonic oscillator using the **new dense output**:
harmonic_oscillator_convergence_plots

In-Place (i) and Out-of-Place (o) convergence tests with and without dense error for the harmonic oscillator using the **standard dense output**:
harm_oscillator_convergence_plots_with3rdOrderHermit

However testing the convergence with the new dense output for ODEs dependent on the first derivative (velocity dependent) results in breaking the `Out-of-Place` implementation in the sense, that the convergence worsen compared to the `In-Place` implementation. The difference can be seen in the plots below:

In-Place (i) and Out-of-Place (o) convergence tests with and without dense error for the damped oscillator using the new dense output:
damped_oscillator_convergence_plots

I can not find the mistake in my `Out-of-Place` implementation.
Looking at the plots I am certain the mistake is not related to the coefficients since both implementations share the same ones and the In-Place on works regardless.

Any help is appreciated! 

The code used for the convergence tests and plots is attached and can directly be pasted in the REPL for easy testing and replication of my results.
[FineRKN5convergenceTests.txt](https://github.com/SciML/OrdinaryDiffEq.jl/files/13319376/FineRKN5convergenceTests.txt)

@ranocha


src/perform_step/rkn_perform_step.jl
```diff
@@ -272,12 +297,33 @@ end
...
+    ku = uprev +
+         dt * (c9 * duprev +
+          dt * (a91 * k1 + a92 * k2 + a93 * k3 + a94 * k4 + a95 * k5 + a97 * k7)) # a96 = a98 = 0
+    kdu = duprev +
+          dt * (abar91 * k1 + abar93 * k3 + abar94 * k4 + abar95 * k5 +
+                abar97 * k7) # abar92 = abar96 = abar98 = 0
+
+    k9 = f.f1(kdu, ku, p, t + dt * c9)
+    k10 = uprev + dt * (duprev + dt * (btilde1_1 * k1 + btilde2_1 * k2 + btilde3_1 * k3 + btilde4_1 * k4 + btilde5_1 * k5 + btilde6_1 * k6 + btilde7_1 * k7 + btilde8_1 * k8 + btilde9_1 * k9)) # higher order Evaluations for ytilde_{n+1} (as donated in corresponding Paper)
+    k11 = uprev + dt * (duprev * 0.5 + dt * (btilde1_12 * k1 + btilde2_12 * k2 + btilde3_12 * k3 + btilde4_12 * k4 + btilde5_12 * k5 + btilde6_12 * k6 + btilde7_12 * k7 + btilde8_12 * k8 + btilde9_12 * k9)) # higher order Evaluations for ytilde_{n+1/2} (as donated in corresponding Paper)
```
ChrisRackauckas, line 317: Can these be done lazily like is done with Vern?

PR #2059, opened 2023-11-14
New Algorithm SharpFineRKN6

HenryLangner:
Implementation of the new algorithm `SharpFineRKN6` mentioned in [#677](https://github.com/SciML/OrdinaryDiffEq.jl/issues/677). The Algorithm works on general second order differential equations.

The convergence plots below show a convergence which does not match the convergence rate mentioned in the paper and a large error.

Convergence plots harmonic Oscillator:
SharpFineRKN6_convergence_harm_osz

Convergence plots damped Oscillator:
SharpFineRKN6_convergence_damped_osz

Unfortunately the scan quality of the paper (https://doi.org/10.1016/0377-0427(92)90081-8) leaves some coefficients unrecognizable (mainly `abar84`, `a51`,`a71`, `a71` and `a82` on page 289). With the order conditions (6), (7) and (21) mentioned in the paper, I identified the correct coefficients which can be found in the src/tableaus/rkn_tableaus.jl file. However that did not resolve the issue with the poor convergence and large error. Further investigation shows that order condition (10) mentioned in the paper is not satisfied. Since (7) and (9) are satisfied the error has to be in one ore more of the following coefficients: `a43`, `a53`, `a54`, `a63`, `a64`, `a65`, `a73`, `a74`,`a75`, `a76`. I double checked the coefficients imported in src/tableaus/rkn_tableaus.jl. Everything is correct here. It is possible that I overlooked something and/or there is a typo in the paper.
The error could also be in the perform step function (implemented in src/perform_step/rkn_perform_step.jl). I double checked everything here too nut could not find a mistake. 

Help is appreciated!
 

@ranocha


ChrisRackauckas:
Try flipping a few signs. I have seen more than a few papers forget a `-` on one of the `a` values.

HenryLangner:
> Try flipping a few signs. I have seen more than a few papers forget a `-` on one of the `a` values.

Good idea, I will try that.

PR #2062, opened 2023-11-16
Drop passing p and t to the JacVec operator

avik-pal:


src/derivative_utils.jl
```diff
@@ -302,9 +302,22 @@ function SciMLOperators.update_coefficients!(W::WOperator,
     dtgamma = nothing,
     transform = nothing)
     if (u !== nothing) && (p !== nothing) && (t !== nothing)
-        update_coefficients!(W.J, u, p, t)
+        if W.J isa FunctionOperator
+            update_coefficients!(W.J, u, ifelse(W.J.p === nothing, nothing, p),
+                ifelse(W.J.t === nothing, nothing, t))
```
ChrisRackauckas, line 307: what's this all about? Why doesn't the `update_coefficients!` handle this correctly?

src/derivative_utils.jl
```diff
@@ -874,7 +887,7 @@ function build_J_W(alg, u, uprev, p, t, dt, f::F, ::Type{uEltypeNoUnits},
         # be overridden with concrete_jac.
 
         _f = islin ? (isode ? f.f : f.f1.f) : f
-        jacvec = JacVec(UJacobianWrapper(_f, t, p), copy(u), p, t;
```
ChrisRackauckas, line 877: how is it going to update the coefficients correctly without this?

avik-pal:
That was a mistake. But now all tests are passing. Also we don't carry around `p` and `t` at multiple levels of the wrapper. It is only present in the FunctionOperator.

Issue #2068, opened 2023-11-20
`BS3` docs document `OwrenZen5` instead


oscardssmith:
The [BS3 documentation](https://github.com/SciML/OrdinaryDiffEq.jl/blob/da70b267a1dacda7839fb4bbb50950b618e24222/src/algorithms/explicit_rk.jl#L257C1-L257C1) documents OwrenZen5 (presumably due to a bad copy paste. This is especially bad since BS3 is one of our recommended algorithms (for high tolerance non-stiff)

PR #2069, opened 2023-11-21
fix docstring of BS3

ranocha:
Closes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2068

I copied the first sentence of the old docstring in

https://github.com/SciML/OrdinaryDiffEq.jl/blob/b50390f89bf2d11cbe6b64ee0dcc952fc0044a8a/src/algorithms.jl#L4256-L4262


ChrisRackauckas:
oh oops thanks for finding this.

ChrisRackauckas:
@avik-pal things are back to green!

Issue #2054, opened 2023-11-08
Rosenbrock methods give unreliable results for DAEs with fast algebraic dynamics


topolarity:
This is in essence the same problem as https://github.com/SciML/OrdinaryDiffEq.jl/issues/2045, except that differential variables are present but evolve far too slowly to provide implicit error control for algebraic variables.

```julia
using ModelingToolkit

@variables t
D = Differential(t)

@mtkmodel Foo begin
    @parameters begin
        ω₁
        ω₂
        l
    end
    @variables begin
        x(t) # differential variable
        y(t) # differential variable
        z(t) # algebraic variable
    end
    @equations begin
        D(x) ~  ω₁ * 2pi * y
        D(y) ~ -ω₁ * 2pi * x
        z^l ~ sin(ω₂ * 2pi * t)
    end
end

@mtkbuild foo = Foo();
```

Solve and plot with a variety of solvers:
```julia
using DifferentialEquations: solve, Rosenbrock23, Rodas4, FBDF

prob = ODEProblem(foo,
                  [foo.x => 1.0, foo.y => 0.0, foo.z => 0.0],
                  (0.0, 1.0),
                  [foo.ω₁ => 1.0, foo.ω₂ => 1000.0, foo.l => 1.0])

sol0 = solve(prob, Rosenbrock23())
sol1 = solve(prob, Rodas4())
sol2 = solve(prob, FBDF())

plots = []
for (name, sol) in [
    ("Rosenbrock23", sol0),
    ("Rodas4", sol1),
    ("FBDF", sol2),
]
    plt = plot(sol(sol.t, idxs=foo.z), layout=(2,1), legend=:none, title="$(name) (t=0..1)")
    plot!(plt[2], sol(sol.t[sol.t .<= 2e-3], idxs=foo.z), layout=(2,1), legend=:none, title="$(name) (t=0..2e-3)")
    push!(plots, plt)
end
plots
```

results in:

![image](https://github.com/SciML/OrdinaryDiffEq.jl/assets/84105208/75eee4db-27d7-45f6-ae09-94f3d8b460d0)
![image](https://github.com/SciML/OrdinaryDiffEq.jl/assets/84105208/be7c9da6-c9ae-4731-85db-1fe5db3f8d80)
![image](https://github.com/SciML/OrdinaryDiffEq.jl/assets/84105208/b674a671-8631-44d3-ae3e-c94f03c9350f)

topolarity:
[...]

PR #2079, opened 2023-12-11
Alternative error control on Rosenbrock23 algebraic equations

ChrisRackauckas:
Adds a separate error term for Rosenbrock23 when the mass matrix is not `I` in order to account for the fact that the algebraic conditions are not error controlled. They are added via doing error control on the absolute tolerance of the residual in the algebraic conditions. There isn't a sensible definition of relative tolerance here so leaving it at that.

Note that this technique is not applicable to the Rodas / stiffly stable methods since they always exactly satisfy the algebraic conditions by design, i.e. this term is just zero. Those methods will need to use a different strategy like residual control.

Handles part of https://github.com/SciML/OrdinaryDiffEq.jl/issues/2054.

Fixes https://github.com/SciML/OrdinaryDiffEq.jl/pull/2079

### DIFF
--- a/test/regression/hard_dae.jl
+++ b/test/regression/hard_dae.jl
@@ -221,6 +221,7 @@ refsol = solve(prob, Rodas4(), saveat = 0.1, callback = cb, tstops = [1.0], relt
 for solver in (Rodas4, Rodas4P, Rodas5, Rodas5P, FBDF, QNDF, Rosenbrock23)
     @show solver
     prob = ODEProblem(f, deepcopy(res.zero), (0, 20.0), deepcopy(p_inv))
+    maxiters = solver isa Rosenbrock23 ? 1e6 : 1e5
     sol = solve(prob, solver(), saveat = 0.1, callback = cb, tstops = [1.0], reltol = 1e-12,
         abstol = 1e-16)
     @test sol.retcode == ReturnCode.Success
--- a/test/regression/hard_dae.jl
+++ b/test/regression/hard_dae.jl
@@ -223,7 +223,7 @@ for solver in (Rodas4, Rodas4P, Rodas5, Rodas5P, FBDF, QNDF, Rosenbrock23)
     prob = ODEProblem(f, deepcopy(res.zero), (0, 20.0), deepcopy(p_inv))
     maxiters = solver isa Rosenbrock23 ? 1e6 : 1e5
     sol = solve(prob, solver(), saveat = 0.1, callback = cb, tstops = [1.0], reltol = 1e-12,
-        abstol = 1e-16)
+        abstol = 1e-16, maxiters = maxiters)
     @test sol.retcode == ReturnCode.Success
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
```julia
using ModelingToolkit, Plots

@variables t
D = Differential(t)

@mtkmodel Foo begin
    @parameters begin
        ω₁
        ω₂
        l
    end
    @variables begin
        x(t) # differential variable
        y(t) # differential variable
        z(t) # algebraic variable
    end
    @equations begin
        D(x) ~  ω₁ * 2pi * y    # solution: x(t) = sin(2πω₁t)
        D(y) ~ -ω₁ * 2pi * x    # solution: y(t) = cos(2πω₁t)
        z^l ~ sin(ω₂ * 2pi * t) # solution: z(t) = sin(2πω₂t) when l == 1.0
    end
end

@mtkbuild foo = Foo();

using OrdinaryDiffEq: solve, Rosenbrock23, Rodas4, FBDF

prob = ODEProblem(foo,
                  [foo.x => 1.0, foo.y => 0.0, foo.z => 0.0],
                  (0.0, 1.0),
                  [foo.ω₁ => 1.0, foo.ω₂ => 1000.0, foo.l => 1.0])

sol0 = solve(prob, Rosenbrock23(), maxiters = 1e7)
sol1 = solve(prob, Rodas4())
sol2 = solve(prob, FBDF())

plots = []
for (name, sol) in [
    ("Rosenbrock23", sol0),
    ("Rodas4", sol1),
    ("FBDF", sol2),
]
    z = sol(sol.t, idxs=foo.z)
    plt = plot(z, layout=(2,1), legend=:none, title="$(name) (t=0..1)")
    plot!(plt[2], z[1:findfirst(sol.t .>= 2e-3)], xlim=(0,2.1e-3), legend=:none, title="$(name) (t=0..2e-3)")
    push!(plots, plt)
end
plot(plots...)
```

Before:

rosenbrock_plot_before

After:

rosenbrock_plot

test/regression/hard_dae.jl
```diff
@@ -221,8 +221,9 @@ refsol = solve(prob, Rodas4(), saveat = 0.1, callback = cb, tstops = [1.0], relt
 for solver in (Rodas4, Rodas4P, Rodas5, Rodas5P, FBDF, QNDF, Rosenbrock23)
     @show solver
     prob = ODEProblem(f, deepcopy(res.zero), (0, 20.0), deepcopy(p_inv))
+    maxiters = solver isa Rosenbrock23 ? 1e8 : 1e5
```
oscardssmith, line 224: is it reasonable that `Rosenbrock23` now takes 1000x more steps than the other solvers?

ChrisRackauckas: Something is up with that test, I think I need to bring it local and check. It wouldn't be that unreasonable though since it's not B-stable.

PR #2081, opened 2023-12-12
Remove uses of NLsolve

avik-pal:


src/initialize_dae.jl
```diff
@@ -20,13 +20,12 @@ function BrownFullBasicInit(; abstol = 1e-10, nlsolve = nothing)
...
 
-using SciMLNLSolve
 default_nlsolve(alg, isinplace, u, autodiff = false) = alg
 function default_nlsolve(::Nothing, isinplace, u, autodiff = false)
-    NLSolveJL(autodiff = autodiff ? :forward : :central)
+    TrustRegion(; autodiff = autodiff ? AutoForwardDiff() : AutoFiniteDiff())
 end
 function default_nlsolve(::Nothing, isinplace::Val{false}, u::StaticArray, autodiff = false)
-    SimpleNewtonRaphson(autodiff = autodiff)
+    SimpleNewtonRaphson(autodiff = autodiff ? AutoForwardDiff() : AutoFiniteDiff())
```
avik-pal, line 28: @ChrisRackauckas should we change this to `SimpleTrustRegion` to make it consistent with the non static array version?

ChrisRackauckas: Yes

avik-pal:
failures seem to be same as master

src/initialize_dae.jl
```diff
@@ -564,7 +563,7 @@ function _initialize_dae!(integrator, prob::DAEProblem,
     if alg.nlsolve !== nothing
         nlsolve = alg.nlsolve
     else
-        nlsolve = NewtonRaphson(autodiff = isAD)
+        nlsolve = NewtonRaphson(autodiff = alg_autodiff(integrator.alg))
```
ChrisRackauckas, line 566: We should keep it as type information now that we can use it effectively.

ChrisRackauckas: Yes it changes the type level stuff to a bool as backwards compat support IIRC

avik-pal:
```
function alg_autodiff(alg)
    autodiff = _alg_autodiff(alg)
    if autodiff == Val(false)
        return AutoFiniteDiff()
    elseif autodiff == Val(true)
        return AutoForwardDiff()
    else
        return _unwrap_val(autodiff)
    end
end
```
This should constant propagate and give the correct AD type right?

ChrisRackauckas: We shouldn't rely on constant propagation. It can fail. But let's pull this in and make this better in a follow up.

PR #2093, opened 2023-12-26
Test master

ChrisRackauckas:


ChrisRackauckas:
Well this was more interesting than I thought it would be: https://github.com/JuliaLang/julia/issues/52635

Issue #1842, opened 2023-01-21
Strange behavior of save_at=false


erlebach:
I am getting strange behavior with the different versions of save_at and save_start and save_end. Here is a MWE:
``` 
using DifferentialEquations

function exponen!(du, u, param, t)
	du[1] = -u[1]
end

tspan = (0., 3.)
prob = ODEProblem(exponen!, [2.], tspan)
solve(prob, Tsit5(), saveat=[0.2, 1.3], save_start=false, save_end=false)
```
I expect the solution at the two times 0.2 and 1.3 to be printed. This is what happens. Next, I explicitly add the start and end times to `saveat`, and specify that save_start and save_end are false. I expect the solution at four times values to be printed. However, only three are printed.
```
sol = solve(prob, Tsit5(), saveat=[0., 0.2, 1.3, 3.], save_start=false, save_end=false)
println("sol.t: ", sol.t)
```
This might be a missing edge case.

ArneBouillon:
I also encountered this issue today. When solving an `ODEProblem` and passing a `saveat` that contains `tspan[1]` and/or `tspan[2]`, the following behaviour is observed:
- When `tspan[1]` is included in `saveat` and `save_start = false`, that time point is not contained in `solve`'s solution (and so the solution has one time point less than does `saveat`.
- When `tspan[2]` is included in `saveat` and `save_end = false`, the terminal time point is still included in the solution.

An example is given by the following code.
```julia
function f(du, u, p, t)
    du[1] = -cos(u[1]) * u[1]
end
prob = ODEProblem(f, [10], (0.0, 0.4))

vcat(solve(prob, Tsit5(); saveat = 0:.1:.4).t...)
# =>  [0.0; 0.1; 0.2; 0.3; 0.4]
vcat(solve(prob, Tsit5(); saveat = 0:.1:.4, save_start = true, save_end = true).t...)
# =>  [0.0; 0.1; 0.2; 0.3; 0.4]
vcat(solve(prob, Tsit5(); saveat = 0:.1:.4, save_start = false, save_end = false).t...)
# =>       [0.1; 0.2; 0.3; 0.4]

```

[...]

PR #2095, opened 2023-12-28
Fix save_end overriding behavior

ChrisRackauckas:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1842


src/integrators/integrator_utils.jl
```diff
@@ -145,6 +149,7 @@ postamble!(integrator::ODEIntegrator) = _postamble!(integrator)
 function _postamble!(integrator)
     DiffEqBase.finalize!(integrator.opts.callback, integrator.u, integrator.t, integrator)
     solution_endpoint_match_cur_integrator!(integrator)
+    save
```
ChrisRackauckas, line 152:
```suggestion
```

PR #2099, opened 2023-12-30
Fix FunctionMap idxs handling in interpolation

ChrisRackauckas:
Fixes https://github.com/SciML/SciMLBase.jl/issues/575


isaacsas:
Thanks!

Issue #2101, opened 2023-12-31
Rodas5P does not allow 2nd order (or higher) derivatives


staticfloat:
**Describe the bug 🐞**

Rodas5P, which has a 4th order interpolant, according to [the documentation](https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/), should allow `integrator(t, Val{2})`, but currently doesn't due to missing support.

**Minimal Reproducible Example 👇**

```julia
using OrdinaryDiffEq
using ODEProblemLibrary: prob_ode_linear
callback = DiscreteCallback(
    (u, t, integ) -> begin
        @show integ(t, Val{2})
        return true
    end,
    identity,
)
sol = solve(prob_ode_linear, Rodas5P(); callback)
```

**Error & Stacktrace ⚠️**

```julia
ERROR: Derivative order too high for interpolation order. An interpolation derivative is
only accurate to a certain deriative. For example, a second order interpolation
is a quadratic polynomial, and thus third derivatives cannot be computed (will be
incorrectly zero). Thus use a solver with a higher order interpolation or compute
the higher order derivative through other means.

You can find the list of available ODE/DAE solvers with their documented interpolations at:

* https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/
* https://docs.sciml.ai/DiffEqDocs/stable/solvers/dae_solve/


Some of the types have been truncated in the stacktrace for improved reading. To emit complete information
in the stack trace, evaluate `TruncatedStacktraces.VERBOSE[] = true` and re-run the code.

Stacktrace:
  [1] _ode_interpolant(Θ::Float64, dt::Float64, y₀::Float64, y₁::Float64, k::Vector{…}, cache::OrdinaryDiffEq.Rosenbrock5ConstantCache{…}, idxs::Nothing, T::Type{…}, differential_vars::Nothing)
    @ OrdinaryDiffEq ~/.julia/packages/OrdinaryDiffEq/nm4K9/src/dense/rosenbrock_interpolants.jl:11
  [2] ode_interpolant(Θ::Float64, dt::Float64, y₀::Float64, y₁::Float64, k::Vector{…}, cache::OrdinaryDiffEq.Rosenbrock5ConstantCache{…}, idxs::Nothing, T::Type{…}, differential_vars::Nothing)
    @ OrdinaryDiffEq ~/.julia/packages/OrdinaryDiffEq/nm4K9/src/dense/generic_dense.jl:578
[...]

PR #2092, opened 2023-12-22
New methods Rodas23W / Rodas3P with error test for interpolation, see issue 2054 https://github.com/SciML/OrdinaryDiffEq.jl/issues/2054

gstein3m:
## Checklist

- [ x] Appropriate tests were added: wp precision benchmarks were successful 
- [x ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API

PR #2102, opened 2024-01-01
Rodas5(P) 2nd and third derivatives

ChrisRackauckas:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2101

Second derivative looks fine, third derivative looks weird.

```julia
using OrdinaryDiffEq, ODEProblemLibrary
prob = prob_ode_linear
sol1 = solve(prob, Rodas5P(), dt = 1 // 2^(8), dense = true)
sol2 = solve(prob, Vern9(), dt = 1 // 2^(8), dense = true)

using Plots
t = 0.0:0.01:1.0
plot(t, Array(sol1(t, Val{1})))
plot!(t, Array(sol2(t, Val{1})))

plot(t, Array(sol1(t, Val{2})))
plot!(t, Array(sol2(t, Val{2})))

plot(t, Array(sol1(t, Val{3})))
plot!(t, Array(sol2(t, Val{3})))
```

### DIFF
--- a/src/dense/rosenbrock_interpolants.jl
+++ b/src/dense/rosenbrock_interpolants.jl
@@ -311,4 +311,70 @@ end
                                     y₀[idxs] + y₁[idxs]) / dt
     out
 end
-#-
+
+# Second Derivative
+@muladd function _ode_interpolant(Θ, dt, y₀, y₁, k, cache::Rosenbrock5ConstantCache,
+    idxs::Nothing, T::Type{Val{2}}, differential_vars)
+    @inbounds (-2 * k[1] + 2 * k[2] + Θ * (-6 * k[2] + 6 * k[3] - 12 * Θ * k[3])) / dt^2
+end
+
+@muladd function _ode_interpolant(Θ, dt, y₀, y₁, k, cache::Rosenbrock5Cache, idxs::Nothing,
+    T::Type{Val{2}}, differential_vars)
+    @inbounds @.. broadcast=false (-2 * k[1] + 2 * k[2] +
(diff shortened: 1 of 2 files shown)


gstein3m:
Here is another example which could explain the behaviour:

```
using OrdinaryDiffEq, Plots

const n = 5

function f(du,u,p,t)
    du[1] = n*t^(n-1)
end
tspan = [0.0, 1.0]; u0 = [0.0]
prob1 = ODEProblem(ODEFunction(f), u0, tspan)
  
sol1 = solve(prob1, Rodas5(), dt = 1 // 2^(8), dense = true)
println(sol1.destats)
err_max = maximum(abs.(sol1[1,:] .- sol1.t.^n))
println("err_max:",err_max)
sol2 = solve(prob1, Vern9(), dt = 1 // 2^(8), dense = true)

t = 0.0:0.01:1.0
plot(sol1(t, Val{3}))
plot!(sol2(t, Val{3}))

```
The polynimial of order 5 can be integrated exactly by Rodas5 / 5P.
Since the interpolation error is not checked, only 5 time steps are taken regardless of the choice of dt. The thrid derivative of the fourth order interpolation is a linear function. This cannot give a good approximation of a second order polynomial within only 5 steps.
I therefore think that the proposed implementation is correct.

Independently of this, the evaluation of the first derivation for Rodas4/4P/4P2 does not work for me.
However, this may also be due to my suggested changes in PR #2092

ChrisRackauckas:
I see. We should update that test case sometime then, but for now it does seem correct.

PR #2103, opened 2024-01-01
Redesign default ODE solver to be fully type-grounded

ChrisRackauckas:
This accomplishes a few things:

* Faster precompile times by precompiling less
* Full inference of results when using the automatic algorithm
* Hopefully faster load times by also precompiling less

This is done the same way as

* linearsolve https://github.com/SciML/LinearSolve.jl/pull/307
* nonlinearsolve https://github.com/SciML/NonlinearSolve.jl/pull/238

and is thus the more modern SciML way of doing it. It avoids dispatch by having a single algorithm that always generates the full cache and instead of dispatching between algorithms always branches for the choice.

It turns out, the mechanism already existed for this in OrdinaryDiffEq... it's CompositeAlgorithm, the same bones as AutoSwitch! As such, this reuses quite a bit of code from the auto-switch algorithms but instead of just having two choices it (currently) has 6 that it chooses between. This means that it has stiffness detection and switching behavior, but also in a size-dependent way.

There are still some optimizations to do though. Like LinearSolve.jl, it would be more efficient to have a way to initialize the caches to size zero and then have a way to re-initialize them to the correct size. Right now, it'll generate the same Jacobian N times and it shouldn't need to do that.


ChrisRackauckas:
```julia
1+1
@time using OrdinaryDiffEq
function lorenz!(du,u,p,t)
 du[1] = 10.0(u[2]-u[1])
end
u0 = [1.0;0.0;0.0]
tspan = (0.0,100.0)
prob = ODEProblem(lorenz!,u0,tspan)
@time sol = solve(prob, OrdinaryDiffEq.DefaultODEAlgorithm())

using Test
@inferred solve(prob, OrdinaryDiffEq.DefaultODEAlgorithm())

function lorenz(du,u,p,t)
 du[1] = 10.0(u[2]-u[1])
end

u0 = [1.0;0.0;0.0]
tspan = (0.0,100.0)
alg = OrdinaryDiffEq.DefaultODEAlgorithm()

using ProfileView
ProfileView.view(flamegraph(tinf))
```

ChrisRackauckas:
```
# PR

# Precompile and load
# 47.139365 seconds (10.32 M allocations: 690.241 MiB, 0.47% gc time, 10.08% compilation time: 2% of which was recompilation)
# Just load
# 3.222100 seconds (4.06 M allocations: 249.297 MiB, 2.98% gc time, 1.94% compilation time)
# First solve
# 0.129151 seconds (115.04 k allocations: 8.043 MiB, 98.91% compilation time)

# Master
# 48.996228 seconds (10.33 M allocations: 691.287 MiB, 0.41% gc time, 9.80% compilation time: 2% of which was recompilation
# Just Load
# 3.320052 seconds (4.06 M allocations: 250.294 MiB, 3.17% gc time, 1.94% compilation time)
```

ChrisRackauckas:
Precompilation doesn't drop much more on https://github.com/SciML/OrdinaryDiffEq.jl/pull/2104

src/composite_algs.jl
```diff
@@ -81,3 +69,114 @@ AutoVern6(alg; lazy = true, kwargs...) = AutoAlgSwitch(Vern6(lazy = lazy), alg;
 AutoVern7(alg; lazy = true, kwargs...) = AutoAlgSwitch(Vern7(lazy = lazy), alg; kwargs...)
 AutoVern8(alg; lazy = true, kwargs...) = AutoAlgSwitch(Vern8(lazy = lazy), alg; kwargs...)
 AutoVern9(alg; lazy = true, kwargs...) = AutoAlgSwitch(Vern9(lazy = lazy), alg; kwargs...)
+
+### Default ODE Solver
+
+EnumX.@enumx DefaultSolverChoice begin
+    Tsit5 = 1
+    Vern7 = 2
+    Rosnebrock23 = 3
```
nathanaelbosch, line 78: @ChrisRackauckas seems like a typo: 
```suggestion
    Rosenbrock23 = 3
```

src/composite_algs.jl
```diff
@@ -81,3 +69,114 @@ AutoVern6(alg; lazy = true, kwargs...) = AutoAlgSwitch(Vern6(lazy = lazy), alg;
...
+function stiffchoice(reltol, len)
+    x = if len > MEDIUMSIZE
+        DefaultSolverChoice.KrylovFBDF
+    elseif len > SMALLSIZE
+        DefaultSolverChoice.FBDF
+    else
+        if reltol < LOW_TOL
+            DefaultSolverChoice.Rodas5P
+        else
+            DefaultSolverChoice.Rosnebrock23
```
ChrisRackauckas, line 143:
```suggestion
            DefaultSolverChoice.Rosenbrock23
```

ChrisRackauckas:
Closed as it was rebased in https://github.com/SciML/OrdinaryDiffEq.jl/pull/2184

PR #2105, opened 2024-01-02
make dae initialization work for non eltypable p

oscardssmith:
This was broken by https://github.com/SciML/OrdinaryDiffEq.jl/pull/2100

### DIFF
--- a/src/initialize_dae.jl
+++ b/src/initialize_dae.jl
@@ -163,7 +163,11 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
         end
 
         nlequation! = @closure (out, u, p) -> begin
-            T = p isa SciMLBase.NullParameters ? eltype(u) : Base.promote_type(eltype(u),eltype(p))
+            if p isa AbstractArray{<:Dual}
+                T = Base.promote_type(eltype(u), eltype(p))
+            else
+                T = eltype(u)
+            end
             update_coefficients!(M, u, p, t)
             # f(u,p,t) + M * (u0 - u)/dt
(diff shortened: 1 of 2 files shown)


test/integrators/dae_initialization_tests.jl
```diff
@@ -197,6 +197,14 @@ integrator = init(prob, DABDF2())
 @test integrator.du[1]≈1.0 atol=1e-9
 @test integrator.du[2]≈1.0 atol=1e-9
 
+# test initialization with parameters without eltype/length
+struct A
+end
+probp = DAEProblem(f, du₀, u₀, A(), tspan, differential_vars = differential_vars)
+for initializealg in (ShampineCollocationInit(), BrownFullBasicInit())
+    @test isapprox(init(probp, DABDF2(); initializealg).u, init(prob, DABDF2(); initializealg).u)
```
staticfloat, line 205: Don't you need an `end` here?

src/initialize_dae.jl
```diff
@@ -163,7 +163,11 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
         end
 
         nlequation! = @closure (out, u, p) -> begin
-            T = p isa SciMLBase.NullParameters ? eltype(u) : Base.promote_type(eltype(u),eltype(p))
+            if p isa AbstractArray{<:Dual}
```
ChrisRackauckas, line 166: This isn't correct.

ChrisRackauckas:
Wait, merged too early.

```julia
            if DiffEqBase.anyeltypedual(p) <: Dual
                T = Base.promote_type(eltype(u), eltype(p))
```

instead of `eltype(p)` it should just use the `DiffEqBase.anyeltypedual(p)` in the type promotion.

PR #2110, opened 2024-01-06
Fix DAE initialization AD when autodiff=false

ChrisRackauckas:
Simply needs to still create the dual cache

### DIFF
--- a/src/initialize_dae.jl
+++ b/src/initialize_dae.jl
@@ -154,7 +154,7 @@ function _initialize_dae!(integrator, prob::ODEProblem, alg::ShampineCollocation
             tmp = copy(_u0)
         end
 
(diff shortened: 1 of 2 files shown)


src/initialize_dae.jl
```diff
@@ -483,6 +483,9 @@ function _initialize_dae!(integrator, prob::ODEProblem,
         end
         uu = isAD ? PreallocationTools.get_tmp(_tmp, T) : _tmp
         du_tmp = isAD ? PreallocationTools.get_tmp(_du_tmp, T) : _du_tmp
+        @show isAD
+        @show T, TP
+        @show typeof(x), typeof(p), typeof(du_tmp)
```
avik-pal, line 488: should have been deleted?

ChrisRackauckas: yeah I put in a hotfix to handle that.

PR #2123, opened 2024-01-28
Fix out-of-place Rosenbrock23

YingboMa:
The error estimate needs to be scaled appropriately.

### DIFF
--- a/src/perform_step/rosenbrock_perform_step.jl
+++ b/src/perform_step/rosenbrock_perform_step.jl
@@ -418,7 +418,11 @@ end
         end
         integrator.stats.nsolve += 1
 
-        utilde = dto6 * f.mass_matrix * (k₁ - 2 * k₂ + k₃)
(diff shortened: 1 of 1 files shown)


src/perform_step/rosenbrock_perform_step.jl
```diff
@@ -418,7 +418,7 @@ end
         end
         integrator.stats.nsolve += 1
 
-        utilde = dto6 * (k₁ - 2 * k₂ + k₃)
+        utilde = dto6 * f.mass_matrix * (k₁ - 2 * k₂ + k₃)
```
oscardssmith, line 421:
```suggestion
if  u isa Number
    utilde = dto6 * f.mass_matrix[1,1] * (k₁ - 2 * k₂ + k₃)
else
    utilde = dto6 * f.mass_matrix * (k₁ - 2 * k₂ + k₃)
end
```

PR #2137, opened 2024-02-12
save du and k for DAEs when dense==true

oscardssmith:
## Checklist

## Additional context

depends on https://github.com/SciML/SciMLBase.jl/pull/622 to add the k field to `DAESolution`.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -25,7 +25,7 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
     save_end = nothing,
     callback = nothing,
     dense = save_everystep &&
-                !(alg isa Union{DAEAlgorithm, FunctionMap}) &&
+                !(alg isa FunctionMap) &&
                 isempty(saveat),
     calck = (callback !== nothing && callback !== CallbackSet()) ||
                 (dense) || !isempty(saveat), # and no dense output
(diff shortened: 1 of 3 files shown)


src/solve.jl
```diff
@@ -413,10 +413,17 @@ function DiffEqBase.__init(prob::Union{DiffEqBase.AbstractODEProblem,
     differential_vars = prob isa DAEProblem ? prob.differential_vars : get_differential_vars(f, u)
 
     id = InterpolationData(f, timeseries, ts, ks, alg_choice, dense, cache, differential_vars, false)
-    sol = DiffEqBase.build_solution(prob, _alg, ts, timeseries,
-        dense = dense, k = ks, interp = id,
-        alg_choice = alg_choice,
-        calculate_error = false, stats = stats)
+    if _alg isa DAEAlgorithm
+        sol = DiffEqBase.build_solution(prob, _alg, ts, timeseries, Vector{typeof(du)}(undef,0),
```
ChrisRackauckas, line 417: we should gate it behind a new `save_du` argument, and allow for this in ODESolutions as well

oscardssmith: why? for ODESolution, du is an output rather than input of the function (and specifically, it will always be k1)

ChrisRackauckas: In both cases they are computed variables and not part of the state, so they act the same way computationally. Also, it would eliminate the need for this special casing because it would just act the same.

oscardssmith: > In both cases they are computed variables and not part of the state, so they act the same way

I see how this is true from the silver perspective, but not the user perspective. for the user, du of an ODE is the output, but for DAE it's an input.

ChrisRackauckas: Why does that matter for saving?

oscardssmith: well one reason that it makes a difference is that for all the ODEAlgorithms I've tried so far (Tsit5, Rodas5P, FBDF), du isn't a computed state at all. (e.g. `integ.du==nothing` after initializing.

ChrisRackauckas: It's always a computed state. You probably want to grab the interpolated version for the DAE solver too, so the saving code is the same.

oscardssmith: if it's always a computed state, why don't any of the ODESolvers compute it?

ChrisRackauckas: The ODE solvers do compute it.

oscardssmith:
```
julia> odeprob = ODEProblem((u,p,t)->u, zeros(2), (0.,1.));
julia> integrator = init(odeprob, Tsit5()); solve!(integrator); integrator.du==nothing
true
```
like technically for ODEs, this value will generally be `k1`, but I again state that that none of the integrators ever touch `integrator.du`

ChrisRackauckas: What does that have to do with saving?

oscardssmith: the field this PR saves is `integrator.du`, so it seems to matter to me that none of the current algorithms ever fill in a value for `integrator.du`

ChrisRackauckas:
Missing tests

ChrisRackauckas:
Bump for tests

PR #2141, opened 2024-02-14
Improve callback modifier handling

ChrisRackauckas:


src/integrators/integrator_interface.jl
```diff
@@ -51,6 +51,10 @@ function DiffEqBase.reeval_internals_due_to_modification!(integrator::ODEIntegra
     end
 
     integrator.u_modified = false
+
+    # This will reset fsal if any u_modified is true!
+    integrator.do_error_check = false
```
oscardssmith, line 56: this is wrong.

Issue #2147, opened 2024-02-23
Unexpected memory accumulation when repeatedly running `solve`


thanasibakis:
**Describe the bug 🐞**

Running `solve` repeatedly seems to cause a julia program to accumulate an unexpected amount of memory over time (as reported by `htop`). If a place a call to `solve` in a `while true` loop, I am able to watch the `MEM%` column of `htop` gradually grow.

I have reproduced this issue on julia versions 1.8, 1.9, and 1.10.

**Expected behavior**

I would expect the memory to be reused each iteration of the loop (or some nuanced version of that idea). I would expect to be able to run this infinite loop without eventually receiving an Out of Memory error from my operating system, on a machine with tens of gigabytes of RAM.

**Minimal Reproducible Example 👇**

Run with `julia --threads=8`. Multithreading is not necessary to reproduce, but accelerates the demonstration of the issue.

Adapted from https://docs.sciml.ai/DiffEqDocs/stable/getting_started/#ode_other_types

```julia
using DifferentialEquations, LinearAlgebra

f(du, u, A, t) = mul!(du, A, u)

function main()
    A = [ 1.0  0 0 -5
          4   -2 4 -3
         -4    0 0  1
          5   -2 2  3]

    Threads.@threads for i in 1:8
        while true
            soln = solve(
                ODEProblem{true}(
                    f,
                    rand(4, 2),
                    (1.0, 0.0),
                    A
                ),
                Tsit5();
                save_everystep = false,
                save_start = false,
                reltol = 1e-3,
                abstol = 1e-3
            )
        end
    end
end

main()
```

**Error & Stacktrace ⚠️**

None from julia. However, monitoring the `MEM%` column of `htop` for the julia process will show a gradually accumulating memory usage.

**Environment (please complete the following information):**

  - Output of `using Pkg; Pkg.status()`

```julia
Status `~/.julia/environments/v1.8/Project.toml`
⌅ [0c46a032] DifferentialEquations v7.10.0
  [37e2e46d] LinearAlgebra
```

  - Output of `using Pkg; Pkg.status(; mode = PKGMODE_MANIFEST)`

```julia
Status `~/.julia/environments/v1.8/Manifest.toml`
  [47edcb42] ADTypes v0.2.6
⌅ [79e6a3ab] Adapt v3.7.2
⌅ [ec485272] ArnoldiMethod v0.2.0
⌅ [4fba245c] ArrayInterface v7.5.1
[...]

PR #2148, opened 2024-02-24
fix memory leak on repeated `init(prob)`

oscardssmith:
this does not yet remove the leak.

edit: this does fix the memory leak if you get the order of your condition correct.

fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2147

I'm not sure if this is easily testable.


topolarity:
I don't think this is a proper fix - we should try to avoid this `gensym()` even when using progress

ChrisRackauckas:
We can just remove the unique id names by default. You guys were the ones that wanted it 😅

topolarity:
Yeah, I'm not sure why we wanted the unique identifier. @pepijndevos any problem if we just use a standard Symbol here?

pepijndevos:
The reason is that every simulation in an ensemble needs an unique I'd to get its own progress bar.

oscardssmith:
The problem is that making them symbols is that symbols are permalloced so they never get freed.

pepijndevos:
Could we move the gensym up in the assembly problem?
Can IDs be strings instead?

topolarity:
Do we really need the progress ID in the first place?

This is my first time looking at the progress setup, but it seems like if we reported the progress events as a delta, we wouldn't have to do `total_progress += new_progress[id] - old_progress[id]` (roughly speaking) and could just do `progress += delta`. Even better, wouldn't it be much more efficient to keep an atomic numerator and denominator, and have each simulation interact with those directly?

pepijndevos:
Sure I mainly did this because it didn't require a sizable refactoring across SciML

PR #2151, opened 2024-02-25
Allow for using the initializeprob with OverrideInit

ChrisRackauckas:
Uses the latest addition to SciMLBase so that if an ODEProblem or DAEProblem tags a nonlinear system for the initialization then it will use it.


src/initialize_dae.jl
```diff
@@ -54,19 +64,32 @@ end
...
         alg::DefaultInit, x::Val{false})
-    _initialize_dae!(integrator, prob,
-        BrownFullBasicInit(integrator.opts.abstol), x)
+    if SciMLBase.has_initializeprob(prob.f)
+        _initialize_dae!(integrator, prob,
+            OverrideInit(integrator.opts.abstol), x)
+    else
+        _initialize_dae!(integrator, prob,
+            BrownFullBasicInit(integrator.opts.abstol), x)
+    end
```
oscardssmith, line 84: can't these be combined?

ChrisRackauckas: maybe, since it wasn't already I assumed that it's potentially due to some ambiguity and with such a big change going on I don't want to touch that at the same time.

Project.toml
```diff
@@ -72,7 +72,7 @@ PrecompileTools = "1"
 Preferences = "1.3"
 RecursiveArrayTools = "2.36, 3"
 Reexport = "1.0"
-SciMLBase = "2.26"
+SciMLBase = "2.27"
```
ChrisRackauckas, line 75:
```suggestion
SciMLBase = "2.27.1"
```

src/initialize_dae.jl
```diff
@@ -124,16 +132,20 @@ function _initialize_dae!(integrator, prob::Union{ODEProblem, DAEProblem},
...
     initializeprob = prob.f.initializeprob
     isAD = alg_autodiff(integrator.alg) isa AutoForwardDiff
-    alg = default_nlsolve(alg.nlsolve, isinplace, initializeprob.u0, isAD)
-    sol = solve(initializeprob, alg)
-    @show "here!!!"
+    alg = default_nlsolve(alg.nlsolve, isinplace, initializeprob.u0, initializeprob, isAD)
+    nlsol = solve(initializeprob, alg)
     if isinplace === Val{true}()
-        integrator.u .= prob.f.initializeprobmap(sol.u)
+        integrator.u .= nlsol[Main.ModelingToolkit.unknowns(prob.f.sys)] # prob.f.initializeprobmap(sol)
```
oscardssmith, line 141: shouldn't this depend on SII rather than MTK directly?

Issue #2152, opened 2024-02-28
FBDF not taking any timesteps except at the start and end on RLC circuit


oxinabox:
**Describe the bug 🐞**

I have created a circuit in MSL representing a transmision line with an end point dampening and solved it with FBDF.
The output is just a single line segement, with just times recorded for start and end.
This is wrong as I know the circuit solves  (e.g with Rodas5P or IDA)  to give a sinusioid.

**Expected behavior**

Solution should be a siniusoid and should match other solves

**Minimal Reproducible Example 👇**

Sorry it is not yet minimized
```julia
using ModelingToolkitStandardLibrary.Electrical, ModelingToolkit, OrdinaryDiffEq
using ModelingToolkitStandardLibrary.Blocks: Constant, Sine

connections = Equation[]
add_conn!(x) = (push!(connections, x); x)  # to match spice we want to interleave connections and declarations
systems = ODESystem[]
add_system!(x) = (push!(systems, x); x)

@named ground = Ground()
add_system!(ground)

@named Src = Sine(amplitude=1, frequency=100e6)
add_system!(Src)
@named Vsin = Voltage()
add_system!(Vsin)
connect(Src.output, Vsin.V) |> add_conn!
connect(Vsin.n, ground.g) |> add_conn!

@named Rsrc = Resistor(R=250)
add_system!(Rsrc)
connect(Rsrc.n, Vsin.p) |> add_conn!

c_match = 1.5e-12
l_match = 400e-9
@named Lmatch = Inductor(L=l_match)
add_system!(Lmatch)
connect(Rsrc.p, Lmatch.n) |> add_conn!

@named Cmatch = Capacitor(C=c_match)
add_system!(Cmatch)
connect(Rsrc.p, Cmatch.p) |> add_conn!
connect(ground.g, Cmatch.n) |> add_conn!

function TransmissionLineSegment!(_vin, _ii)
    l=0.05e-9
    c=0.02e-12
    L1 = Inductor(L=l, name=Symbol(:L1_, _ii)) |> add_system!
    _vout = L1.p
    connect(_vin, L1.n) |> add_conn!

    C1 = Capacitor(C=c, name=Symbol(:C1_, _ii)) |> add_system!
    connect(C1.n, ground.g) |> add_conn!
    connect(C1.p, _vout) |> add_conn!

    return _vout
end

ii=0
prev = TransmissionLineSegment!(Lmatch.p,  (ii+=1))
[...]

PR #2153, opened 2024-02-28
Directly enforce algebraic condition correctness in ImplicitEuler

ChrisRackauckas:
Its extrapolation-based error estimator is 0 on algebraic variables, just like Rosenbrock23, so we add the algebraic condition evaluations to the equations just like Rosenbrock23


oxinabox:
Will this fix https://github.com/SciML/OrdinaryDiffEq.jl/issues/2152

ChrisRackauckas:
No, this is on ImplicitEuler

Issue #2159, opened 2024-03-19
New Runge–Kutta pairs of orders 9(8) for use in quadruple precision computations


ranocha:
Kovalnogov VN, Fedorov RV, Karpukhina TV, Simos TE, Tsitouras C. Runge–Kutta pairs of orders 9 (8) for use in quadruple precision computations. Numerical Algorithms. 2023 Aug 24:1-5.

https://doi.org/10.1007/s11075-023-01632-8

cwittens:
shouldn't this issue be closed now that https://github.com/SciML/OrdinaryDiffEq.jl/pull/2166 is merged?

PR #2166, opened 2024-04-02
Added Runge–Kutta pairs of orders 9(8) for use in quadruple precision computations

Theozeud:
## Checklist

## Additional context

I added the Runge-Kutta method from https://doi.org/10.1007/s11075-023-01632-8 . The method seems to work, but I have not yet succeeded in getting a test file to work with good results. I need to investigate this further. The nomenclature of the algorithm is not necessarily right and the functions are not in the right place, as I do not know the whole genealogy of the methods. 

#2159


test/algconvergence/ode_quadruple_precision_tests.jl
```diff
@@ -0,0 +1,59 @@
...
+    @test sim.𝒪est[:final]≈OrdinaryDiffEq.alg_order(alg) atol=testTol
+end
+for prob in test_problems_linear
+    sim = test_convergence(dts, prob, alg)
+    @test sim.𝒪est[:final]≈OrdinaryDiffEq.alg_order(alg) atol=testTol
+end
+for prob in test_problems_nonlinear
+    sim = test_convergence(dts, prob, alg)
+    @test sim.𝒪est[:final]≈OrdinaryDiffEq.alg_order(alg) atol=testTol
+end
```
ChrisRackauckas, line 59: Just add these tests to the same place as the other erks

Project.toml
```diff
@@ -9,6 +9,7 @@ Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
+DiffEqDevTools = "f3b72e0c-5b89-59e1-b016-84e28bfd966d"
```
ChrisRackauckas, line 12: don't add the dep

Project.toml
```diff
@@ -26,6 +27,7 @@ Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 MuladdMacro = "46d2c3a1-f734-5fdb-9937-b9b9aeba4221"
 NonlinearSolve = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
+ODEProblemLibrary = "fdc4e326-1af4-4b90-96e7-779fcce2daa5"
```
ChrisRackauckas, line 30:
```suggestion
```

ChrisRackauckas:
Show `plot(sim)` on the convergence simulation. It has a standard plot which can be quite helpful.

Theozeud:
> Show `plot(sim)` on the convergence simulation. It has a standard plot which can be quite helpful.

I get this for the ode du(t) = cos(t) with u0 = 0.0
image

ChrisRackauckas:
No you're just at the barrier of what Float64 can compute. Bump up dt a bit and see, or change to the BigFloat version of the problem.

Theozeud:
Ok I see. How can I change to the BigFloat version ?

ChrisRackauckas:
`prob_ode_bigfloat2Dlinear` is the one setup to do that. 

`tspan = (0.0, 10.0)` You still had time as `Float64`. Set that to `big.((0.0,10.0))`.

```
julia> eps(BigFloat)
1.727233711018888925077270372560079914223200072887256277004740694033718360632485e-77
```

if all is right it should get about 70 digits before it stalls. This looks pretty good but I would guess that you might have something in a later digit that's off.

ChrisRackauckas:
Use https://github.com/SciML/DiffEqDevTools.jl/blob/master/src/ode_tableaus.jl#L1-L9 to get the tableau out (make it BigFloat) and check the first and second order conditions accuracy, i.e. `sum(b) = 1` and `sum(a_ij)_j = c_i`

Theozeud:
Ok, thanks for the details. Using the ``` deduce_Butcher_tableau ``` function, I found that the conditions are not valid, with an error of the magnitude of the stall (1e-33 - 1e-34). I checked the coefficients of coefficients b and I don't make any mistake compared with those written in the paper or the links given therein.

Theozeud:
> if all is right it should get about 70 digits before it stalls. This looks pretty good but I would guess that you might have something in a later digit that's off.

Isn't it normal not to have better precision since the document gives a method for quadruple precision? If we want better precision, I guess we should recompute the coefficients with better precision.

test/algconvergence/ode_quadruple_precision_tests.jl
Theozeud: Which file(s) exactly ?

ChrisRackauckas:
> Isn't it normal not to have better precision since the paper gives a method for quadruple precision? If we want better precision, I guess we should recompute the coefficients with better precision.

Ahh, I saw that it was rational coefficients so I thought they rationalized to exact.

```julia
julia> using Quadmath

julia> eps(Float128)
1.92592994438723585305597794258492732e-34
```

This looks pretty spot on then!

Remove the lowest `dt`s where it's saturated to fix the convergence test.

ChrisRackauckas:
Given that these will be a bit more bespoke to run at higher precision, I think it's fine to keep them in a separate file. But you need to add a line in https://github.com/Theozeud/OrdinaryDiffEq.jl/blob/Quadruple_precision_RK/test/runtests.jl#L147 so they actually run.

oscardssmith:
would it be easy to tweak the coefficients so it's first order at least for BigFloat?

ChrisRackauckas:
You'd need to do it for all conditions. I think it's fine to do the paper as-is.

test/algconvergence/ode_quadruple_precision_tests.jl
```diff
@@ -0,0 +1,59 @@
...
+test_problem_ssp_inplace = ODEProblem(f_ssp_inplace, rand(3, 3), (0.0, 8.0))
+
+println("QPRK98")
+alg = QPRK98()
+for prob in test_problems_only_time
+    sim = test_convergence(dts, prob, alg)
+    @test sim.𝒪est[:final]≈OrdinaryDiffEq.alg_order(alg) atol=testTol
+end
+for prob in test_problems_linear
```
ChrisRackauckas, line 52: This is missing a regression test on the adaptivity. Solve with adaptivity and check the solution is sufficiently small. Check Vern9 as an example

Theozeud: Ok I will do that. But is there Vern9 example in test, I don't find it ?

ChrisRackauckas: https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/test/regression/ode_unrolled_comparison_tests.jl

Theozeud: Ok Thank you

Theozeud: So I have done some tests with adaptivity and it seems to work fine, but the (automatic) interpolation doesn't, but may be that is normal and another unrelated problem : 
```julia 
f = (u, p, t) -> cos(t)
prob_ode_sin = ODEProblem(
                    ODEFunction(f; analytic = (u0, p, t) -> sin(t)), 
                    Float128(0.0), 
                    (Float128(0.0), Float128(4.0)))

alg = QPRK98()
sol = solve(prob_ode_sin, alg, dt = 1 / 2^10, adaptive = true)

plot(sol)
plot!(LinRange(0,4,1000), sin.(LinRange(0,4,1000)), label = "exact solution")
scatter!(sol.t, sin.(sol.t), label = "Real points computed", mc = :yellow)
```
image

Theozeud: Another remark concerning tests is that the order of convergence is each time higer (from 10 to 11.5 instead of 9) for problems I chose, I don't know if this is normal

Theozeud: > Are you updating fsal first and last to be the derivative estimates at 0 and dt?

I don't implement fsal first and last, I will try to add this

ChrisRackauckas: > Another remark concerning tests is that the order of convergence is each time higer (from 10 to 11.5 instead of 9) for problems I chose, I don't know if this is normal

Yeah that's just super convergence. It's typical that extremely simple ODEs can hit higher orders since the next order conditions can just be zero "accidentally" if certain properties like df/du = 0.

Theozeud: > Are you updating fsal first and last to be the derivative estimates at 0 and dt?

Seems better now :

image

> Yeah that's just super convergence. It's typical that extremely simple ODEs can hit higher orders since the next order conditions can just be zero "accidentally" if certain properties like df/du = 0.

Ok yes I see, thanks for the precision

test/algconvergence/ode_quadruple_precision_tests.jl
```diff
@@ -0,0 +1,60 @@
...
+    sim = test_convergence(BigFloat.(dts), prob, alg)
+    sim.𝒪est[:final]
+    @test sim.𝒪est[:final]≈OrdinaryDiffEq.alg_order(alg)+1 atol=testTol
+end
+
+for prob in test_problems_nonlinear
+    sim = test_convergence(dts, prob, alg)
+    sim.𝒪est[:final]
+    @test sim.𝒪est[:final]≈OrdinaryDiffEq.alg_order(alg)+2.5 atol=testTol
+end
```
ChrisRackauckas, line 60: Regression test the adaptive time stepping

Theozeud: I have just added some test for adaptivity, is this like what you had in mind ?

ChrisRackauckas: save_everysteps true and check the number of steps is reasonable.

ChrisRackauckas:
Just run the formatter.

https://github.com/SciML/SciMLStyle?tab=readme-ov-file#juliaformatter

test/algconvergence/ode_quadruple_precision_tests.jl
Theozeud: Ok, done

Theozeud:
> Just run the formatter.
> 
> https://github.com/SciML/SciMLStyle?tab=readme-ov-file#juliaformatter

Done

ChrisRackauckas (approved):

test/algconvergence/ode_quadruple_precision_tests.jl
```diff
@@ -0,0 +1,75 @@
+using OrdinaryDiffEq, DiffEqDevTools, Test, Random
+import ODEProblemLibrary: prob_ode_linear, prob_ode_2Dlinear, prob_ode_bigfloat2Dlinear
+using Quadmath
+
+Random.seed!(100)
+
+dts = Float128.(1 .// 2 .^ (6:-1:2))
```
ChrisRackauckas, line 7: just use bigfloat to keep the deps down.

ChrisRackauckas:
Codecov is being annoying but diving in it looks like there aren't any regressions here, so done.

PR #2167, opened 2024-04-04
Support NonlinearSolve.jl for the ODE nonlinear solver

oscardssmith:
This is WIP, but it mostly works (for in place ODE only), (and is a lot slower since it's missing all the fancy optimizations still).

### DIFF
--- a/src/nlsolve/newton.jl
+++ b/src/nlsolve/newton.jl
@@ -29,11 +29,10 @@ function initialize!(nlsolver::NLSolver{<:NonlinearSolveAlg, false},
         integrator::DiffEqBase.DEIntegrator)
     @unpack uprev, t, p, dt, opts, f = integrator
     @unpack z, tmp, ztmp, γ, α, iter, cache, method, alg = nlsolver
-    @unpack ustep, tstep, k, invγdt = cache
-
     cache.invγdt = inv(dt * nlsolver.γ)
     cache.tstep = integrator.t + nlsolver.c * dt
 
+    @unpack ustep, tstep, k, invγdt = cache
     if DiffEqBase.has_stats(integrator)
         integrator.stats.nf += cache.cache.nf
         integrator.stats.nnonliniter += cache.cache.nsteps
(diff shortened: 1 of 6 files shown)


oscardssmith:
so the current status is that out of place solves now work (although something is going wrong since it is getting a lot of nonlinear solver convergence failures). In place is broken because `_compute_rhs!` isn't autodiff compatible, and no matter what I do, NonlinearSolve is trying to autodiff through it.

avik-pal:
> In place is broken because _compute_rhs! isn't autodiff compatible

How does the default newton in ordinarydiffeq handle this?

> and is a lot slower since it's missing all the fancy optimizations still

you probably need the `recompute_jacobian` option here to mimic the jacobian reuse behavior https://github.com/SciML/NonlinearSolve.jl/blob/0dfc1358be16478a29e8940b5abca896d30ec9ff/src/core/generalized_first_order.jl#L204-L205

oscardssmith:
with `recompute_jacobian=false` this actually is looking really good! I'm getting better stats than `NLNewton` (admitadly so far I've only been testing on `f(u,p,t)=u`

src/nlsolve/newton.jl
```diff
@@ -235,148 +324,143 @@ end
...
-    @unpack uprev, t, p, dt = integrator
-    @unpack tmp, ztmp, γ, α, cache = nlsolver
-    @unpack ustep, tstep, k, invγdt = cache
-    isdae = TF <: DAEFunction
-    b = if isdae
-        @inbounds @simd ivdep for i in eachindex(z)
-            ztmp[i] = (tmp[i] + α * z[i]) * invγdt
-        end
-        if isdefined(integrator.cache, :u₀)
+@inline function _compute_rhs!(tmp::Array, ztmp::Array, ustep::Array, α, tstep, k,
```
ChrisRackauckas, line 417: Do we actually want to inline these?

ChrisRackauckas:
It's missing tests?

src/nlsolve/newton.jl
oscardssmith: I'll remove them. They were just there because they used to be there. It is one where the function call cost is going to be way less than the linear solve, so I don't know why it was manually annotated in the first place.

ChrisRackauckas:
Match a bit of what NLAnderson has done https://github.com/search?q=repo%3ASciML%2FOrdinaryDiffEq.jl+NLAnderson&type=code

avik-pal:
Are we using the jacobian reuse here? or is that a plan for later

PR #2181, opened 2024-05-04
test(inplace_interpolation.jl): check for approx equality

characat0:
## Checklist

## Additional context

Check for approximate equality instead of exact equality.

```Julia
julia> versioninfo()
Julia Version 1.10.3
Commit 0b4590a5507 (2024-04-30 10:59 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 12 × Apple M2 Pro
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
Threads: 8 default, 0 interactive, 4 GC (on 8 virtual cores)
```


ChrisRackauckas:
Why is this needed? What test was this for?

characat0:
This test errored in my machine, I've added the specs and since another test in the same file also checks for approx equality I think this might have been what the author had intended

ChrisRackauckas:
I see. Yes looking at it, it can be dependent on how your CPU gets optimized since SIMD allows for reordering of floating point operations, and `(x + y) + z` is not equal to `x + (y + z)` in floating point, and this is the cause of the difference.

PR #2183, opened 2024-05-07
Optimistic first time convergence criterion

YingboMa:
CVODE estimates the convergence of the first iteration by reusing the
convergence rate of the previous nonlinear solver iteration.

Reference: https://github.com/LLNL/sundials/blob/2abd63bd6cbc354fb4861bba8e98d0b95d65e24a/src/cvodes/cvodes_nls.c#L325-L331

MWE: https://discourse.julialang.org/t/cvode-bdf-outperforms-julia-solvers-for-stiff-system-biology-model/113936

Master branch:
```julia
julia> sol = solve(oprob_jac, FBDF()).stats
SciMLBase.DEStats
Number of function 1 evaluations:                  311
Number of W matrix evaluations:                    31
Number of linear solves:                           222
Number of Jacobians created:                       2
Number of nonlinear solver iterations:             212
Number of nonlinear solver convergence failures:   0
Number of fixed-point solver iterations:                     0
Number of fixed-point solver convergence failures:           0
Number of accepted steps:                          84
Number of rejected steps:                          2
```

This PR:
```julia
julia> sol = solve(oprob_jac, FBDF()).stats
SciMLBase.DEStats
Number of function 1 evaluations:                  251
Number of W matrix evaluations:                    32
Number of linear solves:                           159
Number of Jacobians created:                       2
Number of nonlinear solver iterations:             149
Number of nonlinear solver convergence failures:   0
Number of fixed-point solver iterations:                     0
Number of fixed-point solver convergence failures:           0
Number of accepted steps:                          86
Number of rejected steps:                          3
```

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -750,7 +750,8 @@ end
         if J isa StaticArray &&
            integrator.alg isa
            Union{
-            Rosenbrock23, Rodas23W, Rodas3P, Rodas4, Rodas4P, Rodas4P2, Rodas5, Rodas5P, Rodas5Pe, Rodas5Pr}
+            Rosenbrock23, Rodas23W, Rodas3P, Rodas4, Rodas4P, Rodas4P2,
+            Rodas5, Rodas5P, Rodas5Pe, Rodas5Pr}
             W = W_transform ? J - mass_matrix * inv(dtgamma) :
                 dtgamma * J - mass_matrix
         else
@@ -775,7 +776,8 @@ end
                 W_full
             elseif len !== nothing &&
                    integrator.alg isa
-                   Union{Rosenbrock23, Rodas23W, Rodas3P, Rodas4, Rodas4P, Rodas4P2, Rodas5, Rodas5P, Rodas5Pe, Rodas5Pr}
+                   Union{Rosenbrock23, Rodas23W, Rodas3P, Rodas4, Rodas4P,
+                Rodas4P2, Rodas5, Rodas5P, Rodas5Pe, Rodas5Pr}
                 StaticWOperator(W_full)
             else
                 DiffEqBase.default_factorize(W_full)
@@ -923,7 +925,8 @@ function build_J_W(alg, u, uprev, p, t, dt, f::F, ::Type{uEltypeNoUnits},
             len = StaticArrayInterface.known_length(typeof(J))
             if len !== nothing &&
                alg isa
-               Union{Rosenbrock23, Rodas23W, Rodas3P, Rodas4, Rodas4P, Rodas4P2, Rodas5, Rodas5P, Rodas5Pe, Rodas5Pr}
+               Union{Rosenbrock23, Rodas23W, Rodas3P, Rodas4, Rodas4P,
+                Rodas4P2, Rodas5, Rodas5P, Rodas5Pe, Rodas5Pr}
                 StaticWOperator(J, false)
             else
                 ArrayInterface.lu_instance(J)
--- a/src/nlsolve/nlsolve.jl
+++ b/src/nlsolve/nlsolve.jl
@@ -61,7 +61,7 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator::DiffEqBase.DEIntegrato
 
         # check divergence (not in initial step)
         if iter > 1
-            θ = ndz / ndzprev
+            θ = nlsolver.prev_θ = max(0.3 * nlsolver.prev_θ, ndz / ndzprev)
 
             # When one Newton iteration basically does nothing, it's likely that we
             # are at the precision limit of floating point number. Thus, we just call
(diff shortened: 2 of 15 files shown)


ChrisRackauckas:
> CVODE estimates the convergence of the first iteration by reusing the convergence rate of the previous nonlinear solver iteration.

wild! I wonder how that effects SDIRK.

YingboMa:
Yeah, this should make SDIRK methods more competitive against BDF methods and Rosenbrock methods as well.

YingboMa:
θ
Here's the evolution of the convergence rate for this particular problem. I guess constant extrapolation isn't a bad guess to capture the plateaus.

YingboMa:
θ-t
Here's theta vs model time.

src/derivative_utils.jl
```diff
@@ -750,7 +750,8 @@ end
         if J isa StaticArray &&
            integrator.alg isa
            Union{
-            Rosenbrock23, Rodas23W, Rodas3P, Rodas4, Rodas4P, Rodas4P2, Rodas5, Rodas5P, Rodas5Pe, Rodas5Pr}
+            Rosenbrock23, Rodas23W, Rodas3P, Rodas4, Rodas4P, Rodas4P2,
```
oscardssmith, line 753: This really scares me. Is there not a supertype we could use here or something so we don't forget to add methods in the future?

YingboMa:
The test failures seem unrelated.

devmotion:
AFAICT the DelayDiffEq test failures are caused by this PR.

ChrisRackauckas:
> type FPSolver has no field prev_θ

Yes I'm very sure that's related to this PR.

ChrisRackauckas:
For reference, only Interface I is failing on master, and I hope to get to that this weekend.

src/nlsolve/nlsolve.jl
```diff
@@ -61,7 +61,7 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator::DiffEqBase.DEIntegrato
 
         # check divergence (not in initial step)
         if iter > 1
-            θ = ndz / ndzprev
+            θ = nlsolver.prev_θ = max(0.3 * nlsolver.prev_θ, ndz / ndzprev)
```
oscardssmith, line 64:
```suggestion
        # check divergence (using previous iterations divergance at the initial step)
        prev_θ = hasproperty(nlsolver, :prev_θ) ? nlsolver.prev_θ :  one(ndz)
        if iter > 1
            θ = prev_θ = max(0.3 * prev_θ, ndz / ndzprev)
```
This (combined with the fix on line 88) will fix the CI issue. The problem is that delay diffeq for better or worse (worse) uses this same code with it's own NLSolver struct. Presumably, that struct should get the same optimization, but I'm not fully sure the math works out in the delay case, and this will preserve the status quo for them.

src/nlsolve/nlsolve.jl
```diff
@@ -84,13 +84,15 @@ function nlsolve!(nlsolver::AbstractNLSolver, integrator::DiffEqBase.DEIntegrato
                 nlsolver.nfails += 1
                 break
             end
+        else
+            θ = min(one(nlsolver.prev_θ), nlsolver.prev_θ)
```
oscardssmith, line 88:
```suggestion
θ = min(one(prev_θ), prev_θ)
```

src/nlsolve/nlsolve.jl
ChrisRackauckas: But if you do this then `nlsolver.prev_θ` never updates? I think the right thing to do is just propagate this eveywhere.

oscardssmith: Is there a good reason why DellayDiffEq needs it's own solver at all? It does seem like the interface here might not be especially well defined...

devmotion: Maybe they could be unified a bit more. But DelayDiffEq is something quite special/different with its NLSolver, it uses it for fixed point iterations of the actual integration step which involve running an ODE integrator: https://github.com/SciML/DelayDiffEq.jl/blob/43201e7ac1169ab0c08bc3a41879e4445ce2c827/src/fpsolve/functional.jl#L96-L114

YingboMa:
Many of the regression tests might be too stringent.

ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/9004907852/job/24739004289?pr=2183#step:6:799

Some of the convergence tests are showing big red flags. I don't think it's just a tolerance thing.

ChrisRackauckas:
Is more logic for step rejections required? I.e. can you use the convergence rate of a rejected step?

oscardssmith:
The DelayDiffEq failures seem real (although the tests that are failing do seem pretty bad)

PR #2184, opened 2024-05-08
Redesign default ODE solver to be type-grounded and lazy

oscardssmith:
This is the version of https://github.com/SciML/OrdinaryDiffEq.jl/pull/2103 that doesn't end up initializing the integrator caches that are unused. The code is fairly janky, but it all works and is type stable.


ChrisRackauckas:
Seems like there was a rebase issue in the cache builds

oscardssmith:
I believe the rebase is now actually correct. Rebasing across the formatting changes is really annoying.

oscardssmith:
@ChrisRackauckas I think your merge was wrong. I'll fix it.

ChrisRackauckas:
Awesome, looks like tests pass sans codecov nonsense. Can you prepare the two big downstream changes as well, the fix to DelayDiffEq and also a change to DifferentialEquations.jl which uses this as the default?

oscardssmith:
I think I can make this work without requiring a DelayDiffEq fix (`current` just needs to initialize to 0). https://github.com/SciML/OrdinaryDiffEq.jl/pull/2185 has all the changes needed for this to be used by default. Want me to merge it in with this PR?

ChrisRackauckas:
Yes 1 PR, I'm a bit confused which one is the one to look at.

ChrisRackauckas:
Can you port something similar to https://github.com/SciML/DifferentialEquations.jl/blob/master/test/default_ode_alg_test.jl over to here so it's tested? Big feature to not have a test for 😅

src/solve.jl
```diff
@@ -532,6 +532,13 @@ function DiffEqBase.__init(
     integrator
 end
 
+function DiffEqBase.__init(prob::ODEProblem, ::Nothing, args...; kwargs...)
+    DiffEqBase.__init(prob, DefaultODEAlgorithm(), args...; kwargs...)
+end
+function DiffEqBase.__solve(prob::ODEProblem, ::Nothing, args...; kwargs...)
+    DiffEqBase.__solve(prob, DefaultODEAlgorithm(), args...; kwargs...)
+end
```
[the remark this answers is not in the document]
oscardssmith: we can delete some but not all of it. DifferentialEquations also deals with DAEs, SDEs etc.

oscardssmith:
Unfortunately the fully grounded approach makes this type of test impossible. I think the only thing we can test is that the number of  steps and f evals roughly match what we would expect. These tests have now been added.

ChrisRackauckas:
You can check the `alg_choice` which is tracked?

oscardssmith:
I think this now all works.

src/perform_step/composite_perform_step.jl
```diff
@@ -1,38 +1,61 @@
...
-@benchmark h2(Val{1})
-mean time:        1.585 ns (0.00% GC)
-@benchmark h3(1)
-mean time:        6.423 ns (0.00% GC)
-
-@generated function foo(tup::Tuple, num)
-  N = length(tup.parameters)
-  :(@nif $(N+1) i->(i == num) i->(tup[i]) i->error("unreachable"))
+function init_ith_default_cache(cache::DefaultCache, algs, i)
+    if i == 1
```
prbzrg, line 2: It can be reduced to
```julia
if 1 <= i <= 6
    s = Symbol("cache$i")
    if !isdefined(cache, s)
        setproperty!(cache, s, alg_cache(algs[i], cache.args...))
    end
end
```

oscardssmith: this code is very specifically written to be type stable, which this suggestion wouldn't be.

test/interface/default_solver_tests.jl
```diff
@@ -25,7 +25,7 @@ function rober(u, p, t)
...
 prob_rober = ODEProblem(rober, [1.0,0.0,0.0],(0.0,1e5),(0.04,3e7,1e4))
 sol = solve(prob_rober)
-rosensol = solve(prob_rober, Rosenbrock23())
-# test that default isn't much worse than Rosenbrock23 (we expect it to use Rosenbrock23 for this).
-@test sol.stats.naccept < rosensol.stats.naccept + 2
-@test sol.stats.nf < rosensol.stats.nf + 20
+rosensol = solve(prob_rober, AutoTsit5(Rosenbrock23()))
+# test that default has the same performance as AutoTsit5(Rosenbrock23()) (which we expect it to use for this).
+@test sol.stats.naccept == rosensol.stats.naccept
+@test sol.stats.nf == rosensol.stats.nf
```
ChrisRackauckas, line 31: Needs more extensive testing, i.e. different tolerances check the alg_choice, stiff check the alg_choice, etc.

oscardssmith: what do you mean by stiff check?

prbzrg:
I wonder if it would be beneficial to test the accuracy of the chosen algorithms by the analytic solution.
It can be found in 
https://github.com/SciML/DiffEqProblemLibrary.jl/blob/8d2cf0c543a734053a339cb1eed2e27fe512b04b/lib/ODEProblemLibrary/src/ode_linear_prob.jl#L1-L21

What if we use the problems that defined in `ODEProblemLibrary`?

oscardssmith:
we could, but I'm not sure that's the best test. figuring out appropriate error levels there is always a little ad hoc, and I'm this case, we know exactly what the solution we expect is (it should just be the same as if we pick the solver ourself)

test/interface/default_solver_tests.jl
```diff
@@ -0,0 +1,45 @@
...
+sol = solve(prob_rober)
+rosensol = solve(prob_rober, AutoTsit5(Rosenbrock23()))
+# test that default has the same performance as AutoTsit5(Rosenbrock23()) (which we expect it to use for this).
+@test sol.stats.naccept == rosensol.stats.naccept
+@test sol.stats.nf == rosensol.stats.nf
+@test unique(sol.alg_choice) == [1,3]
+@test sol.alg_choice[1] == 1
+@test sol.alg_choice[end] == 3
+
+sol = solve(prob_rober, reltol=1e-7, abstol=1e-7)
```
ChrisRackauckas, line 38: Use Bruss with different `N`s to test that the different size handling works correctly.

https://docs.sciml.ai/DiffEqDocs/stable/tutorials/advanced_ode_example/

ChrisRackauckas (approved):

src/solve.jl
```diff
@@ -533,10 +533,12 @@ function DiffEqBase.__init(
 end
 
 function DiffEqBase.__init(prob::ODEProblem, ::Nothing, args...; kwargs...)
-    DiffEqBase.__init(prob, DefaultODEAlgorithm(), args...; kwargs...)
+    alg = DiffEqBase.prepare_alg(DefaultODEAlgorithm(), prob.u0, prob.p, prob)
```
ChrisRackauckas, line 536: Why prepare the alg here? That will happen inside the DiffEqBase pipeline.

src/solve.jl
```diff
@@ -533,10 +533,12 @@ function DiffEqBase.__init(
...
+    DiffEqBase.__init(prob, alg, args...; kwargs...)
```
ChrisRackauckas, line 537: You're bypassing the standard pipeline?

src/solve.jl
oscardssmith: form what I can see of other packages that provide default algorithms, this is the standard way of doing so. This is pretty much copied from both `NonlinearSolve.jl` and `DifferentialEquations.jl`

ChrisRackauckas: No, it doesn't match this https://github.com/SciML/DifferentialEquations.jl/blob/master/src/default_solve.jl#L14

src/solve.jl
```diff
@@ -534,6 +532,15 @@ function DiffEqBase.__init(
     integrator
 end
 
+function DiffEqBase.__init(prob::ODEProblem, ::Nothing, args...; kwargs...)
+    alg = DiffEqBase.prepare_alg(DefaultODEAlgorithm(), prob.u0, prob.p, prob)
```
ChrisRackauckas, line 536: you don't need to (and shouldn't) prepare the alg here.

oscardssmith: ah, I was basing off of https://github.com/SciML/DifferentialEquations.jl/blob/ad7b74d096f2883c5686c48a11f4edd94eef644e/src/default_solve.jl#L8. I'll make sure the tests still pass with it deleted.

src/solve.jl
```diff
@@ -534,6 +532,15 @@ function DiffEqBase.__init(
...
+    DiffEqBase.init(prob, alg, args...; kwargs...)
```
ChrisRackauckas, line 537:
```suggestion
    DiffEqBase.init(prob, DefaultODEAlgorithm(autodiff=false), args...; kwargs...)
```

src/solve.jl
```diff
@@ -534,6 +532,15 @@ function DiffEqBase.__init(
...
+end
+function DiffEqBase.__solve(prob::ODEProblem, ::Nothing, args...; kwargs...)
+    alg = DiffEqBase.prepare_alg(DefaultODEAlgorithm(), prob.u0, prob.p, prob)
+    DiffEqBase.solve(prob, alg, args...; kwargs...)
```
ChrisRackauckas, line 541:
```suggestion
    DiffEqBase.solve(prob, DefaultODEAlgorithm(autodiff=false), args...; kwargs...)
```

oscardssmith:
I believe this is now ready to merge.

ChrisRackauckas:
Test fails

prbzrg:
Before merging this, I think the issues linked to https://github.com/SciML/OrdinaryDiffEq.jl/pull/2103 should be checked and specify if this PR can resolve them.

oscardssmith:
This I believe does all the things that PR requires. Specifically, it is type stable, and roughly the same for precompile. I'm not sure either version actually makes a significant precompile difference, since they both take us from precompiling 6 solvers to  a single solver that calls all the functions that those 6 solvers would have called.

ChrisRackauckas:
What's the test failure from?

oscardssmith:
The test failure was from me being wrong about which algorithm would be used for exrober at the last timestep. (I think due to the same issue that causes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2198). The test has been adjusted.

PR #2185, opened 2024-05-09
Use `DefaultODEAlgorithm` by default

oscardssmith:
This makes it so the default ODE solver is used by default once https://github.com/SciML/OrdinaryDiffEq.jl/pull/2184. Note that this looks a lot bigger than it should because it is a branch on top of https://github.com/oscardssmith/OrdinaryDiffEq.jl/tree/os/default_solver-v2.

Perhaps surprisingly, this works without making any `DifferentialEquations` or `DiffEqBase` changes.


oscardssmith:
Moved into https://github.com/SciML/OrdinaryDiffEq.jl/pull/2184

PR #2186, opened 2024-05-10
refactor: use tunable parameters for cache in dae initialization

sathvikbhagavan:
doing `length(p)` fails when `p` is MTKParameters object and gives warnings like:

```julia
┌ Warning: The supplied DiffCache was too small and was enlarged. This incurs allocations
│     on the first call to `get_tmp`. If few calls to `get_tmp` occur and optimal performance is essential,
│     consider changing 'N'/chunk size of this DiffCache to 12.
└ @ PreallocationTools ~/.julia/packages/PreallocationTools/zJxgA/src/PreallocationTools.jl:191
```

It should be fixed by using the length of the tunable portion.

will also fix https://github.com/SciML/ModelingToolkitNeuralNets.jl/issues/4


Project.toml
```diff
@@ -34,6 +34,7 @@ RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
 SciMLOperators = "c0aeaf25-5076-4817-a8d5-81caf7dfa961"
+SciMLStructures = "53ae85a6-f571-4167-b2af-e1d143709226"
```
ChrisRackauckas, line 37: Missing a compat

sathvikbhagavan: https://github.com/SciML/OrdinaryDiffEq.jl/pull/2191 can be merged for it?

PR #2187, opened 2024-05-10
Newmark beta method

### LINKED ISSUE #411: Numark-beta method
@ahojukka5 requested these for JuliaFEM. They are implicit methods for second order ODEs and shouldn't be very difficult. 

https://en.wikipedia.org/wiki/Newmark-beta_method
https://www.sharcnet.ca/Software/Ansys/17.0/en-us/help/wb_sim/ds_rbd_imp_gen_a.html
https://www.comsol.com/forum/thread/39778/generalized-alpha-or-bdf

termi-official:
Prototype to resolve #411 .

## Checklist


src/alg_utils.jl
```diff
@@ -495,6 +495,7 @@ alg_order(alg::ERKN4) = 4
 alg_order(alg::ERKN5) = 5
 alg_order(alg::ERKN7) = 7
 alg_order(alg::RKN4) = 4
+alg_order(alg::NewmarkBeta) = 1 # or at least 2, depending on the parameters.
```
ranocha, line 498: Since the coefficients are part of the algorithm, can you calculate the order here?

oscardssmith: similar to FBDF, you can just set this to 1 I think

termi-official: Thanks for taking a look Hendrik! Yes, I could compute them here. I will fix such stuff after I get a pendulum example working. I just opened the PR to discuss the progress and possibly better design choices for the new class of solvers.

termi-official:
The last remaining problem is a bit tricky. For a given DynamicalODEFunction Newmark needs to solve for a "future accelleration" by solving the problem $Δtₙ f₁(innertmp₂ + z β Δtₙ², innertmp₁ + z γ Δtₙ,t) = z$ for $z$ with given $innertmp, \alpha,\beta$.

The first issue I run into is passing down the analytical Jacobian:

```julia
using OrdinaryDiffEq
function f1_harmonic(dv, v, u, p, t)
    dv .= -u
end
function f2_harmonic(du, v, u, p, t)
    du .= v
end
harmonic_jac_f1(J, v, u, p, t) = J[1,1] = -1.0
ff_harmonic = DynamicalODEFunction(f1_harmonic, f2_harmonic; jac = harmonic_jac_f1)
DiffEqBase.has_jac(ff_harmonic) # false
DiffEqBase.has_jac(ff_harmonic.f1) # false
ff_f1 = ODEFunction{false}(ODEFunction{false}(f1_harmonic); jac=harmonic_jac_f1!)
ff_harmonic = DynamicalODEFunction(ff_f1, f2_harmonic; jac = harmonic_jac_f1)
DiffEqBase.has_jac(ff_harmonic) # false
DiffEqBase.has_jac(ff_harmonic.f1) # false 
```

The second problem is getting the AD on track. My current strategy is to pass a helper type `ArrayPartitionNLSolveHelper` around, which constructs an `ArrayPartition` when multiplied with a `Vector`. The idea is to provide an interface for inplace evaluation of `f.f1` (DynamicalODEFunction) which is compatible with the evaluation form required by the internal nonlinear solver, such that we can call 
`f₁(innertmp + z \alpha,t)  = z`, where $\alpha$ is the mentioned `ArrayPartitionNLSolveHelper`. Should we head down this road or is there possibly a better design?

src/nlsolve/utils.jl
```diff
@@ -132,27 +133,27 @@ function build_nlsolver(alg, u, uprev, p, t, dt, f::F, rate_prototype,
         ::Type{uEltypeNoUnits},
         ::Type{uBottomEltypeNoUnits},
         ::Type{tTypeNoUnits}, γ, c,
-        iip) where {F, uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
+        iip; tmp = zero(u)) where {F, uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
     build_nlsolver(alg, u, uprev, p, t, dt, f, rate_prototype, uEltypeNoUnits,
         uBottomEltypeNoUnits,
-        tTypeNoUnits, γ, c, 1, iip)
+        tTypeNoUnits, γ, c, 1, iip; tmp=tmp)
```
oscardssmith, line 139:
```suggestion
        tTypeNoUnits, γ, c, 1, iip; tmp)
```

src/perform_step/newmark_perform_step.jl
```diff
@@ -0,0 +1,81 @@
...
+    #   ũ(tₙ₊₁)' = u(tₙ)' + u(tₙ)'' (1.0 - γ) Δtₙ
+    # such that we can compute the solution with the correctors
+    #   u(tₙ₊₁)  = ũ(tₙ₊₁)  + u(tₙ₊₁)'' β Δtₙ²
+    #   u(tₙ₊₁)' = ũ(tₙ₊₁)' + u(tₙ₊₁)'' γ Δtₙ
+
+    mass_matrix = f.mass_matrix
+
+    # Evaluate predictor
+    dduprev = integrator.fsalfirst.x[1]
+    duprev, uprev = integrator.uprev.x
```
oscardssmith, line 29: do we expect there to be a difference between `integrator.fsalfirst.x[2]` and `integrator.uprev.x[1]` (i.e. is there a specific reason we expect `uprev.x[1]` to be a better `duprev` or are they basically the same?)

src/perform_step/newmark_perform_step.jl
```diff
@@ -0,0 +1,81 @@
...
+    upred_full = ArrayPartition(
+        duprev + dt*(1.0 - γ)*dduprev,
+        uprev  + dt*dt*(0.5 - β)*dduprev + dt*duprev
+    )
```
oscardssmith, line 33: this seems like it will allocate a bunch, right?

src/perform_step/newmark_perform_step.jl
```diff
@@ -0,0 +1,81 @@
...
+    #     z         - Δtₙ f₁(ũ(tₙ₊₁) +         z β Δtₙ, ũ(tₙ₊₁)' +         z γ,t) = 0
+    #                 Δtₙ f₁(ũ(tₙ₊₁) +         z β Δtₙ, ũ(tₙ₊₁)' +         z γ,t) = z
+    #   γ̂ = [γ, β Δtₙ]:
+    #                 Δtₙ f₁(ũ(tₙ₊₁) +         z γ̂₂    , ũ(tₙ₊₁)' +         z γ̂₁   ,t) = z
+    #   innertmp = [ũ(tₙ₊₁)', ũ(tₙ₊₁)]:
+    #                 Δtₙ f₁(innertmp₂ +         z β Δtₙ², innertmp₁ +         z γ Δtₙ,t) = z
+    # Note: innertmp = nlsolve.tmp
+    nlsolver.γ = ArrayPartitionNLSolveHelper(γ, β * dt) # = γ̂
+    # nlsolver.γ = ArrayPartitionNLSolveHelper(0.0, β * dt)
+    nlsolver.tmp .= upred_full # TODO check f tmp is potentially modified and if not elimiate the allocation of upred_full
```
oscardssmith, line 52: I believe `nlsolver.tmp` isn't modified, but I'm not 100% sure.

src/perform_step/newmark_perform_step.jl
```diff
@@ -0,0 +1,81 @@
...
+    )
+
+    # _tmp = mass_matrix * @.. broadcast=false (α₁ * uprev+α₂ * uprev2)
+    # nlsolver.tmp = @.. broadcast=false _tmp/(dt * β₀)
+
+    # nlsolve!(...) solves for
+    #   dt⋅f(innertmp + γ̂⋅z, p, t + c⋅dt) + outertmp = z
+    # So we rewrite the problem
+    #     u(tₙ₊₁)'' - f₁(ũ(tₙ₊₁) + u(tₙ₊₁)'' β Δtₙ², ũ(tₙ₊₁)' + u(tₙ₊₁)'' γ Δtₙ,t) = 0
+    #   z = Δtₙ u(tₙ₊₁)'':
```
oscardssmith, line 42: I think we generally put the `Δtₙ` as part of `γ` rather than `z`. Is there a reason not to do so?

oscardssmith (changes requested):

src/integrators/integrator_utils.jl
```diff
@@ -507,6 +507,7 @@ nlsolve_f(f, alg::DAEAlgorithm) = f
 function nlsolve_f(integrator::ODEIntegrator)
     nlsolve_f(integrator.f, unwrap_alg(integrator, true))
 end
+nlsolve_f(f::DynamicalODEFunction, alg::NewmarkBeta) = f.f1 # FIXME
```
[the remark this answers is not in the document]
termi-official: Yea, as you pointed out on Slack we should switch to `SecondOrderODEFunction` IIUC.

src/misc_utils.jl
```diff
@@ -197,3 +197,30 @@ function get_differential_vars(f, u)
...
+    ArrayPartition(γ.γ₁*z, γ.γ₂*z)
+end
+
+Base.:*(γ::ArrayPartitionNLSolveHelper{T}, scalar::T) where T = scalar*γ
+
+function Base.:*(scalar::T, γ::ArrayPartitionNLSolveHelper{T}) where T
+    ArrayPartitionNLSolveHelper(scalar*γ.γ₁, scalar*γ.γ₂)
+end
+
+Base.inv(γ::ArrayPartitionNLSolveHelper) = 1/(γ.γ₁+γ.γ₂)
```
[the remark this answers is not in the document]
termi-official: Because I am not sure how to integrate a pair of weights with the W transformation. This is basically a hotfix to make the example work at all.

If you take another look at the perform step function you will see what I am trying to achive. Basically the nonlinear solver interface assumes that $\gamma$ is a scalar, but for Newmark methods we have one gamme for the first part of the solution (block 1 in the associated ArrayPartition) and one gamma for block 2).

src/nlsolve/newton.jl
```diff
@@ -303,7 +303,7 @@ end
     # page 54.
     γdt = isdae ? α * invγdt : γ * dt
 
-    !(W_γdt ≈ γdt) && (rmul!(dz, 2 / (1 + γdt / W_γdt)))
+    # !(W_γdt ≈ γdt) && (rmul!(dz, 2 / (1 + γdt / W_γdt)))
```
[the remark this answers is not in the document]
termi-official: I could not get it to work with Newmark, because here we have a comparison between a scalar `W_γdt` and a pair `γdt`.

src/nlsolve/newton.jl
```diff
@@ -412,6 +423,21 @@ function _compute_rhs!(tmp, ztmp, ustep, γ, α, tstep, k,
     return _vec(ztmp), ustep
 end
 
+function _compute_rhs!(tmp, ztmp, ustep, γ::ArrayPartitionNLSolveHelper, α, tstep, k,
+    invγdt, method::MethodType, p, dt, f, z)
+    mass_matrix = f.mass_matrix
+    ustep = tmp + γ * z
```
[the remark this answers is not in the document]
termi-official: I need to double check what exactly lead to this change, but I think we do not need this anymore.

src/nlsolve/type.jl
```diff
@@ -157,7 +157,7 @@ mutable struct NLNewtonCache{
     firststage::Bool
     firstcall::Bool
     W_γdt::tType
-    du1::uType
+    du1::rateType
```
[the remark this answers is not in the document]
termi-official: This one triggered an error in the cache construction IIRC. The case where rateType is not directly compatible with uType is not covered by the CI (yet).

src/perform_step/newmark_perform_step.jl
```diff
@@ -0,0 +1,81 @@
+function initialize!(integrator, cache::NewmarkBetaCache)
+    duprev, uprev = integrator.uprev.x
+    integrator.f(cache.fsalfirst, integrator.uprev, integrator.p, integrator.t)
+    integrator.stats.nf += 1
+    integrator.fsalfirst = cache.fsalfirst
+    integrator.fsallast  = cache.fsalfirst
```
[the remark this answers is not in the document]
termi-official: If you ask like this, then I think I am understanding something wrong here. From looking at other FSAL methods I thought that we can just put the same memory into both locations.

src/perform_step/newmark_perform_step.jl
termi-official: Yes, `integrator.uprev.x[1]` contains the solution and `integrator.fsalfirst.x[1]` is the second time derivative of `integrator.uprev.x[1]`.

src/perform_step/newmark_perform_step.jl
termi-official: Indeed. As also pointed out below by you during review we might be able to use `nlsolver.tmp` to get rid of this.

src/perform_step/newmark_perform_step.jl
termi-official: Note that the $\gamma$ here is the one from the Newmark solver while $\hat{\gamma}$ is the one from the Newton solver.

src/perform_step/newmark_perform_step.jl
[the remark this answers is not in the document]
oscardssmith: never mind on this one then. I haven't read through other ones in a while, so this looked like a copy paste error, but if other methods do it also, then I think this is fine.

src/perform_step/newmark_perform_step.jl
oscardssmith: oh. that's awful.

src/integrators/integrator_utils.jl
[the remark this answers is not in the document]
termi-official: Okay the reason why I did this was that there is not `SecondOrderODEFunction`. Can we easily infer from the `DynamicalODEFunction` that the second function is only `f2(du,u,v,p,t) = v`?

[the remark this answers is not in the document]
oscardssmith: oh, that's weird. we have a `SecondOrderODEProblem` but no `SecondOrderODEFunction`.

src/nlsolve/type.jl
```diff
@@ -89,11 +89,11 @@ end
...
 abstract type AbstractNLSolver{algType, iip} end
 
-mutable struct NLSolver{algType, iip, uType, gamType, tmpType, tType,
+mutable struct NLSolver{algType, iip, uType, gamType, tmpType, tmp2Type, tType,
     C <: AbstractNLSolverCache} <: AbstractNLSolver{algType, iip}
     z::uType
-    tmp::uType # DIRK and multistep methods only use tmp
-    tmp2::tmpType # for GLM if neccssary
+    tmp::tmpType # DIRK and multistep methods only use tmp
+    tmp2::tmp2Type # for GLM if neccssary
```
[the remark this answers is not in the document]
termi-official: I think this should go into a different PR.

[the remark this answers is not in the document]
oscardssmith: agreed

lib/OrdinaryDiffEqNewmark/src/OrdinaryDiffEqNewmark.jl
termi-official, line 1: I will strip down the imports to the used ones after we have fixed the remaining parts.

lib/OrdinaryDiffEqNewmark/src/alg_utils.jl
```diff
@@ -0,0 +1,5 @@
+alg_extrapolates(alg::NewmarkBeta) = true
```
termi-official, line 1: I will recover the extrapolation once we have ironed out the remaining parts.

lib/OrdinaryDiffEqNewmark/src/newmark_perform_step.jl
```diff
@@ -0,0 +1,106 @@
...
+        integrator.stats.nf += 1
+        residual = M*(aₙ₊₁ - atmp)
+        # Compute jacobian
+        f.jac(J, vₙ₊₁, uₙ₊₁, (γ*dt, β*dt*dt), p, t)
+        # Solve for increment
+        Δaₙ₊₁ = (M-J) \ residual
+        aₙ₊₁ .-= Δaₙ₊₁ # Looks like I messed up the signs somewhere :')
+        increment_norm = integrator.opts.internalnorm(Δaₙ₊₁, t)
+        increment_norm < 1e-4 && break
+        i == 10 && error("Newton diverged. ||Δaₙ₊₁||=$increment_norm")
```
termi-official, line 82: @oscardssmith I think I need some help here regarding the integration with OrdinaryDiffEqNonlinearSolve.jl .

A critical missing piece is an interface for the evaluation of the Jacobian in a suitable form (especially via ad). For now I have just bypassed this with a custom jacobian function, e.g. for the harmonic oscillator in the test
```
    function harmonic_jac(J, v, u, weights, p, t)
        J[1,1] = weights[1] * (0.0) + weights[2] * (-1.0)
        J[1,2] = weights[1] * (0.0) + weights[2] * ( 0.0)
        J[2,2] = weights[1] * (0.0) + weights[2] * (-1.0)
        J[2,1] = weights[1] * (0.0) + weights[2] * ( 0.0)
    end
```
because many implicit second order ODE methods requires that J = Δtₙ²β  ∂fᵤ +  Δtₙγ  ∂fᵥ . See above for a "full" derivation.

lib/OrdinaryDiffEqNewmark/test/runtests.jl
```diff
@@ -0,0 +1,71 @@
...
+    end
+    function harmonic_jac(J, v, u, weights, p, t)
+        J[1,1] = weights[1] * (0.0) + weights[2] * (-1.0)
+        J[1,2] = weights[1] * (0.0) + weights[2] * ( 0.0)
+        J[2,2] = weights[1] * (0.0) + weights[2] * (-1.0)
+        J[2,1] = weights[1] * (0.0) + weights[2] * ( 0.0)
+    end
+    function f2_harmonic!(du, v, u, p, t)
+        du .= v
+    end
```
termi-official, line 18: What is the intended way to enforce this automatically? We do not have a `SecondOrderODEFunction` analogue to `DynamicalODEFunction`.

src/misc_utils.jl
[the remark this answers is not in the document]
termi-official: I abandoned this idea for now, as it needed too many hacks.

termi-official (commented):
Thanks for the feedback. I have updated the implementation to master and made the tests for linear problems (harmonic and damped oscillator) pass. However, I had to manually unroll the implementation. I left some inline comments here on Github on the code sections where I am stuck (and on which I will resume working after fixing the most pressing issues).

termi-official:
@oscardssmith this is now functional using NonlinearSolve.jl directly to circumvent the issues outlined above with using OrdinaryDiffEqNonlinearSolve.jl .

Edit: Also, we should decide on a name for this subpackage before merging.

lib/OrdinaryDiffEqNewmark/src/OrdinaryDiffEqNewmark.jl
oscardssmith: this should be done (or resolved if it already is)

src/OrdinaryDiffEq.jl
```diff
@@ -257,6 +257,9 @@ export OrdinaryDiffEqExponentialRK, LawsonEuler, NorsettEuler, ETD1, ETDRK2, ETD
        EPIRK4s3B,
        EPIRK5s3, EXPRB53s3, EPIRK5P1, EPIRK5P2, ETD2, Exprb32, Exprb43
 
+using OrdinaryDiffEqNewmark
+export NewmarkBeta
+
```
oscardssmith: we don't want this to be an OrdinaryDiffEq dep. We're about to remove most of the other sub-libraries.

lib/OrdinaryDiffEqNewmark/src/newmark_perform_step.jl
```diff
@@ -0,0 +1,118 @@
...
+    solve!(nlcache)
+    if nlcache.retcode != ReturnCode.Success
+        integrator.force_stepfail = true
+        return
+    end
+    aₙ₊₁ = nlcache.u
+
+    u = ArrayPartition(
+        vₙ + dt * ((1 - γ) * aₙ + γ * aₙ₊₁),
+        uₙ + dt * vₙ + dt^2 / 2 * ((1 - 2β) * aₙ + 2β * aₙ₊₁)
```
oscardssmith: ad some `@..` to reduce allocations?

lib/OrdinaryDiffEqNewmark/src/newmark_perform_step.jl
```diff
@@ -0,0 +1,118 @@
...
+
+    #
+    if integrator.opts.adaptive
+        if integrator.success_iter == 0
+            integrator.EEst = one(integrator.EEst)
+        else
+            # Zienkiewicz and Xie (1991) Eq. 21
+            δaₙ₊₁ = (integrator.fsallast.x[1] - aₙ₊₁)
+            integrator.EEst = dt * dt / 2 * (2 * β - 1 / 3) *
+                              integrator.opts.internalnorm(δaₙ₊₁, t)
```
oscardssmith:
```suggestion
            integrator.EEst = dt * dt * (β - 1 // 6) *
                              integrator.opts.internalnorm(δaₙ₊₁, t)
```

lib/OrdinaryDiffEqNewmark/src/newmark_perform_step.jl
```diff
@@ -0,0 +1,118 @@
...
+end
+
+@muladd function perform_step!(integrator, cache::NewmarkBetaCache, repeat_step = false)
+    @unpack t, dt, f, p = integrator
+    @unpack β, γ, nlcache = cache
+
+    M = f.mass_matrix
+
+    # Evaluate predictor
+    aₙ = integrator.fsalfirst.x[1]
```
oscardssmith: this needs a check for whether the `fsalfirst` is valid (e.g. if there is a u_modified callback, it's not). See other fsal methods for how this is done.

oscardssmith:
@ChrisRackauckas any ideas on naming? Denis doesn't love `OrdinaryDiffEqNewmark` since there could be more solvers of this rough form that aren't Nemark methods added in the future, but we don't have better ideas on naming.

ChrisRackauckas:
Generally Second Order multistep implicit methods?

lib/OrdinaryDiffEqNewmark/src/newmark_perform_step.jl
termi-official: So I assume `@muladd` does not enable reordering then?

oscardssmith: it might, but I prefer doing it the way that I want the first time

termi-official:
Typically these methods are single step methods. So `OrdinaryDiffEqSecondOrderImplicit` maybe and the BDF type methods go into a different package?

lib/OrdinaryDiffEqNewmark/src/newmark_perform_step.jl
termi-official: This catches fire. `DimensionMismatch: array could not be broadcast to match destination`

lib/OrdinaryDiffEqNonlinearSolve/src/nlsolve.jl
termi-official, line 1: I can couple this change out into a separate PR if necessary.

oscardssmith:
The annoying thing about `OrdinaryDiffEqSecondOrderImplicit` is that It's unclear from the name whether Second Order refers to the solver or the ODE. It's a shame that math peeps are so bad at naming things.

termi-official:
This is good to go for now (minus the package name). We are missing an interface to query the manual Jacobian in SciML for further optimizations.

termi-official:
Any suggestions or comments for the package name? I do not have a strong opinion here. I have changed it now to `OrdinaryDiffEqSecondOrderImplicit` but I can also just change it back to `OrdinaryDiffEqNewmark`.

lib/OrdinaryDiffEqNewmark/src/newmark_perform_step.jl
```diff
@@ -0,0 +1,119 @@
...
+    solve!(nlcache)
+    if nlcache.retcode != ReturnCode.Success
+        integrator.force_stepfail = true
+        return
+    end
+    aₙ₊₁ = nlcache.u
+
+    u .= ArrayPartition(
+        vₙ + dt * ((1 - γ) * aₙ + γ * aₙ₊₁),
+        uₙ + dt * vₙ + dt^2 / 2 * ((1 - 2β) * aₙ + 2β * aₙ₊₁),
```
ChrisRackauckas: this is going to allocate each step. it should `@.` equals write into the two parts.

lib/OrdinaryDiffEqNewmark/test/runtests.jl
ChrisRackauckas: There isn't at this time

lib/OrdinaryDiffEqNonlinearSolve/src/nlsolve.jl
ChrisRackauckas: this is fine

lib/OrdinaryDiffEqNewmark/src/newmark_perform_step.jl
termi-official: Good catch.

lib/OrdinaryDiffEqNewmark/src/algorithms.jl
```diff
@@ -32,7 +33,7 @@ end
 function NewmarkBeta(; β = 0.25, γ = 0.5, chunk_size = Val{0}(),
         autodiff = Val{true}(), standardtag = Val{true}(),
         concrete_jac = nothing, diff_type = Val{:forward},
-        nlsolve = NewtonRaphson())
+        nlsolve = NewtonRaphson(), thread = False())
```
oscardssmith: can we use `thread = Val(false)` (to match other thread arguments)?

lib/OrdinaryDiffEqNewmark/src/newmark_perform_step.jl
```diff
@@ -48,9 +43,9 @@ end
             integrator.EEst = one(integrator.EEst)
         else
             # Zienkiewicz and Xie (1991) Eq. 21
-            δaₙ₊₁ = (integrator.fsallast - aₙ₊₁)
+            @.. broadcast=false thread=thread atmp = (integrator.fsallast - aₙ₊₁)
```
oscardssmith: broadcast=false is the default

lib/OrdinaryDiffEqNewmark/Project.toml
```diff
@@ -0,0 +1,48 @@
+name = "OrdinaryDiffEqNewmark"
+uuid = "d204908a-63b9-11ef-18f5-cfec7123a93b"
+authors = ["Dennis Ogiermann <termi-official@users.noreply.github.com>"]
+version = "0.0.1"
```
termi-official, line 4:
```suggestion
version = "0.0.1"
```

What is the versioning policy for SciML repos?

.github/workflows/CI.yml
```diff
@@ -43,6 +43,7 @@ jobs:
           - OrdinaryDiffEqLinear
           - OrdinaryDiffEqLowOrderRK
           - OrdinaryDiffEqLowStorageRK
+          - OrdinaryDiffEqNewmark
```
termi-official, line 46: The CI does not pick up the correct OrdinaryDiffEqCore version for the new package. Is there an easy fix?

PR #2196, opened 2024-05-16
I think this fixes a few minor bugs with nonlinear convergence

oscardssmith:
This code was overly simplified when I was debugging before and wasn't undone before merging. I'm not fully sure what a good test-case is for this yet because you need a case where the difference in error metric really matters.


ChrisRackauckas:
Test failures

oscardssmith:
Tests passing.

PR #2202, opened 2024-05-20
Fix initialization size to not try/catch

ChrisRackauckas:


src/initialize_dae.jl
```diff
@@ -528,9 +528,10 @@ function _initialize_dae!(integrator, prob::ODEProblem,
     if isAD
         csize = count(algebraic_vars)
         if !(p isa SciMLBase.NullParameters) && typeof(_u) !== typeof(u)
-            try
+            if SciMLSturctures.isscimlstructure(p)
```
ChrisRackauckas, line 531:
```suggestion
            if SciMLStructures.isscimlstructure(p)
```

sathvikbhagavan:
I think you would have to import `isscimlstructure` in https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/OrdinaryDiffEq.jl#L111

PR #2209, opened 2024-05-22
fix: remake initialization problem during DAE initialization

AayushSabharwal:
Close https://github.com/SciML/ModelingToolkit.jl/issues/2717

Requires
- https://github.com/SciML/ModelingToolkit.jl/pull/2729

## Checklist

### DIFF
--- a/src/initialize_dae.jl
+++ b/src/initialize_dae.jl
@@ -134,7 +134,16 @@ end
 function _initialize_dae!(integrator, prob::Union{ODEProblem, DAEProblem},
         alg::OverrideInit, isinplace::Union{Val{true}, Val{false}})
     initializeprob = prob.f.initializeprob
-
+    if initializeprob.f.sys !== nothing && prob.f.sys !== nothing
+        initu0vars = variable_symbols(initializeprob)
+        initu0order = variable_index.((initializeprob,), initu0vars)
+        # Variable symbols are not guaranteed to be in order
+        invpermute!(initu0vars, initu0order)
+        initu0 = getu(prob.f.initializeprob, initu0vars)(prob)
+        initp = remake_buffer(initializeprob, parameter_values(initializeprob),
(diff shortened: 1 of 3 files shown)


src/initialize_dae.jl
```diff
@@ -134,7 +134,16 @@ end
...
-
+    if initializeprob.f.sys !== nothing && prob.f.sys !== nothing
+        initu0vars = variable_symbols(initializeprob)
+        initu0order = variable_index.((initializeprob,), initu0vars)
+        # Variable symbols are not guaranteed to be in order
+        invpermute!(initu0vars, initu0order)
+        initu0 = getu(prob.f.initializeprob, initu0vars)(prob)
+        initp = remake_buffer(initializeprob, parameter_values(initializeprob),
+            Dict(sym => getu(prob, sym)(prob) for sym in parameter_symbols(initializeprob)))
+        initializeprob = remake(initializeprob; u0 = initu0, p = initp)
```
ChrisRackauckas, line 145: I'm not sure this all makes sense. First of all, it shouldn't come from `prob` since you can have callback reinitialization, so it should come from `integrator.u`. But also, this is just very heavy and dynamic: can't we just add a `initializeprobsetu` in the ODEFunction which sets up the initalizeprob based on `integrator.u`, and then also have one for `p` that updates based on any `p` updates (and also this should update `t`, which is a psuedo-parameter of the initialization)

AayushSabharwal: That makes sense. Since this involves SciMLBase changes (changing ODEFunction and DAEFunction fields) I'll merge this PR with https://github.com/SciML/OrdinaryDiffEq.jl/pull/2228 since that depends on this PR and modifies those structs anyway.

AayushSabharwal:
Superseded by https://github.com/SciML/OrdinaryDiffEq.jl/pull/2228

PR #2217, opened 2024-05-23
initialize bitstype caches in `DefaultCache`

oscardssmith:
otherwise they were getting uninitialized memory. We also don't have to wory about the extra memory use because if it's a bitstype it will be fairly small (this only occurs for `ConstantCache`s).


ChrisRackauckas:
Merge and release when tests pass

oscardssmith:
I don't have merge rights.

PR #2221, opened 2024-05-24
reduce default precompilation

oscardssmith:
This is an attempt to work around https://discourse.julialang.org/t/modelingtoolkit-takes-forever-to-precompile-on-windows-11/114492/62 which appears to be caused by large sysimages. Also, precompiling the default method should get the most commonly used methods.


ChrisRackauckas:
Merge and release. Ask on discourse for people to update and test

PR #2224, opened 2024-05-24
Disable Default solver on windows

oscardssmith:
until the Julia bug is fixed


prbzrg:
It seems that the problem on Windows is fixed now.

PR #2225, opened 2024-05-24
fix callbacks with default algorithm

oscardssmith:
fixes https://github.com/SciML/DifferentialEquations.jl/issues/1035


oscardssmith:
tests pass.

Issue #2203, opened 2024-05-20
`Rosenbrock23` fails for weird mass matrices


oscardssmith:
```
using OrdinaryDiffEq, LinearAlgebra
function swaplinear(u, p, t)
    [u[2], u[1]].*p
end
swaplinearf = ODEFunction(swaplinear, mass_matrix=ones(2,2)-I(2))
prob_swaplinear = ODEProblem(swaplinearf, rand(2), (0., 1.) 1.01)
sol = solve(prob_swaplinear, Rosenbrock23())
```
produces
```
ERROR: MethodError: no method matching length(::OrdinaryDiffEq.DifferentialVarsUndefined)

Closest candidates are:
  length(::LibGit2.GitBlob)
   @ LibGit2 ~/.julia/juliaup/julia-1.10.3+0.x64.linux.gnu/share/julia/stdlib/v1.10/LibGit2/src/blob.jl:3
  length(::Base.EnvDict)
   @ Base env.jl:224
  length(::DataStructures.EnumerateAll)
   @ DataStructures ~/.julia/packages/DataStructures/95DJa/src/multi_dict.jl:96
  ...

Stacktrace:
  [1] _similar_shape(itr::OrdinaryDiffEq.DifferentialVarsUndefined, ::Base.HasLength)
    @ Base ./array.jl:710
  [2] _collect(cont::UnitRange{Int64}, itr::OrdinaryDiffEq.DifferentialVarsUndefined, ::Base.HasEltype, isz::Base.HasLength)
    @ Base ./array.jl:765
  [3] collect(itr::OrdinaryDiffEq.DifferentialVarsUndefined)
    @ Base ./array.jl:759
  [4] broadcastable(x::OrdinaryDiffEq.DifferentialVarsUndefined)
    @ Base.Broadcast ./broadcast.jl:743
  [5] broadcasted(::typeof(!), ::OrdinaryDiffEq.DifferentialVarsUndefined)
    @ Base.Broadcast ./broadcast.jl:1339
  [6] perform_step!(integrator::OrdinaryDiffEq.ODEIntegrator{…}, cache::OrdinaryDiffEq.Rosenbrock23ConstantCache{…}, repeat_step::Bool)
    @ OrdinaryDiffEq ~/.julia/dev/OrdinaryDiffEq/src/perform_step/rosenbrock_perform_step.jl:446
  [7] perform_step!
[...]

PR #2226, opened 2024-05-26
throw error for `Rosenbrock23` on non-diagonal mass matrices

oscardssmith:
fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2203


ChrisRackauckas (approved):

ChrisRackauckas:
This is needed to be accounted for in the default alg as well?

Issue #1029, opened 2020-02-08
Relaxation Runge-Kutta


ChrisRackauckas:
https://arxiv.org/abs/2001.04826

Seems like a good GSoC.

@ranocha

ranocha:
Thanks for pinging me, @ChrisRackauckas.

The paper above investigates the setting of Hamiltonian problems. Just for reference: The earlier papers
- https://arxiv.org/abs/1905.09847
- https://arxiv.org/abs/1905.09129

describe the general procedure (and contain the basic proofs) and other applications. We are still working on extending and generalizing these results. That's why I haven't made a PR yet. But it definitely seems to be worthwhile to implement such a relatively general approach as used for conservative/Hamiltonian systems.

If a GSoC student is interested in this, I am of course willing to help and assist in mentoring, if you like. Since he is interested in open source projects (and relaxation Runge-Kutta methods, of course), I'll also ping @ketch to keep him up to date.

ketch:
Yes, implementing those methods in JuliaDiffEq would be very nice.  From a software point of view, they require some unusual things that would break most APIs:

- one or two user-defined functions need to be passed in and used inside the time step algorithm
- the step length dt that is actually taken is not the step length requested at the beginning of the step

ranocha:
I think a good first step would be to implement relaxation methods for conservative systems (Hamiltonians, conserved entropies, ...) as callbacks, similar to https://docs.juliadiffeq.org/latest/features/callback_library/#Manifold-Conservation-and-Projection-1. Then, we can pass all necessary information to such a callback and can also update the current time of the integrator. This should be a rather general approach.

For dissipative systems, I don't see a good way to reuse the existing code, since we need such modifications as described by @ketch above.

Biswajitghosh98:
@ChrisRackauckas @ranocha I was working on a low storage SSPRK algorithm mentioned in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/493
I'll try to submit the pull request as soon as possible for the same ( https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/compare/master...Biswajitghosh98:SSPRK53_H ) 
Can I start working on it alongside, and look up possible solutions for the bottlenecks mentioned by @ketch ?

ranocha:
As mentioned above, it could be possible to start with an implementation of relaxation (for conservative systems at first) as callback, along the lines of
```julia
[...]

PR #2227, opened 2024-05-27
[WIP] Relaxation Runge-Kutta

Theozeud:
## Start of the implementation of relaxation Runge-Kutta methods (see #1029)

Instead of writing things with a callback as suggested in the issue, I prefer to propose, I think, a more suitable solution by decoupling perform_step! into several steps one of which will allow the user to do modifications on the computations of the method. I think the problem with the callback is when it occurs : in the loopfooter function where things that have already modified  have been registered as dtnew, t, tprev and others things related to tstops, which could be changed by the relaxation methods. 

For the moment, to test the new structure, I copy the Tsit5 method with another name Tsit5_for_relaxation. I will certainly change some parts of the code and its structure. I am not sure yet if the idea is good, but I think it is a good starting point.

#### From the relaxation method point of view :
- [x] Implement a copy of Tsit5 with the new structure explained above,
- [x] Test if the code run without the relaxation,
- [x] Implement a structure to perform a relaxation and look for (a) good package(s) and method(s) to do the needed optimization step,
- [x] Implement FSAL-R,
- [x] See if it is possible to implement easily R-FSAL,
- [x] Look for suitable problems for the relaxation method,
- [x] Make the relaxation method running,
- [x] Make the relaxation method working,
- [x] Make benchmark.

#### From an interface point of view :
- [ ] See which variables are really needed to be added to the structure integrator,
- [x] See if the splitting in three steps of the perform_step! function is a good choice or if it is better to do only two,
- [x] Investigate all things needed to meticulously take care of time as matching tstops ect,
- [x] Find out what need to be changed in DiffEqBase.jl if we want to add a solver option (because it looks like something needs to be changed)


ChrisRackauckas:
Alerting @ranocha who may be interested in keeping on top of this.

Theozeud:
Two comments on the current implementation :
-  The new structure seems to work at least without any modification of the calculation by a user.
-   I divided perform_step ! into three steps:
     - computation!
     - modify_step !
     - finalize_step !
    
     The first is to do the mathematical side of the method, the second is to allow the user to make modifications to the calculation and the solver to track it, and the last is to do all the work related to the DiffEq solver.
But it is not easy to divide the tasks between computation ! and finalize_step ! In particular, we need to reload some objects to do this, so I am not sure it is really necessary, except perhaps to recalculate fsal or things like that if u changes.

ranocha (commented):
Interesting development. If you want to achieve best performance with step size control, you can have a look at our recent preprint https://arxiv.org/abs/2311.14050

Feel free to ask me if you have any specific questions while working on this

JoshuaLampert:
Regarding the step
>  Implement a structure to perform a relaxation and look for a good package/method to do the needed optimization step

in your TODO list above, I have an implementation of the relaxation step (there it is implemented as a callback) in https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/a7eb1307169d1278f0c4e1afeae55b3e9a8cea64/src/callbacks_step/relaxation.jl, which might or might not help you.

Theozeud:
> Interesting development. If you want to achieve best performance with step size control, you can have a look at our recent preprint https://arxiv.org/abs/2311.14050
> 
> Feel free to ask me if you have any specific questions while working on this

> Regarding the step
> 
> > Implement a structure to perform a relaxation and look for a good package/method to do the needed optimization step
> 
> in your TODO list above, I have an implementation of the relaxation step (there it is implemented as a callback) in https://github.com/JoshuaLampert/DispersiveShallowWater.jl/blob/a7eb1307169d1278f0c4e1afeae55b3e9a8cea64/src/callbacks_step/relaxation.jl, which might or might not help you.

Ok, thanks to both of you, I will take a look!

Theozeud:
Update on my work : 
- The new structure with relaxation runs and works partially. I need to fix things related to tstops matching.
- I am reading the article mentionned by Hendrik Ranochahttps://arxiv.org/abs/2311.14050 about better step size control.

Theozeud:
@ChrisRackauckas, @ranocha  I have a small question if someone can help me. I compare my implementation Tsit5_for_relaxtion with the existing Tsit5 on a simple problem  without relaxation such that results must be the same. I encounter a small problem of interpolation which could be due to my code, but  I wonder if it could be due to the fact I do not define _ode_interpolant for Tsit5_for_relaxation.

image

oscardssmith:
>  I wonder if it could be due to the fact I do not define _ode_interpolant for Tsit5_for_relaxation.

that is the reason.

src/OrdinaryDiffEq.jl
```diff
@@ -77,6 +77,8 @@ import DiffEqBase: resize!, deleteat!, addat!, full_cache, user_cache, u_cache,
                    add_saveat!, set_reltol!,
                    set_abstol!, postamble!, last_step_failed,
                    isautodifferentiable
+                   
+export change_u!, change_dt!, apriori_bounds_dt
```
oscardssmith, line 81: do these need to be exported?

oscardssmith:
The fact that this requires ~1000 lines to add relaxation to a single algorithm seems like it should be improved. It seems like we should be able to duplicate less of the steps between the regular solvers and the relaxation ones. Currently, it seems like the chance of us making improvements to the regular and those changes not getting reflected in the relaxation solver is pretty high.

Theozeud:
> The fact that this requires ~1000 lines to add relaxation to a single algorithm seems like it should be improved. It seems like we should be able to duplicate less of the steps between the regular solvers and the relaxation ones. Currently, it seems like the chance of us making improvements to the regular and those changes not getting reflected in the relaxation solver is pretty high.

Thanks for your review. In fact, it does not require as many lines at all. I have replicated the Tsit5 solver as Tsit5_for_relaxation so as not to "break" the initial solver and make comparisons. But what I had in mind would be to modify the perform_step! function for all the methods for which we can do relaxations, into a general perform_step! function for all the methods. Instead, each method would have a computation! function and other functions such as the calculation of error estimates. I am aware that this is a big change to the general solver, but I can't see any other way of including relaxation before the loopfooter! function. (So we wouldn't be able to use all the variants of relaxation, including the R-FSAL method). The other solution would be to add a condtion like  "if relaxation do relaxation end"
inside each method, with a lot of copy and paste, but this does not seem viable for me in the long term, for example if there are other different methods than relaxations.  I therefore propose this structure, which of course may not be desirable for reasons I am unaware of due to lack of experience, so I wanted to discuss it with you before going any further.

In the 1000 lines, there are several test files that I have made so that you can test them and see that the code works with relaxation, as well as a jupiter notebook that I should remove that shows that the code is not slowed down by the new perform_step! structure. For the tests, I have taken the examples I found in the https://arxiv.org/abs/2311.14050 article.

oscardssmith:
Closing in favor of https://github.com/SciML/OrdinaryDiffEq.jl/pull/2283

PR #2233, opened 2024-05-30
Optimistic first time convergence criterion (without messing with DelayDiffEq)

oscardssmith:
basically just https://github.com/SciML/OrdinaryDiffEq.jl/pull/2183 but made to not use the prev_theta value if it doesn't exist in the integrator. I made a new PR rather than just pushing to yingbo's because I don't want to break his history if I end up needing a rebase.

### DIFF
--- a/src/integrators/controllers.jl
+++ b/src/integrators/controllers.jl
@@ -16,6 +16,12 @@ end
 
 @inline function step_reject_controller!(integrator, alg)
     step_reject_controller!(integrator, integrator.opts.controller, alg)
+    cache = integrator.cache
+    if hasfield(typeof(cache), :nlsolve)
+        nlsolve = cache.nlsolve
+        nlsolve.prev_θ = one(nlsolve.prev_θ)
+    end
+    return nothing
 end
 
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/9310615565/job/25628330734?pr=2233#step:6:1081

Make this just a normal broadcast? And I think the other test failure is just a tolerance thing?

oscardssmith:
> Make this just a normal broadcast?

isn't that a totally unrelated fastbroadcast bug? This code doesn't change anything about how the bdf methods work...

ChrisRackauckas:
But you need the BDF tests running to have this PR

oscardssmith:
re-tunning ci now that it is fixed, but assuming ci is green, this should be good to merge.

ChrisRackauckas:
InterfaceV failure seems real.

oscardssmith:
The InterfaceV failure was in fact real. The problem was that if the nonlinear solve was accepted after a single iteration, but the step was rejected, we weren't resetting prev_θ leading to an infinite loop of accepting the nlsolve after a single iteration, but rejecting the integrator step.

PR #2235, opened 2024-06-03
Check master

ChrisRackauckas:


ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/9351844044/job/25738579832?pr=2235#step:6:1024 @oscardssmith what did this one come from?

oscardssmith:
not sure. I think the SciMLBase instability changes might be related since an earlier version of that pr caused it to fall, but I got it to pass before. it is a problem that is right on the edge of sol ability though. it required taking a step of floating point solution once or twice.

ChrisRackauckas:
Can you do a day of greening?

PR #2236, opened 2024-06-04
Partial revert for ROW docs and fix of tableau docstrings

termi-official:
xref https://github.com/SciML/OrdinaryDiffEq.jl/commit/f8f8307cf3717ae985c6ef165ef2bc8d60947bfc#r142723908 for the discussion and thanks for pointing this out @cwittens . Can you double check that we get it right this time?


cwittens:
If I am not mistaken the W in W-methods doesnt not stand for Wanner.
As far as I see it, you can have Rosenbrock-Wanner-methods or short Rosenbrock-methods see german [Wiki](https://de.wikipedia.org/wiki/Rosenbrock-Wanner-Verfahren).
And special kinds of  Rosenbrock-Wanner-methods (or short Rosenbrock-methods) are called W-methods, after Wolfbrandt, if they fullfill another condition (one can use an approximation of the Jacobian without order loss). See the Introduction of this [Paper](https://www.sciencedirect.com/science/article/pii/S0377042709006396).

So all the methods are ROW (Rosenbrock-Wanner), but for example ROS2S is also a W-method, but ROS2 isnt.

@ranocha can you maybe confirm / deny my explanation.

oscardssmith:
oh, that's awful. They really should be called ROWW or something to avoid confusion.

ChrisRackauckas:
Wolfbrandt and Wanner should just have a cage match for naming rights.

termi-official:
> If I am not mistaken the W in W-methods doesnt not stand for Wanner. As far as I see it, you can have Rosenbrock-Wanner-methods or short Rosenbrock-methods see german [Wiki](https://de.wikipedia.org/wiki/Rosenbrock-Wanner-Verfahren).

Wait, now I am a confused. I thought he difference between Rosenbrock and Rosenbrock-Wanner methods is that the latter reuses the same linear system in all stages while the former builds a new linear system for each stage. The interpretation of these methods is as in doing a single Newton vs a single Quasi-Newton step and taking the result as the solution approximation.

> And special kinds of Rosenbrock-Wanner-methods (or short Rosenbrock-methods) are called W-methods, after Wolfbrandt, if they fullfill another condition (one can use an approximation of the Jacobian without order loss). See the Introduction of this [Paper](https://www.sciencedirect.com/science/article/pii/S0377042709006396).
> 
> So all the methods are ROW (Rosenbrock-Wanner), but for example ROS2S is also a W-method, but ROS2 isnt.
I see. Thanks for pointing this out.

ChrisRackauckas:
> Wait, now I am a confused. I thought he difference between Rosenbrock and Rosenbrock-Wanner methods is that the latter reuses the same linear system in all stages while the former builds a new linear system for each stage. The interpretation of these methods is as in doing a single Newton vs a single Quasi-Newton step and taking the result as the solution approximation.

Yes. That's the technical difference. I never realized it didn't stand for Wanner though. What a phony.

termi-official:
So, next round of fixes I guess.

`rosenbrock_wanner_docstring` -> `rosenbrock_wolfbrandt_docstring`

and

https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/doc_utils.jl#L135 -> `"Rosenbrock-W(olfbrandt) Method. "`

?


> oh, that's awful. They really should be called ROWW or something to avoid confusion.

Seems like Collin already started :) https://github.com/SciML/OrdinaryDiffEq.jl/blob/762a742762b8238a045b2da54a5e599a0a9194ee/src/generic_rosenbrock.jl#L1267

ranocha:
As far as I can tell, @cwittens is right. If you want full clarity, you could use

`rosenbrock_wanner_docstring` -> `rosenbrock_wanner_wolfbrandt_docstring` or  `rosenbrock_wanner_wmethod_docstring`

or something like that.

PR #2243, opened 2024-06-10
work around FastBroadcast bugfix

oscardssmith:
hopefully this fixes CI. tests currently in progress locally.
## Checklist

### DIFF
--- a/src/perform_step/bdf_perform_step.jl
+++ b/src/perform_step/bdf_perform_step.jl
@@ -890,8 +890,9 @@ function perform_step!(integrator, cache::QNDFCache{max_order},
         u₀[i] = s + uprev[i]
     end
     @.. broadcast=false ϕ=zero(u)
+    vecϕ = _vec(ϕ)
     for i in 1:k
-        @views @.. ϕ+=γₖ[i] * D[:, i]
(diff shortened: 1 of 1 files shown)


oscardssmith:
oh, things are far more broken than we thought. The problem seems to be real and the fact that it broke when FastBroadcast changed seems to be entirely coincidental????

oscardssmith:
So the problem is that `D[:,i]` is a length 8 vector, while u is a 4x2 matrix. So presumably at some time D changed from being matrix like to vector like but only in the in place version??

oscardssmith:
Oh, this is very funny. version 0.3.0 of Fastbroadcast fixed a bug that this code was relying on. Previously it was allowing broadcasts between 4x2 matrices and length 8 vectors even though that isn't supposed to work.

oscardssmith:
well this was very dumb. I think it should be fixed now.

ChrisRackauckas (approved):

oscardssmith:
Let's see if https://github.com/YingboMa/FastBroadcast.jl/pull/62 fixed this

oscardssmith:
Local testing shows it did, but I closed and opened before the registry was actually updated.

Issue #2242, opened 2024-06-08
`ERROR: UndefVarError: a71 not defined` when using `Rodas5P` with `ContinuousCallback`


vancleve:
Stacktrace starts here:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/7f0eb18c635fdbe7972f1ddb1ce4dab3bd28d79e/src/dense/stiff_addsteps.jl#L1173

Minimish working example
```julia
using ComponentArrays
using Parameters: @unpack
using DifferentialEquations

default_params = ComponentVector(
    rb = 1.0,
    μB = 1.0,
    βgi = 0.0025,
    βvi = 0.0025,
    γ = 1.5,
    σ = 1.5,
    ϵ = 0.1,
    Φ = 0.5,
    ω_D = 0.001,
    ωI = 0.0001,
    θ = 0.03,
    D_threshold = 0.85,
    ρ = 1,
    η = -0.005,
    Bx = 1000,
    s = 0.5,
    xe = 0.5,
    Kb = 50000,
    KI = 1000
)

default_u0 = ComponentVector(
    B_g = 20,
    B_v = 1,
    V_D = 1,
    V_I = 1,
    I = 1,
    D = 0.001
)

function virulence_ode!(du, u, p, t)
    @unpack B_g, B_v, V_D, V_I, I, D = u
    @unpack rb, μB, βgi, βvi, γ, σ, ϵ, Φ, ω_D, ωI, θ, ρ, η, Bx, s, xe, Kb, KI = p
    
    du.B_g = rb * B_g * ( 1 - ((B_g + B_v) / Kb)) - (βgi * I * B_g) - μB * B_g * (1 / (1 + exp(η * (B_g - Bx)))) ## dB_g/dt
    du.B_v = rb * exp(-ρ * γ) * B_v * ( 1 - ((B_g + B_v) / Kb)) - (βvi * I * B_v) + μB * B_g * (1 / (1 + exp(η * (B_g - Bx)))) ## dB_v / dt
    du.V_D = (1-s)*γ*B_v - ϵ* V_D
    du.V_I = s*γ*B_v - ϵ*V_I - V_I*I
    du.I = Φ * (B_g + B_v) * (1-(I/KI)) - (σ*V_I * I) - (xe * βgi*I*B_g) - (xe * βvi*I*B_v)  ## dI / dt
[...]

PR #2244, opened 2024-06-10
fix `Rodas5P` dense output for non `Array` inplace

oscardssmith:
fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2242
I really don't like the fact that we have custom methods for `Array` since they dramatically increase the risk of untested typos in the case of inplace with non-array types. We really should be testing to see if we can delete these cases.


oscardssmith:
I'm very unsure whether this second commit is good or not, but it removes the specialized `<:Array` implimentations for `_ode_addsteps!` of Rosenbrock matrices on the theory that most of the time should be in the linearsolve, and the broadcasted versions should be fast anyway. This significantly reduces the risk of future bugs of this type by making it so the in place array code and in place non-array code are identical.

oscardssmith:
if the second commit here is ok, this is good to merge.

PR #2245, opened 2024-06-11
fix default solver interpolation

oscardssmith:
Fixes issue observed by @TorkelE where the default solver wasn't interpolating properly.

### DIFF
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -601,7 +601,7 @@ function ode_interpolation!(vals, tvals, id::I, idxs, deriv::D, p,
             cache_i₊ = cache.cache2
         elseif current_alg == 3
             cache_i₊ = cache.cache3
-        elseif cerrent_alg == 4
+        elseif current_alg == 4
             cache_i₊ = cache.cache4
         elseif current_alg == 5
             cache_i₊ = cache.cache5
--- a/src/dense/generic_dense.jl
+++ b/src/dense/generic_dense.jl
@@ -593,21 +593,6 @@ function ode_interpolation!(vals, tvals, id::I, idxs, deriv::D, p,
     if cache isa CompositeCache
         current_alg = id.alg_choice[i₊]
         cache_i₊ = cache.caches[current_alg]
-    elseif cache isa DefaultCache
-        current_alg = id.alg_choice[i₊]
-        if current_alg == 1
-            cache_i₊ = cache.cache1
-        elseif current_alg == 2
-            cache_i₊ = cache.cache2
-        elseif current_alg == 3
-            cache_i₊ = cache.cache3
-        elseif current_alg == 4
-            cache_i₊ = cache.cache4
-        elseif current_alg == 5
-            cache_i₊ = cache.cache5
-        elseif current_alg == 6
-            cache_i₊ = cache.cache6
-        end
     else
         cache_i₊ = cache
     end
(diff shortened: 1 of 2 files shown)


src/dense/generic_dense.jl
```diff
@@ -459,25 +649,32 @@ function ode_interpolation!(vals, tvals, id::I, idxs, deriv::D, p,
...
+                    cache_i₊ = cache.cache2
+                elseif current_alg == 3
+                    cache_i₊ = cache.cache3
+                elseif cerrent_alg == 4
+                    cache_i₊ = cache.cache4
+                elseif current_alg == 5
+                    cache_i₊ = cache.cache5
+                elseif current_alg == 6
+                    cache_i₊ = cache.cache6
+                end
```
ChrisRackauckas, line 667: Is this actually used anywhere?

src/dense/generic_dense.jl
```diff
@@ -415,9 +590,24 @@ function ode_interpolation!(vals, tvals, id::I, idxs, deriv::D, p,
...
         cache_i₊ = cache.caches[current_alg]
+    elseif cache isa DefaultCache
+        current_alg = id.alg_choice[i₊]
+        if current_alg == 1
+            cache_i₊ = cache.cache1
+        elseif current_alg == 2
+            cache_i₊ = cache.cache2
+        elseif current_alg == 3
+            cache_i₊ = cache.cache3
+        elseif cerrent_alg == 4
```
ranocha, line 604:
```suggestion
        elseif current_alg == 4
```

src/dense/generic_dense.jl
```diff
@@ -459,25 +649,32 @@ function ode_interpolation!(vals, tvals, id::I, idxs, deriv::D, p,
...
+        elseif cache isa DefaultCache
+            if current_alg != id.alg_choice[i₊] # switched algorithm
+                current_alg = id.alg_choice[i₊]
+                if current_alg == 1
+                    cache_i₊ = cache.cache1
+                elseif current_alg == 2
+                    cache_i₊ = cache.cache2
+                elseif current_alg == 3
+                    cache_i₊ = cache.cache3
+                elseif cerrent_alg == 4
```
ranocha, line 661:
```suggestion
                elseif current_alg == 4
```

src/dense/generic_dense.jl
oscardssmith: yeah. It's used further down in the function. Note that I removed the `else` so that it now runs for all cache types.

ChrisRackauckas: Don't you need to use it in the same part of the for loop to avoid the dynamism?

oscardssmith: I guess that would be better. I was following the design for Composite, but I guess that doesn't make as much sense for default.

ChrisRackauckas (approved):

ChrisRackauckas:
Merge when tests pass.

tomaklutfu (commented):
`C` is unused here. Did you intend to do `cache::Tuple{C}`?

ChrisRackauckas:
Oh this right here? 

https://github.com/SciML/OrdinaryDiffEq.jl/pull/2245/files/c5e292415127e344d7084acd1cee86001e05bd36#diff-880197818e0d2bb2dab8f58cc3be7df428ac18e981932f5de86923a298405d66R496

Yeah that looks like a typo

oscardssmith:
good catch. Not a problem, but unintended and easy to fix.https://github.com/SciML/OrdinaryDiffEq.jl/pull/2248

PR #2247, opened 2024-06-12
update `dolinsolve` to new `precs` interface.

oscardssmith:
No one was passing these arguments as far as I can tell, which meant that if the user provided a `precs` to the solver, it would get `nothing` for `du, u, p, t`.


oscardssmith:
This will wait until https://github.com/SciML/LinearSolve.jl/pull/514 is done (at which point we can remove `precs` from `OrdinaryDiffEq` entirely since it will just be a `linsolve` argument)

PR #2251, opened 2024-06-14
fix `get_tmp_cache` for Default solver (and some other default solver DAE fixes)

oscardssmith:
Previously, this was assuming that `cache.cache1` was filled in which wasn't always true for DAEs.


ChrisRackauckas (approved):
MWTP

oscardssmith:
@ChrisRackauckas should we tag a new release with this?

Issue #2197, opened 2024-05-17
Out of place `linsolve=KrylovJL_GMRES` is broken


oscardssmith:
```
julia> using LinearSolve, OrdinaryDiffEq
julia> function rober(u, p, t)
           y₁, y₂, y₃ = u
           k₁, k₂, k₃ = p
           [-k₁ * y₁ + k₃ * y₂ * y₃,
             k₁ * y₁ - k₃ * y₂ * y₃ - k₂ * y₂^2,
             k₂ * y₂^2]
       end
rober (generic function with 1 method)

julia> prob = ODEProblem(rober, [1.0,0.0,0.0],(0.0,1e3),(0.04,3e7,1e4));

julia> solve(prob, FBDF(linsolve=KrylovJL_GMRES()))
[...]

PR #2252, opened 2024-06-17
fix out of place `linsolve=KrylovJL*`

oscardssmith:
fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2197. This works, but I'm not 100% sure why we are treating `StaticWOperator` differently here, but it is just copying from the in place method.

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -682,9 +682,7 @@ function calc_W!(W, integrator, nlsolver::Union{Nothing, AbstractNLSolver}, cach
     end
 
     # calculate W
-    if W isa AbstractSciMLOperator && !(W isa Union{WOperator, StaticWOperator})
-        update_coefficients!(W, uprev, p, t; transform = W_transform, dtgamma)
-    elseif W isa WOperator
+    if W isa WOperator
         if isnewton(nlsolver)
             # we will call `update_coefficients!` for u/p/t in NLNewton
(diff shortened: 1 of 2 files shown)


oscardssmith:
@ChrisRackauckas does https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/9550147409/job/26321432771?pr=2252 look related to you? I can't tell.

ChrisRackauckas:
Yes, that looks related. It looks like it may just be a tolerance issue though, so the tests there may just have too strict of a tolerance. But definitely related.

ChrisRackauckas:
`StaticWOperator` is immutable and is rebuilt each time instead of modifying, so it doesn't update coefficients in place but instead just makes a new operator.

ChrisRackauckas:
The DDE test looks relevant.

oscardssmith:
The Delay issues are unrelated (but very concerning). I can reproduce the failures without this PR.

ChrisRackauckas:
Okay, follow up with downstream for a bit because it seems DDE and SDE have gotten out of sync with the last few changes.

prbzrg:
When does this change get a release? I'm still getting the error.

prbzrg:
I'm using `v6.84.0` and it seems that this PR isn't included in it.
https://github.com/SciML/OrdinaryDiffEq.jl/compare/v6.84.0...master

PR #2269, opened 2024-07-01
Don't use prev_theta for non-adaptive solves

oscardssmith:
found by @bradcarman. The early exit here relies on implicit feedback from the  solver to prevent it from overly aggressively terminating the nonlinear solve as successful. As such, we disable prev_theta for non-adaptive algorithms.

### DIFF
--- a/src/nlsolve/nlsolve.jl
+++ b/src/nlsolve/nlsolve.jl
@@ -104,7 +104,7 @@ function nlsolve!(nlsolver::NL, integrator::DiffEqBase.DEIntegrator,
         η = DiffEqBase.value(θ / (1 - θ))
         # don't trust θ for non-adaptive on first iter because the solver doesn't provide feedback
         # for us to know whether our previous nlsolve converged sufficiently well
-        check_η_convergance = (iter > 1 || (isnewton(nlsolver) && isadaptive(nlsolver)))
+        check_η_convergance = (iter > 1 || (isnewton(nlsolver) && isadaptive(integrator.alg)))
         if (iter == 1 && ndz < 1e-5) ||
            (check_η_convergance && η >= zero(η) && η * ndz < κ)
             nlsolver.status = Convergence
--- a/src/nlsolve/nlsolve.jl
+++ b/src/nlsolve/nlsolve.jl
@@ -88,8 +88,7 @@ function nlsolve!(nlsolver::NL, integrator::DiffEqBase.DEIntegrator,
                 break
             end
         else
-            # don't use prev_θ for non-adaptive because we own't know if it actually converged
-            if !integrator.opts.adaptive || !integrator.accept_step
+            if !integrator.accept_step
                 prev_θ = one(prev_θ)
             end
             θ = prev_θ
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
This will need a test.

oscardssmith:
Agreed. It's a little tricky to test for since it requires a sufficiently nonlinear problem that actually runs into it. e.g. none of the algconvergence tests caught the issue.

ChrisRackauckas:
Can we kick out an anonymized form of the model that's being tested with? ODEProblemExpr and then obfuscate variable names? Or `modelingtoolkitize` it then ODEProblemExpr?

oscardssmith:
Theoretically, we probably could, but I would like to have a test for a situation where I actually believe that the solution is correct.

ChrisRackauckas:
ahh that's good

oscardssmith:
should we merge this?

ChrisRackauckas:
Rebase, tests should be passing

oscardssmith:
rebased.

ChrisRackauckas:
Some tests need to be adjusted

oscardssmith:
This does suggest a bug since this should be strictly more accurate for non-adaptive methods...

oscardssmith:
I think I see the problem with the previous version. Lets see if this passes tests.

src/nlsolve/nlsolve.jl
```diff
@@ -102,8 +102,11 @@ function nlsolve!(nlsolver::NL, integrator::DiffEqBase.DEIntegrator,
 
         # check for convergence
         η = DiffEqBase.value(θ / (1 - θ))
+        # don't trust θ for non-adaptive on first iter because the solver doesn't provide feedback
+        # for us to know whether our previous nlsolve converged sufficiently well
+        check_η_convergance = (iter > 1 || (isnewton(nlsolver) && isadaptive(solver)))
```
ChrisRackauckas, line 107:
```suggestion
        check_η_convergance = (iter > 1 || (isnewton(nlsolver) && isadaptive(nlsolver)))
```

src/nlsolve/nlsolve.jl
```diff
@@ -102,8 +102,11 @@ function nlsolve!(nlsolver::NL, integrator::DiffEqBase.DEIntegrator,
 
         # check for convergence
         η = DiffEqBase.value(θ / (1 - θ))
+        # don't trust θ for non-adaptive on first iter because the solver doesn't provide feedback
+        # for us to know whether our previous nlsolve converged sufficiently well
+        check_η_convergance = (iter > 1 || (isnewton(nlsolver) && isadaptive(nlsolver)))
```
ChrisRackauckas, line 107:
```suggestion
        check_η_convergance = (iter > 1 || (isnewton(nlsolver) && isadaptive(integrator.alg)))
```

bradcarman:
Did we add a test to ensure this works moving forward?  Maybe we should at least add a benchmark that failed previously but works now?

ChrisRackauckas:
this is very difficult to target a specific test towards. At best if we can anonymize an integration test that would help.

oscardssmith:
we do not. while it's easy (in hindsight) to see the issue here, coming up with an example ODE that observes the behavior as previously mentioned is somewhat nontrivial. specifically, we need an ODE where the Newton iteration transitions from converging with a high order (ideally 10 or higher) but which then in the very next step has the error jump substantially (but less than the order of the Newton solver) . furthermore, the accumulated error from this single step needs to throw of the solution significantly.

PR #2274, opened 2024-07-04
Added Verner Methods

ParamThakkar123:
## Checklist


ParamThakkar123:
@ChrisRackauckas Should I also shift Tsit, BS5, AutoVern methods to the OrdinaryDiffEqVerner lib ?? They are almost always defined together in some files

ChrisRackauckas:
Make an explicit RK group for those, though Tsit5 should be in its own because it's very popular so it should have the lowest dependency setup.

ParamThakkar123:
> @ChrisRackauckas Should I also shift Tsit, BS5, AutoVern methods to the OrdinaryDiffEqVerner lib ?? They are almost always defined together in some files

These methods are defined together in composite_algs.jl file. What should be done for that ??

ChrisRackauckas:
The default algorithm will need to move to an OrdinaryDiffEqDefault which depends on everything it uses in the default algorithm.

ChrisRackauckas:
The default solver needs to be its own thing. For now, just but the imports required for it before its definition.

ParamThakkar123:
What changes must be made here ?? I will make those

lib/OrdinaryDiffEqDefault/src/composite_algs.jl
```diff
@@ -1,6 +1,7 @@
 ### AutoSwitch
 ### Designed to switch between two solvers, stiff and non-stiff
 
+using OrdinaryDiffEq: Vern6, Vern7, Vern8, Vern9
```
ChrisRackauckas, line 4: Autoswitch itself should live in the core package. 73-79 should live with the methods that use.  The default package only needs 83 onwards.

ParamThakkar123:
@ChrisRackauckas I am getting a @fold error here. I saw that it's present in MacroTools. I tried to include it but it made no difference.

ParamThakkar123:
@ChrisRackauckas Meanwhile I am done with Rosenbrock methods.

ParamThakkar123:
@ChrisRackauckas There are some precision errors now. Due to which the tests are failing

lib/OrdinaryDiffEqFeagin/src/alg_utils.jl
```diff
@@ -3,4 +3,4 @@ alg_order(alg::Feagin12) = 12
 alg_order(alg::Feagin14) = 14
 
 alg_adaptive_order(alg::Feagin10) = 8
-alg_adaptive_order(alg::Feagin14) = 12
\ No newline at end of file
+alg_adaptive_order(alg::Feagin14) = 12
```
ChrisRackauckas, line 6: don't reformat all of the other ones. Your formatting here is incorrect. It's already corrected on master.

ChrisRackauckas:
For some reason CI didn't trigger, but that should be good to go with the last change.

src/OrdinaryDiffEq.jl
```diff
@@ -410,7 +410,8 @@ export constructDormandPrince
 
 export FunctionMap, Euler, Heun, Ralston, Midpoint, RK4, ExplicitRK, OwrenZen3, OwrenZen4,
        OwrenZen5,
-       BS3, BS5, DP5, Tsit5, DP8, TanYam7, TsitPap8, CompositeAlgorithm, Anas5, RKO65, FRK65, PFRK87,
+       BS3, BS5, DP5, Tsit5, DP8, TanYam7, TsitPap8, CompositeAlgorithm, Anas5, RKO65,
+       FRK65, PFRK87,
```
ChrisRackauckas, line 414: are you sure you're on the right formatter version?

ParamThakkar123: I am using this version of JuliaFormatter

[98e50ef6] JuliaFormatter v1.0.56

ParamThakkar123:
@ChrisRackauckas I started a PR for Rosenbrock Methods #2278

PR #2282, opened 2024-07-13
WIP: Make Enzyme discrete adjoints work

ChrisRackauckas:
MWE now works:

```julia
using Enzyme, OrdinaryDiffEq, StaticArrays

Enzyme.EnzymeCore.EnzymeRules.inactive_type(::Type{SciMLBase.DEStats}) = true
Enzyme.EnzymeCore.EnzymeRules.inactive(::typeof(OrdinaryDiffEq.increment_nf!), args...) = true
Enzyme.EnzymeCore.EnzymeRules.inactive(::typeof(OrdinaryDiffEq.increment_nf_from_initdt!), args...) = true
Enzyme.EnzymeCore.EnzymeRules.inactive(::typeof(OrdinaryDiffEq.fixed_t_for_floatingpoint_error!), args...) = true
Enzyme.EnzymeCore.EnzymeRules.inactive(::typeof(OrdinaryDiffEq.increment_accept!), args...) = true
Enzyme.EnzymeCore.EnzymeRules.inactive(::typeof(OrdinaryDiffEq.increment_reject!), args...) = true
Enzyme.EnzymeCore.EnzymeRules.inactive(::typeof(DiffEqBase.fastpow), args...) = true
Enzyme.EnzymeCore.EnzymeRules.inactive(::typeof(OrdinaryDiffEq.increment_nf_perform_step!), args...) = true
Enzyme.EnzymeCore.EnzymeRules.inactive(::typeof(OrdinaryDiffEq.check_error!), args...) = true
Enzyme.EnzymeCore.EnzymeRules.inactive(::typeof(OrdinaryDiffEq.log_step!), args...) = true

function lorenz!(du, u, p, t)
    du[1] = 10.0(u[2] - u[1])
end

const _saveat =  SA[0.0,0.25,0.5,0.75,1.0,1.25,1.5,1.75,2.0,2.25,2.5,2.75,3.0]

function f(y::Array{Float64}, u0::Array{Float64})
    tspan = (0.0, 3.0)
    prob = ODEProblem{true, SciMLBase.FullSpecialize}(lorenz!, u0, tspan)
    sol = DiffEqBase.solve(prob, Tsit5(), saveat = _saveat, sensealg = DiffEqBase.SensitivityADPassThrough())
    y .= sol[1,:]
    return nothing
end;
u0 = [1.0; 0.0; 0.0]
d_u0 = zeros(3)
y  = zeros(13)
dy = zeros(13)

Enzyme.autodiff(Reverse, f,  Duplicated(y, dy), Duplicated(u0, d_u0));
```

Core issues to finish this:

1. I shouldn't have to pull all of the logging out to a separate function, but there seems to be a bug in enzyme with int inactivity https://github.com/EnzymeAD/Enzyme.jl/issues/1636
2. `saveat` has issues because it uses Julia ranges, which can have a floating point fix issue https://github.com/EnzymeAD/Enzyme.jl/issues/274
3. adding the zero(u), zero(u) is required because Enzyme does not seem to support non-fully initialized types (@wsmoses is that known?) and segfaults when trying to use the uninitialized memory. So making the inner constructor not use undef is and easy fix to that. But that's not memory optimal. It would take a bit of a refactor to make it memory optimal, but it's no big deal and it's probably something that improves the package anyways.
4. Inactivating fastpow directly probably isn't a good idea. But the step controller should be inactive, we already do the dual/tracker dropping on that for the other AD methods. So we can move that inactivity to the right spot before merging.

### DIFF
--- a/src/integrators/controllers.jl
+++ b/src/integrators/controllers.jl
@@ -65,7 +65,7 @@ end
         q = inv(qmax)
     else
         expo = 1 / (get_current_adaptive_order(alg, integrator.cache) + 1)
-        qtmp = DiffEqBase.fastpow(EEst, expo) / gamma
+        qtmp = ^(EEst, expo) / gamma
         @fastmath q = DiffEqBase.value(max(inv(qmax), min(inv(qmin), qtmp)))
         # TODO: Shouldn't this be in `step_accept_controller!` as for the PI controller?
         integrator.qold = DiffEqBase.value(integrator.dt) / q
@@ -138,8 +138,8 @@ end
     if iszero(EEst)
         q = inv(qmax)
     else
-        q11 = DiffEqBase.fastpow(EEst, float(beta1))
-        q = q11 / DiffEqBase.fastpow(qold, float(beta2))
+        q11 = ^(EEst, float(beta1))
+        q = q11 / ^(qold, float(beta2))
         integrator.q11 = q11
         @fastmath q = max(inv(qmax), min(inv(qmin), q / gamma))
     end
@@ -412,7 +412,7 @@ end
             fac = min(gamma, (1 + 2 * maxiters) * gamma / (iter + 2 * maxiters))
         end
         expo = 1 / (get_current_adaptive_order(alg, integrator.cache) + 1)
-        qtmp = DiffEqBase.fastpow(EEst, expo) / fac
+        qtmp = ^(EEst, expo) / fac
         @fastmath q = DiffEqBase.value(max(inv(qmax), min(inv(qmin), qtmp)))
         integrator.qold = q
     end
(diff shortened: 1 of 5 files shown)


wsmoses:
@ChrisRackauckas I don't follow, what's the error with non-fully initialized types?

wsmoses:
Presumably not known, so open an issue

ChrisRackauckas:
> @ChrisRackauckas I don't follow, what's the error with non-fully initialized types?

Julia allows for a type to be partially defined by having an inner type constructor which only puts a value on a subset of arguments. The others are set to `undef` and need to be defined before access. It seems that Enzyme does not do well with that.

src/integrators/controllers.jl
```diff
@@ -65,7 +65,7 @@ end
         q = inv(qmax)
     else
         expo = 1 / (get_current_adaptive_order(alg, integrator.cache) + 1)
-        qtmp = DiffEqBase.fastpow(EEst, expo) / gamma
+        qtmp = ^(EEst, expo) / gamma
```
oscardssmith, line 68: shouldn't these have a `@fastmath` (and Float32)?

ChrisRackauckas: Yes, I was just seeing what's required in order to make Enzyme work on this.

ChrisRackauckas:
All of the pieces are in now, Enzyme now works on the explicit solvers!

Issue #2287, opened 2024-07-18
Default algorithm with StaticArray state and a ContinuousCallback errors out.


abavoil:
**Describe the bug 🐞**

Solving an ODEProblem with a StaticArray state, a ContinuousCallback and the default integration algorithm raises tries applies setindex! to a StaticArray and errors.

**Expected behavior**

It should properly integrate and solve the callback.

**Minimal Reproducible Example 👇**

```julia
using Pkg
Pkg.activate(; temp=true)
Pkg.add(["OrdinaryDiffEq", "DiffEqCallbacks", "StaticArrays"])

using OrdinaryDiffEq, DiffEqCallbacks, StaticArrays

cb = ContinuousCallback((u, t, integrator) -> t - 1, (integrator) -> nothing)
ode_problem = ODEProblem((u, p, t) -> zero(u), [0], 2, Float64[])
SA_ode_problem = remake(ode_problem, u0=SA[0])

solve(ode_problem; callback=cb)  # mutable u0: OK
solve(SA_ode_problem, Tsit5(); callback=cb)  # give algorithm: OK
solve(SA_ode_problem; callback=cb)  # KO
```

**Error & Stacktrace ⚠️**

```julia
ERROR: setindex!(::SVector{1, Float64}, value, ::Int) is not defined.
 Hint: Use `MArray` or `SizedArray` to create a mutable static array
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] setindex!(a::SVector{1, Float64}, value::Float64, i::Int64)
    @ StaticArrays ~/.julia/packages/StaticArrays/MSJcA/src/indexing.jl:3
  [3] macro expansion
    @ ~/.julia/packages/StaticArrays/MSJcA/src/broadcast.jl:159 [inlined]
  [4] _broadcast!
    @ ~/.julia/packages/StaticArrays/MSJcA/src/broadcast.jl:143 [inlined]
  [5] _copyto!
    @ ~/.julia/packages/StaticArrays/MSJcA/src/broadcast.jl:70 [inlined]
  [6] copyto!
    @ ~/.julia/packages/StaticArrays/MSJcA/src/broadcast.jl:63 [inlined]
  [7] materialize!
    @ ./broadcast.jl:914 [inlined]
  [8] materialize!
    @ ./broadcast.jl:911 [inlined]
  [9] fast_materialize!
[...]

PR #2288, opened 2024-07-18
fix `is_constant_cache(::DefaultCache)

oscardssmith:
fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2287

### DIFF
--- a/src/cache_utils.jl
+++ b/src/cache_utils.jl
@@ -16,7 +16,7 @@ function is_constant_cache(cache::DefaultCache)
     elseif current == 6
         is_constant_cache(cache.cache6)
     else
-        errror("This should not occur (please report a bug)")
+        error("This should not occur (please report a bug)")
     end
 end
 
@@ -40,6 +40,8 @@ function DiffEqBase.unwrap_cache(integrator::ODEIntegrator, is_stiff)
             return cache.cache5
         elseif current == 6
             return cache.cache6
+        else
+            error("This should not occur (please report a bug)")
         end
     elseif alg.choice_function isa AutoSwitch
         num = is_stiff ? 2 : 1
--- a/src/cache_utils.jl
+++ b/src/cache_utils.jl
@@ -1,24 +1,7 @@
-is_constant_cache(cache::OrdinaryDiffEqConstantCache) = true
-is_constant_cache(cache::OrdinaryDiffEqCache) = false
-is_constant_cache(cache::CompositeCache) = is_constant_cache(cache.caches[1])
-function is_constant_cache(cache::DefaultCache)
-    current = cache.current
-    if current == 1
-        is_constant_cache(cache.cache1)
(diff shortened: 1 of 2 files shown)


src/cache_utils.jl
```diff
@@ -1,6 +1,24 @@
...
+function is_constant_cache(cache::DefaultCache)
+    current = cache.current
+    if current == 1
+        is_constant_cache(cache.cache1)
+    elseif current == 2
+        is_constant_cache(cache.cache2)
+    elseif current == 3
+        is_constant_cache(cache.cache3)
+    elseif current == 4
+        is_constant_cache(cache.cache4)
```
ChrisRackauckas, line 13: They should all always give the same answer? Being able to constant prop may be required for type stability?

oscardssmith: they won't give the same answer because if you look up one that hasn't been initialized, it can segfault. It also shouldn't be needed for type stability since all branches are type stable and return a `Bool`

ChrisRackauckas: but it needs to constant prop to true/false to constant prop for the type choices after the function.

oscardssmith: Good catch. The new verison of this will now work.

ChrisRackauckas:
Just use the type

oscardssmith:
> Just use the type

good point. That is how fieldtype works...

abavoil:
Your fix still gives me an error. Should the answer be something like this ?  `cache` is needed to access `cache1` or the first element of `caches`.

```julia
is_constant_cache(::OrdinaryDiffEqConstantCache) = true
is_constant_cache(::OrdinaryDiffEqCache) = false
is_constant_cache(cache::CompositeCache) = is_constant_cache(cache.caches[1])
is_constant_cache(cache::D) where {D<:DefaultCache} = is_constant_cache(cache.cache1)
```

oscardssmith:
the reason that doesn't work is that `cache1` might be `undef` because we wanted to save memory. I think I have a fix here though.

oscardssmith:
ok this seems to now be working.

src/cache_utils.jl
```diff
@@ -1,6 +1,7 @@
-is_constant_cache(cache::OrdinaryDiffEqConstantCache) = true
-is_constant_cache(cache::OrdinaryDiffEqCache) = false
-is_constant_cache(cache::CompositeCache) = is_constant_cache(cache.caches[1])
+is_constant_cache(::OrdinaryDiffEqConstantCache) = true
+is_constant_cache(::OrdinaryDiffEqCache) = false
+is_constant_cache(::CompositeCache) = is_constant_cache(cache.caches[1])
```
ChrisRackauckas, line 3:
```suggestion
is_constant_cache(cache::CompositeCache) = is_constant_cache(cache.caches[1])
```

PR #2291, opened 2024-07-22
fix out of place `WOperator`

oscardssmith:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/2252 made it so there were some cases in which we weren't updating J correctly leading to convergence failures. Debugging this was hell, and all of this code needs to be ripped out in the nearish future and replaced with DifferentiationInterface and a sane layout. While I was at it, I also fixed a couple places where the stats were completely broken for out of place. Tests incoming.

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -90,13 +90,12 @@ function calc_J(integrator, cache, next_step::Bool = false)
             J = jacobian(uf, uprev, integrator)
         end
 
-        integrator.stats.njacs += 1
-
         if alg isa CompositeAlgorithm
             integrator.eigen_est = constvalue(opnorm(J, Inf))
         end
     end
 
+    integrator.stats.njacs += 1
     J
 end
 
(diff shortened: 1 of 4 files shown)


oscardssmith:
https://github.com/SciML/StochasticDiffEq.jl/pull/574 fixes the StochasticDiffEq test failure (which was caused by testing that you can call calc_W! on an out of place ODE which you just can't do). The other test failures look unrelated. I think this is finally ready to merge.

src/derivative_utils.jl
```diff
@@ -860,6 +824,7 @@ function update_W!(nlsolver::AbstractNLSolver,
     nothing
 end
 
+import StaticArrays: StaticMatrix
```
ChrisRackauckas, line 827: should go to the top level

src/derivative_utils.jl
```diff
@@ -908,42 +873,36 @@ function build_J_W(alg, u, uprev, p, t, dt, f::F, ::Type{uEltypeNoUnits},
         else
             deepcopy(f.jac_prototype)
         end
-        __f = if IIP
-            (du, u, p, t) -> _f(du, u, p, t)
+        W = if J isa StaticMatrix && alg isa OrdinaryDiffEqRosenbrockAdaptiveAlgorithm
```
ChrisRackauckas, line 876: why is this Rosenbrock specific?

oscardssmith: not sure. I was vaguely following how this worked previously

PR #2295, opened 2024-07-26
use StaticW more often

oscardssmith:
as pointed out by @ChrisRackauckas in https://github.com/SciML/OrdinaryDiffEq.jl/pull/2291, we should at least try this for the non-Rosenbrock solvers.

### DIFF
--- a/src/derivative_utils.jl
+++ b/src/derivative_utils.jl
@@ -896,7 +896,7 @@ function build_J_W(alg, u, uprev, p, t, dt, f::F, ::Type{uEltypeNoUnits},
             J
         elseif IIP
             similar(J)
-        elseif J isa StaticMatrix && alg isa OrdinaryDiffEqRosenbrockAdaptiveAlgorithm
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Needs tests

oscardssmith:
Don't existing tests cover this? What additional test would you like to see?

oscardssmith:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/4fd5afb8b5b3a105103dcb6f2430acc75b67862d/test/interface/linear_solver_test.jl#L191 which we just added hits it.

ChrisRackauckas:
Alright then merge if tests pass

oscardssmith:
interesting... there is an explicit test that `calc_W` returns a `StaticArrays.LU` but I'm unclear why that would be tested...

ChrisRackauckas:
Legacy. That can be removed if we see the Static operator is a faster approach

oscardssmith:
test removed

PR #2297, opened 2024-07-26
Make W_transform always true

oscardssmith:
Rosenbrock23 and Rosenbrock32 need updating to the W_transform versions, but other than that, this PR removes ~200 lines of complexity for free.


ChrisRackauckas:
Yeah this is a good idea. The same would need to be done on StochasticDiffEq for this to go through.

oscardssmith:
Now that Rosenbrock23 is updated, I think this should be ready to go. Let's see if CI agrees.

oscardssmith:
ah it looks like ODEExtrapolation also is currently written with transform false.

oscardssmith:
CI green. Merge?

ChrisRackauckas (approved):

PR #2300, opened 2024-07-27
Finish DAE

ChrisRackauckas:


src/nlsolve/newton.jl
```diff
@@ -176,7 +176,7 @@ Equations II, Springer Series in Computational Mathematics. ISBN
     ndz = opts.internalnorm(atmp, t)
     # NDF and BDF are special because the truncation error is directly
     # proportional to the total displacement.
-    if integrator.alg isa QNDF
+    if has_special_newton_error(integrator.alg)
```
ChrisRackauckas, line 179: @YingboMa is there a reason this is only QNDF and not FBDF?

YingboMa: I am not sure about the error derivation of FBDF, but I think it depends on the history dts as well. Maybe not strong enough to still approximate it as proportional?

ChrisRackauckas: approximately proportional seems like a reasonable assumption to me

PR #2306, opened 2024-08-01
make Rosenbrock23 use W transform

oscardssmith:
This is not quite working yet (it only is converging to order 1), but it's close.

### DIFF
--- a/src/perform_step/rosenbrock_perform_step.jl
+++ b/src/perform_step/rosenbrock_perform_step.jl
@@ -374,7 +374,7 @@ end
         linsolve_tmp = @.. (integrator.fsallast - linsolve_tmp +
                        c₃₂ * f₁ + 2 * integrator.fsalfirst + dt * dT)
     end
-    k₃ = _reshape(W \ _vec(linsolve_tmp), axes(uprev)) * invnegdtγ
+    k₃ = _reshape(W \ _vec(linsolve_tmp), axes(uprev)) * neginvdtγ
     integrator.stats.nsolve += 1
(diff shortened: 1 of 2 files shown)


oscardssmith:
I believe this is now working. I don't think I've done it quite optimally, but it's good enough to not get in the way of https://github.com/SciML/OrdinaryDiffEq.jl/pull/2297

src/perform_step/rosenbrock_perform_step.jl
```diff
@@ -80,7 +80,7 @@ end
 
     @.. broadcast=false linsolve_tmp=f₁ - tmp
 
-    linres = dolinsolve(integrator, linres.cache; b = _vec(linsolve_tmp))
+    linres = dolinsolve(integrator, linres.cache; b = _vec(linsolve_tmp))/dtγ
```
ChrisRackauckas, line 83: this is causing an allocation?

ChrisRackauckas: Isn't it easier to just update the coeficients?

oscardssmith: yeah will fix.

oscardssmith: > Isn't it easier to just update the coeficients?

Computationally yes, but only once after writing the corrected version without changing the coefficients.

oscardssmith: redundant divisions removed.

ChrisRackauckas (approved):

PR #2313, opened 2024-08-05
Split out OrdinaryDiffEqCore and make dependencies into a tree

ChrisRackauckas:


ChrisRackauckas:
We can remove all of the buildkite runners and make them github action runners if that makes it easier @thazhemadam

thazhemadam:
> We can remove all of the buildkite runners and make them github action runners if that makes it easier

It should be fine to run them on Buildkite itself. Will GHA even be able to run them (the non-GPU ones at least)?

PR #2316, opened 2024-08-07
Function to strip solutions and interpolation of functions

jClugstor:
## Checklist

- [x ] All documentation related to code changes were updated
- [x ] The new code follows the
  
## Additional context

Adds a function to strip `InterpolationData` objects of their functions, and strip `ODEsolution` objects of their functions, in order to accommodate serialization better.

### DIFF
--- a/src/interp_func.jl
+++ b/src/interp_func.jl
@@ -61,7 +61,7 @@ function InterpolationData(id::InterpolationData, f)
 end
 
 # strip interpolation of function information
-function strip_interpolation(id::InterpolationData)
+function SciMLBase.strip_interpolation(id::InterpolationData)
     InterpolationData(nothing, id.timeseries,
         id.ts,
         id.ks,
--- a/src/interp_func.jl
+++ b/src/interp_func.jl
@@ -59,3 +59,15 @@ function InterpolationData(id::InterpolationData, f)
         id.differential_vars,
         id.sensitivitymode)
 end
+
+# strip interpolation of function information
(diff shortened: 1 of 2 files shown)


src/misc_utils.jl
```diff
@@ -130,3 +130,16 @@ function get_differential_vars(f, u)
 end
 
 isnewton(::Any) = false
+
+function strip_solution(sol::ODESolution)
```
ChrisRackauckas, line 134: This goes in SciMLBase, everything used here is defined there.

src/interp_func.jl
```diff
@@ -59,3 +59,15 @@ function InterpolationData(id::InterpolationData, f)
         id.differential_vars,
         id.sensitivitymode)
 end
+
+# strip interpolation of function information
+function strip_interpolation(id::InterpolationData)
```
ChrisRackauckas, line 64: This needs to extend not shadow from SciMLBase, so make sure to import.

ChrisRackauckas:
Rebase this PR onto https://github.com/SciML/OrdinaryDiffEq.jl/pull/2313, i.e. just add that one function

src/interp_func.jl
jClugstor: is `SciMLBase.strip_interpolation ... ` fine, or should I do an explicit `import SciMLBase: strip_interpolation

ChrisRackauckas: namespacing is fine.

Issue #1066, opened 2020-03-13
Magnus Methods


ChrisRackauckas:

Biswajitghosh98:
MagnusGL6 implemented in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1183
And MagnusNC6 implemented in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1184

Biswajitghosh98:
MagnusNC8 implemented in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1188
and Magnus GL8 implemented in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1186

Biswajitghosh98:
Formally, given an IVP defined by the linear differential equation **dX/dt = A(t)X**  , and **X(t<sub>0</sub>) = I** , where **I** is a matrix containing the inital values of the dependent variables at **t<sub>0</sub>**, we can obtain a solution of the IVP that can be written in the form 
**X(t,t<sub>0</sub>) = e<sup>Ω(t,t<sub>0</sub>)</sup>**.

Now, **Ω** is obtained as an infinite series **Ω(t,t<sub>0</sub>) = Σ Ω<sub>k</sub>(t,t<sub>0</sub>)** , wherein the k<sup>th</sup> term, or **Ω<sub>k</sub>** in the series is a multiple integral of nested commutators containing k operators **A(t)**.

What each of the implemeted algoritm does is, an algorithm of order **n**, say **MagnusGL6**, where **n=6** numerically approximates the integral terms with quadratures so as to obtain the required accuracy of order **O(dt<sup>n+1</sup>)**

To use any of these, one can simply :
- Define an update function for the coefficient matrix **A(t)**
```
function update_func(A,u,p,t)
    A[1,1] = cos(t)
    A[2,1] = sin(t)
    A[1,2] = -sin(t)
    A[2,2] = cos(t)
end
```
- Formulate the ODE Problem 
```
A = DiffEqArrayOperator(ones(2,2),update_func=update_func)
prob = ODEProblem(A, ones(2), (1.0, 6.0))
```
- Use any of the listed algorithms to solve the problem, depending on the convergence order needed, which in our case was **MagnusGL6()**
```
sol = solve(prob,MagnusGL6(),dt=1/10)
```

The generated solution object `sol` can now be handled separately.

PR #2317, opened 2024-08-07
docs v7 rework

ArnoStrouwen:
Is this template what you had in mind @ChrisRackauckas?


docs/src/explicit/Tsit5.md
```diff
@@ -0,0 +1,39 @@
+# OrdinaryDiffEqTsit5
+
+Recommended solver for most non-stiff problems.
```
ChrisRackauckas, line 3:
```suggestion
Recommended solver for most non-stiff problems at default and higher tolerance.
```

ChrisRackauckas:
Yes exactly.

ArnoStrouwen:
I came up with a way to automate the same example over all the libs, but it's a pretty ugly hack.

ChrisRackauckas:
Looks like it works, I'll take it.

ArnoStrouwen:
The current dev version of docs removed documentation about threading, while this is still there on stable. Is this a regression?
dev
stable

ChrisRackauckas:
That's a regression

ArnoStrouwen:
Are the `Anas*` methods really `OrdinaryDiffEqAlgorithm` and not `OrdinaryDiffEqAdaptiveAlgorithm`?
Same question for:
```jl
RKO65
RKM
MSRK*
PSRK*
Alshina6
```

lib/OrdinaryDiffEqLowOrderRK/src/algorithms.jl
```diff
@@ -276,7 +273,7 @@ function FRK65(stage_limiter!, step_limiter! = trivial_limiter!; omega = 0.0)
     FRK65(stage_limiter!, step_limiter!, False(), omega)
 end
 
-@doc explicit_rk_docstring("TBD", "RKM")
+@doc explicit_rk_docstring("To be done", "RKM")
```
ArnoStrouwen, line 276: @ChrisRackauckas please write a sentence describing this method.

ChrisRackauckas:
> Are the Anas* methods really OrdinaryDiffEqAlgorithm and not OrdinaryDiffEqAdaptiveAlgorithm?

IIRc they are non-adaptive. You can check the perform_step! on whether they calculate an EEst.

lib/OrdinaryDiffEqLowOrderRK/src/algorithms.jl
ChrisRackauckas: Optimal Runge–Kutta Methods for First Order Pseudospectral Operators

https://www.sciencedirect.com/science/article/abs/pii/S0021999199962605

An explicit Runge-Kutta method which is designed to have good stability properties when applied to pseudospectral discretizations of hyperbolic partial differential equaitons.

ArnoStrouwen:
How does` HighOrderRK` fit into the story? If I understand correctly, the recommended solver page skips over them in favor of `Vern` for lower tolerances.

ChrisRackauckas:
Yes they aren't recommended for most things. DP8 is legacy fallback to dop853 but Verner is just generally better unless you really need a non-lazy interpolation.

lib/OrdinaryDiffEqLowStorageRK/src/algorithms.jl
```diff
@@ -890,7 +890,7 @@ end
 
 #SSP Optimized Runge-Kutta Methods
 
-@doc explicit_rk_docstring("TBD",
+@doc explicit_rk_docstring("To be done",
```
ArnoStrouwen, line 893: @ChrisRackauckas, write another sentence, please.

ChrisRackauckas: "Optimal Strong-Stability-Preserving Runge–Kutta Time Discretizations for Discontinuous Galerkin Methods" by Kubatko, Yeager, Ketcheson (2014)

ArnoStrouwen: So why is this solver in `OrdinaryDiffEqLowStorageRK `and not in `OrdinaryDiffEqSSPRK`, since I'm guessing SSPRK stands for strong stability preserving runge kutta.

ChrisRackauckas: It is both SSP and a low storage method IIUC? @ranocha

lib/OrdinaryDiffEqSSPRK/src/algorithms.jl
```diff
@@ -347,7 +347,7 @@ function KYKSSPRK42(stage_limiter!, step_limiter! = trivial_limiter!)
         False())
 end
 
-@doc explicit_rk_docstring("TBD", "SHLDDRK52")
+@doc explicit_rk_docstring("To be done", "SHLDDRK52")
```
ArnoStrouwen, line 350: @ChrisRackauckas These above 3 solvers also need more info.

ChrisRackauckas: The LDDRK methods are from https://www.sciencedirect.com/science/article/abs/pii/S0021999198959861

> The explicit Runge-Kutta (RK)
methods are widely used to discretize the time derivative because of their advantages that
include flexibility, large stability limits, and ease of programming. Hu and co-workers [2]
showed that the dissipation and dispersion properties of the RK methods depend on their
coefficients and optimized them for the convective wave equation, obtaining what they
called low-dissipation and dispersion Runge-Kutta (LDDRK) methods. These methods are
more efficient than classical ones, in terms of work required for a given accuracy, for wave
propagation problems.

lib/OrdinaryDiffEqSSPRK/src/algorithms.jl
```diff
@@ -333,7 +333,7 @@ function SHLDDRK_2N(stage_limiter!, step_limiter! = trivial_limiter!)
         False())
 end
 
-@doc explicit_rk_docstring("TBD",
+@doc explicit_rk_docstring("To be done",
     "KYKSSPRK42")
 Base.@kwdef struct KYKSSPRK42{StageLimiter, StepLimiter, Thread} <: OrdinaryDiffEqAlgorithm
```
ChrisRackauckas, line 338: https://link.springer.com/article/10.1007/s10915-013-9796-7

> Discontinuous Galerkin (DG) spatial discretizations are often used in a method-of-lines approach with explicit strong-stability-preserving (SSP) Runge–Kutta (RK) time steppers for the numerical solution of hyperbolic conservation laws. The time steps that are employed in this type of approach must satisfy Courant–Friedrichs–Lewy stability constraints that are dependent on both the region of absolute stability and the SSP coefficient of the RK method. While existing SSPRK methods have been optimized with respect to the latter, it is in fact the former that gives rise to stricter constraints on the time step in the case of RKDG stability. Therefore, in this work, we present the development of new “DG-optimized” SSPRK methods with stability regions that have been specifically designed to maximize the stable time step size for RKDG methods of a given order in one space dimension. These new methods represent the best available RKDG methods in terms of computational efficiency, with significant improvements over methods using existing SSPRK time steppers that have been optimized with respect to SSP coefficients. Second-, third-, and fourth-order methods with up to eight stages are presented, and their stability properties are verified through application to numerical test cases.

ChrisRackauckas:
> Seems weird to me the 5 from AB5 does not occur anywhere in the description.

5th Order Adams-Bashforth method. History is prestarted using Ralston's 3rd order Runge-Kutta method.

ArnoStrouwen:
I'm not sure what of this `JVODE` stuff needs a docstring:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/dc0e1e733fdd79a6c643cbd0c4f4be03c934cb91/lib/OrdinaryDiffEqNordsieck/src/algorithms.jl#L7-L20
And even less sure what then should be in that string.

ChrisRackauckas:
It should have a warning as experimental. JVODE is an adaptive BDF implementation in Nordsieck form, similar to VODE. However, FBDF is recommended in almost all scenarios, and so this is more of a development artifact.

docs/src/explicit/Nordsieck.md
```diff
@@ -0,0 +1,17 @@
+# OrdinaryDiffEqNordsieck
+
+This article is a stub.
```
ArnoStrouwen, line 3: @ChrisRackauckas write a sentence here  what Nordsieck is about.

ChrisRackauckas: The Nordsieck form is an alternative representation of multistep methods which, instead of representing and saving past step values in a history vector, it uses a derivative list (like a Taylor expansion) for the computation of the next point. The Nordsieck form was pioneered by early implementations of BDF methods such LSODE, VODE, and finally CVODE. It can have some advantages in terms of restartability as the full Nordsieck vector can be instantiated given only the information of `f` and its derivatives after discontinuities, but the higher derivative representations can also introduce numerical instabilities of their own.

The Nordsieck implementations here are considered experimental implementations of the LSODE non-fixed leading coefficient form and are generally considered inferior to the fixed-leading history-based BDF implementation of FBDF, and thus for all standard usage we recommend FBDF. However, this algorithm is kept for experimental research and development purposes with the possibility of one day becoming a more discontinuity-aware BDF implementation.

docs/src/explicit/LowStorageRK.md
```diff
@@ -0,0 +1,57 @@
+# OrdinaryDiffEqLowStorageRK
+
+Solvers specialized on certain PDE types.
```
ChrisRackauckas, line 3:
```suggestion
These methods are designed to have reduced register requirements, allowing for larger sets of ODEs to more
easily fit into RAM. For example, while the 5th order Tsit5 requires around 9 concurrent instantiations of the
ODE state `u`, `RDPK3Sp510` can achieve 5th order with 3 registers, effectively requiring 1/3 of the memory.
However, there are some efficiency trade-offs used in the design of the low-storage RK methods, and thus they
are generally only recommended in situations which are RAM bound, like large-scale PDE discretizations.
```

docs/src/explicit/Nordsieck.md
ArnoStrouwen: Is this true of AN5 also?

docs/src/explicit/SSPRK.md
```diff
@@ -0,0 +1,32 @@
+# OrdinaryDiffEqSSPRK
+
+Solvers specialized on certain PDE types.
```
ChrisRackauckas, line 3: SSPRK methods are Runge-Kutta methods which support the "strongly preserving property" (SSP). They are
designed for the use in discretizations of hyperbolic partial differential equations and conservation laws
as they have extra stability properties ( e.g., stability with respect to total variation, the maximum norm, or other convex functionals) when step-size restrictions are respected. In particular, these properties are granted if the step-size is kept to a level where the CFL coefficients are less than the SSP coefficient.

Note that for SSPRK methods, a algorithm utility `OrdinaryDiffEqCore.ssp_coefficient(alg)` exists that allows for querying the SSP coefficient for use in step size calculations.

docs/src/explicit/Nordsieck.md
ChrisRackauckas: Yes, AN5 is just the Adam's version, so it's similar to the non-stiff run of LSODE rather than the stiff BDF one.

lib/OrdinaryDiffEqNordsieck/src/algorithms.jl
```diff
@@ -16,5 +22,13 @@ function JVODE(algorithm = :Adams; bias1 = 6, bias2 = 6, bias3 = 10,
         addon = 1 // 10^6)
     JVODE(algorithm, bias1, bias2, bias3, addon)
 end
+"""
+!!! warning "Experimental"
+    `JVODE` is experimental, the solver `FBDF` is generally preferred.
```
ChrisRackauckas, line 27:
```suggestion
    `JVODE` is experimental, the solver `VCABM` is generally preferred.
```

lib/OrdinaryDiffEqNordsieck/src/algorithms.jl
```diff
@@ -1,9 +1,15 @@
...
+    "Adaptive step size Adams explicit Method",
+    "",
+    "",
+    "")
+struct AN5 <: OrdinaryDiffEqAdaptiveAlgorithm end
 """
-AN5: Adaptive step size Adams explicit Method
-An adaptive 5th order fixed-leading coefficient Adams method in Nordsieck form.
+!!! warning "Experimental"
+    `JVODE` is experimental, the solver `FBDF` is generally preferred.
```
ChrisRackauckas, line 11:
```suggestion
    `JVODE` is experimental, the solver `VCABM` is generally preferred.
```

docs/src/explicit/Feagin.md
```diff
@@ -0,0 +1,17 @@
+# OrdinaryDiffEqFeagin
+
+Preferred solvers for non-stiff problems at very low tolerance, `<1e-30`.
+Best combined with preciser than `Float64` number types for the state, such as the `BigFloat` number type.
```
ChrisRackauckas, line 4:
```suggestion
Best combined with preciser than `Float64` number types for the state, such as the `BigFloat` number type.
Note that the Feagin methods have a less robust error estimator than the Verner methods, and thus even for
very low tolerance problems the Verner methods (`Vern9`) may still be more efficient. In addition, at extremely
low tolerances the explicit extrapolation methods allow for arbitrarily high variable order stepping which will
also outperform the Feagin methods. As such, the Feagin methods may be useful in the Float128 precision range
but should be tested against other algorithms.
```

ChrisRackauckas:
We should also audit to ensure that every algorithm has its source reference in the docstring. I think almost all do, but a few may be missing some.

ArnoStrouwen:
> We should also audit to ensure that every algorithm has its source reference in the docstring. I think almost all do, but a few may be missing some.

Some of the historic ones do miss a reference:

Euler
Heun
Ralston
Midpoint

Then also:
Anas5
FRK65
TsitPap8
PFRK87

many CKLLSRK*

ChrisRackauckas:
For the historic ones we can just point to Wikipedia or Hairer I

ChrisRackauckas:
Anas5: https://github.com/SciML/OrdinaryDiffEq.jl/issues/330 https://www.sciencedirect.com/science/article/pii/S0377042704002432

FRK65: https://github.com/SciML/OrdinaryDiffEq.jl/pull/1033 https://www.researchgate.net/publication/325858674_Fitted_modifications_of_Runge-Kutta_pairs_of_orders_65

ChrisRackauckas:
TsitPap8 https://github.com/SciML/DiffEqDevTools.jl/blob/master/src/ode_tableaus.jl#L4654

ChrisRackauckas:
BTW, I really wonder if we should move the tableaus to a lib OrdinaryDiffEqExplicitRKTableaus and document those here as well.

ChrisRackauckas:
PFRK87 https://github.com/SciML/OrdinaryDiffEq.jl/pull/1047 https://www.sciencedirect.com/science/article/pii/S037704271730095X

ChrisRackauckas:
CKLLSRK  https://github.com/SciML/OrdinaryDiffEq.jl/pull/675 https://www.sciencedirect.com/science/article/pii/S0168927499001415

ArnoStrouwen:
Most of the explicit solvers have been done now.

The other solvers will be more challenging.
Since we need to figure out how to get a docstring inside loops like this:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/OrdinaryDiffEqRosenbrock/src/algorithms.jl#L91

lib/OrdinaryDiffEqSymplecticRK/src/algorithms.jl
```diff
@@ -1,6 +1,10 @@
+@doc generic_solver_docstring("DESCRIPTION TBD",
```
ArnoStrouwen, line 1: Quite a bit to still fill in for each symplectic integrators @ChrisRackauckas.

ChrisRackauckas:
> Since we need to figure out how to get a docstring inside loops like this:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/OrdinaryDiffEqRosenbrock/src/algorithms.jl#L91

IIRC Symbolics or SymbolicUtils figured out how to tag docstrings onto live objects. So @shashi or @YingboMa know how to do it. I can't find the snippet that does it though.

lib/OrdinaryDiffEqLowStorageRK/src/algorithms.jl
ranocha: Yes, but I think it should be in OrdinaryDiffEqSSPRK.jl, see https://github.com/SciML/OrdinaryDiffEq.jl/issues/2373
Many SSP methods are low-storage schemes, but we still want to collect them in the SSP sub-package

lib/OrdinaryDiffEqCore/src/alg_utils.jl
```diff
@@ -412,3 +412,12 @@ is_mass_matrix_alg(alg::Union{OrdinaryDiffEqAlgorithm, DAEAlgorithm}) = false
...
 is_mass_matrix_alg(alg::NewtonAlgorithm) = !isesdirk(alg)
+
+# All algorithms should be shown using their keyword definition, and not as structs
+function Base.show(io::IO, alg::OrdinaryDiffEqAlgorithm)
+    print(io, String(typeof(alg).name.name), "(;")
+    for fieldname in fieldnames(typeof(alg))
+        print(io, " ", fieldname, " = ", getfield(alg, fieldname), ",")
+    end
+    print(io, ")")
+end
```
ArnoStrouwen, line 423: Should this be further upstreamed/generalized to SciMLAlgorithm?

ChrisRackauckas: If it'scorrect? I don't think it is as a 2-arg show. I think it needs to be on the 3-arg with a text mimetype since I don't think it round trips.

ArnoStrouwen: This is the `show` method that is used in the latest release.
This commit just moves it from a wrong location in OrdinaryDiffEqLowStorageRK to OrdinaryDiffEqCore.

I could swap it to `Base.show(io::IO, ::MIME"text/plain", alg::OrdinaryDiffEqAlgorithm)`, I have no strong opinions.

ChrisRackauckas: That dispatch would be better.

lib/OrdinaryDiffEqFIRK/src/algorithms.jl
```diff
@@ -1,20 +1,69 @@
...
+    publisher={Elsevier}}"""
+
+extra_keyword_description = """
+    - `extrapolant`: TBD
+    - `fast_convergence_cutoff`: TBD
+    - `new_W_γdt_cutoff`: TBD
+    - `controller`: TBD
+    - `κ`: TBD
+    - `maxiters`: TBD
+    - `smooth_est`: TBD
```
ArnoStrouwen, line 49: @ChrisRackauckas these need to be filled in.

docs/src/implicit/FIRK.md
```diff
@@ -0,0 +1,14 @@
+# OrdinaryDiffEqFIRK
+
+This article is a stub.
```
ArnoStrouwen, line 3: @ChrisRackauckas write a sentence what FIRK is all about. The current solver recommendations don't seem to feature these heavily.

lib/OrdinaryDiffEqPRK/src/algorithms.jl
```diff
@@ -1,10 +1,10 @@
...
-These methods utilize multithreading on the f calls to parallelize the problem.
-This requires that simultaneous calls to f are thread-safe.
-"""
+@doc generic_solver_docstring("A 5 parallel, 2 processor method of 5th order.
+        ",
+        "KuttaPRK2p5",
+        "Explicit Runge-Kutta Method",
+        "REFERENCE TBD",
+        "- `threading`: TBD",
+        "threading = true,")
```
ArnoStrouwen, line 7: @ChrisRackauckas reference and description of threading needed.
I'm confused what the difference is between `threading` here and `thread` in e.g. Tsit85.

docs/src/implicit/PDIRK.md
```diff
@@ -0,0 +1,12 @@
+# OrdinaryDiffEqPDIRK
+
+This article is a stub.
```
ArnoStrouwen, line 3: @ChrisRackauckas write a sentence what PDIRK is about.

docs/src/explicit/Extrapolation.md
```diff
@@ -0,0 +1,18 @@
...
+
+```@docs
+AitkenNeville
+ImplicitEulerExtrapolation
+ExtrapolationMidpointDeuflhard
+ImplicitDeuflhardExtrapolation
+ExtrapolationMidpointHairerWanner
+ImplicitHairerWannerExtrapolation
+ImplicitEulerBarycentricExtrapolation
+```
```
ArnoStrouwen, line 18: @ChrisRackauckas write a sentence describing what Extrapolation is all about.

All 7 of the solvers are lacking a reference and description of their keywords.

ArnoStrouwen: Also, putting these solvers under explicit isn't correct, since there are a few implicit ones also.
Would it not make more sense to split this up into ExplicitExtrapolation and ImplicitExtrapolation?

ChrisRackauckas: https://github.com/SciML/OrdinaryDiffEq.jl/issues/2371#issuecomment-2283654640

ChrisRackauckas: > Would it not make more sense to split this up into ExplicitExtrapolation and ImplicitExtrapolation?

it would.

ChrisRackauckas: The extrapolation paper describes it all in too much detail. https://ieeexplore.ieee.org/abstract/document/9926357 . Euler is first order, midpoint and HairerWanner are second order. Deuflhard vs HairerWanner vs Barycentric is somewhat an implementation detail in how the coalescing is performed.

docs/src/implicit/PDIRK.md
ChrisRackauckas: PDIRK methods are parallel DIRK methods. SDIRK methods, or singly-diagonally implicit methods, have to build and solve a factorize a Jacobian of the form W = I-gamma*J where gamma is dependent on the chosen method. PDIRK methods use multiple different choices of gamma, i.e. W_i = I-gamma_i*J, which are all used in the update process. There are some advantages to this, as no SDIRK method can be a higher order than 5 while DIRK methods generally can have arbitrarily high order and lower error coefficients, leading to lower errors at larger `dt` sizes. With the right construction of the tableau, these matrices can be factorized and the underlying steps can be computed in parallel, which is why these are the parallel DIRK methods.

Note that this set is currently experimental as there are no parallel DIRK tableaus that achieve good performance in the literature.

lib/OrdinaryDiffEqPRK/src/algorithms.jl
ChrisRackauckas: It's the same. It's an oversight that these were named differently and we should correct this with the v7

lib/OrdinaryDiffEqStabilizedRK/src/algorithms.jl
```diff
@@ -62,66 +67,74 @@ for Alg in [:ESERK4, :ESERK5, :RKC]
...
+    - `eigen_est`: function of the form
+        `(integrator) -> integrator.eigen_est = upper_bound`,
+        where `upper_bound` is an estimated upper bound on the spectral radius of the Jacobian matrix.
+        If `eigen_est` is not provided, `upper_bound` will be estimated using the power iteration.
+    """,
+    """
+    controller = :PI
+    eigen_est = nothing,
+    """)
 struct SERK2{E} <: OrdinaryDiffEqAdaptiveAlgorithm
```
ArnoStrouwen, line 138: @ChrisRackauckas This docstring is new, and still has a blank description and reference.

docs/src/implicit/PDIRK.md
ArnoStrouwen: If there are no good DIRK tableaus, why is Kvaerno5 one of the recommended solvers?
https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/#Stiff-Problems

ChrisRackauckas: It's an (E)SDIRK tableau.

docs/src/implicit/FIRK.md
ChrisRackauckas: FIRK methods are fully implicit Runge-Kutta methods. They can have special properties, like be symplectic integrators, and can achieve higher order for the same number of stage in comparison to diagonal methods. However, the fully implicit methods have a larger implicit system to solve and thus have a higher linear algebra cost. This can be useful in some contexts to promote more parallelism, but also since the size of the factorization is cubic and the dominant cost for large equations, multiplying O(n^3) operations to O((sn)^3) can be a considerable cost increase for FIRK tableaus where s, the number of stages, is particularly large. That said, the restriction to diagonal implicitness imposes order restrictions, such as SDIRK methods having a maximum order of 5, which can restrict the problems best suited for SDIRK methods.

The most common FIRK method in production are those based on RadauIIA tableaus, which is an ODE representation of Gaussian collocation. Like Gaussian collocation, it achieves higher order convergence than its stages, namely order 2s+1 for s stages. Thus RadauIIA FIRK methods tend to be some of the highest order methods (excluding extrapolation methods). This means that high order RadauIIA methods are recommended in the same scenarios that high-order explicit Runge-Kutta methods are recommended simply with the restriction of being a stiff equation. Such scenarios include cases like very low tolerances: RadauIIA methods can be the best performing methods for scenarios where tolerances are `1e-9` and below. Additionally, for ODE systems of size less than 200, the increased size of the Jacobian is mitigated by improved multithreading, since BLAS implementations are only good at multithreading LU factorizations after a certain matrix size. For this reason, RadauIIA methods tend to be recommended in cases where ODE size is small to intermediate and very accurate solutions are required.

They should be tested against the parallel implicit extrapolation which also specialize in this regime.

lib/OrdinaryDiffEqFIRK/src/algorithms.jl
ChrisRackauckas: I'm hoping we can remove those soon as we swap to the new nonlinear solvers. I'd leave them undocumented for now so it's not public API.

lib/OrdinaryDiffEqStabilizedRK/src/algorithms.jl
ChrisRackauckas: https://github.com/SciML/OrdinaryDiffEq.jl/pull/732 https://onlinelibrary.wiley.com/doi/10.1002/num.21704

The papers show that it should be close to ROCK2 but generally we've seen the ROCK integrators perform a bit better.

lib/OrdinaryDiffEqSymplecticRK/src/algorithms.jl
ArnoStrouwen: @ChrisRackauckas bump
Description for:
VelocityVerlet
VelocityLeapfrog
PseudoVerletLeapfrog
McAte2
McAte3
Ruth3
CandyRoz4
McAte4
CalvoSanz4
Yoshida6
KahanLi6
McAte8
KahanLi8
SofSpa10

Descriptions and references for:
SymplecticEuler
McAte42
McAte5

lib/OrdinaryDiffEqBDF/src/algorithms.jl
```diff
@@ -1,11 +1,61 @@
-"""
-E. Alberdi Celayaa, J. J. Anza Aguirrezabalab, P. Chatzipantelidisc. Implementation of
-an Adaptive BDF2 Formula and Comparison with The MATLAB Ode15s. Procedia Computer Science,
-29, pp 1014-1026, 2014. doi: https://doi.org/10.1016/j.procs.2014.05.091
+function BDF_docstring(description::String,
```
ArnoStrouwen, line 1: BDF fill in

Description for:
SBDF

Reference for:
MEBDF2

Both for:
DImplicitEuler
DABDF2
DFBDF

lib/OrdinaryDiffEqPDIRK/src/algorithms.jl
```diff
@@ -1,10 +1,17 @@
...
-
-#################################################
-"""
-PDIRK44: Parallel Diagonally Implicit Runge-Kutta Method
-A 2 processor 4th order diagonally non-adaptive implicit method.
-"""
+@doc differentiation_rk_docstring("A 2 processor 4th order diagonally non-adaptive implicit method.",
+    "PDIRK44",
+    "Parallel Diagonally Implicit Runge-Kutta Method.";
+    references = "TBD",
```
ArnoStrouwen, line 4: Might have looked over it in your general explanation of PDIRK, but the method still needs a reference, @ChrisRackauckas.

ChrisRackauckas: https://github.com/SciML/OrdinaryDiffEq.jl/pull/815 https://academic.oup.com/imajna/article-abstract/10/4/463/697696?redirectedFrom=fulltext&login=false

lib/OrdinaryDiffEqBDF/src/algorithms.jl
ChrisRackauckas: DImplicitEuler is Implicit Euler for implicit DAE form. It uses an apriori error estimator for adaptivity based on a finite differencing approximation from SPICE.

DABDF2 is a fully implicit implementation of BDF2 which uses the adaptivity scheme from https://www.sciencedirect.com/science/article/pii/S1877050914002683

DFBDF is a fully implicit implementation of FBDF based on Shampine's https://www.degruyter.com/document/doi/10.1515/JNMA.2002.291/html?lang=en

ChrisRackauckas: MEBDF2 is https://www.sciencedirect.com/science/article/pii/S0377042700004635

ChrisRackauckas: SBDF methods are implicit-explicit (IMEX) methods designed for SplitODEFunction equations which reduce the size of the implicit handling to a subset of the equations. It's similar to the additive Runge-Kutta methods in splitting mode, like KenCarp4, but instead using a multistep BDF approach. https://epubs.siam.org/doi/10.1137/0732037

lib/OrdinaryDiffEqPRK/src/algorithms.jl
ArnoStrouwen: @ChrisRackauckas you did not give a reference.

lib/OrdinaryDiffEqSymplecticRK/src/algorithms.jl
ChrisRackauckas: VelocityVerlet, PseudoVerletLeapfrog, and VelocityLeapfrog are just the well known schemes https://en.wikipedia.org/wiki/Verlet_integration in their subforms.

McAte methods and references to a lot of the other 4th order and below are in table 2 from https://www.researchgate.net/publication/230979099_The_Accuracy_of_Symplectic_Integrators https://iopscience.iop.org/article/10.1088/0951-7715/5/2/011

ChrisRackauckas: https://scicomp.stackexchange.com/questions/29149/what-does-symplectic-mean-in-reference-to-numerical-integrators-and-does-scip should be used in the summary text for the symplectic integrators page.

lib/OrdinaryDiffEqRosenbrock/src/OrdinaryDiffEqRosenbrock.jl
```diff
@@ -48,15 +48,21 @@ function rosenbrock_wanner_docstring(description::String,
         extra_keyword_default = "",
         with_step_limiter = false)
     keyword_default = """
+    chunk_size = Val{0}(),
```
ArnoStrouwen, line 51: @ChrisRackauckas Rosenbrock methods that lack a reference:
ROS34PW1a
ROS34PW2
ROS34PW3

lib/OrdinaryDiffEqSymplecticRK/src/algorithms.jl
ChrisRackauckas: SymbplecticEuler is https://en.wikipedia.org/wiki/Semi-implicit_Euler_method

ChrisRackauckas: For some of the descriptions you can just use the one liner from https://docs.sciml.ai/DiffEqDocs/stable/solvers/dynamical_solve/#Symplectic-Integrators

lib/OrdinaryDiffEqPRK/src/algorithms.jl
ChrisRackauckas: IIRC it's https://www.jstor.org/stable/2158288

lib/OrdinaryDiffEqSSPRK/src/algorithms.jl
```diff
@@ -320,7 +320,16 @@ function SSPRK33(stage_limiter!, step_limiter! = trivial_limiter!)
         step_limiter!, False())
 end
 
-@doc explicit_rk_docstring("TBD", "SHLDDRK_2N")
+@doc explicit_rk_docstring("Low dissipation and dispersion Runge-Kutta schemes for computational acoustics", "SHLDDRK_2N",
```
ArnoStrouwen, line 323: SDIRK missing stuff @ChrisRackauckas 

Reference:
ImplicitEuler
ImplicitMidpoin

Description:
ESDIRK436L2SA2
ESDIRK437L2SA
ESDIRK547L2SA2
ESDIRK659L2SA

Reference and description:
SDIRK22
SSPSDIRK2
CFNLIRK3
SFSDIRK4 - 8
ESDIRK54I8L2SA

lib/OrdinaryDiffEqRosenbrock/src/OrdinaryDiffEqRosenbrock.jl
ChrisRackauckas: https://www.sciencedirect.com/science/article/pii/S0377042716303934 has a lot of references in table 4.  These particular methods are from https://link.springer.com/article/10.1007/s10543-005-0035-y

lib/OrdinaryDiffEqSSPRK/src/algorithms.jl
ChrisRackauckas: SSPSDIRK is an SSP-optimized SDIRK method, so it's an implicit SDIRK method for handling stiffness but if the `dt` is below the SSP coefficient * dt, then the SSP property of the SSP integrators (the other page) is satisified. https://www.davidketcheson.info/assets/papers/implicit_ssp_postprint.pdf As such this is a method which is expected to be good on advection-dominated cases where an explicit SSP integrator would be used, but where reaction equations are sufficient stiff to justify implicit integration.

ChrisRackauckas: SFSDIRK methods are https://www.sciencedirect.com/science/article/pii/S0168927407001651 "with special nonlinear stability properties indicated by the terms total-variation-diminishing (TVD), strongly stable and monotonic".

ChrisRackauckas: https://github.com/SciML/OrdinaryDiffEq.jl/pull/1113

ChrisRackauckas: CFNLIRK3  is the third order method of https://github.com/SciML/OrdinaryDiffEq.jl/issues/1093 https://www.sciencedirect.com/science/article/pii/S0168927400000611

ChrisRackauckas: SDIRK22 is from https://ntrs.nasa.gov/api/citations/20160005923/downloads/20160005923.pdf

lib/OrdinaryDiffEqCore/src/doc_utils.jl
```diff
@@ -62,15 +60,48 @@ function explicit_rk_docstring(description::String,
...
+    extra_keyword_description::String = "",
+    extra_keyword_default::String = "")
+    keyword_default = """
+    chunk_size = Val{0}(),
+    autodiff = true,
+    standardtag = Val{true}(),
+    concrete_jac = nothing,
+    diff_type = Val{:forward},
+    linsolve = nothing,
+    precs = DEFAULT_PRECS,
```
ArnoStrouwen, line 90: @ChrisRackauckas I'll make these private for each solver (not only here) unless you object.

lib/OrdinaryDiffEqBDF/src/algorithms.jl
```diff
@@ -1,11 +1,61 @@
...
+    an Adaptive BDF2 Formula and Comparison with The MATLAB Ode15s. Procedia Computer Science,
+    29, pp 1014-1026, 2014. doi: https://doi.org/10.1016/j.procs.2014.05.091
+    """,
+    extra_keyword_description = """
+    - `κ`: TBD
+    - `tol`: TBD
+    - `nlsolve`: TBD
+    - `smooth_est`: TBD
+    - `extrapolant`: TBD
+    - `controller`: TBD
```
ArnoStrouwen, line 47: @ChrisRackauckas which of these need to go private for now?

docs/src/explicit/SymplecticRK.md
```diff
@@ -0,0 +1,35 @@
...
+Because of discretization error, when it is solving a Hamiltonian system it doesn't get exactly the correct trajectory on the manifold.
+Instead, that trajectory itself is perturbed `O(Δtn)` for the order n from the true trajectory.
+Then there's a linear drift due to numerical error of this trajectory over time
+Normal integrators tend to have a quadratic (or more) drift, and do not have any good global guarantees about this phase space path (just local).
+What means is that symplectic integrators tend to capture the long-time patterns better than normal integrators because of this lack of drift and this almost guarantee of periodicity. 
+
+```@eval
+first_steps = evalfile("./common_first_steps.jl")
+first_steps("OrdinaryDiffEqSymplecticRK", "McAte2")
+```
```
ArnoStrouwen, line 13: The generic lorenz example probably is not a good fit for symplectic?

docs/src/explicit/RKN.md
```diff
@@ -0,0 +1,30 @@
+# OrdinaryDiffEqRKN
+
+Second order solvers.
+
+```@eval
+first_steps = evalfile("./common_first_steps.jl")
+first_steps("OrdinaryDiffEqRKN", "Nystrom4")
```
ArnoStrouwen, line 7: The generic lorenz example probably is not a good fit for second order?

lib/OrdinaryDiffEqLinear/src/algorithms.jl
```diff
@@ -1,34 +1,76 @@
...
+    (:CG4a, " Fourth order Crouch-Grossman method.", "ref TBD")]
+    @eval begin
+        @doc generic_solver_docstring($Description,
+            $(string(Alg)),
+            "Semilinear ODE solver",
+            $Ref,
+            """
+            - `krylov`: TBD
+            - `m`: TBD
+            - `iop`: TBD
```
ArnoStrouwen, line 27: Not sure if can copy the keyword description of LinearExponential.
krylov definitely takes on different values.

docs/src/semilinear/Linear.md
```diff
@@ -0,0 +1,48 @@
+# OrdinaryDiffEqLinear
+
+Methods for semi-linear differential equations.
+
+```@eval
+first_steps = evalfile("./common_first_steps.jl")
+first_steps("OrdinaryDiffEqLinear", "LieRK4")
```
ArnoStrouwen, line 7: The typical example is probably not appropriate for linear solvers.

lib/OrdinaryDiffEqLinear/src/algorithms.jl
```diff
@@ -1,34 +1,76 @@
...
+@doc generic_solver_docstring("First order method using Cayley transformations.",
+    "CayleyEuler",
+    "Semilinear ODE solver",
+    "ref TBD", "", "")
 struct CayleyEuler <: OrdinaryDiffEqAlgorithm end
 
+@doc generic_solver_docstring("Exact solution formula for linear, time-independent problems.",
+    "LinearExponential",
+    "Semilinear ODE solver",
+    "ref TBD",
```
ArnoStrouwen, line 58: @ChrisRackauckas ref

lib/OrdinaryDiffEqLinear/src/algorithms.jl
```diff
@@ -1,34 +1,76 @@
...
+@doc generic_solver_docstring("Fourth Order Adaptive Magnus method.",
+    "MagnusAdapt4",
+    "Semilinear ODE solver",
+    "ref TBD", "", "")
 struct MagnusAdapt4 <: OrdinaryDiffEqAdaptiveAlgorithm end
 
+@doc generic_solver_docstring("First order method using Cayley transformations.",
+    "CayleyEuler",
+    "Semilinear ODE solver",
+    "ref TBD", "", "")
```
ArnoStrouwen, line 52: @ChrisRackauckas ref

lib/OrdinaryDiffEqLinear/src/algorithms.jl
```diff
@@ -1,34 +1,76 @@
...
         iop::Int
+        end
     end
     @eval $Alg(; krylov = false, m = 30, iop = 0) = $Alg(krylov, m, iop)
 end
 
+@doc generic_solver_docstring("Fourth Order Adaptive Magnus method.",
+    "MagnusAdapt4",
+    "Semilinear ODE solver",
+    "ref TBD", "", "")
```
ArnoStrouwen, line 46: @ChrisRackauckas ref

lib/OrdinaryDiffEqExponentialRK/src/algorithms.jl
```diff
@@ -22,11 +48,27 @@ end
...
 const ETD1 = NorsettEuler # alias
 
-for Alg in [:Exprb32, :Exprb43]
-    @eval struct $Alg{CS, AD, FDT, ST, CJ} <:
-                 OrdinaryDiffEqAdaptiveExponentialAlgorithm{CS, AD, FDT, ST, CJ}
-        m::Int
-        iop::Int
+for (Alg, Description, Ref) in [
+            (:Exprb32, "3rd order adaptive Exponential-Rosenbrock scheme.", "ref TBD"),
+            (:Exprb43, "4th order adaptive Exponential-Rosenbrock scheme.", "ref TBD")]
```
ArnoStrouwen, line 53: @ChrisRackauckas refs

lib/OrdinaryDiffEqExponentialRK/src/algorithms.jl
```diff
@@ -37,12 +79,37 @@ for Alg in [:Exprb32, :Exprb43]
...
-        m::Int
-        iop::Int
+for (Alg, Description, Ref) in [
+    (:Exp4, "4th order EPIRK scheme.", "REF TBD")
+    (:EPIRK4s3A, "4th order EPIRK scheme with stiff order 4.", "REF TBD")
+    (:EPIRK4s3B, "4th order EPIRK scheme with stiff order 4.", "REF TBD")
+    (:EPIRK5s3, "5th order “horizontal” EPIRK scheme with stiff order 5. Broken.", "REF TBD")
+    (:EXPRB53s3, "5th order EPIRK scheme with stiff order 5.", "REF TBD")
+    (:EPIRK5P1, "5th order EPIRK scheme", "REF TBD")
+    (:EPIRK5P2, "5th order EPIRK scheme", "REF TBD")
```
ArnoStrouwen, line 89: @ChrisRackauckas refs

docs/src/explicit/SymplecticRK.md
ChrisRackauckas: It needs to have a special function form. Grab the symplectic example from the test.

docs/src/explicit/RKN.md
ChrisRackauckas: Yeah it needs a SecondOrderODEProblem, IIRC there's one in the test.

lib/OrdinaryDiffEqLinear/src/algorithms.jl
```diff
@@ -1,34 +1,76 @@
...
+    (:MagnusGL6, "Sixth order Magnus method approximated using Gauss-Legendre quadrature.", "ref TBD"),
+    (:MagnusGL8, "Eighth order Magnus method approximated using Newton-Cotes quadrature.", "ref TBD"),
+    (:MagnusNC8, "Eighth order Magnus method approximated using Gauss-Legendre quadrature.", "ref TBD"),
+    (:MagnusGL4, "Fourth order Magnus method approximated using Gauss-Legendre quadrature.", "ref TBD"),
+    (:RKMK2, "Second order Runge–Kutta–Munthe-Kaas method.", "ref TBD"),
+    (:RKMK4, "Fourth order Runge–Kutta–Munthe-Kaas method.", "ref TBD"),
+    (:LieRK4, "Fourth order Lie Runge-Kutta method.", "ref TBD"),
+    (:CG2, "Second order Crouch–Grossman method.", "ref TBD"),
+    (:CG3, "Third order Crouch-Grossman method.", "ref TBD"),
+    (:CG4a, " Fourth order Crouch-Grossman method.", "ref TBD")]
```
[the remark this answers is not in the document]
ChrisRackauckas: https://github.com/SciML/OrdinaryDiffEq.jl/issues/1066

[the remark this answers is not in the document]
ChrisRackauckas: https://github.com/SciML/OrdinaryDiffEq.jl/issues/1064

lib/OrdinaryDiffEqExponentialRK/src/algorithms.jl
ChrisRackauckas: https://github.com/SciML/OrdinaryDiffEq.jl/issues/394 has the complete set of references.

lib/OrdinaryDiffEqLinear/src/algorithms.jl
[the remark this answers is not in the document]
ArnoStrouwen: What is Hairer III? I can't find a third volume.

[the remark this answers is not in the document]
ArnoStrouwen: MagnusMidpoint and MagnusLeapfrog are still open.

lib/OrdinaryDiffEqLinear/src/algorithms.jl
ArnoStrouwen: I copied them from ODEExponentialRK. Reopen if that is not accurate.

lib/OrdinaryDiffEqStabilizedIRK/src/algorithms.jl
```diff
@@ -1,3 +1,12 @@
+@doc generic_solver_docstring("Description TBD",
+    "IRKC",
+    "Stabalized Implicit Runge Kutta method.",
+    "REF TBD",
```
ArnoStrouwen, line 4: @ChrisRackauckas ref and description for this implicit stabilized method.

docs/src/implicit/BDF.md
```diff
@@ -0,0 +1,45 @@
...
+IMEXEulerARK
+```
+
+### DAE
+
+```@docs
+DImplicitEuler
+DABDF2
+DFBDF
+```
```
ArnoStrouwen, line 45: It is a bit weird that BDF houses these exotic solvers.

docs/src/imex/IMEXMultistep.md
```diff
@@ -0,0 +1,10 @@
+# IMEX Multistep Methods
+
+This article is a stub and likely needs a specialized example.
```
ArnoStrouwen, line 3: @ChrisRackauckas

lib/OrdinaryDiffEqIMEXMultistep/src/algorithms.jl
```diff
@@ -21,6 +25,10 @@ function CNAB2(; chunk_size = Val{0}(), autodiff = Val{true}(), standardtag = Va
         extrapolant)
 end
 
+@doc generic_solver_docstring("TBD",
+    "CNLF2",
+    "IMEX Multistep method.",
+    "REF TBD", "", "")
```
ArnoStrouwen, line 31: CNAB2 and CNLF2 description and reference @ChrisRackauckas.

docs/src/explicit/QPRK.md
```diff
@@ -0,0 +1,14 @@
+# OrdinaryDiffEqQPRK
+
+This article is a stub.
```
ArnoStrouwen, line 3: Q is quadruple?

docs/src/semilinear/Linear.md
ChrisRackauckas: There is an example in the docs right now IIRC that you can copy.

lib/OrdinaryDiffEqLinear/src/algorithms.jl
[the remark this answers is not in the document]
ChrisRackauckas: Hairer III = Hairer Geometric Integration https://link.springer.com/book/10.1007/3-540-30666-8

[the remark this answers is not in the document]
ChrisRackauckas: MagnusMidpoint and MagnusLeapfrog might also be in that book? I need to double check.

docs/src/explicit/QPRK.md
ChrisRackauckas: Quad Precision, it's specialized RK methods for quad-precision arithmetic, so values in the Float128 range.

lib/OrdinaryDiffEqIMEXMultistep/src/algorithms.jl
ChrisRackauckas: Crank-Nicholson Adams-Bashforth 2, and Crank-Nicholson Leapfrong 2. For the first, https://global-sci.org/intro/article_detail/ijnamB/229.html and https://onlinelibrary.wiley.com/doi/epdf/10.1002/fld.2035. https://www.sciencedirect.com/science/article/abs/pii/S0893965920302445 and https://www.sciencedirect.com/science/article/pii/S0377042714004336 for the other.

docs/src/implicit/BDF.md
ChrisRackauckas: ImplicitEuler is equivalent to BDF 1.  https://en.wikipedia.org/wiki/Backward_differentiation_formula . SBDF are just IMEX BDF methods which treat a part of it explicitly.

lib/OrdinaryDiffEqLinear/src/algorithms.jl
ChrisRackauckas: I believe this is also in Hairer Geometric.

lib/OrdinaryDiffEqLinear/src/algorithms.jl
ChrisRackauckas: Will need to double check against the code, but I believe it was this one https://link.springer.com/article/10.1007/s10483-008-0901-5 IIRC.

lib/OrdinaryDiffEqLinear/src/algorithms.jl
ArnoStrouwen: It is this one:
https://www.cambridge.org/core/journals/acta-numerica/article/abs/liegroup-methods/856125FF1EAF7762DEF6E37EEBA9CA5F

docs/src/imex/IMEXMultistep.md
ArnoStrouwen: The example in the tests was pretty meh, not really having a stiff and non-stiff part, but will have to do for now.

lib/OrdinaryDiffEqStabilizedIRK/src/algorithms.jl
ArnoStrouwen: @ChrisRackauckas you missed this one.

ChrisRackauckas: Stabilized Explicit Runge-Kutta methods, like Runge-Kutta-Chebyshev methods and ROCK methods, are explicit methods which chain together many stages in a specific way to get large stability regions. they are made in such a way to converge to a large stability region, and thus suitable to stiff equations. However, they converge to having a large stability region in the direction of the negative real axis, and thus are only stable on a subset of stiff equations which are not dominated by large complex eigenvalues in the Jacobian. Stabilized implicit methods try to mitigate this problem by being an IMEX type scheme, requiring a SplitODEProblem where the splitting is designed to treat the large complex eigenvalues implicitly while treating the large real eigenvalues using a fast explicit stabilized RK type of method.

lib/OrdinaryDiffEqBDF/src/algorithms.jl
ChrisRackauckas: These should all be standard and documented? IIRC They are all already mentioned in the DifferentialEquations.jl docs.

lib/OrdinaryDiffEqStabilizedIRK/src/algorithms.jl
ArnoStrouwen: Still needs a ref.

docs/src/implicit/BDF.md
```diff
@@ -0,0 +1,46 @@
...
+IMEXEuler
+IMEXEulerARK
+```
+
+### DAE
+
+```@docs
+DImplicitEuler
+DABDF2
+DFBDF
```
ChrisRackauckas, line 45: Seems a bit rough to have the fully implicit DAE methods so deep in here, since they are a bit special

lib/OrdinaryDiffEqCore/src/doc_utils.jl
```diff
@@ -62,15 +60,48 @@ function explicit_rk_docstring(description::String,
...
+        references::String = "",
+        extra_keyword_description::String = "",
+        extra_keyword_default::String = "")
+    keyword_default = """
+    chunk_size = Val{0}(),
+    autodiff = true,
+    standardtag = Val{true}(),
+    concrete_jac = nothing,
+    diff_type = Val{:forward},
+    linsolve = nothing,
```
ArnoStrouwen, line 89: @ChrisRackauckas current docs don't describe what the default value of `nothing` does for `linsolve`.
https://docs.sciml.ai/DiffEqDocs/stable/features/linear_nonlinear/#Linear-Solvers:-linsolve-Specification

ChrisRackauckas: `nothing` just uses the `DefaultLinearSolver`

lib/OrdinaryDiffEqFIRK/src/algorithms.jl
```diff
@@ -1,20 +1,41 @@
...
+    - `step_limiter!`: TBD"""
+extra_keyword_default = """
+    extrapolant = :dense,
+    fast_convergence_cutoff = 1 // 5,
+    new_W_γdt_cutoff = 1 // 5,
+    controller = :Predictive,
```
ArnoStrouwen, line 25: @ChrisRackauckas 
For FIRK methods, this page only implements Predictive controller.
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/OrdinaryDiffEqFIRK/src/controllers.jl
Am I correct to assume that any other controller, such as PI, will thus not work?

Also, I don't see these methods in the `contollers.jl` file used anywhere else in `OrdinaryDiffEqFIRK`?

lib/OrdinaryDiffEqPDIRK/src/algorithms.jl
```diff
@@ -1,10 +1,26 @@
...
+    year={1990},
+    publisher={Oxford University Press}}""",
+    extra_keyword_description = """
+    - `nlsolve`: TBD,
+    - `extrapolant`: TBD,
+    - `threading`: TBD,
+    """,
+    extra_keyword_default = """
+    nlsolve = NLNewton(),
+    extrapolant = :constant,
```
ArnoStrouwen, line 21: @ChrisRackauckas similar comment to the controller, it is hard to know what other values might be allowed, since `extrapolant` does not seem to take local action in `OrdinaryDiffEqPDIRK`.

ChrisRackauckas: Some methods have a specialized controller that is highly recommended to be used. Other controllers will still work, but not as well.

ArnoStrouwen:
@mortenpi we have some large pages in these docs. On Windows I need to do `size_threshold_ignore = ["implicit\\Rosenbrock.md"])`, instead of `size_threshold_ignore = ["implicit/Rosenbrock.md"])`, is this expected behavior?

mortenpi:
> On Windows I need to do `size_threshold_ignore = ["implicit\\Rosenbrock.md"])`, instead of `size_threshold_ignore = ["implicit/Rosenbrock.md"])`, is this expected behavior?

No. We should be normalizing the paths so they work cross-platform. Do you mind opening an issue?

Just to clarify: if you use the Unix-y path (`"implicit/Rosenbrock.md"`), it then doesn't work on Windows, right? A workaround you can do to get the PR passing would be to set is as `joinpath("implicit", "Rosenbrock.md")`.

lib/OrdinaryDiffEqSSPRK/src/algorithms.jl
```diff
@@ -334,10 +367,19 @@ function KYKSSPRK42(stage_limiter!, step_limiter! = trivial_limiter!)
         False())
 end
 
-@doc explicit_rk_docstring("TBD",
```
ChrisRackauckas, line 337: Looks like this got deleted.

lib/OrdinaryDiffEqSSPRK/src/algorithms.jl
```diff
@@ -320,8 +320,41 @@ function SSPRK33(stage_limiter!, step_limiter! = trivial_limiter!)
         step_limiter!, False())
 end
 
-@doc explicit_rk_docstring("TBD",
-    "KYKSSPRK42")
+@doc explicit_rk_docstring(
+    "Low dissipation and dispersion Runge-Kutta schemes for computational acoustics",
+    "SHLDDRK_2N",
```
ChrisRackauckas, line 325: This should have been moved to low order RK.

PR #2375, opened 2024-08-12
Remove the InteractiveUtils dependency by reimplementing subtypes

ChrisRackauckas:
This cuts out a decent loading dependency


oscardssmith:
Can this really not just use a list? This seems like a fairly internal function to re-implement.

lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
```diff
@@ -53,6 +51,7 @@ else
     struct OrdinaryDiffEqTag end
 end
 
+include("utils.jl")
```
ChrisRackauckas, line 54:
```suggestion
```

PR #2382, opened 2024-08-15
Autocompat

ChrisRackauckas:
Using:

```julia
function autocompat(ctx=Pkg.Types.Context(); io=nothing)
    io = something(io, ctx.io)
    updated_deps = String[]
    for dep_list in (ctx.env.project.deps, ctx.env.project.weakdeps, ctx.env.project.extras),
        (dep, uuid) in dep_list

        compat_str = Pkg.Operations.get_compat_str(ctx.env.project, dep)
        isnothing(compat_str) || continue
        if uuid in ctx.env.manifest
            v = ctx.env.manifest[uuid].version
            v === nothing && (v = "<0.0.1, 1")
        else
            try
                pkg_versions = Pkg.Versions.VersionSpec([Pkg.Operations.get_all_registered_versions(
                    ctx, uuid)...])
                if isempty(pkg_versions)
                    @warn "No versions of $(dep) are registered. Possibly a Standard Library package."
                    v = "<0.0.1, 1"
                else
                    latest_version = Pkg.Operations.get_latest_compatible_version(
                        ctx, uuid, pkg_versions)
                    v = latest_version
                end
            catch err
                @error "Encountered Error $(err) while processing $(dep). Skipping."
                continue
            end
        end
        Pkg.Operations.set_compat(ctx.env.project, dep, string(v)) ||
            Pkg.Types.pkgerror("invalid compat version specifier \"$(string(v))\"")
        push!(updated_deps, dep)
    end
    if isempty(updated_deps)
        Pkg.printpkgstyle(io, :Info, "no misssing compat entries found. No changes made.";
            color=Base.info_color())
    elseif length(updated_deps) == 1
        Pkg.printpkgstyle(io, :Info,
            "new entry set for $(only(updated_deps)) based on its current version";
            color=Base.info_color())
    else
        Pkg.printpkgstyle(io,
            :Info,
            "new entries set for $(join(updated_deps, ", ", " and ")) based on their current versions";
            color=Base.info_color())
    end
    Pkg.Types.write_env(ctx.env)
    return Pkg.Operations.print_compat(ctx; io)
end

for x in readdir("/Users/chrisrackauckas/.julia/dev/OrdinaryDiffEq/lib")
    Pkg.activate("/Users/chrisrackauckas/.julia/dev/OrdinaryDiffEq/lib/$x")
    autocompat()
end
```


ChrisRackauckas:
This would have been much easier if https://github.com/JuliaLang/Pkg.jl/pull/3266 existed!

lib/OrdinaryDiffEqExponentialRK/Project.toml
```diff
@@ -4,30 +4,49 @@ authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
...
+ExponentialUtilities = "1.26.1"
+FastBroadcast = "0.3.5"
+LinearAlgebra = "<0.0.1, 1"
+LinearSolve = "2.32.0"
+MuladdMacro = "0.2.4"
+OrdinaryDiffEqCore = "<0.0.1, 1"
+OrdinaryDiffEqDifferentiation = "<0.0.1, 1"
+OrdinaryDiffEqSDIRK = "<0.0.1, 1"
+OrdinaryDiffEqTsit5 = "<0.0.1, 1"
+OrdinaryDiffEqVerner = "<0.0.1, 1"
```
avik-pal, line 32: Forgot to mention this on slack, for packages without a registered version it assumes stdlib and puts this in compat

ChrisRackauckas: That compat actually works fine though since the first version for everything is 1.0.

PR #2389, opened 2024-08-17
Split and inactivate increment functions

ChrisRackauckas:


ChrisRackauckas:
This is one step towards really finishing https://github.com/SciML/OrdinaryDiffEq.jl/pull/2282

PR #2393, opened 2024-08-19
Add cache stripping to solution stripping 

jClugstor:
## Checklist

## Additional context

Related to https://github.com/SciML/SciMLBase.jl/issues/629
I still need to add tests. I need to make a problem where the cache has `jac_config` and `grad_config`. And test that it strips the cache correctly. 

This way I don't have to make dispatches for every individual cache that has `jac_config` or `grad_config`. I had to use `eval(nameof(typeof(cache)))` since `typeof` has the parametric type information, so I can't put all `nothing`s into `typeof` of a concrete cache type.

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/interp_func.jl
+++ b/lib/OrdinaryDiffEqCore/src/interp_func.jl
@@ -59,3 +59,24 @@ function InterpolationData(id::InterpolationData, f)
         id.differential_vars,
         id.sensitivitymode)
 end
+
+# strip interpolation of function information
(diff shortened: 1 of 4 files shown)


jClugstor:
Actually, I don't need to use eval. These changes should work better.

jClugstor:
I believe this should be it for most of https://github.com/SciML/SciMLBase.jl/issues/629

I think I'll need to add a test to SciMLBase making sure that `strip_solution` on a solution from an in place problem using an implicit method actually goes through and strips everything.

Issue #233, opened 2017-12-29
Use more metaprogramming in `src/integrators`


YingboMa:
Currently, the code in `src/integrators` is very repetitive and I think it can be cleaned up by some metaprogramming techniques. One concern that @ChrisRackauckas has about metaprogramming in the integrators is that it cannot handle zeros in tableaux automatically because the tableaux are not visible at parse time as they are in a struct. I think we can avoid this problem by just using matrices to store the tableaux, so the compiler can perform constant folding. Also, we can implement fast tableau methods from this. Here is an example.

```julia
T = T2 = Float64
a21 = T(0.5)
a32 = T(0.75)
a41 = T(0.2222222222222222)
a42 = T(0.3333333333333333)
a43 = T(0.4444444444444444)
c1 = T2(0.5)
c2 = T2(0.75)
btilde1 = T(0.06944444444444445)
btilde2 = T(-0.08333333333333333)
btilde3 = T(-0.1111111111111111)
btilde4 = T(0.125)

A = [
 0    0    0    0
 a21  0    0    0
 0    a32  0    0
 a41  0    a43  0
]
C = [ c1, c2, 0 ]
B = [ btilde1, btilde2, btilde3, btilde4 ]
macro p(ex::Symbol, expr::Expr)
    Expr(:call, :push!, :($ex.args), expr)
end

tight_loop(ex) = Expr(:block,
                      Expr(:macrocall, Symbol("@tight_loop_macros"), LineNumberNode,
                           Expr(:for, :(i = uidx),
                                Expr(:block))))

applyA(n) = Expr(:call, :+, :(uprev[i]),
                 Expr(:call, :*, :dt,
                      Expr(:call, :+, [:($(A[n+1, j]) * $(Symbol(:k, j))[i]) for j in 1:n]...)))

#function perform_step!(integrator, tab, cache, repeat_step, Val{true})
    #A, B, C = tab
    body = Expr(:macrocall, Symbol("@inbounds"), LineNumberNode,
                Expr(:macrocall, Symbol("@muladd"), LineNumberNode,
                     Expr(:block)))
    ex   = body.args[end].args[end]
    @p ex Expr(:macrocall, Symbol("@unpack"), LineNumberNode, :((t, dt, uprev, u, f) = integrator))
    @p ex :( k1 = integrator.fsalfirst )
    @p ex :( uidx = eachindex(uprev) )
    @p ex :( tmp = similar(uprev) )
[...]

PR #2430, opened 2024-08-25
Perform step refactor for rosenbrock

ParamThakkar123:
## Checklist

## Additional context

solves a part of #233


ParamThakkar123:
image

@ChrisRackauckas @oscardssmith  I am getting a error here saying RosenbrockCache has no field du1. This error occurs in derivative_utils.jl. I changed the line to dus so that it matches with the fields present in RosenbrockCache but that creates and error in NLNewtonCache. 

Just had a question, how do I make a change so that this error gets solve and doesn't propagate or create errors in other caches ?

ParamThakkar123:
> image
> 
> @ChrisRackauckas @oscardssmith I am getting a error here saying RosenbrockCache has no field du1. This error occurs in derivative_utils.jl. I changed the line to dus so that it matches with the fields present in RosenbrockCache but that creates and error in NLNewtonCache.
> 
> Just had a question, how do I make a change so that this error gets solve and doesn't propagate or create errors in other caches ?

I managed to solve the other ones. But needed your suggestion on this one

ChrisRackauckas:
`du1` is used in the Jacobian calculation, so it's required for all methods that have a `build_jac_config`.

ParamThakkar123:
> `du1` is used in the Jacobian calculation, so it's required for all methods that have a `build_jac_config`.
Should I be adding that explicitly as a field there ??

ChrisRackauckas:
yes

ParamThakkar123:
Got it. I think I made similar mistakes in tsit. I will make a fix there too.

ChrisRackauckas:
> Got it. I think I made similar mistakes in tsit. I will make a fix there too.

No, as mentioned in other places, Tsit is completely different. It already has ExplicitRK. It's more about feature completeness of ExplicitRK and its performance.

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_perform_step.jl
```diff
@@ -1363,9 +1363,11 @@ end
             end
         end
 
-        stage_limiter!(u, integrator, p, t + c[i] * dt)
-        f(du1, u, p, t + c[i] * dt)
-        OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
+        if i >= 1 && i <= 4
```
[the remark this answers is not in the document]
ParamThakkar123: yes

oscardssmith:
Out of place is now working.

oscardssmith:
OK. I think this is ready. Lets see if CI agrees.

oscardssmith:
@ParamThakkar123 I just undid your most recent push. `_ode_addsteps` doesn't have the integrator, so the reltol needs to come from the cache.

ParamThakkar123:
> @ParamThakkar123 I just undid your most recent push. `_ode_addsteps` doesn't have the integrator, so the reltol needs to come from the cache.

Yes. I saw your update. Got it.

ParamThakkar123:
@oscardssmith just a question. How do I know whether an integrator exists for a solver or not from the code itself ?? I can keep this in mind while refactoring further solvers

ParamThakkar123:
@oscardssmith What changes must be made in order to deal with the fastbroadcast error coming up ??

oscardssmith:
> ow do I know whether an integrator exists for a solver or not from the code itself 

This is only a thing in the `add_steps!` portions of the code.

oscardssmith:
The fast_broadcast issue means that there is a `_vec` missing (probably when copying from the ks to the us.

ParamThakkar123:
> The fast_broadcast issue means that there is a `_vec` missing (probably when copying from the ks to the us.

image

The error occurs in stiff_addsteps here. the `_vec` is already present

ParamThakkar123:
image

@oscardssmith tried the fixes suggested for this but still found the tests failing at that same point

ParamThakkar123:
@oscardssmith Any changes to be made on this PR from my end ?? I see that there are still some tests failing with the precision errors

oscardssmith:
I think I can finish it from here. the remaining errors are obscure and complicated. (currently fixing up resizing)

ParamThakkar123:
Okay, @ChrisRackauckas @oscardssmith I will start working on other solvers then. Which solvers should be the next one ?? Though I started working on SDIRK and Tsit on my own but just wanted to know if you have any priorities with respect to solvers regarding which ones should be done first

oscardssmith:
stats look identical, and performance looks marginally better somehow? I think this is ready to merge.
```
function rober(u, p, t)
   y₁, y₂, y₃ = u
   k₁, k₂, k₃ = p
   [-k₁ * y₁ + k₃ * y₂ * y₃,
       k₁ * y₁ - k₃ * y₂ * y₃ - k₂ * y₂^2,
       k₂ * y₂^2]
end
prob_rober = ODEProblem(rober, [1.0, 0.0, 0.0], (0.0, 1e3), (0.04, 3e7, 1e4))

# PR
julia> @benchmark sol = solve(prob_rober, Rodas4())
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  190.246 μs … 127.780 ms  ┊ GC (min … max):  0.00% … 99.72%
 Time  (median):     220.651 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   281.783 μs ±   1.310 ms  ┊ GC (mean ± σ):  13.86% ±  8.56%

   ██▅▄▃▂▁▂▃▃▁                                                  ▂
  █████████████▇▇▆▇▆▅▆▅▄▃▄▁▃▁▄▄▃▁▃▄▁▃▄▃▄▄▁▅▄▄▄▅▅▄▆▆▅▄▆▅▄▆▇▆▅▆▆▆ █
  190 μs        Histogram: log(frequency) by time       1.02 ms <
# v 6.89
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  203.744 μs … 129.049 ms  ┊ GC (min … max):  0.00% … 99.74%
 Time  (median):     234.780 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   277.780 μs ±   1.306 ms  ┊ GC (mean ± σ):  12.14% ±  8.67%

      ▄▆██▇▆▅▄▃▂▂▁▁              ▁▂▁▁▁▁▂▁                       ▂
  ▆▆▆███████████████▇▆▇██▇▆▅▆▆▅▆██████████▇█▇▇▅▄▄▅▁▃▄▄▃▅▄▃▄▄▁▄▅ █
  204 μs        Histogram: log(frequency) by time        447 μs <

 Memory estimate: 452.92 KiB, allocs estimate: 5914.
 ```

**Comment by ParamThakkar123** [comment]
@oscardssmith there's an undef error saying du is not defined in stiff_addsteps

**Comment by ParamThakkar123** [comment]
Should I push a commit to fix it now

**Comment by oscardssmith** [comment]
@ParamThakkar123 where was that error? I didn't see it...

**Comment by ParamThakkar123** [comment]
image

@oscardssmith here's a snapshot of the error. It can be seen in the buildkite tests inside integratorII_v1 tests

**Comment by oscardssmith** [comment]
@ParamThakkar123 please stop pushing incorrect fixes. `du` is not supposed to be in the `ConstantCache`.

**Comment by ParamThakkar123** [comment]
@oscardssmith okay sorry for that

**Comment by oscardssmith** [comment]
@ChrisRackauckas good to merge?

File: lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_caches.jl [hunk 2]
```diff
@@ -663,99 +697,23 @@ end
...
-struct Rodas4ConstantCache{TF, UF, Tab, JType, WType, F, AD} <: RosenbrockConstantCache
+struct Rodas4ConstantCache{TF, UF, Tab, JType, WType, F, AD, rateType} <: RosenbrockConstantCache
     tf::TF
     uf::UF
     tab::Tab
     J::JType
     W::WType
     linsolve::F
     autodiff::AD
+    ks::Vector{rateType}
```
ChrisRackauckas, line 708: why would this cache?

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_caches.jl
```diff
@@ -764,190 +722,62 @@ function alg_cache(alg::Rodas4, u, rate_prototype, ::Type{uEltypeNoUnits},
     J, W = build_J_W(alg, u, uprev, p, t, dt, f, uEltypeNoUnits, Val(false))
     linprob = nothing #LinearProblem(W,copy(u); u0=copy(u))
     linsolve = nothing #init(linprob,alg.linsolve,alias_A=true,alias_b=true)
+    ks = Vector{typeof(rate_prototype)}(undef, 6)
     Rodas4ConstantCache(tf, uf,
-        Rodas4Tableau(constvalue(uBottomEltypeNoUnits),
+        tabtype(alg)(constvalue(uBottomEltypeNoUnits),
```
ChrisRackauckas, line 727: does this infer correctly?

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_perform_step.jl
```diff
@@ -1246,138 +1228,89 @@ end
...
+        end
+        linsolve_tmp = @.. du + dtd[stage] * dT + linsolve_tmp1
 
-    if mass_matrix === I
-        linsolve_tmp = du + (dtC61 * k1 + dtC62 * k2 + dtC65 * k5 + dtC64 * k4 + dtC63 * k3)
-    else
-        linsolve_tmp = du +
-                       mass_matrix *
-                       (dtC61 * k1 + dtC62 * k2 + dtC65 * k5 + dtC64 * k4 + dtC63 * k3)
+        ks[stage] = _reshape(W \ -_vec(linsolve_tmp), axes(uprev))
```
ChrisRackauckas, line 1258: mutation of this form is going to break downstream autodiff with some algorithms.

ChrisRackauckas: It needs to be something like a map.

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_caches.jl
oscardssmith: this was cached because otherwise we needed to make this vector new at each step. (since before we had separate variables `k1`, `k2` etc but we needed them to exist in an indexable collection.

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_caches.jl
oscardssmith: yes. each alg is it's own type.

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_perform_step.jl
oscardssmith: this can't be something like a map (since each `k` needs the previous `ks`). Would using a Tuple with Accessors.jl work?

ChrisRackauckas: Yes

oscardssmith: fixed.

oscardssmith:
ok, ready to merge?

ChrisRackauckas:
There's still a lot more to do of course with Rosenbrock, since there's now two generic caches (generic_rosenbrock + this one), and this only got the 4th order.

ParamThakkar123:
Yes @ChrisRackauckas . I am working on that part too. Will start with a new PR addressing this in an hour or two.

PR #2431, opened 2024-08-25
Perform step refactor for Tsit

ParamThakkar123:
## Checklist

## Additional context

solves a part of #233


oscardssmith:
how does the performance compare?

ChrisRackauckas:
Explicit RK is a bit tougher. It has a lot harder performance-wise. It needs a good understanding of https://github.com/SciML/OrdinaryDiffEq.jl/issues/233#issuecomment-2081665059

ChrisRackauckas:
I don't understand the point of this PR. As mentioned in the other threads and the comment above, the general tableau for ERK methods is about feature-completeness and correctness of ExplicitRK and subsuming other RKs into that.

ParamThakkar123:
> I don't understand the point of this PR. As mentioned in the other threads and the comment above, the general tableau for ERK methods is about feature-completeness and correctness of ExplicitRK and subsuming other RKs into that.

Actually, all I did here is refactor of perform_step implementations focusing only on the tsit solvers.  The Tsit tableau was implemented in the library for this solvers. So I just refactored it that way. Focusing only on this.

ChrisRackauckas:
Yes, that does not make sense given ExplicitRK exists.

ParamThakkar123:
> Yes, that does not make sense given ExplicitRK exists.

So should I be doing ExplicitRK first and then combining tsit into it ??

ChrisRackauckas:
ExplicitRK exists and has a Tsit5 tableau already. Its mostly about feature-completeness and performance now.

ParamThakkar123:
So what should be done further as far as this PR is concerned ?

ParamThakkar123:
@ChrisRackauckas should I move on with the refactoring of ExplicitRK first ??

ParamThakkar123:
> No it's already "refactored". It's about improving it and feature completing it. Testing it against the other implementations of RKs and making sure it floating point matches, and making sure default controller values are set appropriately.

Got it. Should I work towards feature completing it. What should be the first thing that should be done to start off with that ?

ChrisRackauckas:
Run Tsit5 and the tableau form of Tsit5 and get the times array to match exactly. I think the controller params are probably the place to start.

ParamThakkar123:
> Run Tsit5 and the tableau form of Tsit5 and get the times array to match exactly. I think the controller params are probably the place to start.

@ChrisRackauckas I ran the Tsit5 and its tableau form on a sample DE as follows : 
image

and the times arrays seem to match exactly

ChrisRackauckas:
What about with stiffness detection and switching?

ParamThakkar123:
> What about with stiffness detection and switching?

image

@ChrisRackauckas

ChrisRackauckas:
I mean build an autoswitch alg that uses the ExplicitRK version, does that match the Tsit5 behavior?

PR #2436, opened 2024-08-27
Fix and add test coverage for custom jacobians in DAEFunctions.

termi-official:


lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
```diff
@@ -899,7 +901,11 @@ function build_J_W(alg, u, uprev, p, t, dt, f::F, ::Type{uEltypeNoUnits},
         end
     else
         J = if !IIP && DiffEqBase.has_jac(f)
-            f.jac(uprev, p, t)
+            if f isa DAEFunction
+                f.jac(uprev, uprev, p, 1.0, t)
```
ChrisRackauckas, line 905:
```suggestion
                f.jac(uprev, uprev, p, one(t), t)
```

PR #2439, opened 2024-08-28
canonical docstrings

ArnoStrouwen:
Make docstrings show up properly.


ArnoStrouwen:
I failed to make `QNDF` show up properly, there is something else wrong with that one.

PR #2443, opened 2024-08-29
Add check whether relaxation is better than the full step

SouthEndMusic:
Not sure whether this should be upstreamed to `LineSearches.jl`

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
@@ -456,8 +456,11 @@ function relax!(dz, nlsolver::AbstractNLSolver, integrator::DEIntegrator, f::TF,
         end
         α0 = one(eltype(ustep))
         ϕ0, dϕ0 = ϕdϕ(zero(α0))
-        α, _ = linesearch(ϕ, dϕ, ϕdϕ, α0, ϕ0, dϕ0)
-        @.. dz = dz * α
+        α, ϕα = linesearch(ϕ, dϕ, ϕdϕ, α0, ϕ0, dϕ0)
+        # Check whether relaxation is better than the full step
+        if ϕα < ϕ(1)
+            @.. dz = dz * α
+        end
         return dz
     end
 end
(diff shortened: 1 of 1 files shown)


oscardssmith:
Presumably it should be upstreamed.

SouthEndMusic:
@oscardssmith I still feel some work on this side is required as well. Most line-search algorithms from `LineSearches.jl` error when `ϕ'(0) < 0` but that shouldn't stop the ODE solver. The relaxation can be refactored so that line-search based relaxation is only performed when that condition is met. I think that fixes the problem I'm encountering. Then upstream this check should be added as wel to `BackTracking`.

SouthEndMusic:
Hm, on closer inspection the condition for `BackTracking` is somewhat more complicated, so it can not be done that generically. I suppose putting the line-search in try-catch is not desirable?

SouthEndMusic:
Maybe the method for computing the derivative of the residual with respect to the relative step size is just inaccurate for my particular use case, which would justify my current version of this PR again,

oscardssmith:
one other question: If you use the (still somewhat undocumented NonlinearSolveAlg) does this bug still appear?

ChrisRackauckas:
We really should just be using NonlinearSolveAlg and pushing that forward rather than small improvements to the other.

SouthEndMusic:
I'm trying to run the ODE solver with `nlsolve = NonlinearSolveAlg(NewtonRaphson(; linesearch = BackTracking(), autodiff = AutoForwardDiff())` but I get an error because my rhs is called with `eltype(du) = Float64` and `eltype(u) = ForwardDiff.Dual{...}`.

oscardssmith:
oh good point, I need to fix that...

SouthEndMusic:
Looking at how `NonlinearSolve.jl` deals with this [here](https://github.com/SciML/NonlinearSolve.jl/blob/e457a984e6d2b9e27580322cb7a2a2288119773e/src/globalization/line_search.jl#L200-L222), that looks a lot like what I propose above. So indeed, using `NonlinearSolveAlg` will probably solve the problem. I'll close the issue + PR once I have been able to test it.

SouthEndMusic:
@oscardssmith btw, currently my run fails because `remake` is not defined here:

https://github.com/SciML/OrdinaryDiffEq.jl/blob/1eef9db17600766bb71e7dce0cb105ae5f99b2a5/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl#L71

PR #2444, opened 2024-08-30
reuse jacobians in NonlinearSolveAlg

oscardssmith:
This seems to just work. I think I only turned off reuse because it was making debugging the other bugs in this harder.


ChrisRackauckas:
Test that the old and new Newton get the same number of steps and Jacobians?

PR #2448, opened 2024-09-01
Perform_step! refactor for Rosenbrock part 2

ParamThakkar123:
## Checklist

## Additional context

solves a part of #233


lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_perform_step.jl
```diff
@@ -1600,66 +1600,35 @@ end
...
     integrator.k[1] = dense1
     integrator.k[2] = dense2
     integrator.k[3] = dense3
 end
 
-@muladd function perform_step!(integrator, cache::Rosenbrock5Cache, repeat_step = false)
-    @unpack t, dt, uprev, u, f, p = integrator
-    @unpack du, du1, du2, k1, k2, k3, k4, k5, k6, k7, k8, dT, J, W, uf, tf, linsolve_tmp, jac_config, atmp, weight, stage_limiter!, step_limiter! = cache
-    @unpack a21, a31, a32, a41, a42, a43, a51, a52, a53, a54, a61, a62, a63, a64, a65, C21, C31, C32, C41, C42, C43, C51, C52, C53, C54, C61, C62, C63, C64, C65, C71, C72, C73, C74, C75, C76, C81, C82, C83, C84, C85, C86, C87, gamma, d1, d2, d3, d4, d5, c2, c3, c4, c5 = cache.tab
+@muladd function perform_step!(integrator, cache::RosenbrockCache, repeat_step = false)
```
oscardssmith, line 1613: this should be deleted (since we already wrote this perform step above)

ParamThakkar123: But it that case I need to merge both of them right away

ParamThakkar123: @oscardssmith Should I merge them ??

oscardssmith: I don't think the one for Rodas4 should need any changes to work for Rodas5.

ParamThakkar123: @oscardssmith But I saw that the length of the dense vector in case of Rodas5 is 3 while in Rodas4 it's 2. Will any changes be required to be made to address that ?

oscardssmith: I believe that should be dealt with automatically. I made the code loop from `1:length(integrator.k)` so it should just work.

lib/OrdinaryDiffEqRosenbrock/src/stiff_addsteps.jl
```diff
@@ -558,54 +558,22 @@ function _ode_addsteps!(k, t, uprev, u, dt, f, p, cache::Rosenbrock5ConstantCach
     nothing
 end
 
-function _ode_addsteps!(k, t, uprev, u, dt, f, p, cache::Rosenbrock5Cache,
```
oscardssmith, line 561: This function also should be deleted in it's entirety (for the same reason as perform step).

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_tableaus.jl
```diff
@@ -339,167 +339,43 @@ function Rodas4P2Tableau(T, T2)
...
-    h47 = convert(T, -5.792426076169686)
-    h48 = convert(T, -5.32503859794143)
-
+    A = T[
+        0                  0                 0                   0                  0 0
+        2.0                0                 0                   0                  0 0
+        3.040894194418781  1.041747909077569 0                   0                  0 0
+        2.576417536461461  1.622083060776640 -0.9089668560264532 0                  0 0
+        2.760842080225597  1.446624659844071 -0.3036980084553738 0.2877498600325443 0 0
+        -14.09640773051259 6.925207756232704 -41.47510893210728  2.343771018586405  24.13215229196062 0
```
oscardssmith, line 358: This needs an extra row of `-14.09640773051259 6.925207756232704 -41.47510893210728 2.343771018586405 24.13215229196062 1 0` (and the rest of the matrix needs an extra column of `0`)

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_tableaus.jl
```diff
@@ -339,167 +339,43 @@ function Rodas4P2Tableau(T, T2)
...
+        -10.31323885133993   0                   0                    0                   0                    0                    0 0
+        -21.04823117650003   -7.234992135176716  0                    0                   0                    0                    0 0
+        32.22751541853323    -4.943732386540191  19.44922031041879     0                   0                    0                    0 0
+        -20.69865579590063   -8.816374604402768  1.260436877740897     -0.7495647613787146 0                    0                    0 0
+        -46.22004352711257   -17.49534862857472  -289.6389582892057    93.60855400400906   318.3822534212147    0                    0 0
+        34.20013733472935    -14.15535402717690  57.82335640988400     25.83362985412365   1.408950972071624    -6.551835421242162   0 0
+        42.57076742291101    -13.80770672017997  93.98938432427124     18.77919633714503   -31.58359187223370   -6.685968952921985   -5.810979938412932 0
+    ]
+    c = T2[0, 0.38, 0.3878509998321533, 0.4839718937873840, 0.4570477008819580]
```
oscardssmith, line 370: This is missing (I believe 2) `1`s at the end.

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_tableaus.jl
```diff
@@ -339,167 +339,43 @@ function Rodas4P2Tableau(T, T2)
...
+        -10.31323885133993   0                   0                    0                   0                    0                    0 0
+        -21.04823117650003   -7.234992135176716  0                    0                   0                    0                    0 0
+        32.22751541853323    -4.943732386540191  19.44922031041879     0                   0                    0                    0 0
+        -20.69865579590063   -8.816374604402768  1.260436877740897     -0.7495647613787146 0                    0                    0 0
+        -46.22004352711257   -17.49534862857472  -289.6389582892057    93.60855400400906   318.3822534212147    0                    0 0
+        34.20013733472935    -14.15535402717690  57.82335640988400     25.83362985412365   1.408950972071624    -6.551835421242162   0 0
+        42.57076742291101    -13.80770672017997  93.98938432427124     18.77919633714503   -31.58359187223370   -6.685968952921985   -5.810979938412932 0
+    ]
+    c = T2[0, 0.38, 0.3878509998321533, 0.4839718937873840, 0.4570477008819580]    
+    d = T[gamma, -0.1823079225333714636, -0.319231832186874912, 0.3449828624725343, -0.377417564392089818]
```
oscardssmith, line 371: This is missing (I believe 3) `0`s at the end.

ParamThakkar123:
Okay. Working on the changes

oscardssmith:
the interpolates still need some fixing.

ParamThakkar123:
@oscardssmith what can be a possible fix here ?? merge the constant caches of rodas4 and rodas5 ??

ParamThakkar123:
@ChrisRackauckas What changes do I need to make to the interpolant parts exactly ?? What can be a possible fix here ??

oscardssmith:
You should merge the constant caches as well, and then you need to change the ode_interpolant code to look at the alg order rather than the cache type (and ideally also be generic over alg order)

ParamThakkar123:
got it.

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_tableaus.jl
oscardssmith: Extra row still missing.

ParamThakkar123:
@oscardssmith I think there is a slight discrepancy with respect to the size of the vectors and matrices in Rodas4 and Rodas5. When I was merging the two the sizes of vectors and matrices in case of Rodas5 was 1 more than those in rodas4. The tests failing here seem to point to the same thing. For one solver the size of the vectors and matrices is appropriate while for the other solver it's smaller by one.

oscardssmith:
The Rodas5 matrices should all be bigger than those for Rodas4. If you get them to be the right size, however the math should all work out.

ParamThakkar123:
> The Rodas5 matrices should all be bigger than those for Rodas4. If you get them to be the right size, however the math should all work out.

Yeah. I think they should be brought to the same sizes in such a way that, the math that goes inside is not affected. But what values can be added ?? 1 or 0 ??

oscardssmith:
You accidentally had added the last stage of `Rodas5` to Rodas5P`. Deleting that error (and removing the extra `k`) fixes it.

ParamThakkar123:
image


@oscardssmith What changes must be made here ? The size of the k vector is found to be 2 but it accesses the 3rd element. and I didn't find where the k is defined other than as a parameter.

ParamThakkar123:
> image
> 
> @oscardssmith What changes must be made here ? The size of the k vector is found to be 2 but it accesses the 3rd element. and I didn't find where the k is defined other than as a parameter.

do I need to add some check for this ?? @oscardssmith

oscardssmith:
the problem is you're mashing together the 4th and 5th order interpolants, You need to make this method check the order and do the appropriate interpolation (see the 4th order ones higher up in the file).

ParamThakkar123:
@oscardssmith and just one more question (sorry for so many question, just trying to understand everything well), any algorithm has two cache, the normal one and the constant cache. How they should be handled ?? because algorithm is same, so will be the alg_order, but caches and their corresponding interpolants are different at some places

oscardssmith:
the interpolants are the same. The difference is just whether it's using an in place or out of place computation.

ParamThakkar123:
@oscardssmith anything left on this PR to be done ?? I saw only precision based errors where the tests are failing

oscardssmith:
It's not yet ready to merge, but I think I can take it from here. There is something fascinating going wrong with `Rodas5`. The thing that makes very little sense to me at this point is that from what I can see, the problem only exists for in place `Rodas5` (but not out of place `Rodas5` or in place  `Rodas5P`...)

ParamThakkar123:
@oscardssmith meanwhile, should I start fixing the other solvers ?

oscardssmith:
If you want to make another PR for the 3rd order rosenbrock methods that would be great (make the PR on top of this one)

ParamThakkar123:
@oscardssmith a new PR started for refactor of 3rd order rosenbrock #2463

ParamThakkar123:
@oscardssmith anything fixes to make here ??

oscardssmith:
I think I need to talk with @ChrisRackauckas tomorrow about some of the details of this PR.

PR #2453, opened 2024-09-04
ci(CompatHelper): fix CompatHelper

thazhemadam:
CompatHelper has been broken for a long time now. Fix it by specifying the right `subdirs`.
* `test/down` should be `test/downstream`
* `diagrams` project does not exist


ChrisRackauckas:
what about the lib directories?

PR #2458, opened 2024-09-06
Better warning on lack of initialization algorithm and slim Rosenbrock

ChrisRackauckas:
This fixes the overly large Rosenbrock by giving and erroring mechanism to tell users that adding a nonlinear solver is required for DAEs. Before, we simply put it as a dep to Rosenbrock to fix the issues with the split PR with a note that this would be removed shortly. This does the removal and instead throws a contextual error saying you just need to add that library if you want that functionality. This should make Rosenbrocks a lot leaner for ODEs, and this will make cases where DAE intiialization is required on an explicit method give a much better error. In some cases, an error will go away since the trivial case is now contained in Core, meaning that if a nonlinear solver isn't actually required (analytically solved initialization) it will just handle it without erroring whereas before it would require the nonlinear solver libraries (and then not use them).


oscardssmith:
This appears to have broken precompilation.

ChrisRackauckas:
How? CI doesn't see it.

SebastianM-C:
looks like method overwriting
```
WARNING: Method definition default_nlsolve(Nothing, Base.Val{true}, Any, SciMLBase.NonlinearProblem{uType, isinplace, P, F, K, PT} where PT where K where F where P where isinplace where uType) in module OrdinaryDiffEqCore at /home/sebastian/.julia/packages/OrdinaryDiffEqCore/rRqPh/src/initialize_dae.jl:123 overwritten in module OrdinaryDiffEqNonlinearSolve at /home/sebastian/.julia/packages/OrdinaryDiffEqNonlinearSolve/0gbhD/src/initialize_dae.jl:1.
ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
```

SebastianM-C:
Using the current master I see
```
Precompiling project...
  ? OrdinaryDiffEqNonlinearSolve
  ? OrdinaryDiffEqPDIRK
  ? OrdinaryDiffEqStabilizedIRK
  ? OrdinaryDiffEqSDIRK
  ? OrdinaryDiffEqFIRK
  ? OrdinaryDiffEqIMEXMultistep
  ? OrdinaryDiffEqBDF
  ? OrdinaryDiffEqDefault
  ? OrdinaryDiffEq
```
I'm not sure what the `?` means

PR #2464, opened 2024-09-08
Make examples runnable

ArnoStrouwen:
Most of the lorenz examples actually don't work if you try them.
Let's turn them into proper `@example`s.


ArnoStrouwen:
Not sure why Documenter suddenly starts complaining about missing docstrings again.
By adding the libs to the doc project it seems like the stable versions are grabbed instead of dev.

PR #2472, opened 2024-09-19
fix `ForwardDiff` `jacobian!` stats

oscardssmith:
found while investigating https://github.com/SciML/OrdinaryDiffEq.jl/pull/2448


oscardssmith:
StochasticDiff failure is unrelated (caused by https://github.com/SciML/OrdinaryDiffEq.jl/pull/2297 and not picked up in tests for weird reasons, and fixed by https://github.com/SciML/StochasticDiffEq.jl/pull/585)

PR #2481, opened 2024-09-29
EnzymeExt Use inactive_noinl

wsmoses:
This will improve performance, if it still passes.


oscardssmith:
What's the difference?

PR #2487, opened 2024-10-08
fix initialize! of  FBDF

oscardssmith:
fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2476. This presumably should be tested, but it appears that none of the ODE bdfs have any tests at all????


oscardssmith:
CI looks very unhappy but possibly for only unrelated reasons?

PR #2488, opened 2024-10-09
fix convergence test CI failure

oscardssmith:
order is 8, not 8.5 apparently.

@Shreyas-Ekanathan can you confirm that this is expected?


oscardssmith:
you say that, but setting both to 8 seems to have passed tests...

ChrisRackauckas:
Is this a hitting floating point error issue?

Shreyas-Ekanathan:
I think so, thats why we used `BigFloat` on the adaptive version

PR #2489, opened 2024-10-10
Fix cache_strip for default algs

jClugstor:
## Checklist

- [x ] Appropriate tests were added
- [ x] All documentation related to code changes were updated
- [ x] The new code follows the
- [ x] Any new documentation only uses public API
  
## Additional context

This should get rid of all of the stuff inside of the cache, but keeping the main type of the cache. Also fixes cache stripping for default algorithms.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEq"
 uuid = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
-version = "6.89.1"
+version = "6.89.0"
 
 [deps]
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEq"
 uuid = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
-version = "6.89.0"
+version = "6.89.1"
```
SebastianM-C, line 4: Don't we want to bump `OrdinaryDiffEqCore`? I think that if we want a bump in `OrdinaryDiffEq`, that would also need to bump the minimum compat on `OrdinaryDiffEqCore` in `OrdinaryDiffEq` besides bumping the version number.

jClugstor: Ah yeah, you're right.

Issue #2492, opened 2024-10-11
Floating point error using adjoint_sensitivities when tspan includes 16f0 in v6.89.0


jesseylin:
**Describe the example**

Adjoint solve immediately fails with a floating point precision error exactly at 16f0, regardless of the ODEFunction, solver, etc.
Seems to not occur in `v6.59.3`.

(Originally from https://github.com/SciML/SciMLSensitivity.jl/issues/1133)

**Minimal Reproducible Example 👇**

```julia
using Random
using OrdinaryDiffEq
using SciMLSensitivity
using LinearAlgebra

using Test

# MWE with OrdinaryDiffEq v6.59.3
flow = function (u, p, t)
    -u + p
end

rng = Xoshiro(1234)
ps = randn(rng, 11)
s0 = randn(rng, 11)

tspan = (0.0f0, prevfloat(16.0f0))
prob = ODEProblem(flow, s0, tspan, ps)
sol = solve(prob, Tsit5())
sensealg = GaussAdjoint(autojacvec=ReverseDiffVJP())

# at prevfloat(16f0)
_, gs = adjoint_sensitivities(sol, Tsit5(); sensealg, g=(u, p, t) -> norm(u))
gs
@test !iszero(gs)

# at 16f0
tspan = (0.0f0, 16.0f0)
prob = ODEProblem(flow, s0, tspan, ps)
sol = solve(prob, Tsit5())
_, gs = adjoint_sensitivities(sol, Tsit5(); sensealg, g=(u, p, t) -> norm(u))
@test !iszero(gs)
```

**Working deps**
```julia
Status `~/src/mwe/Project.toml`
⌃ [1dea7af3] OrdinaryDiffEq v6.59.3
⌃ [1ed8b502] SciMLSensitivity v7.51.0
  [37e2e46d] LinearAlgebra
  [9a3f8284] Random
```

**Broken deps**
```julia
Status `~/src/mwe_broken/Project.toml`
  [1dea7af3] OrdinaryDiffEq v6.89.0
  [1ed8b502] SciMLSensitivity v7.68.0
  [37e2e46d] LinearAlgebra
  [9a3f8284] Random
```

**Working manifest**
```julia
Status `~/src/mwe/Manifest.toml`
⌅ [47edcb42] ADTypes v0.2.7
  [621f4979] AbstractFFTs v1.5.0
  [7d9f7c33] Accessors v0.1.38
[...]

PR #2493, opened 2024-10-12
fix initdt calculation

oscardssmith:
fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2492.
That said, @ChrisRackauckas our initdt selection does seem like we never considered the case where `iszero(u0)`. When this happens, our `d₀` is 0, so our `dt₀ = smalldt`, which puts a hard cap on our return value of `100*smalldt`, which is often incredibly pessimistic. Would it be reasonable to have the 2nd stage of this algorithm loop when `100dt₀ < dt₁` (i.e. we will do an extra function call at  `t + dt₁` to see if our `dt₁` guess was reasonable. This would cost an extra function eval, but in this case, our `dt₁` is 300x bigger than our `d₀`, which will require taking multiple steps with basically no progress before our timestepping is reasonable.

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/initdt.jl
+++ b/lib/OrdinaryDiffEqCore/src/initdt.jl
@@ -8,8 +8,8 @@
     oneunit_tType = oneunit(_tType)
     dtmax_tdir = tdir * dtmax
 
-    dtmin = nextfloat(integrator.opts.dtmin)
-    smalldt = convert(_tType, oneunit_tType * 1 // 10^(6))
+    dtmin = nextfloat(max(integrator.opts.dtmin, eps(t)))
+    smalldt = max(dtmin, convert(_tType, oneunit_tType * 1 // 10^(6)))
 
     if integrator.isdae
         return tdir * max(smalldt, dtmin)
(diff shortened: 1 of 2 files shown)


lib/OrdinaryDiffEqCore/src/initdt.jl
```diff
@@ -235,7 +235,7 @@ end
     oneunit_tType = oneunit(_tType)
     dtmax_tdir = tdir * dtmax
 
-    dtmin = nextfloat(integrator.opts.dtmin)
+    dtmin = nextfloat(max(integrator.opts.dtmin, eps(t)))
     smalldt = convert(_tType, oneunit_tType * 1 // 10^(6))
```
ChrisRackauckas, line 239:
```suggestion
    smalldt = max(dtmin, convert(_tType, oneunit_tType * 1 // 10^(6)))
```

ChrisRackauckas:
> Would it be reasonable to have the 2nd stage of this algorithm loop when 100dt₀ < dt₁ (i.e. we will do an extra function call at t + dt₁ to see if our dt₁ guess was reasonable. This would cost an extra function eval, but in this case, our dt₁ is 300x bigger than our d₀, which will require taking multiple steps with basically no progress before our timestepping is reasonable

Just doing some starting acceleration might make more sense. But that doesn't sound so unreasonable.

PR #2495, opened 2024-10-12
Bump delaydiffeq.jl test tolerance

ChrisRackauckas:
This seems to be due to improved SIMD in some operations which leads to a difference in the time stepping value between the two.

### DIFF
--- a/test/downstream/delaydiffeq.jl
+++ b/test/downstream/delaydiffeq.jl
@@ -24,7 +24,7 @@ using Test
         @test sol.errors[:l∞] < error
 
         sol_scalar = solve(prob_scalar, ddealg)
-        @test sol.t≈sol_scalar.t atol=1e-3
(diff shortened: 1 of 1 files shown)


test/downstream/delaydiffeq.jl
```diff
@@ -24,7 +24,7 @@ using Test
         @test sol.errors[:l∞] < error
 
         sol_scalar = solve(prob_scalar, ddealg)
-        @test sol.t≈sol_scalar.t atol=1e-6
+        @test sol.t≈sol_scalar.t atol=1e-3
```
ChrisRackauckas, line 27:
```suggestion
        @test sol.t≈sol_scalar.t atol=1e-2
```

ChrisRackauckas:
@oscardssmith this one seems to be a bit too far. Can you investigate it? I wonder if there's a divergence in the init dt algorithms or something in the two forms that's being found here.

PR #2503, opened 2024-10-25
Use Aliasing API 

jClugstor:
## Checklist

- [ x] Appropriate tests were added
- [ x] All documentation related to code changes were updated
- [ x] The new code follows the
- [ x] Any new documentation only uses public API
  
## Additional context
This makes use of the Aliasing API in SciMLBase https://github.com/SciML/SciMLBase.jl/pull/830

The default of `alias_u0` is preserved here, and it isn't breaking because it still respects when `alias_u0` is put in as a kwarg to solve.


lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -162,8 +167,13 @@ function DiffEqBase.__init(
     p = prob.p
 
     # Get the control variables
+    # If alias kwarg is just default, use alias_u0, which is false by default, or is set by a kwarg to solve
+    # If alias_u0 is not nothing, use the alias_u0 provided by the user
+    if isnothing(alias.alias_u0)
+        alias = ODEAliases(alias_u0)
+    end
```
ranocha, line 174: This is breaking since it does not do what is documented in older versions if users just set `alias_u0`.

lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -71,6 +71,7 @@ function DiffEqBase.__init(
         initialize_integrator = true,
         alias_u0 = false,
         alias_du0 = false,
+        alias = ODEAliases(),
```
ChrisRackauckas, line 74:
```suggestion
        alias = ODEAliases(alias_u0 = alias_u0, alias_du0=false),
```

It needs the deprecation path.

lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -119,6 +120,10 @@ function DiffEqBase.__init(
         @warn("Dense output is incompatible with saveat. Please use the SavingCallback from the Callback Library to mix the two behaviors.")
     end
 
+    if !(alias isa ODEAliases)
```
ChrisRackauckas, line 123: Should check if the other alias kwargs are used and throw a depwarn in that case.

And we should allow / handle boolean `alias` specification as described in the higher level issue.

lib/OrdinaryDiffEqCore/src/solve.jl
jClugstor: If users set `alias_u0` in `solve`, and `ODEAliases.alias_u0` is `nothing`, it uses the `alias_u0` from the solve call.  So the only way that the `alias_u0` from the solve call isn't used is if an explicit `ODEAliases` with `ODEAliases.alias_u0` either true or false is put in the call to solve.

jClugstor:
This just needs to use to use the version of SciMLBase with the AliasSpecifiers once that's released, then it should work and pass tests.

lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -158,19 +157,62 @@ function DiffEqBase.__init(
...
+        
+        aliases 
+
+    else
+         # If alias isa Bool, all fields of ODEAliases set to alias
+        if alias isa Bool
+            aliases = ODEAliasSpecifier(alias = alias)
+        elseif isnothing(alias) || alias isa ODEAliasSpecifier 
+            aliases = alias
+        end
```
ChrisRackauckas, line 193: Canonicalize this in the DifEqBase part?

jClugstor: I'm not sure what you mean. This is here so that if `alias = true` or `alias = false` in `solve` that gets passed to the `ODEAliasSpecifier` constructor so that every variable is aliased / not aliased.

jClugstor: I'm not sure what you mean. This is here so that if `alias = true` or `alias = false` in `solve` that gets passed to the `ODEAliasSpecifier` constructor so that every variable is aliased / not aliased.

jClugstor:
Oh yeah, this will require a release of DiffEqBase so that the `alias` keyword is accepted.

jClugstor:
@ChrisRackauckas any chance of a new SciMLBase release coming soon? Also, I documented the `alias` keyword as a common argument in SciMLBase, and every AbstractAliasSpecifier type has docstrings. Should I also put something in DiffEqDocs?

ChrisRackauckas:
Done.

> Should I also put something in DiffEqDocs?

Yes in the DiffEqBase/solve.jl docstring.

ChrisRackauckas (approved):

ChrisRackauckas (approved):

PR #2507, opened 2024-10-31
Allow configuration of the initialization done after interpolation while applying an affect

BenChung:
## Checklist

## Additional context

Part of https://github.com/SciML/DiffEqBase.jl/pull/1102 and https://github.com/SciML/SciMLBase.jl/pull/842

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl
@@ -3,7 +3,7 @@
 # Hence, we need to have two separate functions.
 
 function _change_t_via_interpolation!(integrator, t,
-        modify_save_endpoint::Type{Val{T}}) where {T}
+        modify_save_endpoint::Type{Val{T}}, reinitialization_method=nothing) where {T}
     # Can get rid of an allocation here with a function
     # get_tmp_arr(integrator.cache) which gives a pointer to some
     # cache array which can be modified.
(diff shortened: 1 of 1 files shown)


lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl
```diff
@@ -3,7 +3,7 @@
 # Hence, we need to have two separate functions.
 
 function _change_t_via_interpolation!(integrator, t,
-        modify_save_endpoint::Type{Val{T}}) where {T}
+        modify_save_endpoint::Type{Val{T}}, reinitialization_method=nothing) where {T}
```
ChrisRackauckas, line 6:
```suggestion
        modify_save_endpoint::Type{Val{T}}, initialize_alg=nothing) where {T}
```

ChrisRackauckas: either this or callback_initializealg? The choice here doesn't follow the naming conventions or the other names.

BenChung: How about `reinitalize_alg`? I'm not super happy with simply calling it `initialize_alg` because it doesn't answer the question of "why does something that occurs mid-solve need to do initialization?" While "reinitialize" doesn't capture that super clearly, I think it's more obvious then just `initialize_alg`.

PR #2508, opened 2024-11-01
Switch to using ADTypes for AD choice

jClugstor:
## Checklist

## Additional context

This will switch over to using ADTypes for the `autodiff` keyword in the construction of the solver types. 
- Changes the default `autodiff` keyword to be `AutoForwardDiff()` instead of `true`.
- Adds a deprecation path so that using `true` or `false` for `autodiff` uses `AutoForwardDiff()` and `AutoFiniteDiff()` respectively.
- Also ensures that if the `chunk_size` and `diff_type` kwargs are used with a Bool `autodiff`, they are respected.


jClugstor:
@oscardssmith It's able to precompile in this state, so at least anything in the precompile workloads will work. I'm pretty sure some tests won't pass though, and I still need to add some tests as well.

jClugstor:
@oscardssmith All CI tests are passing now. Some integration tests are broken.

oscardssmith:
The StochasticDiffEq error looks very odd. @gdalle any idea what would cause this?
```
 In-place methods: Test Failed at /home/runner/work/OrdinaryDiffEq.jl/OrdinaryDiffEq.jl/downstream/test/complex_tests.jl:42
  Expression: solve(prob, alg)
    Expected: ArgumentError
      Thrown: DivideError
      DivideError: integer division error
      Stacktrace:
        [1] div
          @ ./int.jl:295 [inlined]
        [2] div
          @ ./div.jl:310 [inlined]
        [3] div
          @ ./div.jl:365 [inlined]
        [4] cld
          @ ./div.jl:322 [inlined]
        [5] generate_chunked_partials(x::Matrix{ComplexF64}, colorvec::UnitRange{Int64}, cs::Val{0})
          @ SparseDiffTools ~/.julia/packages/SparseDiffTools/m4eae/src/differentiation/compute_jacobian_ad.jl:80
        [6] SparseDiffTools.ForwardColorJacCache(f::SciMLBase.UJacobianWrapper{true, SciMLBase.SDEFunction{true, SciMLBase.FullSpecialize, Main.var"##Complex Number Tests#233".var"#f#2", Main.var"##Complex Number Tests#233".var"#f#2", LinearAlgebra.UniformScaling{Bool}, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED), Nothing, Nothing, Nothing}, Float64, SciMLBase.NullParameters}, x::Matrix{ComplexF64}, _chunksize::Val{0}; dx::Nothing, tag::Type, colorvec::UnitRange{Int64}, sparsity::Nothing)
          @ SparseDiffTools ~/.julia/packages/SparseDiffTools/m4eae/src/differentiation/compute_jacobian_ad.jl:37
        [7] ForwardColorJacCache
          @ ~/.julia/packages/SparseDiffTools/m4eae/src/differentiation/compute_jacobian_ad.jl:22 [inlined]
```

gdalle:
It appears you have a chunksize of `0` somewhere. Maybe an ill-advised choice in `AutoForwardDiff(; chunksize=...)`? The default is `nothing` and not `zero`.

jClugstor:
Yeah, my guess is that it has something to do with the chunksize,

jClugstor:
I think it's because the fallback `_get_fwd_chunksize(AD) = Val(0)`. Since the Stochastic algs aren't using ADTypes yet, it falls back to that inside of `get_chunksize`.

gdalle:
If anything, that fallback should be `Val(1)`. Unless we're explicitly trying to throw an error by using zero.

oscardssmith:
What's happening here is that we were using `Val(0)` to mean automatic chunksize construction. Does DI have a good way to ask for that?

gdalle:
That's what `AutoForwardDiff(; chunksize=nothing)` does.

jClugstor:
Actually, I think I'm wrong, for a Stochastic algorithm, it should be getting the chunksize from one of these.

```julia
OrdinaryDiffEq.get_chunksize(alg::StochasticDiffEqNewtonAlgorithm{CS,AD,FDT,ST,CJ,Controller}) where {CS,AD,FDT,ST,CJ,Controller} = Val(CS)
OrdinaryDiffEq.get_chunksize(alg::StochasticDiffEqNewtonAdaptiveAlgorithm{CS,AD,FDT,ST,CJ,Controller}) where {CS,AD,FDT,ST,CJ,Controller} = Val(CS)
OrdinaryDiffEq.get_chunksize(alg::StochasticDiffEqJumpNewtonAdaptiveAlgorithm{CS,AD,FDT,ST,CJ,Controller}) where {CS,AD,FDT,ST,CJ,Controller} = Val(CS)
OrdinaryDiffEq.get_chunksize(alg::StochasticDiffEqJumpNewtonDiffusionAdaptiveAlgorithm{CS,AD,FDT,ST,CJ,Controller}) where {CS,AD,FDT,ST,CJ,Controller} = Val(CS)
```

gdalle:
Note that depending on the semantics of chunk size in these libraries (which I don't know), setting `nothing` as the default may not work either. All I can comment on is ADTypes and DifferentiationInterface:

- In ADTypes, `Auto(Polyester)ForwardDiff` is the only backend with a configurable chunk size, and it can be either `nothing` or an integer `> 0`
- In DI, I leverage the choice of chunk size in `AutoForwardDiff` and set it to `1` for every other backend (except Enzyme which has a custom rule). Chunk size handling is internal right now because I'm not yet sure I have nailed the right API.

gdalle:
To get a clearer idea of which tests are actually failing and which tests are just Codecov complaining about your uploads, you can temporarily set this setting to `false`:

https://github.com/SciML/OrdinaryDiffEq.jl/blob/ce43d57d9aa617ca402d31baf5d91b82c28d7eab/.github/workflows/CI.yml#L97

jClugstor:
Oh thanks, that's helpful.

jClugstor:
@oscardssmith all CI tests pass, and all of the failing integration tests are failing on master as well.

oscardssmith (approved):

oscardssmith:
looks good! Thanks so much

ChrisRackauckas:
Yeah and we need to delete it repo by repo for ones that don't use the high level CI description 😱 . At least there aren't too many repos left that need to do it.

lib/OrdinaryDiffEqBDF/src/algorithms.jl
```diff
@@ -113,14 +113,17 @@ struct ABDF2{CS, AD, F, F2, P, FDT, ST, CJ, K, T, StepLimiter} <:
...
 end
-function ABDF2(; chunk_size = Val{0}(), autodiff = true, standardtag = Val{true}(),
+function ABDF2(; chunk_size = Val{0}(), autodiff = AutoForwardDiff(), standardtag = Val{true}(),
         concrete_jac = nothing, diff_type = Val{:forward},
         κ = nothing, tol = nothing, linsolve = nothing, precs = DEFAULT_PRECS,
         nlsolve = NLNewton(),
         smooth_est = true, extrapolant = :linear,
         controller = :Standard, step_limiter! = trivial_limiter!)
+
+    AD_choice = _process_AD_choice(autodiff, chunk_size, diff_type)
```
ChrisRackauckas, line 123: what about standardtag?

lib/OrdinaryDiffEqCore/src/alg_utils.jl
```diff
@@ -441,3 +457,8 @@ function Base.show(io::IO, ::MIME"text/plain", alg::OrdinaryDiffEqAlgorithm)
     end
     print(io, ")")
 end
+
+
+#function get_chunksize(alg::AutoForwardDiff{CS}) where {CS}
+#    Val(CS)
+#end
```
ChrisRackauckas, line 464: ?

lib/OrdinaryDiffEqCore/src/algorithms.jl
```diff
@@ -88,7 +88,8 @@ struct CompositeAlgorithm{CS, T, F} <: OrdinaryDiffEqCompositeAlgorithm
     algs::T
     choice_function::F
     function CompositeAlgorithm(algs::T, choice_function::F) where {T, F}
-        CS = mapreduce(alg -> has_chunksize(alg) ? get_chunksize_int(alg) : 0, max, algs)
+        #CS = mapreduce(alg -> has_chunksize(alg) ? get_chunksize_int(alg) : 0, max, algs)
```
ChrisRackauckas, line 91: ?

lib/OrdinaryDiffEqCore/src/misc_utils.jl
```diff
@@ -130,3 +130,36 @@ function get_differential_vars(f, u)
...
+        return _bool_to_ADType(Val{true}(), chunksize, diff_type)
+    end
+
+    typeof(ad_alg)
+end
+
+
+_ADType_to_Bool(::Type{T}) where {T <: AutoForwardDiff} = true
+
+_ADType_to_Bool(::Type{T}) where {T <: AutoFiniteDiff} = false
```
ChrisRackauckas, line 165: Why only those two?

lib/OrdinaryDiffEqCore/src/misc_utils.jl
```diff
@@ -130,3 +130,36 @@ function get_differential_vars(f, u)
...
+end
+
+# Functions to get ADType type from Bool or ADType object, or ADType type
+_process_AD_choice(ad_alg::Bool, chunksize, diff_type) = _bool_to_ADType(Val(ad_alg), chunksize, diff_type)
+
+function _process_AD_choice(ad_alg::AbstractADType, chunksize, diff_type) 
+    # need a path for if just chunksize is specified in the Algorithm construction
+    if !(chunksize === Val{0}())
+        return _bool_to_ADType(Val{true}(), chunksize, diff_type)
+    end
```
ChrisRackauckas, line 157: This is dropping the other information?

lib/OrdinaryDiffEqCore/src/misc_utils.jl
```diff
@@ -130,3 +130,36 @@ function get_differential_vars(f, u)
...
+function _bool_to_ADType(::Val{true}, chunksize, diff_type)
+    Base.depwarn("Using a `Bool` for keyword argument `autodiff` is deprecated. Please use an `ADType` specifier.", :_bool_to_ADType)
+    chunksize = SciMLBase._unwrap_val(chunksize) == 0 ? nothing : SciMLBase._unwrap_val(chunksize)
+    typeof(AutoForwardDiff(chunksize = chunksize))
+end
+
+function _bool_to_ADType(::Val{false}, chunksize, diff_type)
+    Base.depwarn("Using a `Bool` for keyword argument `autodiff` is deprecated. Please use an `ADType` specifier.", :_bool_to_ADType)
+    if diff_type isa Type
+        return typeof(AutoFiniteDiff(fdtype = diff_type()))
```
ChrisRackauckas, line 144: Why is this allowed?

lib/OrdinaryDiffEqDifferentiation/src/alg_utils.jl
```diff
@@ -2,26 +2,40 @@
...
+#function alg_autodiff(alg)
+#    autodiff = _alg_autodiff(alg)
+#    if autodiff == Val(false)
+#        return AutoFiniteDiff()
+#    elseif autodiff == Val(true)
+#        return AutoForwardDiff()
+#    else
+#        return _unwrap_val(autodiff)
+#    end
+#end
```
ChrisRackauckas, line 26: ?

lib/OrdinaryDiffEqCore/src/misc_utils.jl
oscardssmith: these are the only 2 ADTypes supported by this PR. The followup to add full DI support is coming later

ChrisRackauckas: Throw an error on the others then? You have a branch for the others.

lib/OrdinaryDiffEqCore/src/misc_utils.jl
```diff
@@ -130,3 +130,36 @@ function get_differential_vars(f, u)
 end
 
 isnewton(::Any) = false
+
+
+function _bool_to_ADType(::Val{true}, chunksize, diff_type)
+    Base.depwarn("Using a `Bool` for keyword argument `autodiff` is deprecated. Please use an `ADType` specifier.", :_bool_to_ADType)
+    chunksize = SciMLBase._unwrap_val(chunksize) == 0 ? nothing : SciMLBase._unwrap_val(chunksize)
+    typeof(AutoForwardDiff(chunksize = chunksize))
```
gdalle, line 138: why return the type and not the ADTypes object itself?

lib/OrdinaryDiffEqDifferentiation/src/alg_utils.jl
```diff
@@ -2,26 +2,40 @@
...
+        return AutoForwardDiff()
+    elseif autodiff == Val(false)
         return AutoFiniteDiff()
-    elseif autodiff == Val(true)
+    elseif nameof(autodiff) == :AutoForwardDiff
         return AutoForwardDiff()
-    else
-        return _unwrap_val(autodiff)
+    elseif nameof(autodiff) == :AutoFiniteDiff
+        return AutoFiniteDiff()
```
gdalle, line 38: why drop the information contained in `alg`?

lib/OrdinaryDiffEqCore/src/misc_utils.jl
ChrisRackauckas: Yes that's the same question I have above in https://github.com/SciML/OrdinaryDiffEq.jl/pull/2508#discussion_r1844930312. Why is the type ever used and why not just use the ADTypes object everywhere? I think there's two clear major changes to make here:

1. Do not accept types, i.e. `autodiff=AutoForwardDiff` shouldn't be an acceptable choice, just `autodiff=AutoForwardDiff()`
2. Never change instances to types internally. There's no need to.
3. Delete all of the code doing weird stuff on types. If you always have instances you never have to do that and you never have to guess any runtime elements.

If ADTypes stays as objects rather than their types, all of these problems go away. Keep life simple.

lib/OrdinaryDiffEqCore/src/misc_utils.jl
jClugstor: This is in case some one uses the `chunksize` keyword arg, but doesn't use a Bool `autodiff` argument. If they just use `chunksize` the AD used is AutoForwardDiff with that chunksize. 
I should probably have a more specific warning about this route though, especially if they try to use an ADType for `autodiff` and the `chunksize` kwarg at the same time.

lib/OrdinaryDiffEqCore/src/misc_utils.jl
jClugstor: Yeah, I was trying to make this work with as little changes possible, and I didn't think that the ADType objects could be a type parameter. 
But it looks like the algorithms will need an `autodiff` field anyway, so might as well do that now?

Accepting types thing was something weird I had to do because of `remake` .

jClugstor:
I'm not sure what's going on with the DelayDiffEq errors

lib/OrdinaryDiffEqBDF/src/algorithms.jl
jClugstor: We can't do the tags in the creation of the alg, since they use the eltype of `u`. They need to be done where that's accessible. For the full DI integration I'm moving the tag creation to `prepare_alg` for the implicit algs, but that's not needed for this.

jClugstor:
At this point, the only tests that are failing different from the ones on master are the DelayDiffEqIntegration tests. Those are failing because the tests aren't using the dev versions of a couple of packages they should be, **https://github.com/SciML/OrdinaryDiffEq.jl/issues/2553

jClugstor:
I believe all the tests that are failing here are also failing on master.

jClugstor:
@ChrisRackauckas @oscardssmith I think this is ready

lib/OrdinaryDiffEqBDF/Project.toml
```diff
@@ -1,9 +1,10 @@
 name = "OrdinaryDiffEqBDF"
 uuid = "6ad6398a-0878-4a85-9266-38940aa047c8"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "1.1.2"
+version = "1.2"
```
ChrisRackauckas, line 4:
```suggestion
version = "1.2.0"
```

ChrisRackauckas (approved):

ChrisRackauckas:
Can you make sure to do the downstream in StochasticDiffEq and DelayDiffEq?

jClugstor:
Yes, will do.

gdalle:
hot damn, this is a great christmas gift! thanks for all your hard work

Issue #2289, opened 2024-07-19
Rosenbrock-W method docstrings


termi-official:
The doc categorization for Rosenbrock-W (after Wolfbrandt) are messed up on the latest release and need to be rolled back before https://github.com/SciML/OrdinaryDiffEq.jl/commit/f8f8307cf3717ae985c6ef165ef2bc8d60947bfc see https://github.com/SciML/OrdinaryDiffEq.jl/pull/2236#issuecomment-2148234155 for more discussion and https://github.com/SciML/OrdinaryDiffEq.jl/pull/2236#issuecomment-2149637232 for the concrete action to take for the doc strings functions.

gstein3m:
Maybe Table 1 in https://github.com/hbrs-cse/RosenbrockMethods/blob/main/paper/JuliaPaper.pdf can help in deciding whether it is a W method. Note that the W property usually only refers to ODEs and not to DAEs. 
However, the W property does not appear to be utilized in the implementation. The user must then specify an approximated Jacobian matrix themselves.

PR #2512, opened 2024-11-05
Fix Rosenbrock docstrings for Wolfbrandt-property

termi-official:
Closes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2289 . I synced the doc strings against the following paper  https://github.com/hbrs-cse/RosenbrockMethods/blob/main/paper/JuliaPaper.pdf - thanks @gstein3m !

@cwittens do you have time to double check if this is correct now? I could not find time to go though all the papers and I am also not anywhere close to be an expert on these methods.


cwittens:
I think besides Rodas4P everything seems correct 👍

termi-official:
Thanks for double checking! Fixed the oopsie in the latest commit.

ChrisRackauckas (approved):

PR #2531, opened 2024-11-15
clean up radau tableau generation

oscardssmith:
This adds tableau caching for arbitrary order tableaus (and simplifies the tableau generation slightly)
@Shreyas-Ekanathan can you review this?


oscardssmith:
Also, @Shreyas-Ekanathan it would be really nice if we can optimize the error control tableau generation a little more. The fact that the higher order ones require exponentially more symbolic codegen is pretty bad (for reference, 5th order takes 0.01 seconds, 9th order takes 0.3 seconds, and 13th order takes 35 seconds on my computer).

oscardssmith:
Closed in favor of https://github.com/SciML/OrdinaryDiffEq.jl/pull/2543

PR #2554, opened 2024-12-09
Move DAE init alg error message further down the dependency stack

Ickaser:
## Checklist

## Additional context

Still working on #2513 ; this pull request attempts to address that issue in a non-breaking way, whereas #2514 solves it more elegantly but with a breaking change to the default initialization algorithm. (That will probably need to be rebased on this pull request once this works as desired.)

## Attempted approach

`BrownFullBasicInit` and `ShampineCollocationInit` both require `OrdinaryDiffEqNonlinearSolve`. The original implementation had an error message when the nonlinear solve is attempted if `ODENLSolve` is not loaded, but that nonlinear solve happens in a method of `_initialize_dae!` which is defined only in `ODENLSolve`, so the error message needs to be placed in `ODECore`. This attempt currently checks in the `DefaultInit` method for `_initialize_dae!` whether `ODENLSolve` is loaded, and if it is, proceeds to use default algorithms as done previously. If not loaded, then an error is thrown.

I don't feel like enough of an expert to say authoritatively what the best way is to check if `ODENLSolve` is loaded. From what I could find of discussions on Discourse, etc., it seems like the recommended way might be to do
```julia
if !isdefined(Main, :OrdinaryDiffEqNonlinearSolve)
```
so that is what I have done here.

Two other approaches I turned up (but that seem brittle or discouraged) might be something along the lines of
```julia
if !hasmethod(_initialize_dae!, Tuple{Any, ODEProblem, BrownFullBasicInit, typeof(isinplace))
```
or checking if `ODENLSolve` is in `Base.loaded_modules`.

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/initialize_dae.jl
+++ b/lib/OrdinaryDiffEqCore/src/initialize_dae.jl
@@ -50,7 +50,8 @@ function _initialize_dae!(integrator, prob::ODEProblem,
     if SciMLBase.has_initializeprob(prob.f)
         _initialize_dae!(integrator, prob,
             OverrideInit(integrator.opts.abstol), x)
-    elseif !isdefined(Main, :OrdinaryDiffEqNonlinearSolve)
+    elseif !applicable(_initialize_dae!, integrator, prob,
+        BrownFullBasicInit(integrator.opts.abstol), x)
         error("`OrdinaryDiffEqNonlinearSolve` is not loaded, which is required for the default initialization algorithm (`BrownFullBasicInit` or `ShampineCollocationInit`). To solve this problem, either do `using OrdinaryDiffEqNonlinearSolve` or pass `initializealg = CheckInit()` to the `solve` function. This second option requires consistent `u0`.")
     else
         _initialize_dae!(integrator, prob,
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

ChrisRackauckas (approved):

ChrisRackauckas:
Looks like this is a bit too trigger happy in the tests.

Ickaser:
The reason this was too trigger-happy:
```julia
if !isdefined(Main, :OrdinaryDiffEqNonlinearSolve)
```
fails if the user (or tester) loads `ODENLSolve` via `using OrdinaryDiffEq` or `using DifferentialEquations`. We could try to make a list of the packages that all eventually load `ODENLSolve` and check if any of those are loaded, but then if a user makes their own package and `@reexport`s ODENLSolve (which is something I have done myself) then this breaks that. That might look like the following:
```julia
const nlsolve_umbrella_pkgs = [:OrdinaryDiffEqNonlinearSolve, :OrdinaryDiffEq, :DifferentialEquations]
if all([!isdefined(Main, pkg) for pkg in nlsolve_umbrella_pkgs])
```
As I said, though, that breaks if someone tries to reexport in their own package. So in the end I think directly checking if the methods that get dispatched from the default algorithm exist is the only non-breaking way to introduce this informative error. That ends up looking like the following, where we basically check if the method exists, and call it if it does:
```julia
    elseif !applicable(_initialize_dae!, integrator, prob,
        BrownFullBasicInit(integrator.opts.abstol), x)
        error("`OrdinaryDiffEqNonlinearSolve` is not loaded, which is required for the default initialization algorithm (`BrownFullBasicInit` or `ShampineCollocationInit`). To solve this problem, either do `using OrdinaryDiffEqNonlinearSolve` or pass `initializealg = CheckInit()` to the `solve` function. This second option requires consistent `u0`.")
    else
            BrownFullBasicInit(integrator.opts.abstol), x)
``

Unlike my first approach, this seems to actually allow tests to pass, at least as far as the tests have gotten on my machine while I write.

Ickaser:
If this new approach looks alright and works on existing tests, then I suppose a couple tests should get added somewhere to make sure that this error works as intended? If so, those tests will need to be removed by the v7 commit that addresses this issue more elegantly.

Ickaser:
@ChrisRackauckas The remaining test errors look unrelated to me, although I could be wrong. 

Should there be new tests making sure this error triggers as expected? They will essentially require loading only part of OrdinaryDiffEq, attempting a DAE solve and seeing if the appropriate error is hit, so perhaps it makes sense for such a test to live in its own file, where its dependency can be obvious; should that file essentially imitate what I see in the following file? 
https://github.com/SciML/OrdinaryDiffEq.jl/blob/44014a7a83ba9b663e38630ef968b2a33a0d7a48/test/interface/dae_initialization_tests.jl#L1

Ickaser:
Bump!
Since this is intended to provide an informative error, then get superceded by #2514 in v7, is it worth introducing new tests? Testing a partial load of the package seems like something that's not already in the test suite, so I'm not sure where such a test should even go.

PR #2556, opened 2024-12-13
Added threading to ABM algorithms using @..

mmesiti:
1. in *_perform_step.jl in functions using @.. in unpacking of the "cache" object add a "thread" variable
2. in *_caches.jl for the cache types affected (determined in 1.)
   - add parametric type Thread
   - and a field thread::Thread
   - in the alg_cache function, add 'alg.thread' as argument to the cache type constructor
3. in *_algorithms.jl, for the alg type affected (determined in 2.)
   - add parametric type Thread
   - add field thread::Thread
   - add outer constructor taking a thread argument with default value False(), that builds the respective algorithm object forwarding the thread parameter to it.

## Checklist

## Additional context

Referring to this discussion: https://discourse.julialang.org/t/differentialequations-jl-mpi-jl-pencilarrays-jl-lack-of-scaling-observed/122998/18?u=mmesiti

### DIFF
--- a/lib/OrdinaryDiffEqAdamsBashforthMoulton/src/adams_bashforth_moulton_caches.jl
+++ b/lib/OrdinaryDiffEqAdamsBashforthMoulton/src/adams_bashforth_moulton_caches.jl
@@ -4,7 +4,7 @@ get_fsalfirstlast(cache::ABMMutableCache, u) = (cache.fsalfirst, cache.k)
 function get_fsalfirstlast(cache::ABMVariableCoefficientMutableCache, u)
     (cache.fsalfirst, cache.k4)
 end
-@cache mutable struct AB3Cache{uType, rateType} <: ABMMutableCache
+@cache mutable struct AB3Cache{uType, rateType, Thread} <: ABMMutableCache
     u::uType
     uprev::uType
     fsalfirst::rateType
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
This is set as a draft, but it seems like it's correct? What's the incomplete part here?

mmesiti:
> This is set as a draft, but it seems like it's correct? What's the incomplete part here?

Still need to make sure all the boxes can be ticked!

mmesiti:
> If you can add some regression tests for the threading choices then I think that's sufficient.

Where should those tests go?  Should they be included in the `runtests.jl` of the ABM subpackage (as I have done now) or should they go somewhere else, where they are guaranteed to run with multiple threads?

ChrisRackauckas:
https://github.com/mmesiti/OrdinaryDiffEq.jl/blob/multithreaded-ABM/lib/OrdinaryDiffEqAdamsBashforthMoulton/test/abm_convergence_tests.jl extend the convergence tests to have a few of the multithreaded choices.

ChrisRackauckas (approved):

mmesiti:
Have I forgot to bump the version of the ABM package? I'm not seeing this code when I do `Pkg.add("OrdinaryDiffEqAdamsBashforthMoulton")`.

PR #2559, opened 2024-12-17
Optimize `VerletLeapfrog` method

efaulhaber:
## Checklist

- [x] Appropriate tests were added (does not require additional tests, method is already tested)
- [x] All documentation related to code changes were updated (no docs changes necessary, method and API didn't change)
- [x] Any new documentation only uses public API (no new docs)
  
## Additional context

The `VerletLeapfrog` method is currently using the generic `Symplectic2Cache` implementation:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/4b29babd6bbe9a4eadd1ffef10b90a812db9aa3c/lib/OrdinaryDiffEqSymplecticRK/src/symplectic_perform_step.jl#L222-L246
This contains 3 calls to `f1` and 2 calls to `f2`. The kick-drift-kick scheme of the Leapfrog method only requires a single call to `f1` and a single call to `f2` in each step, which is the main selling point of the method, making the current implementation basically useless.

I implemented an optimized step for this method, following the `VelocityVerlet` method, which also has its own step function.

### DIFF
--- a/lib/OrdinaryDiffEqSymplecticRK/src/algorithms.jl
+++ b/lib/OrdinaryDiffEqSymplecticRK/src/algorithms.jl
@@ -30,7 +30,7 @@ struct VelocityVerlet <: OrdinaryDiffEqPartitionedAlgorithm end
     verlet1967, "", "")
 struct VerletLeapfrog <: OrdinaryDiffEqPartitionedAlgorithm end
 
-default_linear_interpolation(alg::VerletLeapfrog, prob) = true
+OrdinaryDiffEqCore.default_linear_interpolation(alg::VerletLeapfrog, prob) = true
 
 @doc generic_solver_docstring("2nd order explicit symplectic integrator.",
     "PseudoVerletLeapfrog",
--- a/lib/OrdinaryDiffEqSymplecticRK/src/algorithms.jl
+++ b/lib/OrdinaryDiffEqSymplecticRK/src/algorithms.jl
@@ -30,6 +30,8 @@ struct VelocityVerlet <: OrdinaryDiffEqPartitionedAlgorithm end
     verlet1967, "", "")
 struct VerletLeapfrog <: OrdinaryDiffEqPartitionedAlgorithm end
 
+default_linear_interpolation(alg::VerletLeapfrog, prob) = true
+
 @doc generic_solver_docstring("2nd order explicit symplectic integrator.",
     "PseudoVerletLeapfrog",
     "Symplectic Runge-Kutta Methods",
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Yeah that looks right, seems it just got over simplified. 

The interpolation should be specialized though so that it goes to linear instead of Hermite then, since the derivative would be off. Or a special centered Hermite would need to be used.

efaulhaber:
I'll be honest, I have no idea what I need to change to make this use a linear interpolation. Could you point me to the code that I will have to change?

ChrisRackauckas:
I just realized we were doing that a bit bespoke. Set this branch up on top of https://github.com/SciML/OrdinaryDiffEq.jl/pull/2560 with the trait and I think it's good to go.

efaulhaber:
Like so?

lib/OrdinaryDiffEqSymplecticRK/src/symplectic_perform_step.jl
```diff
@@ -171,7 +174,7 @@ end
     # v(t+Δt) = v(t) + 1/2*(a(t)+a(t+Δt))*Δt
     du = duprev + dt * (half * ku + half * kdu)
 
-    OrdinaryDiffEqCore.increment_nf!(integrator.stats, 2)
+    OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
```
efaulhaber, line 177: Note that this was a bug with the `VelocityVerlet` method.

lib/OrdinaryDiffEqSymplecticRK/src/algorithms.jl
```diff
@@ -30,6 +30,8 @@ struct VelocityVerlet <: OrdinaryDiffEqPartitionedAlgorithm end
     verlet1967, "", "")
 struct VerletLeapfrog <: OrdinaryDiffEqPartitionedAlgorithm end
 
+default_linear_interpolation(alg::VerletLeapfrog, prob) = true
```
ChrisRackauckas, line 33:
```suggestion
OrdinaryDiffEqCore.default_linear_interpolation(alg::VerletLeapfrog, prob) = true
```

ChrisRackauckas:
Rebase onto the latest master and I think this is good to go.

ChrisRackauckas (approved):

ChrisRackauckas:
Test fails.

efaulhaber:
Tests related to my changes seem to be running now. Everything else seems to be unrelated.

ChrisRackauckas:
yeah I need to do some test greening during the holidays. It's all small little things but it has built up a few of them...

anyways this looks good.

PR #2575, opened 2025-01-16
feat: support initialization of `DiscreteProblem`

AayushSabharwal:
## Checklist

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/initialize_dae.jl
+++ b/lib/OrdinaryDiffEqCore/src/initialize_dae.jl
@@ -98,7 +98,7 @@ function _initialize_dae!(integrator, prob::DiscreteProblem,
     alg::DefaultInit, x::Union{Val{true}, Val{false}})
     if SciMLBase.has_initializeprob(prob.f)
         # integrator.opts.abstol is `false` for `DiscreteProblem`.
-        _initialize_dae!(integrator, prob, OverrideInit(one(eltype(prob.u0)) * 1e-6), x)
+        _initialize_dae!(integrator, prob, OverrideInit(one(eltype(prob.u0)) * 1e-12), x)
     end
 end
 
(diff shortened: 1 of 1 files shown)


isaacsas:
Doesn’t this need to be in SciMLBase or DiffEqBase? OrdinaryDiffEq is not a dependency of JumpProcesses so this would never get loaded for a pure JumpProblem?

AayushSabharwal:
This is for `DiscreteProblem`, not `JumpProblem`. `JumpProblem` wraps another problem, and `__init(jprob::JumpProblem)` calls `__init(jprob.prob)` so the initialization there would run.

ChrisRackauckas:
But there are methods for DiscreteProblem in for example SimpleDiffEq.jl. I think this dispatch might make sense in SciMLBase since it's not specific to any method here?

AayushSabharwal:
It's specific to the integrator and the initializealg. Trying to solve a `DiscreteProblem` with `FunctionMap` will run into the `__init` defined which will call `initialize_dae!` with `DefaultInit` as the initializealg. SimpleDiffEq doesn't hook into initialization yet, so that PR is separate.

lib/OrdinaryDiffEqCore/src/initialize_dae.jl
```diff
@@ -94,6 +94,14 @@ function _initialize_dae!(integrator, prob::DAEProblem,
     end
 end
 
+function _initialize_dae!(integrator, prob::DiscreteProblem,
+    alg::DefaultInit, x::Union{Val{true}, Val{false}})
+    if SciMLBase.has_initializeprob(prob.f)
+        # integrator.opts.abstol is `false` for `DiscreteProblem`.
+        _initialize_dae!(integrator, prob, OverrideInit(one(eltype(prob.u0)) * 1e-6), x)
```
ChrisRackauckas, line 101:
```suggestion
        _initialize_dae!(integrator, prob, OverrideInit(one(eltype(prob.u0)) * 1e-12), x)
```

PR #2577, opened 2025-01-21
Fix hang in BuildKite CI

lte678:
I believe that due to the bug documented at https://github.com/JuliaLang/julia/issues/56458, the buildkite CI is hanging indefinitely causing the tests to fail.
This seems to be triggered by the use of the `--threads=auto` flag.

This merge request simply removes the flag for now, since a fix does not yet seem to be available.

https://github.com/JuliaCI/julia-test-buildkite-plugin/blob/main/hooks/command does not seem to support adding the flag at the `julia --threads=auto` invocation level, where it would again be permitted.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -38,6 +38,7 @@ steps:
     env:
       BUILDKITE_PLUGIN_JULIA_VERSION: "{{matrix.version}}"
       GROUP: "{{matrix.group}}"
+      JULIA_NUM_THREADS: 2
     plugins:
       - JuliaCI/julia#v1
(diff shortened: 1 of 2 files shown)


lte678:
The pipeline looks good to me. I

Since this is an upstream bug that will presumably be fixed, I would just drop the option for now and perhaps add an issue to track it, since I assume that this change has a negative effect on the CI runtime and is worth reinstating at some point.

ChrisRackauckas:
But what about the multithreading tests? How is that testing the thread safety?

ChrisRackauckas:
Thanks!

PR #2585, opened 2025-01-28
prepare for switching to Linsolve Interface

oscardssmith:
This is part 1 of a redesign of https://github.com/SciML/OrdinaryDiffEq.jl/pull/2318 to make it actually mergable. Specifically, this PR updates OrdinaryDiffEqRosenbrock and OrdinaryDiffEqNonlinearSolve to pass the `du`, `u`, `p`, and `t` via the NonlinearProblem rather than to `dolinsolve`. In a followup PR, I will then remove these fields from `dolinsolve`, but I think doing things this way will make it so we can make the OrdinaryDiffEqDifferentiation change require a new enough version of the solver libraries to not get unexpected inputs.


oscardssmith:
Time to start up on this again.

oscardssmith:
I think this now works (with deprecations for the old precs interface). WP diagram incoming.

oscardssmith:
Looks like CI isn't using the deved versions here. Any idea why?

ChrisRackauckas:
Just on LTS

oscardssmith:
oh, is LTS CI not deving the sources properly?

ChrisRackauckas:
yeah LTS ignores sources. That's a Pkg thing.

test/interface/linear_nonlinear_tests.jl
```diff
@@ -44,28 +44,28 @@ solref = @test_nowarn solve(prob,
     TRBDF2(autodiff = AutoFiniteDiff(), linsolve = KrylovJL_GMRES(),
         smooth_est = false));
 @test length(sol.t) < 20
-sol = @test_nowarn solve(prob,
+sol = solve(prob,
     TRBDF2(autodiff = AutoFiniteDiff(), linsolve = KrylovJL_GMRES(),
         precs = precsl, smooth_est = false, concrete_jac = true));
```
ChrisRackauckas, line 49: instead update the preconditioners being tested? Also the docs tutorial on this will need an update.

PR #2587, opened 2025-02-05
cleanup `step!` and make it return whether the step was sucessful

oscardssmith:
As pointed out by https://discourse.julialang.org/t/how-can-i-check-if-step-was-successful/125581, there isn't currently a good way to check whether `step!` succeeds. This makes it return a bool for whether it worked, and also removes a bunch of `@inbounds` statements that don't do anything since `@inbounds` doesn't traverse function boundaries.


ChrisRackauckas (approved):

oscardssmith:
The only question here to me is whether we want to return a bool or the retcode.

ChrisRackauckas:
A retcode would be good if we have it at this point?

oscardssmith:
merge?

PR #2608, opened 2025-02-20
fix: rerun `initialize_dae!` in `reinit!`

AayushSabharwal:
Close https://github.com/SciML/ModelingToolkit.jl/issues/3375


lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl
```diff
@@ -406,6 +407,12 @@ function DiffEqBase.reinit!(integrator::ODEIntegrator, u0 = integrator.sol.prob.
         auto_dt_reset!(integrator)
     end
 
+    if reinit_dae &&
+       (integrator.isdae || SciMLBase.has_initializeprob(integrator.sol.prob.f))
+        DiffEqBase.initialize_dae!(integrator)
+        update_uprev!(integrator)
+    end
```
ChrisRackauckas, line 414: It should probably be after reiniting the cache?

AayushSabharwal: I did it here because in `solve` we do it before `initialize_callbacks!`, and running it after seems like it can lead to weird issues

PR #2609, opened 2025-02-21
[WIP] Explicit Taylor methods

tansongchen:
## Checklist

- [ ] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API

PR #2620, opened 2025-03-04
Explicit Taylor solvers

tansongchen:
## Checklist

## Additional context

Splitted from https://github.com/SciML/OrdinaryDiffEq.jl/pull/2609 , commits are squashed to make it easier to handle conflicts when DAETS comes in.

### DIFF
--- a/lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_perform_step.jl
+++ b/lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_perform_step.jl
@@ -1,0 +1,100 @@
+using TaylorDiff: TaylorDiff, extract_derivative, extract_derivative!
+
+@inline make_taylor(all::Vararg{X, P}) where {P, X <: AbstractArray} = TaylorArray(
+    Base.first(all), Base.tail(all))
+@inline make_taylor(all::Vararg{X, P}) where {P, X} = TaylorScalar(all)
+
+function initialize!(integrator, cache::ExplicitTaylor2ConstantCache)
+    integrator.kshortsize = 3
+    integrator.k = typeof(integrator.k)(undef, integrator.kshortsize)
+end
+
+@muladd function perform_step!(
+        integrator, cache::ExplicitTaylor2ConstantCache, repeat_step = false)
+    @unpack t, dt, uprev, u, f, p = integrator
+    k1 = f(uprev, p, t)
+    u1 = make_taylor(uprev, k1)
+    t1 = TaylorScalar{1}(t, one(t))
+    k2 = f(u1, p, t1).partials[1]
+    u = @.. uprev + dt * k1 + dt^2 / 2 * k2
+    OrdinaryDiffEqCore.increment_nf!(integrator.stats, 3)
+    integrator.k[1] = k1
+    integrator.k[2] = k2
+    integrator.u = u
+end
+
+function initialize!(integrator, cache::ExplicitTaylor2Cache)
+    integrator.kshortsize = 3
+    resize!(integrator.k, integrator.kshortsize)
+    # Setup k pointers
+    integrator.k[1] = cache.k1
+    integrator.k[2] = cache.k2
+    integrator.k[3] = cache.k3
+    return nothing
+end
+
+@muladd function perform_step!(integrator, cache::ExplicitTaylor2Cache, repeat_step = false)
+    @unpack t, dt, uprev, u, f, p = integrator
+    @unpack k1, k2, k3, utilde, tmp = cache
+
+    # The following code is written to be fully non-allocating
+    f(k1, uprev, p, t)
+    u1 = make_taylor(uprev, k1)
+    t1 = TaylorScalar{1}(t, one(t))
+    out1 = make_taylor(k1, k2)
+    f(out1, u1, p, t1)
+    @.. u = uprev + dt * k1 + dt^2 / 2 * k2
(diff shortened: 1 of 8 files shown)


tansongchen:
In reply to this comment,
image

when `make_taylor`, it's just a wrapper over several existing arrays, no new arrays are allocated.

lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_perform_step.jl
```diff
@@ -0,0 +1,103 @@
...
+
+@muladd function perform_step!(integrator, cache::ExplicitTaylorConstantCache{P}, repeat_step = false) where P
+    @unpack t, dt, uprev, u, f, p = integrator
+    us = typeof(u)[]
+    integrator.k[1] = f(uprev, p, t)
+    push!(us, integrator.k[1])
+    u = @.. uprev + dt * us[1]
+    dti = dt
+    for i in 1:P-1
+        ui = make_taylor(uprev, us...)
```
ChrisRackauckas, line 62: Splatting arrays is really not a good idea. Did you mean for this to be a tuple?

lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_perform_step.jl
```diff
@@ -0,0 +1,103 @@
...
+
+    # The following code is written to be fully non-allocating
+    f(ks[1], uprev, p, t)
+    @.. us[1] .= ks[1]
+    @.. u = uprev + dt * us[1]
+    dti = dt
+    for i in 1:P-1
+        ui = make_taylor(uprev, us[1:i]...)
+        ti = TaylorScalar{i}(t, one(t))
+        outi = make_taylor(ks[1:i+1]...)
```
ChrisRackauckas, line 95: Same here, it's an array?

tansongchen: `ks` and `us` are tuples, size parameterized by the order `P`.

image

ChrisRackauckas: If it's parameterized it can't be adaptive?

lib/OrdinaryDiffEqTaylorSeries/src/alg_utils.jl
```diff
@@ -0,0 +1,55 @@
+alg_order(::ExplicitTaylor2) = 2
+alg_stability_size(alg::ExplicitTaylor2) = 1
+
+alg_order(::ExplicitTaylor{P}) where {P} = P
+alg_stability_size(alg::ExplicitTaylor) = 1
+
+JET_CACHE = IdDict()
```
oscardssmith, line 7: `const`?

oscardssmith: also thread safety?

oscardssmith (approved):
I'm a bit concerned about JET_CACHE, but otherwise, this looks good to me.

lib/OrdinaryDiffEqTaylorSeries/src/alg_utils.jl
```diff
@@ -0,0 +1,55 @@
...
+    jet = build_function(u, u0, t0; expression = Val(false), cse = true)
+    if !haskey(JET_CACHE, f)
+        JET_CACHE[f] = []
+    end
+    push!(JET_CACHE[f], (order, p, jet))
+    return jet
+end
+
+# evaluate using Qin Jiushao's algorithm
+@generated function evaluate_polynomial(t::TaylorScalar{T, P}, z) where {T, P}
```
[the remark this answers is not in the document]
tansongchen: It's pretty like the built-in macro `@evalpoly`, which unroll the iteration such that it doesn't need to go through a loop to use Horner's algorithm to evaluate polynomials

lib/OrdinaryDiffEqTaylorSeries/src/alg_utils.jl
tansongchen: How to make this thread-safe? Do I need to switch to other data structures

oscardssmith: if you're going to use a global cache, you need to synchronize mutation to it. Alternatively, is there a way to make the cache local?

ChrisRackauckas: rewriting you are guaranteed to have the same function right? So I don't think it's unsafe. What might happen is you might compile more times than you need to, but that would still be correct.

oscardssmith: no. If you have concurent modification of an IDDict without synchronization, that's a data race (which is UB). You can get torn writes, or other arbitrarily wrong results getting written.

ChrisRackauckas: Okay torn writes would be bad. So it just needs a lock on write?

oscardssmith: I think the best approach would be to make the cache non-global. As long as you have a separate cache per function (which should be totally fine) there won't be multithreading to cause problems.

ChrisRackauckas:
🎉

PR #2622, opened 2025-03-12
feat: add initialize_dae method for ImplicitDiscreteProblem

vyudu:
## Checklist

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/initialize_dae.jl
+++ b/lib/OrdinaryDiffEqCore/src/initialize_dae.jl
@@ -85,6 +85,21 @@ function _initialize_dae!(integrator, prob::DiscreteProblem,
     end
 end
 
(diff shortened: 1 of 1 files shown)


lib/OrdinaryDiffEqCore/src/initialize_dae.jl
```diff
@@ -85,6 +85,21 @@ function _initialize_dae!(integrator, prob::DiscreteProblem,
     end
 end
 
+function _initialize_dae!(integrator, prob::ImplicitDiscreteProblem,
```
ChrisRackauckas, line 88: This isn't ODE specific, I don't understand why it would be here not in the implicit discrete solver library, or https://github.com/SciML/SciMLBase.jl/blob/master/src/initialization.jl

vyudu: ah the DiscreteProblem method is here so I assumed it went here, will move

PR #2630, opened 2025-03-17
ImplicitDiscreteSolve: handle case where u is nothing

vyudu:
Sometimes there will be no unknowns in the implicit system (e.g. callbacks where all the updates are explicit), and u is nothing. This handles this case.

### DIFF
--- a/lib/ImplicitDiscreteSolve/src/cache.jl
+++ b/lib/ImplicitDiscreteSolve/src/cache.jl
@@ -1,5 +1,5 @@
 mutable struct ImplicitDiscreteState{uType, pType, tType}
-    u::Vector{uType}
+    u::uType
     p::pType
     t_next::tType
 end
@@ -16,7 +16,7 @@ function alg_cache(alg::SimpleIDSolve, u, rate_prototype, ::Type{uEltypeNoUnits}
         dt, reltol, p, calck,
         ::Val{true}) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
 
-    state = ImplicitDiscreteState(similar(u), p, t)
+    state = ImplicitDiscreteState(isnothing(u) ? nothing : similar(u), p, t)
     SimpleIDSolveCache(u, uprev, state, nothing)
 end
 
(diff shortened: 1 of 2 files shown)


lib/ImplicitDiscreteSolve/src/cache.jl
```diff
@@ -16,7 +16,7 @@ function alg_cache(alg::SimpleIDSolve, u, rate_prototype, ::Type{uEltypeNoUnits}
         dt, reltol, p, calck,
         ::Val{true}) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
 
-    state = ImplicitDiscreteState(similar(u), p, t)
+    state = ImplicitDiscreteState(isnothing(u) ? nothing : similar(u), p, t)
```
ChrisRackauckas, line 19:
```suggestion
    state = ImplicitDiscreteState(isnothing(u) ? nothing : zero(u), p, t)
```

lib/ImplicitDiscreteSolve/src/cache.jl
```diff
@@ -31,7 +31,7 @@ function alg_cache(alg::SimpleIDSolve, u, rate_prototype, ::Type{uEltypeNoUnits}
         dt, reltol, p, calck,
         ::Val{false}) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
 
-    state = ImplicitDiscreteState(similar(u), p, t)
+    state = ImplicitDiscreteState(isnothing(u) ? nothing : similar(u), p, t)
```
ChrisRackauckas, line 34:
```suggestion
    state = ImplicitDiscreteState(isnothing(u) ? nothing : zero(u), p, t)
```

ChrisRackauckas:
Tests?

lib/ImplicitDiscreteSolve/src/solve.jl
```diff
@@ -33,7 +33,8 @@ end
 
 function _initialize_dae!(integrator, prob::ImplicitDiscreteProblem,
         alg::DefaultInit, x::Union{Val{true}, Val{false}})
-    atol = one(eltype(prob.u0)) * 1e-12
+    isnothing(prob.u0) && return
+    atol = 1e-12
```
ChrisRackauckas, line 37: this is eltype sensitive?

Issue #2632, opened 2025-03-19
Rodas4 solver allocates while Tsit5 does not


gbene:
**Question❓**

Hi! 
I was trying out different solvers on the Lorenz equation system: 

```julia
using DifferentialEquations
using BenchmarkTools



 function lorenz!(du, u, p, t)
       x, y, z = u
       σ, ρ, β = p
      
       du[1] = σ*(y-x)
       du[2] = x*(ρ-z)-y
       du[3] = x*y - β*z

end


u₀ = [1.0, 0.0, 0.0]
tₛ = (0.0,10.0)
p = [10.0, 28.0, 8 / 3]
                       

eᵣ = 1e-6
eₐ = 1e-6

problem = ODEProblem(lorenz!, u₀, tₛ, p) 

```

I noticed that if I benchmark using Tsit5() and Rodas4(), with Rodas, memory is also allocated to something other than the integration cache, hence the GC hits (i think, I am still learning the ropes). Tsit5 does not (as per [this page](https://docs.sciml.ai/SciMLTutorialsOutput/html/introduction/03-optimizing_diffeq_code.html)) :

```julia


solver1 = Tsit5()

b1 = @benchmark solve(problem, solver1, reltol = eᵣ, abstol = eₐ, save_everystep=false)

solver2 = Rodas4()

b2 = @benchmark solve(problem, solver2, reltol = eᵣ, abstol = eₐ, save_everystep=false)

julia> b1
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  42.331 μs … 106.459 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     44.378 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   44.510 μs ±   1.223 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                         ▂▃▇▇█▇▅▃▁                              
  ▂▂▂▂▂▂▂▂▂▂▁▁▂▁▂▂▂▂▂▃▄▅▇█████████▇▆▅▄▄▃▃▃▃▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  42.3 μs         Histogram: frequency by time         46.7 μs <

 Memory estimate: 4.34 KiB, allocs estimate: 85.

julia> b2
BenchmarkTools.Trial: 7993 samples with 1 evaluation per sample.
[...]

PR #2634, opened 2025-03-21
fix rosenbrock allocations

oscardssmith:
This doesn't remove all the allocations since `lu!` still allocates a `Vector` for the permutations, but it removes a bunch of them.

fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2632


ChrisRackauckas:
> This doesn't remove all the allocations since lu! still allocates a Vector for the permutations, but it removes a bunch of them.

LinearSolve.jl avoids that if you use the MKL or apple BLASes, which most things default to? And RecursiveFactorization.jl. It's just the OpenBLAS usage that will have that.

PR #2643, opened 2025-03-26
support ForwardDiff 1.0

oscardssmith:


oscardssmith:
ok we finally have enough to start testing this.

pjaap:
I really looking forward to this. This currently blocks the upgrade to `ForwardDiff v1` in `VoronoiFVM`.

ChrisRackauckas:
The biggest issue is likely with differentiation of interpolations because we want to make sure it doesn't change the discontinuity handling behavior. So there's a bit to do here.

oscardssmith:
I think this might be ready now. Let's see if tests agree. The only change needed was a change to the interpolation tests to use a left-derivative since ForwardDiff@1.0 makes the derivative a right derivative by default.

ChrisRackauckas:
Is it breaking to our users that the ForwardDiff 1.0 uses the right derivative by default? We have a continuity argument, when using ForwardDiff that argument is no longer respected? Is that breaking? Or is that ForwardDiff behavior that is breaking which the user only gets if they bump to v1 as well so it's okay?

ChrisRackauckas:
Should we be correcting the forwarddiff branch in our continuity handling?

oscardssmith:
> Is it breaking to our users that the ForwardDiff 1.0 uses the right derivative by default?

I don't think so.
> We have a continuity argument, when using ForwardDiff that argument is no longer respected?

This would be breaking if our interpolations used ForwardDiff internally, but they don't, so it's not a problem.

JoshuaLampert:
Is there anything left to do or can this be merged? I think there are plenty of repos waiting for this, such that they can test ForwardDiff v1 in their repos.

oscardssmith:
I need to find an actual fix to `test/regression/hard_dae` but that is the last remaining item. The problem is in the test, but finding a version of this test that is valid for both `ForwardDiff@0.10` and `ForwardDiff@1` is a little tricky

oscardssmith:
lets see if the rebase magically fixes the failure? I'm very confused here because this is exactly the test I was running locally and it was passing there...

oscardssmith:
so good news, bad news. The good news is that of the failing tests, IntegrationTest / DelayDiffEq.jl/Interface/1.10 is `@test_broken` that start working. The bad news is that `hard_dae` is failing, and I have no idea why since it isn't failing locally...

Issue #2653, opened 2025-04-02
Sparsity pattern modification should not happen outside of DI pipeline


gdalle:
**Describe the bug 🐞**

In #2657, it seems that the sparsity pattern is modified using the `mass_matrix` outside of the DI pipeline (see https://github.com/SciML/OrdinaryDiffEq.jl/pull/2567#discussion_r2024482430). This is an issue because:
1. during preparation, the `sparsity_detector` from `AutoSparse` won't be aware of this change
2. during preparation, the `coloring_algorithm` from `AutoSparse` will compute a coloring for the unmodified matrix, and prepare decompression accordingly
3. during execution, the function `SparseMatrixColorings.decompress!` will either error or return incorrect results because it expected the unmodified sparsity pattern

This problem was detected by https://github.com/Deltares/Ribasim/pull/2201

**Expected behavior**

Any modification of the sparsity pattern should be done in a way that allows the `sparsity_detector` to see it.

**Minimal Reproducible Example 👇**

Not available yet.

**Error & Stacktrace ⚠️**

The typical stack trace could look something like:
```julia
ERROR: DimensionMismatch: all inputs to eachindex must have the same indices, got Base.OneTo(28) and Base.OneTo(26)
Stacktrace:
  [1] throw_eachindex_mismatch_indices(::IndexLinear, ::Base.OneTo{Int64}, ::Vararg{Base.OneTo{Int64}})
    @ Base ./abstractarray.jl:325
  [2] eachindex
    @ ./abstractarray.jl:393 [inlined]
  [3] eachindex
    @ ./abstractarray.jl:382 [inlined]
  [4] decompress!
    @ ~/.julia/packages/SparseMatrixColorings/BED6Y/src/decompression.jl:361 [inlined]
  [5] _sparse_jacobian_aux!(::Tuple{typeof(Ribasim.water_balance!), Vector{Float64}}, 
  [6] jacobian!
    @ ~/.julia/packages/DifferentiationInterface/7eD1K/ext/DifferentiationInterfaceSparseMatrixColoringsExt/jacobian.jl:237 [inlined]
```

**Environment (please complete the following information):**

  - Output of `using Pkg; Pkg.status()`

Not available yet

  - Output of `using Pkg; Pkg.status(; mode = PKGMODE_MANIFEST)`

Not available yet

  - Output of `versioninfo()`

Not available yet

**Additional context**

[...]

PR #2662, opened 2025-04-03
Fix sparsity pattern mismatch and test

jClugstor:
## Checklist

## Additional context

Adds test to make sure that the issues brought up in 
https://github.com/SciML/OrdinaryDiffEq.jl/issues/2653#issuecomment-2774941317 are taken care of. 

This will also need to come with a bump to SparseMatrixColorings to a version that's not out yet, I'll need to add the bump when the new version comes out.

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/src/alg_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/alg_utils.jl
@@ -106,7 +106,7 @@ function prepare_user_sparsity(ad_alg, prob)
     sparsity = prob.f.sparsity
 
     if !isnothing(sparsity) && !(ad_alg isa AutoSparse)
-        if sparsity isa SparseMatrixCSC
+        if sparsity isa SparseMatrixCSC && !DiffEqBase.has_jac(prob.f)
             if prob.f.mass_matrix isa UniformScaling
                 idxs = diagind(sparsity)
                 @. @view(sparsity[idxs]) = 1
--- a/lib/OrdinaryDiffEqDifferentiation/src/alg_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/alg_utils.jl
@@ -106,7 +106,7 @@ function prepare_user_sparsity(ad_alg, prob)
     sparsity = prob.f.sparsity
 
     if !isnothing(sparsity) && !(ad_alg isa AutoSparse)
-        if sparsity isa SparseMatrixCSC && !DiffEqBase.has_jac(prob.f)
+        if sparsity isa SparseMatrixCSC && !SciMLBase.has_jac(prob.f)
             if prob.f.mass_matrix isa UniformScaling
                 idxs = diagind(sparsity)
                 @. @view(sparsity[idxs]) = 1
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
@@ -205,9 +205,9 @@ function calc_J!(J, integrator, cache, next_step::Bool = false)
             if !isnothing(integrator.f.jac_prototype) &&
                integrator.f.jac_prototype isa SparseMatrixCSC
                
-                integrator.f.jac_prototype.nzval .= 1.0
-                J .= 1.0 .* integrator.f.jac_prototype
-                J.nzval .= 0.0
+                integrator.f.jac_prototype.nzval .= true
+                J .= true .* integrator.f.jac_prototype
+                J.nzval .= false
                 f.jac(J, uprev, p, t)
             else 
                 f.jac(J, uprev, p, t)
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
@@ -177,7 +177,9 @@ function calc_J!(J, integrator, cache, next_step::Bool = false)
 
             # we need to set all nzval to a non-zero number
             # otherwise in the following line any zero gets interpreted as a structural zero
-            if !isnothing(integrator.f.jac_prototype)
+            if !isnothing(integrator.f.jac_prototype) && 
+                integrator.f.jac_prototype isa SparseMatrixCSC
+
                 integrator.f.jac_prototype.nzval .= 1.0
                 J .= 1.0 .* integrator.f.jac_prototype
                 J.nzval .= 0.0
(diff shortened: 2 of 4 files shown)


ChrisRackauckas:
Fails because of missing import.

This would require a bump to the minimum in order to pass?

jClugstor:
Yes, it will need to use https://github.com/gdalle/SparseMatrixColorings.jl/pull/224, which hasn't been released yet. I jumped the gun on this one a little bit.

ChrisRackauckas:
Okay, update this PR to do all of that when it's out and then ping me or @oscardssmith for review. I'm going to ignore this for now.

jClugstor:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2653

lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
```diff
@@ -172,7 +172,18 @@ function calc_J!(J, integrator, cache, next_step::Bool = false)
...
-            f.jac(J, duprev, uprev, p, uf.α * uf.invγdt, t)
+            # need to do some jank here to account for sparsity pattern of W
+            # https://github.com/SciML/OrdinaryDiffEq.jl/issues/2653
+
+            # we need to set all nzval to a non-zero number
+            # otherwise in the following line any zero gets interpreted as a structural zero
+            integrator.f.jac_prototype.nzval .= 1.0
+            J .= 1.0 .* integrator.f.jac_prototype
+            J.nzval .= 0.0
+            f.jac(J, uprev, p, t)
```
ChrisRackauckas, line 183: That's not the right Jac call. This is the DAE branch.

lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
```diff
@@ -183,7 +194,17 @@ function calc_J!(J, integrator, cache, next_step::Bool = false)
...
+            # https://github.com/SciML/OrdinaryDiffEq.jl/issues/2653
+
+            # we need to set all nzval to a non-zero number
+            # otherwise in the following line any zero gets interpreted as a structural zero
+            integrator.f.jac_prototype.nzval .= 1.0
+            J .= 1.0 .* integrator.f.jac_prototype
+            J.nzval .= 0.0
             f.jac(J, uprev, p, t)
+            MM = integrator.f.mass_matrix isa UniformScaling ? integrator.f.mass_matrix(length(integrator.u)) : integrator.f.mass_matrix
+            J .= J .+ MM
```
ChrisRackauckas, line 207: This doesn't go here.

lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
jClugstor: Does the DAE branch not need this as well?

jClugstor: Oh yes, just a copy paste error

lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
jClugstor: Right, this will cause the mass matrix to be applied twice, won't it.

This might need to go in `jacobian2W!`? Looks like that might slow things down though...

test/interface/sparsediff_tests.jl
```diff
@@ -82,3 +85,37 @@ for f in [f_oop, f_ip]
     end
 end
 
+# test for https://github.com/SciML/OrdinaryDiffEq.jl/issues/2653#issuecomment-2778430025
```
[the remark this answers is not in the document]
jClugstor: Oh yes, good idea.

lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
ChrisRackauckas: `jacobian2W!` already does this?

jClugstor: Right, looks like taking away those lines makes it work without applying the MM twice. I thought there would be issues in `jacobian2W!` if the sparisty pattern of `J` was still the same as the `jac_prototype`s, but apparently not.

lib/OrdinaryDiffEqDifferentiation/src/alg_utils.jl
```diff
@@ -106,7 +106,7 @@ function prepare_user_sparsity(ad_alg, prob)
     sparsity = prob.f.sparsity
 
     if !isnothing(sparsity) && !(ad_alg isa AutoSparse)
-        if sparsity isa SparseMatrixCSC
+        if sparsity isa SparseMatrixCSC && !DiffEqBase.has_jac(prob.f)
```
ChrisRackauckas, line 109:
```suggestion
        if sparsity isa SparseMatrixCSC && !SciMLBase.has_jac(prob.f)
```

lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
```diff
@@ -172,7 +172,21 @@ function calc_J!(J, integrator, cache, next_step::Bool = false)
...
+            # https://github.com/SciML/OrdinaryDiffEq.jl/issues/2653
+
+            # we need to set all nzval to a non-zero number
+            # otherwise in the following line any zero gets interpreted as a structural zero
+            if !isnothing(integrator.f.jac_prototype) && 
+                integrator.f.jac_prototype isa SparseMatrixCSC
+
+                integrator.f.jac_prototype.nzval .= 1.0
+                J .= 1.0 .* integrator.f.jac_prototype
+                J.nzval .= 0.0
```
gdalle, line 185:
```suggestion
                integrator.f.jac_prototype.nzval .= true
                J .= true .* integrator.f.jac_prototype
                J.nzval .= false
```
These are "minimally-typed" zeros and ones

lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
```diff
@@ -183,7 +197,21 @@ function calc_J!(J, integrator, cache, next_step::Bool = false)
...
+            # https://github.com/SciML/OrdinaryDiffEq.jl/issues/2653
+
+            # we need to set all nzval to a non-zero number
+            # otherwise in the following line any zero gets interpreted as a structural zero
+            if !isnothing(integrator.f.jac_prototype) &&
+               integrator.f.jac_prototype isa SparseMatrixCSC
+               
+                integrator.f.jac_prototype.nzval .= 1.0
+                J .= 1.0 .* integrator.f.jac_prototype
+                J.nzval .= 0.0
```
gdalle, line 210:
```suggestion
                integrator.f.jac_prototype.nzval .= true
                J .= true .* integrator.f.jac_prototype
                J.nzval .= false
```

ChrisRackauckas:
The failures are internal ITP shifting by one ulp and enzyme. I'll handle those separately.

Issue #2659, opened 2025-04-03
Undefined names in OrdinaryDiffEqDifferentiation


gdalle:
**Describe the bug 🐞**

Some names are undefined in OrdinaryDiffEqDifferentiation, as caught by JET:

**Expected behavior**

`SparseArrays.spzeros` should be imported here

https://github.com/SciML/OrdinaryDiffEq.jl/blob/f3d421860418f4df57989a3ddad40ab1a051fffb/lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl#L16

`res` in `JVPCache` should be replaced with the actual thing we want to return (not sure what that is, otherwise I would have opened the PR myself)

https://github.com/SciML/OrdinaryDiffEq.jl/blob/f3d421860418f4df57989a3ddad40ab1a051fffb/lib/OrdinaryDiffEqDifferentiation/src/operators.jl#L45-L48

**Minimal Reproducible Example 👇**

```julia
julia> JET.test_package(OrdinaryDiffEqDifferentiation, target_defined_modules=true)
JET-test failed at /Users/guillaumedalle/.julia/packages/JET/sQkqi/src/JETBase.jl:1071
  Expression: (JET.report_package)(OrdinaryDiffEqDifferentiation; toplevel_logger = nothing, target_defined_modules = true)
  ═════ 3 possible errors found ═════
  ┌ build_J_W(alg::Any, u::Any, uprev::Any, p::Any, t::Any, dt::Any, f::F, jac_config::Any, ::Type{…}, ::Val{…}) where {…} @ OrdinaryDiffEqDifferentiation /Users/guillaumedalle/Documents/GitHub/Julia/OrdinaryDiffEq.jl/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl:767
  │ `OrdinaryDiffEqDifferentiation.spzeros` is not defined: OrdinaryDiffEqDifferentiation.spzeros
  └────────────────────
  ┌ build_J_W(alg::Any, u::Any, uprev::Any, p::Any, t::Any, dt::Any, f::F, jac_config::Any, ::Type{…}, ::Val{…}) where {…} @ OrdinaryDiffEqDifferentiation /Users/guillaumedalle/Documents/GitHub/Julia/OrdinaryDiffEq.jl/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl:800
[...]

PR #2664, opened 2025-04-03
Fix undefined res and spzeros 

jClugstor:
## Checklist

## Additional context

I thought about adding some of the JET tests in this PR as well, but getting it to test all the packages will be a lot, and after testing it turns out that some of the things it warns about aren't "real", so I think it can be a separate PR so we can think about the best way to add JET tests. 

This should fix https://github.com/SciML/OrdinaryDiffEq.jl/issues/2659

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/src/operators.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/operators.jl
@@ -42,10 +42,6 @@ function JVPCache(f::DiffEqBase.AbstractDiffEqFunction, du, u, p, t; autodiff)
     return JVPCache{eltype(du)}(jvp_op, f, du, u, p, t)
 end
 
-function (op::JVPCache)(v, u, p, t)
-    return op.jvp_op(op.du, v, u, p, t) 
-end
-
 function (op::JVPCache)(Jv, v, u, p, t)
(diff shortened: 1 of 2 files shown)


lib/OrdinaryDiffEqDifferentiation/src/operators.jl
```diff
@@ -43,8 +43,7 @@ function JVPCache(f::DiffEqBase.AbstractDiffEqFunction, du, u, p, t; autodiff)
 end
 
 function (op::JVPCache)(v, u, p, t)
-    op.jvp_op(op.du, v, u, p, t)
-    return res
+    return op.jvp_op(op.du, v, u, p, t)
```
ChrisRackauckas, line 46: the out of place call is mutating?

jClugstor: We actually don't need an out of place version of this in the first place, we never use this on OOP problems, so getting rid of this is best.

ChrisRackauckas:
Can you add the mwe?

jClugstor:
What kind of MWE does this one need?

ChrisRackauckas:
oh wait the JET tests are a separate PR

Issue #2663, opened 2025-04-03
Add JET testing for some static analysis in tests


jClugstor:
This would be nice so that we can avoid issues like this one: https://github.com/SciML/OrdinaryDiffEq.jl/issues/2659

PR #2665, opened 2025-04-03
Add JET and Aqua Tests to Subpackages

jClugstor:
## Checklist

## Additional context

This adds JET `test_package` https://aviatesk.github.io/JET.jl/dev/jetanalysis/#JET.test_package in the typo mode to catch typos in every subpackage. 

https://github.com/SciML/OrdinaryDiffEq.jl/issues/2663


ChrisRackauckas:
Can you also do a full Aqua QA test?

gdalle:
Note that you should probably be careful with JET versioning, for instance `JET.test_package` is not yet fully supported by JET v0.10 (see https://github.com/aviatesk/JET.jl/releases/tag/v0.10.0)

jClugstor:
Many of these are legitimate JET/Aqua failures. Adding these tests without taking care of those issues (if we can) will add a lot of noise when looking at test failures. I can disable the failing tests for each package though to get it set up.

oscardssmith:
IMO it's probably at least worth a try to fix the issues.

ChrisRackauckas:
Do the same as what was done for the other repos. Cut it down to subsets, disable what will take too much time for now and file issues on those, but get what we can https://github.com/SciML/ModelingToolkit.jl/pull/2364/files#diff-77acaaefd44e4d00398e2700bc87be028acc6625d6c01c480214e2b992929c0aR3

jClugstor:
Jet complains about `DAECache` not being defined in OrdinaryDiffEqCore, and I can't find any reference to it besides here: https://github.com/SciML/OrdinaryDiffEq.jl/blob/550948230a176ff428875d66356be600e5a59988/lib/OrdinaryDiffEqCore/src/solve.jl#L500C11-L500C33. Where is `DAECache` defined?

oscardssmith:
nowhere! this is the fun part about JET. It catches your totally broken code.

jClugstor:
I'm guessing there are no tests with `recompile_flag == false` and `!(_alg isa OrdinaryDiffEqAlgorithm)`. Is that just a really uncommon scenario?

oscardssmith:
both halves of that are unlikely yeah.

jClugstor:
Most of the failures are from the `pre` tests since JET 0.9 doesn't work on Julia 1.12. Is there a nice way to specify that for Julia 1.12 we should use Jet 0.10? Or should I just turn off the JET tests based on the VERSION ?

lib/ImplicitDiscreteSolve/Project.toml
```diff
@@ -23,10 +23,14 @@ SymbolicIndexingInterface = "0.3.38"
 Test = "1.10.0"
 UnPack = "1.0.2"
 julia = "1.10"
+JET = "0.9.18"
```
oscardssmith, line 26:
```suggestion
JET = "0.9.18, 0.10"
```
This should support Julia 1.12

jClugstor: I thought about that but wouldn't that make it always use JET 0.10, unless JET has a compat restriction for Julia?

gdalle: Keep in mind that JET 0.10 is marketed as experimental, especially for `test_package`

jClugstor: This just uses it for typos, so hopefully that would be fine. Anyways, it doesn't work because it needs `Julia 1.12.0-beta1.11`, but `pre` only uses `Julia 1.12.0-beta1`. Very cool.

oscardssmith: It's `JET's job to provide appropriate compats (and I believe it does so).

> Julia 1.12.0-beta1.11, but pre only uses Julia 1.12.0-beta1. Very cool.

never mind. That sounds like a JET bug...

jClugstor: Is Julia 1.12.0-beta1.11 even a thing?

oscardssmith: no

oscardssmith: my buest guess is there's a missing comma?

jClugstor: It does ask you to use a specific branch at https://github.com/aviatesk/JET.jl/blob/master/CHANGELOG.md#0103: 
"However, to use JET v0.10.3, you need to use Julia version v"1.12.0-beta1.11" or later. As of April 19, 2025, you need to build Julia from https://github.com/JuliaLang/julia/pull/58009 for using this 1.12 beta version."

jClugstor: Ok, so probably just a typo there. The real issue is that `test_package` available in JET 0.10.4, even though the changelog says it should be.

jClugstor:
I fixed as many of the small things like typos, compat bounds etc. that I could. The rest of the JET things detected seem like they will take some more involved work to fix.

ChrisRackauckas:
Rebase? It looks like there's some tests I fixed that are failing here.

Also seems like there are JET errors still https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/14862925186/job/41732330999?pr=2665

jClugstor:
It should be up to date, last commit before I rebased last was the `fixmaster` one. 

Yes, some of the JET failures are things I don't know how to fix or should probably go in a separate PR. There are only a few like that and I'll file issues for each of them.

ChrisRackauckas:
Okay anything failing needs to be properly marked `@test_broken` so that CI is always green.

Issue #2621, opened 2025-03-04
Add new Verner methods


oscardssmith:
Vern has published updated versions of his tableaus that are more efficient/have better interpolants.
https://www.sfu.ca/~jverner/RKV76.IIa.Efficient.000003389335684.240711.FLOAT6040OnWeb
https://www.sfu.ca/~jverner/RKV87.IIa.Efficient.000000011182-240510.FLOAT6040OnWeb
https://www.sfu.ca/~jverner/RKV98.IIa.Efficient.00000034399.240407.BetterEfficientonWeb

PR #2693, opened 2025-05-01
WIP: add fixed-step RKV87-IIa (not yet working)

Irura-Coder:
This PR adds the Verner 8/7 (RKV87-IIa) method following issue #2621.
It compiles without errors, but RKV87 doesn’t show up in the REPL and I haven’t gotten the test to pass yet.
Would appreciate help debugging what’s missing — I suspect it’s not being registered properly.
@ChrisRackauckas


lib/OrdinaryDiffEqVerner/src/alg_utils.jl
```diff
@@ -12,4 +12,10 @@ alg_stability_size(alg::Vern7) = 4.6400
 alg_stability_size(alg::Vern8) = 5.8641
 alg_stability_size(alg::Vern9) = 4.4762
 
+# ── Verner RKV87-IIa traits ────────────────────────────────────────
+alg_order(::RKV87)       = 8        # primary solution order
+embedded_order(::RKV87)  = 7        # will matter once adaptivity is added
+isfsal(::RKV87)          = false    # pair is NOT FSAL
+has_dense_output(::RKV87)= false    # we haven’t wired the interpolant yet
```
ChrisRackauckas, line 19: that's not a trait, also it has a fallback to linear.

lib/OrdinaryDiffEqVerner/src/verner_rk_perform_step.jl
```diff
@@ -1013,6 +1013,18 @@ end
...
+#  RKV87-IIa wrappers
+# ------------------------------------------------------------------
+
+initialize!(integrator, cache::RKV87ConstantCache) = nothing
+
+# The tableau we built is an ExplicitRKTableau with 13 stages
+perform_step!(integrator,
+              cache::RKV87ConstantCache,
+              tab::ExplicitRKTableau{13}) =
+    verner_lowstorage_step!(integrator, cache, tab)
```
ChrisRackauckas, line 1026: This function doesn't exist? Is this outputted by an LLM?

PR #2695, opened 2025-05-03
Don't double recurse init with default algorithm

ChrisRackauckas:
Fixes https://github.com/SciML/SciMLSensitivity.jl/issues/1165


oscardssmith:
should we get a test for  SciML/ModelingToolkit.jl#3327?

PR #2699, opened 2025-05-05
use [sources]

oscardssmith:
This is the OrdinaryDiffEq followup to https://github.com/SciML/NonlinearSolve.jl/pull/571.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -85,7 +85,7 @@ jobs:
         run: |
           julia --project=. -e '
               using Pkg;
-              Pkg.develop(map(path ->Pkg.PackageSpec.(;path="$(@__DIR__)/lib/$(path)"), readdir("./lib")));
+              VERSION < v"1.11" && Pkg.develop(map(path ->Pkg.PackageSpec.(;path="$(@__DIR__)/lib/$(path)"), readdir("./lib")));
           '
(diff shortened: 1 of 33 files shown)


ChrisRackauckas (approved):

ranocha:
Looking at https://pkgdocs.julialang.org/v1.10/toml-files/ and https://pkgdocs.julialang.org/v1.11/toml-files/, it seems like the `sources` section is only available in Julia 1.11, not in the LTS 1.10. Is this correct?

oscardssmith:
that is correct. on 1.10 and under, this change won't do anything (but everything will still work)

ranocha:
Thanks! Just wanted to confirm this.

oscardssmith:
@avik-pal what did you do to make this work for CI on NonlinearSolve?

avik-pal:
it pretty much just worked :sweat_smile:. NonlinearSolve does have a per subpackage CI workflow. Can you try and change https://github.com/oscardssmith/OrdinaryDiffEq.jl/blob/7a35128936dd2fbbf759db93545cf87b3fc0a867/.github/workflows/CI.yml#L84 to run only on 1.10?

oscardssmith:
lets try it...

oscardssmith:
that seems to work. Thanks!

PR #2715, opened 2025-05-19
fix `sources` in Ci for 1.11 and up

oscardssmith:
Why do we have both `.buildkite` and `.github` ci?


ChrisRackauckas:
buildkite is for GPUs, and multithreading, and test sets that were just long. We can probably move some back to Github Actions by now.

ChrisRackauckas:
Github actions are still borked.

PR #2718, opened 2025-05-21
fix NonlinearSolve tests (broken by #2573)

oscardssmith:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/2573 added an `abs` that wasn't supposed to be there. Removing it fixes the test.


ChrisRackauckas:
Still lots of sources issues...

oscardssmith:
yeah. I still need to figure out why https://github.com/SciML/OrdinaryDiffEq.jl/pull/2717 didn't work, but https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/15188283323/job/42714110714 shows that this fixes the red for OrdinaryDiffEQNonlinearSolve (which means that our LTS CI is going through completely finally)

PR #2731, opened 2025-05-31
Bump DI to v0.7

ErikQQY:
Test with https://github.com/SciML/BoundaryValueDiffEq.jl/issues/337


ChrisRackauckas:
https://buildkite.com/julialang/ordinarydiffeq-dot-jl/builds/7143#01972b03-5175-4b7c-ba82-465dee614697/996-1785

lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
```diff
@@ -309,8 +309,8 @@ function resize_jac_config!(cache, integrator)
             ad_left = autodiff_alg
         end
 
-        cache.jac_config = ([DI.prepare!_jacobian(
-                                   uf, cache.du1, config, ad, integrator.u)
+        SciMLBase.@reset cache.jac_config = ([DI.prepare!_jacobian(
```
ErikQQY, line 312: Is this an effective way to update `jac_config` in `cache`?

ChrisRackauckas: No because matrices can't be resized.

lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
```diff
@@ -268,12 +268,12 @@ function build_jac_config(alg, f::F1, uf::F2, du1, uprev,
                 autodiff_alg_reverse = dir_reverse
             end
 
-            jac_config_forward = DI.prepare_jacobian(uf, du1, autodiff_alg_forward, u)
-            jac_config_reverse = DI.prepare_jacobian(uf, du1, autodiff_alg_reverse, u)
+            jac_config_forward = DI.prepare_jacobian(uf, du1, autodiff_alg_forward, u, strict=Val(false))
```
oscardssmith, line 271: is the `strict` argument new in DI 0.7? If so, this won't work with DI 0.6, right?

lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
oscardssmith: what does `SciMLBase.@reset` actually do?

ChrisRackauckas: It's Acessors.jl

lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
ChrisRackauckas: I don't know

oscardssmith: ah looks like it already existed and the change was just switching the default from false to true. https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/799/files

oscardssmith:
I'm investigating why this wasn't working with `strict` and it looks like the issue is that we're calling `build_*_config` before `FunctionWrappersWrapers`ing the `f`.

ChrisRackauckas:
It definitely wraps before that because it wraps in the DiffEqBase part before it ever hits OrdinaryDiffEq.jl. The question would instead be, who unwraps it.

Issue #2501, opened 2024-10-24
Pseudospectral method for semilinear PDE (Split-Step Fourier, RKIP)


Azercoco:
Hi,

I am doing a PhD in nonlinear optics in which we encounter several semilinear parabolic PDE like the Nonlinear Schrödinger Equation (NLSE) and its variant.

Most of these PDEs are semilinear meaning they can be expressed as 

$$\partial_t y = \hat{D}(\omega)y + N(y)$$ where $\hat{D}$ is a constant stiff linear operator but which can be easily computed in Fourier space. This allows to create efficient integrator as $\exp(A \times dt)$ can be easily computed.

For now, in the available algorithms, only exponential integrator (ETDRK) are able to take profit of this structure but several other algorithms exist:

- Split-Step Fourier : this is a implementation of the Strang splitting ([based on the Baker-Campbell-Haussdorf formula](https://en.wikipedia.org/wiki/Baker%E2%80%93Campbell%E2%80%93Hausdorff_formula)). This method is the most commonly used. 

The idea is to make the approximation $\exp( h  (\hat{D} + N) ) \approx \exp( h \hat{D}) \exp(h N) + \mathcal{O}(h^2)$. For the NLSE, $exp( h \hat{D})$ is easily computed in the spectral/frequency domain while $\exp(h N)$ can be easily computed in the time domain. 

There is numerous improvement of this algorithm in the literature with adaptative time stepping, higher order ...

- RKIP (RK in the interaction picture also sometimes refer as the integrating factor IF in exponential integrator):  

The idea is to make the transformation $y_I(t) = exp(t\hat{D})y$ then after transformation the equation become:

$$\partial_t y_I = N_I(t) y_t$$ with $N_I(t) =  exp(t\hat{D}) N exp(-t\hat{D})$. If we write a single Runge-Kutta step, we can see that we actually only compute $exp(h\hat{D})$. This allows to use any explicit Runge-Kutta method with their order and adaptive time stepping.

These algorithms could be really useful for fast integration of semilinear PDE with adaptative time stepping.  Because they use FTTs, they also greatly benefit from GPU acceleration. 

** Implentation difficulties

Split-Step Fourier method could be implemented within a more generalized Strang splitting for Split ODE Problem with possibility of using both $exp(f_1)$ and $exp(f_2)$ 

[...]

PR #2736, opened 2025-06-10
[New algorithm] Runge-Kutta in the interaction picture (RKIP)

Azercoco:
See #2501.

The RKIP is a method to solve `SplitODEProblem` where `f1` is a `AbstractSciMLOperator` which has an efficient way of computing/storing `exp(A*t)`. It performs a change of variable to turn the (stiff) problem into a non-stiff one which can be solved with any Runge-Kutta methods.

The RKIP method is well suited for pseudospectral method for solving elliptic PDE (such as the ones from [FourierFlows.jl](https://github.com/FourierFlows/FourierFlows.jl)). 

This method is similar to the exponential Runge-Kutta (like ETDRK4) but allows for adaptative time stepping and arbitrary order. In counterpart, it requires a constant linear operator (so no Jacobian can be used).

### DIFF
--- a/lib/OrdinaryDiffEqRKIP/Project.toml
+++ b/lib/OrdinaryDiffEqRKIP/Project.toml
@@ -1,0 +1,44 @@
+name = "OrdinaryDiffEqRKIP"
+uuid = "a4daff8c-1d43-4ff3-8eff-f78720aeecdc"
+authors = ["Azercoco <20425137+Azercoco@users.noreply.github.com>"]
+version = "1.0.0"
+
+
+[sources]
+OrdinaryDiffEqCore = {path = "../OrdinaryDiffEqCore"}
+
+[deps]
+DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
+DiffEqDevTools = "f3b72e0c-5b89-59e1-b016-84e28bfd966d"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MaybeInplace = "bb5d69b7-63fc-4a16-80bd-7e42200c7bdb"
+OrdinaryDiffEqCore = "bbf590c4-e513-4bbe-9b18-05decba2e5d8"
+SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
+SciMLOperators = "c0aeaf25-5076-4817-a8d5-81caf7dfa961"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
+
+[compat]
+DiffEqBase = "6.175"
+DiffEqDevTools = "2.48"
+MaybeInplace = "0.1.4"
+OrdinaryDiffEqCore = "1.26"
+SciMLBase = "2.99"
+SciMLOperators = "1.3.1"
+StaticArrays = "1.9.13"
+UnPack = "1.0.2"
+LinearAlgebra = "1.11"
+CUDA = "5.5.2"
+FFTW = "1.8.0"
+SafeTestsets = "0.1.0"
+
+julia = "1.11"
+
+[extras]
+FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
+SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+
+[targets]
+test = ["FFTW", "Test", "SafeTestsets", "CUDA"]
--- a/lib/OrdinaryDiffEqRKIP/src/OrdinaryDiffEqRKIP.jl
+++ b/lib/OrdinaryDiffEqRKIP/src/OrdinaryDiffEqRKIP.jl
@@ -1,0 +1,23 @@
+module OrdinaryDiffEqRKIP
+
+using LinearAlgebra: ldiv!, exp, axpy!, norm, mul!
+using SciMLOperators: AbstractSciMLOperator
+using UnPack: @pack!, @unpack
+using MaybeInplace: @bb
+using SciMLBase: isinplace
+using DiffEqBase: ExplicitRKTableau
+using DiffEqDevTools: constructDormandPrince6
+
+import OrdinaryDiffEqCore: OrdinaryDiffEqAdaptiveExponentialAlgorithm, alg_adaptive_order,
+    alg_order, alg_cache, @cache, SplitFunction, get_fsalfirstlast, initialize!, perform_step!,
+    has_dtnew_modification, calculate_residuals, calculate_residuals!, increment_nf!, OrdinaryDiffEqAdaptiveAlgorithm, OrdinaryDiffEqMutableCache, dtnew_modification
+
+
+include("rkip_cache.jl")
+include("algorithms.jl")
+include("rkip_utils.jl")
+include("rkip_perform_step.jl")
+
+export RKIP
+
+end
--- a/lib/OrdinaryDiffEqRKIP/src/OrdinaryDiffEqRKIP.jl
+++ b/lib/OrdinaryDiffEqRKIP/src/OrdinaryDiffEqRKIP.jl
@@ -14,7 +14,7 @@ import OrdinaryDiffEqCore: OrdinaryDiffEqAdaptiveExponentialAlgorithm, alg_adapt
                            has_dtnew_modification, calculate_residuals,
                            calculate_residuals!, increment_nf!,
                            OrdinaryDiffEqAdaptiveAlgorithm, OrdinaryDiffEqMutableCache,
-                           dtnew_modification
+                           dtnew_modification, generic_solver_docstring
 
 include("rkip_cache.jl")
 include("algorithms.jl")
--- a/lib/OrdinaryDiffEqRKIP/src/OrdinaryDiffEqRKIP.jl
+++ b/lib/OrdinaryDiffEqRKIP/src/OrdinaryDiffEqRKIP.jl
@@ -9,9 +9,12 @@ using DiffEqBase: ExplicitRKTableau
 using DiffEqDevTools: constructDormandPrince6
 
 import OrdinaryDiffEqCore: OrdinaryDiffEqAdaptiveExponentialAlgorithm, alg_adaptive_order,
-    alg_order, alg_cache, @cache, SplitFunction, get_fsalfirstlast, initialize!, perform_step!,
-    has_dtnew_modification, calculate_residuals, calculate_residuals!, increment_nf!, OrdinaryDiffEqAdaptiveAlgorithm, OrdinaryDiffEqMutableCache, dtnew_modification
-
+                           alg_order, alg_cache, @cache, SplitFunction, get_fsalfirstlast,
+                           initialize!, perform_step!,
+                           has_dtnew_modification, calculate_residuals,
+                           calculate_residuals!, increment_nf!,
+                           OrdinaryDiffEqAdaptiveAlgorithm, OrdinaryDiffEqMutableCache,
+                           dtnew_modification
 
 include("rkip_cache.jl")
 include("algorithms.jl")
--- a/lib/OrdinaryDiffEqRKIP/src/algorithms.jl
+++ b/lib/OrdinaryDiffEqRKIP/src/algorithms.jl
@@ -1,0 +1,155 @@
+using OrdinaryDiffEqCore: OrdinaryDiffEqCore
+
+mutable struct RKIP_ALG{tableauType<:ExplicitRKTableau,elType,dtType<:AbstractVector{elType}} <: OrdinaryDiffEqAdaptiveAlgorithm
+    tableau::tableauType
+    dt_for_expÂ_caching::dtType
+    clamp_lower_dt::Bool
+    clamp_higher_dt::Bool
+    use_ldiv::Bool
+    cache::Union{Nothing,RKIPCache}
+end
+
+alg_order(alg::RKIP_ALG) = alg.tableau.order
+alg_adaptive_order(alg::RKIP_ALG) = alg.tableau.adaptiveorder
+
+has_dtnew_modification(alg::RKIP_ALG) = true
+
+function dtnew_modification(alg::RKIP_ALG{tableauType,elType,dtType}, dtnew) where {tableauType,elType,dtType}
+    @unpack dt_for_expÂ_caching = alg
+    if first(alg.dt_for_expÂ_caching) > dtnew && alg.clamp_lower_dt
+        dtnew = first(alg.dt_for_expÂ_caching)
+    elseif last(alg.dt_for_expÂ_caching) < dtnew && alg.clamp_higher_dt
+        dtnew = last(alg.dt_for_expÂ_caching)
+    else
+        dtnew = alg.dt_for_expÂ_caching[searchsortedfirst(alg.dt_for_expÂ_caching, dtnew)]
+    end
+    return dtnew
+end
+
+dtnew_modification(_, alg::RKIP_ALG, dtnew) = dtnew_modification(alg, dtnew)
+
+
+"""
+Runge-Kutta in the Interaction Picture (RIKP) method.
+
+This is suited for solving semilinear problem of the form:
+
+```math
+\frac{du}{dt} =  Au + f(u,p,t)
+```
+where A is possibly stiff constant linear operator whose scaled exponential exp(Ah) can be calculated efficiently for any h.
+
+The problem is first transformed into the Interaction Picture.
+
+```math
+u_I(t) = \exp(At) u_0
+\frac{du_I}{dt} =  f_I(u_I,p,t)
+f_I(u_I,p,t) = f(exp(-At)u_I, p, t)
+```
+This new system is then solved with an explicit Runge-Kutta method.
+
+This type of problem is often encountered in semilinear parabolic PDE: heat diffusion, Schrödinger equation ...
+
+This method is only implemented for semilinear problem: `SplitODEProblem` when the first function `f1` is a `AbsractSciMLOperator` A.
+	
+The algorithm needs for A to have the following traits of the `AbsractSciMLOperator` implemented
+
+```julia
+LinearAlgebra.exp(A, t) = #...
+A(du, u, v, p, t) # for in-place problem
+A(u, v p, t) # for out-of-place problem 
+```
+
+Moreover, the resulting ```expA = exp(A, t)``` must also implement ```(expA)(u, v, p, t)``` and/or ```(expA)(du, u, v, p, t)```.
+The algorithm only works for constant `A` (does no depend on v, p, t).
+
+The algorithm will cache and resue the computed operator-exponential for a set of steps. 
+
+# Arguments
+- `dtmin::T`: the smallest step `dt` for which `exp(A*dt)` will be cached. Default is `1e-3`
+- `dtmax::T`: the largest step `dt` for which `exp(A*dt)` will be cached. Default is `1.0`
+
+The set of cached steps size `dt_for_expA_caching` will follow a geometric progression such that `first(dt_for_expA_caching) = dtmin` and `last(dt_for_expA_caching) = dtmax`.
+
+Time stepping can still happen outside the bonds but this will results in no cache (as `exp(A*h)` will be computed each step) degrading the performances. The time step can be clamped within the cache range through the args `clamp_lower_dt`  and `clamp_higher_dt`.
+
+# Optional Arguments
+- `nb_of_cache_step::Integer`: the number of steps. Default is `100`.
+- `tableau::ExplicitRKTableau`: the Runge-Kutta Tableau to use, default is `constructDormandPrince6()`.
+- `clamp_lower_dt::Bool`: weither to clamp proposed step to the smallest cached step in order to force the use of cached exponential, improving performance. 
+	This may prevent reaching the desired tolerance. Default is `false`.
+- `clamp_higher_dt::Bool`: weither to clamp proposed step to the largest cached step in order to force the use of cached exponential, improving performance. 
+	This can cause performance degredation if `integrator.dtmax` is too small. Default is `true`.
+- `use_ldiv::Bool`: weither, to use `ldiv(exp(A, t), v)` instead of caching `exp(A, -t)*v`. Reduce the memory usage but slightly less efficient. `ldiv` must be implemented. Only works for in-place problem. Default is `false`.
+
+Cached operator exponentials are stored in the alorithm meaning that in the following code:
+
+```julia
+rkip = RKIP()
+solve(ode_prob_1, rkip, t1)
+solve(ode_prob_2, rkip, t2)
+````
+
+will reuse the cache of the first solve. This can be useful when needed to solve several time the same problem but one must be sure that `A` does not change. 
+
+"""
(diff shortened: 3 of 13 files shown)


lib/OrdinaryDiffEqRKIP/src/algorithms.jl
```diff
@@ -0,0 +1,178 @@
...
+
+will reuse the precomputed exponential cached during the first `solve` call.
+This can be useful for solving several times successively problems with a common `A`.
+
+"""
+REFERENCE = """Zhongxi Zhang, Liang Chen, and Xiaoyi Bao, "A fourth-order Runge-Kutta in the interaction picture method for numerically solving the coupled nonlinear Schrödinger equation," Opt. Express 18, 8261-8276 (2010)"""
+
+KEYWORD_DESCRIPTION = """
+- `nb_of_cache_step::Integer`: the number of steps. Default is `100`.
+- `tableau::ExplicitRKTableau`: the Runge-Kutta Tableau to use. Default is `constructDormandPrince6()`.
```
oscardssmith: why is this the default? The Dormand Prince integrators tend to not be very good.

Azercoco: I need to recheck but from my benchmark on NLSE-like equations, it seems to perform the best. What others methods would you suggest ?

oscardssmith: Tsit5 or Vern6/7 tend to be good see https://docs.sciml.ai/SciMLBenchmarksOutput/stable/NonStiffODE/FitzhughNagumo_wpd/ for details.

Azercoco: I redid my benchmark on NLSE-like equation with 2048 points. Here are the results:

```
 --- Tableau benchmark result for reltol = 1.0e-02 ---
BogakiShampine5 : 0.44 sec | nf = 29328
Dverk : 0.519 sec | nf = 34462
DormandLockyerMcCorriganPrince6 : 0.584 sec | nf = 37395
PapakostasPapaGeorgiou5 : 0.621 sec | nf = 40968
CashKarp : 0.666 sec | nf = 44853
SharpVerner6 : 0.715 sec | nf = 48073
SharpSmart5 : 0.726 sec | nf = 48333
Verner6 : 0.761 sec | nf = 45852
TsitourasPapakostas6 : 0.833 sec | nf = 50918
RKF4 : 0.855 sec | nf = 57412
RKF5 : 0.855 sec | nf = 51392
DormandPrince6 : 0.862 sec | nf = 56120
Papakostas6 : 0.877 sec | nf = 60157
Tsitouras5 : 1.118 sec | nf = 69153
VernerEfficient6 : 1.216 sec | nf = 82045
Verner9162 : 1.307 sec | nf = 91051
VernerRobust6 : 1.384 sec | nf = 89436
ClassicVerner6 : 1.471 sec | nf = 92704
BogakiShampine3 : 1.724 sec | nf = 113520
``` 



```
 --- Tableau benchmark result for reltol = 1.0e-04 ---
BogakiShampine5 : 0.757 sec | nf = 49558
CashKarp : 0.799 sec | nf = 52757
Verner6 : 1.013 sec | nf = 63260
Dverk : 1.018 sec | nf = 63362
PapakostasPapaGeorgiou5 : 1.058 sec | nf = 64503
SharpVerner6 : 1.076 sec | nf = 72735
RKF5 : 1.133 sec | nf = 75195
DormandPrince6 : 1.17 sec | nf = 77064
Papakostas6 : 1.217 sec | nf = 77561
TsitourasPapakostas6 : 1.276 sec | nf = 83796
DormandLockyerMcCorriganPrince6 : 1.309 sec | nf = 88410
SharpSmart5 : 1.41 sec | nf = 83643
RKF4 : 1.515 sec | nf = 92612
VernerEfficient6 : 1.581 sec | nf = 108132
ClassicVerner6 : 1.731 sec | nf = 113597
VernerRobust6 : 1.769 sec | nf = 121261
Verner9162 : 1.916 sec | nf = 124833
Tsitouras5 : 2.059 sec | nf = 132078
BogakiShampine3 : 2.699 sec | nf = 177924
``` 

```
 --- Tableau benchmark result for reltol = 1.0e-06 ---
Dverk : 0.852 sec | nf = 53451
CashKarp : 0.932 sec | nf = 57216
DormandLockyerMcCorriganPrince6 : 0.967 sec | nf = 59264
RKF5 : 1.001 sec | nf = 66355
Verner6 : 1.04 sec | nf = 69278
SharpVerner6 : 1.042 sec | nf = 65439
BogakiShampine5 : 1.055 sec | nf = 66031
PapakostasPapaGeorgiou5 : 1.201 sec | nf = 79173
TsitourasPapakostas6 : 1.308 sec | nf = 88556
Papakostas6 : 1.312 sec | nf = 84040
DormandPrince6 : 1.34 sec | nf = 84017
Tsitouras5 : 1.344 sec | nf = 90348
SharpSmart5 : 1.396 sec | nf = 90768
RKF4 : 1.646 sec | nf = 110575
VernerEfficient6 : 1.725 sec | nf = 115599
VernerRobust6 : 1.768 sec | nf = 124605
ClassicVerner6 : 1.781 sec | nf = 119632
Verner9162 : 1.852 sec | nf = 128101
BogakiShampine3 : 4.5 sec | nf = 283458
```

```
 --- Tableau benchmark result for reltol = 1.0e-08 ---
Dverk : 0.842 sec | nf = 54250
CashKarp : 0.883 sec | nf = 57606
DormandLockyerMcCorriganPrince6 : 0.954 sec | nf = 60385
Verner6 : 1.045 sec | nf = 70162
BogakiShampine5 : 1.06 sec | nf = 65742
SharpVerner6 : 1.082 sec | nf = 66161
RKF5 : 1.089 sec | nf = 68227
PapakostasPapaGeorgiou5 : 1.216 sec | nf = 81378
Papakostas6 : 1.271 sec | nf = 84477
TsitourasPapakostas6 : 1.314 sec | nf = 88981
Tsitouras5 : 1.343 sec | nf = 91473
DormandPrince6 : 1.367 sec | nf = 84289
SharpSmart5 : 1.4 sec | nf = 92538
RKF4 : 1.707 sec | nf = 114755
VernerEfficient6 : 1.742 sec | nf = 115960
ClassicVerner6 : 1.774 sec | nf = 119938
VernerRobust6 : 1.786 sec | nf = 124662
Verner9162 : 1.86 sec | nf = 128272
BogakiShampine3 : 5.204 sec | nf = 331950
``` 

From this, I think the default tableau should be either `Dverk`, `CashKarp` or `BogakiShampine5` . What do you think ?

ChrisRackauckas: the tableau methods do not have the same performance and that isn't showing accuracy, just tolerance.

Azercoco: I just benchmarked with higher order tableau

```
 --- Tableau benchmark result for reltol = 1.0e-02 ---
BogakiShampine5 : 0.403 sec | nf = 26846
RKF8 : 0.425 sec | nf = 23358
Dverk : 0.435 sec | nf = 27679
CashKarp : 0.553 sec | nf = 37300
DormandLockyerMcCorriganPrince6 : 0.564 sec | nf = 36217
...

 --- Tableau benchmark result for reltol = 1.0e-04 ---
BogakiShampine5 : 0.681 sec | nf = 45529
RKF8 : 0.699 sec | nf = 42339
CashKarp : 0.742 sec | nf = 50365
VernerRobust7 : 0.776 sec | nf = 51789
Verner6 : 0.899 sec | nf = 59469
...

 --- Tableau benchmark result for reltol = 1.0e-06 ---
RKF8 : 0.697 sec | nf = 41907
Dverk : 0.765 sec | nf = 50646
BogakiShampine5 : 0.77 sec | nf = 51751
DormandLockyerMcCorriganPrince6 : 0.812 sec | nf = 54305
CashKarp : 0.823 sec | nf = 55045
...

 --- Tableau benchmark result for reltol = 1.0e-08 ---
RKF8 : 0.683 sec | nf = 40746
Dverk : 0.799 sec | nf = 51122
DormandLockyerMcCorriganPrince6 : 0.813 sec | nf = 54419
CashKarp : 0.821 sec | nf = 55305
VernerRobust7 : 0.908 sec | nf = 53679
...
```
and `RKF8` seems the best except at low relative tolerance where it is slightly outperformed by `BogakiShampine5`.

Azercoco: > the tableau methods do not have the same performance and that isn't showing accuracy, just tolerance.

I am not sure to understand. How would you benchmark which tableau is the best ?

ChrisRackauckas: Work-precision as Oscar showed. Otherwise it's not showing the true error.

Azercoco: > Work-precision as Oscar showed. Otherwise it's not showing the true error.

The NLSE-like equation I am testing in on has chaotic dynamics. Is it still possible to have an accuracy measurement on it ?

ChrisRackauckas: No, for a chaotic system it's a bit difference because they all have O(1) error, so you might as well always use the lowest order method that is stable (Euler) since you will always get a shadow trajectory anyways.

Azercoco: > No, for a chaotic system it's a bit difference because they all have O(1) error, so you might as well always use the lowest order method that is stable (Euler) since you will always get a shadow trajectory anyways.

It may depends whether the integration time exceed the Lyapunov time I would guess. Currently, my benchmarking strategy is just to reduce the integration time until I have agreement between the solver.

ChrisRackauckas: Then it's not chaotic.

Azercoco: image

Here are the results I got. `CashKarp`, `Verner6` and `BogakiShampine5` seems to be the best except at very high precision where `Verner7`become better.

ChrisRackauckas: Is the system only like 5 times differentiable?

Azercoco: No it is C-inf but higher order tableau were less performant so I did not include them in the final plot.

oscardssmith: The thing that's surprising here is that all of these solvers have roughly the same slope (while we would expect higher order solvers to have a shallower slope since they should have higher order)

Azercoco: image

Comparison with lower order method. The slopes seem all correct. The last plot has possibly a too long integration time compared to Lyapunov time. The only thing strange is that the order 4 and 5 method seems to be have the same slope. But from this, I would suggest having either `CashKarp` (for medium accuracy) or `Verner7` (for high accuracy) as the default tableau.

ChrisRackauckas: It's very odd for Cash Karp to be good, I wonder if it has some odd super convergence on this problem 😅

Azercoco: That's entirely possible that it is a specific of this problem. Also, the error control is made on the transformed equation so the change of coordinate can increase (or decrease) the error. So you would suggest to set Verner6 as default ?

ChrisRackauckas: I think that would be a generally good choice.

Azercoco: Verner6 is now the default tableau.

src/OrdinaryDiffEq.jl
```diff
@@ -208,6 +208,10 @@ export LawsonEuler, NorsettEuler, ETD1, ETDRK2, ETDRK3, ETDRK4, HochOst4, Exp4,
        EPIRK4s3B,
        EPIRK5s3, EXPRB53s3, EPIRK5P1, EPIRK5P2, ETD2, Exprb32, Exprb43
 
+using OrdinaryDiffEqRKIP
+export RKIP
```
```suggestion
```

No need to add this here. In fact, OrdinaryDiffEq will go the other direction and drop deps soon anyways.

lib/OrdinaryDiffEqRKIP/test/type_test.jl
```diff
@@ -0,0 +1,86 @@
...
+
+# @testset "Out-of-place Matrix 1x1 Operator SVector Adaptative" begin
+#     test((μ) -> MatrixOperator([μ;;]), u0 -> SVector(u0), false)
+# end
+
+# # Fails for a strange reason as calling ldiv!(a, B, c) on a MatrixOperator dispatch toward an inexisting method
+
+# @testset "In-Place MatrixOperator Vector Float Adaptative Ldiv" begin
+#     test(μ -> MatrixOperator([μ;;]), u0 -> [u0], true; use_ldiv=true)
+# end
```
ChrisRackauckas: Set as `@test_broken`

lib/OrdinaryDiffEqRKIP/test/runtests.jl
```diff
@@ -0,0 +1,6 @@
+using SafeTestsets
+
+@safetestset "Type Safety Tests" include("type_test.jl")
+@safetestset "Cache Test" include("cache_recycling_test.jl")
+@safetestset "Fourier Semilinear PDE Tests" include("semilinear_pde_test_cpu.jl")
+@safetestset "CUDA Test" include("semilinear_pde_test_cuda.jl")
```
ChrisRackauckas, line 6: This needs to move to the CUDA queue

lib/OrdinaryDiffEqRKIP/src/algorithms.jl
```diff
@@ -0,0 +1,176 @@
...
+
+# Arguments
+- `dtmin::T`: the smallest step `dt` for which `exp(A*dt)` will be cached. Default is `1e-3`
+- `dtmax::T`: the largest step `dt` for which `exp(A*dt)` will be cached. Default is `1.0`
+
+The fixed steps will follow a geometric progression.
+Time stepping can still happen outside the bounds (for the end step for e.g) but no cache will occur (`exp(A*dt)` getting computed each step) degrading the performances.
+The time step can be forcibly clamped within the cache range through the keywords `clamp_lower_dt` and `clamp_higher_dt`.
+
+The cached operator exponentials are also directly stored in the alorithm such that:
```
ChrisRackauckas:
```suggestion
The cached operator exponentials are also directly stored in the algorithm such that:
```

lib/OrdinaryDiffEqRKIP/src/algorithms.jl
```diff
@@ -0,0 +1,176 @@
...
+
+```julia
+rkip = RKIP()
+solve(ode_prob_1, rkip, t1)
+solve(ode_prob_2, rkip, t2)
+````
+
+will reuse the precomputed exponential cached during the first `solve` call.
+This can be useful for solving several times successively problems with a common `A`.
```
ChrisRackauckas, line 54: `t1` and `t2` aren't defined here. And this is interface breaking: it should just use `init` and `reinit!` / `solve!` and it'll do this without the extra work anyways.

lib/OrdinaryDiffEqRKIP/src/algorithms.jl
```diff
@@ -0,0 +1,176 @@
...
+\frac{du_I}{dt} &=  f_I(u_I,p,t) \\
+f_I(u_I,p,t) &= f(exp(-At)u_I, p, t) \\
+\end{aligned}
+```
+and is then solved with an explicit (adaptive) Runge-Kutta method.
+
+This solver is only implemented for semilinear problem: `SplitODEProblem` when the first function `f1` is a `AbstractSciMLOperator` A implementing:
+
+```julia
+LinearAlgebra.exp(A, t) # = exp(A*t)
```
ChrisRackauckas, line 27: This should instead be using ExponentialUtilities.jl algorithm choice for `exponential!`

Azercoco: As the implementation of the exponential is very operator dependent, are you sure that is should not use the `AbstractSciMLOperator` interface instead ? 

cf : https://github.com/SciML/SciMLOperators.jl/blob/7ba386430a229776b41f481ff352eacd7c9f09d4/src/interface.jl#L382C1-L382C60

I checked the code of ExponetialUtilities and it seems to be useful when the exponential matrix output is dense. But in that case, caching the matrix as made here does not make a lot of sense and one would be better to used an ETD method with Krylov based expmv.

lib/OrdinaryDiffEqRKIP/test/runtests.jl
Azercoco: This mean moving the test to OrdinaryDiffEq/test/gpu  ?

ChrisRackauckas: yes

lib/OrdinaryDiffEqRKIP/src/algorithms.jl
ChrisRackauckas: > As the implementation of the exponential is very operator dependent, are you sure that is should not use the AbstractSciMLOperator interface instead ?

ExponentialUtilities.jl will specialize on claimed properties like symmetric.

> I checked the code of ExponetialUtilities and it seems to be useful when the exponential matrix output is dense. But in that case, caching the matrix as made here does not make a lot of sense and one would be better to used an ETD method with Krylov based expmv.

You mean sparse? `expv` is for the sparse case. But `exponential!` is for the dense case, and it will be faster for standard dense matrices.

Azercoco: What I mean is the RKIP method is only efficient if the operator exponential `exp(A*dt)` can be both computed and stored efficiently, like the differential operator from discretized PDE, as they are diagonal in FourierDomain.

While `expv` can be used for exp vector product, the method efficiency comes from _caching the exponentiated operator_  `exp(A*dt)` in matrix-free manner, which is only efficient for low memory operator.

`exponential!` -> efficient dense matrix exponential but not a generic matrix-free operator exponential method
`expv` -> API for `exp(A)` vector product but does not cache the exp-operator
`exp` -> SciMLOperators API for computing matrix-free exponential of operators and storing them, which is what the RKIP does.

Now, it would be possible to disable the caching of  `exp(A)` and use `expv`will a fallback on ExponentialUtilities for dense matrix but in that case, exponential integrators like ETDRK4 would generally be more suited.

ChrisRackauckas:
@oscardssmith you were reviewing this?

src/OrdinaryDiffEq.jl
```diff
@@ -208,6 +208,8 @@ export LawsonEuler, NorsettEuler, ETD1, ETDRK2, ETDRK3, ETDRK4, HochOst4, Exp4,
        EPIRK4s3B,
        EPIRK5s3, EXPRB53s3, EPIRK5P1, EPIRK5P2, ETD2, Exprb32, Exprb43
 
+
```
oscardssmith: remove random newlines

ChrisRackauckas:
It was missing a change to the CI file so that the new tests actually run. If those pass then we merge.

lib/OrdinaryDiffEqRKIP/Project.toml
```diff
@@ -0,0 +1,44 @@
...
+[compat]
+DiffEqBase = "6.175"
+DiffEqDevTools = "2.48"
+MaybeInplace = "0.1.4"
+OrdinaryDiffEqCore = "1.26"
+SciMLBase = "2.99"
+SciMLOperators = "1.3.1"
+StaticArrays = "1.9.13"
+UnPack = "1.0.2"
+LinearAlgebra = "1.11"
```
ChrisRackauckas:
```suggestion
LinearAlgebra = "1.10"
```

lib/OrdinaryDiffEqRKIP/Project.toml
```diff
@@ -0,0 +1,44 @@
...
+CUDA = "5.5.2"
+FFTW = "1.8.0"
+SafeTestsets = "0.1.0"
+
+julia = "1.11"
```
ChrisRackauckas:
```suggestion
julia = "1.10"
```

src/OrdinaryDiffEq.jl
```diff
@@ -208,6 +208,8 @@ export LawsonEuler, NorsettEuler, ETD1, ETDRK2, ETDRK3, ETDRK4, HochOst4, Exp4,
...
+
```
```suggestion
```

lib/OrdinaryDiffEqRKIP/Project.toml
```diff
@@ -0,0 +1,43 @@
...
+SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
+SciMLOperators = "c0aeaf25-5076-4817-a8d5-81caf7dfa961"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
+
+[compat]
+DiffEqBase = "6.175"
+DiffEqDevTools = "2.48"
+MaybeInplace = "0.1.4"
+OrdinaryDiffEqCore = "1.26"
```
ChrisRackauckas:
```suggestion
OrdinaryDiffEqCore = "2"
```

lib/OrdinaryDiffEqRKIP/src/OrdinaryDiffEqRKIP.jl
```diff
@@ -0,0 +1,26 @@
+module OrdinaryDiffEqRKIP
+
+using LinearAlgebra: ldiv!, exp, axpy!, norm, mul!
+using SciMLOperators: AbstractSciMLOperator
+using UnPack: @pack!, @unpack
```
devmotion:
```suggestion
```

lib/OrdinaryDiffEqRKIP/Project.toml
```diff
@@ -0,0 +1,43 @@
...
+[deps]
+DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
+DiffEqDevTools = "f3b72e0c-5b89-59e1-b016-84e28bfd966d"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MaybeInplace = "bb5d69b7-63fc-4a16-80bd-7e42200c7bdb"
+OrdinaryDiffEqCore = "bbf590c4-e513-4bbe-9b18-05decba2e5d8"
+SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
+SciMLOperators = "c0aeaf25-5076-4817-a8d5-81caf7dfa961"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
```
devmotion:
```suggestion
```

lib/OrdinaryDiffEqRKIP/Project.toml
```diff
@@ -0,0 +1,43 @@
...
+
+[compat]
+DiffEqBase = "6.175"
+DiffEqDevTools = "2.48"
+MaybeInplace = "0.1.4"
+OrdinaryDiffEqCore = "2"
+SciMLBase = "2.99"
+SciMLOperators = "1.3.1"
+StaticArrays = "1.9.13"
+UnPack = "1.0.2"
```
devmotion:
```suggestion
```

ChrisRackauckas:
Finished and merged in https://github.com/SciML/OrdinaryDiffEq.jl/pull/2938

PR #2743, opened 2025-06-24
Reset fsal when necessary for first tstep

oscardssmith:
fixes https://github.com/SciML/DifferentialEquations.jl/issues/1092


ChrisRackauckas (approved):

ChrisRackauckas:
Actually this is missing a test. Can you add that?

PR #2744, opened 2025-06-24
add tests for https://github.com/SciML/OrdinaryDiffEq.jl/pull/2743

oscardssmith:


ChrisRackauckas:
fails 😅

oscardssmith:
this makes very little sense. It's working locally... Is CI not picking up the dev version somehow?

oscardssmith:
Oh, It looks like it is only failing on v1.10, so it likely is a CI issue. I'll investigate.

PR #2753, opened 2025-07-15
fix stats of stabilized RK methods

ranocha:
## Checklist

## Additional context

I was wondering why an algorithm became so much slower when supplying an `eigen_est` although both the number of steps and the number of function evaluations reported in the stats reduced significantly 🙈

### DIFF
--- a/lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
+++ b/lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
@@ -313,7 +313,6 @@ end
     @.. broadcast=false uᵢ₋₁=uprev + (dt * recf[ccache.start]) * fsalfirst
     ccache.mdeg < 2 && (@.. broadcast=false u=uᵢ₋₁)
     # for the second to the ccache.mdeg th stages
-    @show ccache.mdeg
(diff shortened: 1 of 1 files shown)


lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -310,10 +313,12 @@ end
     @.. broadcast=false uᵢ₋₁=uprev + (dt * recf[ccache.start]) * fsalfirst
     ccache.mdeg < 2 && (@.. broadcast=false u=uᵢ₋₁)
     # for the second to the ccache.mdeg th stages
+    @show ccache.mdeg
```
ChrisRackauckas:
```suggestion
```

ChrisRackauckas (approved):

ChrisRackauckas:
> I was wondering why an algorithm became so much slower when supplying an eigen_est although both the number of steps and the number of function evaluations reported in the stats reduced significantly 🙈

I've seen this before and started looking into the adaptivity algorithm and thought it must have possible improvements but couldn't figure out why it was behaving like that. This explains a lot 😅

ranocha:
Thanks for your review - and for catching the debug statement I forgot.

PR #2755, opened 2025-07-16
fix allocations in ROCK4

ranocha:
## Checklist

## Additional context

I observed some weird allocations when using `ROCK4()`. After some time, I found the culprit (missing boradcast statements). Since these were not so obvious, I improved the readability of the code to help figure this out.

I already bumped the version in the Project.toml so that you can immediately release a new version after merging this.

### DIFF
--- a/lib/OrdinaryDiffEqStabilizedRK/Project.toml
+++ b/lib/OrdinaryDiffEqStabilizedRK/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqStabilizedRK"
 uuid = "358294b1-0aab-51c3-aafe-ad5ab194a2ad"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "1.1.1"
+version = "1.1.2"
 
 [sources]
 OrdinaryDiffEqCore = {path = "../OrdinaryDiffEqCore"}


lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -385,14 +405,16 @@ end
     f(k, uᵢ₋₃, p, tᵢ₋₂)
     OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
     @.. broadcast=false u+=B₄ * k
-    integrator.opts.adaptive && (tmp += B̂₄ * k)
+    if integrator.opts.adaptive
+        @.. broadcast=false tmp += B̂₄ * k
+    end
```
ranocha, line 410: broadcasting was missing here

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -385,14 +405,16 @@ end
...
 
     f(k, u, p, t + dt)
     OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
 
     #Error estimate (embedded method of order 3)
     if integrator.opts.adaptive
-        tmp += B̂₅ * k
+        @.. broadcast=false tmp += B̂₅ * k
```
ranocha, line 417: and here

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
oscardssmith: note that the `broadcast=false` can (and probably should) be removed. It's been the default for a while.

ranocha: Shall I remove it everywhere in these methods?

PR #2759, opened 2025-07-18
Fixed adaptive out-of-place Stepanov5.

toma-vlad:
Hello, 

This is my first PR. Hopefully, I am doing this correctly. 

While benchmarking solvers, I noticed `Stepanov5()` had an error for out-of-place ODEProblems. The following MRE should result in error for at least versions `v6.99.0` `v6.98.0` of `OrdinaryDiffEq.jl`.  

```julia 
using OrdinaryDiffEq #using OrdinaryDiffEqLowOrderRK

function lorenz(u, p, t)
	dx = 10.0 * (u[2] - u[1])
        dy = u[1] * (28.0 - u[3]) - u[2]
        dz = u[1] * u[2] - (8 / 3) * u[3]
           
	[dx, dy, dz]
end

u0 = [1.0; 0.0; 0.0]
tspan = (0.0, 100.0)
prob = ODEProblem(lorenz, u0, tspan)

solve(prob, Stepanov5())
```

The issue was in the implementation of `perform_step!(integrator, cache::Stepanov5ConstantCache, repeat_step = false)`, probably a transcription error, which I've addressed in this PR.

It appears that the issue has passed through testing because the tests for `OrdinaryDiffEqLowOrderRK.jl` do not cover adaptive time stepping! 

I did not add new tests because this appears to be a generalized issue in all submodules of OrdinaryDifferentialEq.jl and should probably be addressed higher up. 

## Checklist

  [COLPRAC](https://github.com/SciML/COLPRAC). - Hopefully? :)

### DIFF
--- a/lib/OrdinaryDiffEqLowOrderRK/src/low_order_rk_perform_step.jl
+++ b/lib/OrdinaryDiffEqLowOrderRK/src/low_order_rk_perform_step.jl
@@ -1553,11 +1553,11 @@ function perform_step!(integrator, cache::Stepanov5ConstantCache, repeat_step = 
     OrdinaryDiffEqCore.increment_nf!(integrator.stats, 6)
 
     if integrator.opts.adaptive
-        @.. broadcast=false utilde=dt * (btilde1 * k1 + btilde2 * k2 +
+        utilde=dt * (btilde1 * k1 + btilde2 * k2 +
                                     btilde3 * k3 + btilde4 * k4 +
                                     btilde5 * k5 + btilde6 * k6 +
                                     btilde7 * k7)
-        calculate_residuals!(atmp, utilde, uprev, u, integrator.opts.abstol,
+        atmp = calculate_residuals(utilde, uprev, u, integrator.opts.abstol,
             integrator.opts.reltol, integrator.opts.internalnorm, t)
         integrator.EEst = integrator.opts.internalnorm(atmp, t)
     end


oscardssmith (approved):

oscardssmith:
thanks! We really need better tests for this sort of thing. I'll think about the right approach, but this PR is good to go as is.

ChrisRackauckas:
There's no analytical solution on what adaptive time steps are supposed to be 😅 , so most algorithms have a small little thing right after the convergence test that's just a regression test, just a `solve` that's adaptive and then capping `length(sol.t) < reasonable value` after hand-testing that the adaptivity makes sense. I haven't thought of a better way of testing the adaptivty metrics but I'd take recommendations if you've got one. You could in theory use `init` and then test `integrator.EEst` is sufficiently small when dt is small, but then with PI adaptivity you have the delay term etc. so it gets complicated to turn those into unit tests, hence the high level integration regression testing mostly covers it. But it's definitely not a fantastic way to cover that.

oscardssmith:
Couldn't we do our convergence tests with adaptivity on, and test that as we decrease `abstol`/`reltol` that our `f` calls scale as expected?

toma-vlad:
In regards to testing adaptive stepping, I think it's a two-fold issue. Currently, the adaptive stepping functions aren't explicitly, called by the `test_convergence` functions, if I understand correctly how `test_convergence` works.

So, firstly, you could just call the solve, for a problem which is known to converge, and then if the solve returns `retcode: Success`, you know that the solver is working and you haven't broken anything. 

Secondly, on the problem of convergence, I don't necessarily feel qualified to give suggestions, but certainly, running the solver against a `hard` problem with an analytic solution, such as the [figure-8 periodic motion](https://github.com/SebastianM-C/ElectronDynamicsModels.jl/blob/main/poster/poster.pdf), could give a you a baseline idea of whether your time steps are small enough? And if the steps get too small, the error is going to increase again?

ChrisRackauckas:
> Currently, the adaptive stepping functions aren't explicitly, called by the test_convergence functions, if I understand correctly how test_convergence works.

That is correct. test convergence requires fixed dt.

> Secondly, on the problem of convergence, I don't necessarily feel qualified to give suggestions, but certainly, running the solver against a hard problem with an analytic solution, such as the [figure-8 periodic motion](https://github.com/SebastianM-C/ElectronDynamicsModels.jl/blob/main/poster/poster.pdf), could give a you a baseline idea of whether your time steps are small enough? And if the steps get too small, the error is going to increase again?

That's what I mentioned we currently do, for example that's what the length check line is doing in the convergence tests like in https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/OrdinaryDiffEqRosenbrock/test/ode_rosenbrock_tests.jl#L51 . They should all have it, so if there's a convergence test set that is missing the adaptivity regression test then we should fix that up.

> So, firstly, you could just call the solve, for a problem which is known to converge, and then if the solve returns retcode: Success, you know that the solver is working and you haven't broken anything.

I think we just need to add a `@test SciMLBase.successful_retcode(sol)` right after each length test. That's probably a quick Claude PR.

ChrisRackauckas:
We also need to split out the Enzyme parts from pre 😅

ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/7393265211cb3c891791ca1b09524e592ff19a52/test/regression/ode_adaptive_tests.jl#L17 Stepanov5 actually had a test for this, but only in-place 😅. We should probably just make that also have an out of place version.

toma-vlad:
> https://github.com/SciML/OrdinaryDiffEq.jl/blob/7393265211cb3c891791ca1b09524e592ff19a52/test/regression/ode_adaptive_tests.jl#L17
> Stepanov5 actually had a test for this, but only in-place 😅. We should probably just make that also have an out of place version.

I'm sorry, this project is very large and I didn't realize those tests were in a different place (not in `OrdinaryDiffEqLowOrderRK`). But yes, this was my suggestion; seems the tests just need to be extended a bit.

ChrisRackauckas:
That's fine. It happens. There is still a bit of a mess because it was only a few months ago that we split the whole package into a bunch of subpackages, and some tests like this it's necessarily clear if it would go into the subpackage or the whole package 😅

PR #2762, opened 2025-07-19
Make DI unstrict due to not having view support

ChrisRackauckas:
Fixes Integrators test on Buildkite

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
@@ -267,12 +267,12 @@ function build_jac_config(alg, f::F1, uf::F2, du1, uprev,
                 autodiff_alg_reverse = dir_reverse
             end
 
-            jac_config_forward = DI.prepare_jacobian(uf, du1, autodiff_alg_forward, u)
-            jac_config_reverse = DI.prepare_jacobian(uf, du1, autodiff_alg_reverse, u)
+            jac_config_forward = DI.prepare_jacobian(uf, du1, autodiff_alg_forward, u, strict = Val(false))
+            jac_config_reverse = DI.prepare_jacobian(uf, du1, autodiff_alg_reverse, u, strict = Val(false))
 
             jac_config = (jac_config_forward, jac_config_reverse)
         else
-            jac_config1 = DI.prepare_jacobian(uf, du1, autodiff_alg, u)
+            jac_config1 = DI.prepare_jacobian(uf, du1, autodiff_alg, u, strict = Val(false))
             jac_config = (jac_config1, jac_config1)
         end
 


oscardssmith:
@gdalle anything DI could do to make strict less painful for this kind of thing? it feels like we shouldn't have to care about the array type of the length and eltype match

gdalle:
> it feels like we shouldn't have to care about the array type if the length and eltype match

For some backends, that's true. ForwardDiff and Zygote most likely won't yell at you if the types are a mismatch. On the other hand, Enzyme and Mooncake are much more picky about the type of arrays you can pass. See the MWE below:

```julia
using DifferentiationInterface
import Enzyme, ForwardDiff, Mooncake, Zygote

function test_mismatched_prep(backend; strict=Val(false))
    x0 = ones(2)
    x = view(ones(2), 1:2)
    prep = prepare_jacobian(identity, backend, x0; strict)
    return jacobian(identity, prep, backend, x)
end
```

```julia
julia> test_mismatched_prep(AutoForwardDiff())
2×2 Matrix{Float64}:
 1.0  0.0
 0.0  1.0

julia> test_mismatched_prep(AutoZygote())
2×2 Matrix{Float64}:
 1.0  0.0
 0.0  1.0

julia> test_mismatched_prep(AutoEnzyme())
ERROR: MethodError: no method matching EnzymeCore.BatchDuplicated(::SubArray{Float64, 1, Vector{…}, Tuple{…}, true}, ::Tuple{Vector{…}, Vector{…}})
The type `EnzymeCore.BatchDuplicated` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  EnzymeCore.BatchDuplicated(::T1, ::NTuple{N, T1}) where {T1, N}
   @ EnzymeCore ~/.julia/packages/EnzymeCore/lmG5F/src/EnzymeCore.jl:135
  EnzymeCore.BatchDuplicated(::T1, ::NTuple{N, T1}, ::Bool) where {T1, N}
   @ EnzymeCore ~/.julia/packages/EnzymeCore/lmG5F/src/EnzymeCore.jl:135

Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/Enzyme/9dy6G/src/sugar.jl:717 [inlined]
...

julia> test_mismatched_prep(AutoMooncake())
ERROR: ArgumentError: signature of arguments, Tuple{Mooncake.CoDual{typeof(identity), Mooncake.NoFData}, Mooncake.CoDual{SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Vector{Float64}}}, not equal to signature required by rule, Tuple{Mooncake.CoDual{typeof(identity), Mooncake.NoFData}, Mooncake.CoDual{Vector{Float64}, Vector{Float64}}}.
Stacktrace:
  [1] __verify_sig(rule::Mooncake.DerivedRule{…}, fx::Tuple{…})
    @ Mooncake ~/.julia/packages/Mooncake/qKNoG/src/interface.jl:31
...
```

In this case, Enzyme's Jacobian preparation pre-allocates the basis vectors `dx0` with the same type as `x0`, and then complains when we construct something like `Duplicated(x, dx0)`. Meanwhile, Mooncake builds rules for a specific signature, which error when arguments change type.

For this reason, preparation can only be correct and error-free if you pass the exact same types at execution time as you did at preparation time.
DI v0.6 did specify that in the documentation but without enforcing it. DI v0.7 introduces the strict setting by default, so that users get more informative errors:

```julia
julia> test_mismatched_prep(AutoEnzyme(); strict=Val(true))
ERROR: PreparationMismatchError (inconsistent types between preparation and execution):
  - f: ✅
  - backend: ✅
  - x: ❌
    - prep: Vector{Float64}
    - exec: SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}
  - contexts: ✅
If you are confident that this check is superfluous, you can disable it by running preparation with the keyword argument `strict=Val(false)` inside DifferentiationInterface.
```

gdalle:
Here's a pure-Enzyme MWE, to show that it's not just DI being picky for the sake of fun ^^

```julia
julia> using Enzyme

julia> x0 = ones(2);

julia> x = view(ones(2), 1:2);

julia> shadows = Enzyme.onehot(x0);  # Enzyme's equivalent for preparation

julia> Enzyme.jacobian(Enzyme.Forward, identity, x; shadows)
ERROR: MethodError: no method matching BatchDuplicated(::SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, ::Vector{Float64})
The type `BatchDuplicated` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  BatchDuplicated(::T1, ::NTuple{N, T1}) where {T1, N}
   @ EnzymeCore ~/.julia/packages/EnzymeCore/lmG5F/src/EnzymeCore.jl:135
  BatchDuplicated(::T1, ::NTuple{N, T1}, ::Bool) where {T1, N}
   @ EnzymeCore ~/.julia/packages/EnzymeCore/lmG5F/src/EnzymeCore.jl:135
```

ChrisRackauckas:
Yeah we can get to that after JuliaCon, for now I just want to get the odd but reasonable edge cases working again.

PR #2774, opened 2025-07-25
New method Rodas6P

gstein3m:
## Checklist

## Additional context

The new Rodas6P method was designed similarly to Rodas5P. It has order 6 for ODEs and index-1 DAEs with integrated order 5 interpolation. 
232 order conditions must be satisfied, plus 175 for the interpolation. 
Therefore, the number of stages is s=19, with the last 3 stages only required for the interpolation. 
For large systems of stiff ODEs or DAEs with high accuracy requirements, the efficiency may be better than Rodas5P.


lib/OrdinaryDiffEqRosenbrock/src/algorithms.jl
```diff
@@ -85,6 +85,9 @@ University of Geneva, Switzerland.
  Preprint 2024
  https://github.com/hbrs-cse/RosenbrockMethods/blob/main/paper/JuliaPaper.pdf
 
+ #### Rodas56
+- Steinebach G.   Construction of Rosenbrock–Wanner method Rodas6P , to prepare
```
ranocha:
```suggestion
 #### Rodas6P
- Steinebach G.   Construction of Rosenbrock–Wanner method Rodas6P , to prepare
```

?

lib/OrdinaryDiffEqRosenbrock/src/algorithms.jl
```diff
@@ -138,6 +142,11 @@ Rosenbrock23
     "Rodas5P", with_step_limiter = true)
 Rodas5P
 
+@doc rosenbrock_docstring(
+    "Efficient for medium and high tolerance stiff problems. A 6th order A-stable and stiffly stable embedded Rosenbrock method for differential-algebraic problems.",
```
ranocha:
```suggestion
    "Efficient for medium and strict tolerance stiff problems. A 6th order A-stable and stiffly stable embedded Rosenbrock method for differential-algebraic problems.",
```

Strictly speaking, `1.0e5` is a high tolerance, but I guess you refer to small tolerances like `1.0e-10`. Since this is always confusing, it may be better to speak about strict vs. loose tolerances.

ChrisRackauckas:
I thought there was a 5th order limit barrier, is that only with A-stability?

ChrisRackauckas:
Looks right and passing tests, so merging.

gstein3m:
This A-stability barrier is only for multistep methods. Rodas6P is A-stable.
Thank you for merging.

ChrisRackauckas:
I thought there was a thing where any SDIRK method has a 5th order barrier where it's only possible to be L-stable and have a single gamma if it's 5th order? That was something in Hairer II IIRC, and I thought that carried over to Rosenbrocks as well. 

BTW, one interesting set of methods would be Rosenbrock methods that have multiple different gammas, since for the size Rosenbrocks tend to do well, what we could do is multithread across the factorizations like what is done in https://ieeexplore.ieee.org/abstract/document/9926357/ (https://arxiv.org/abs/2207.08135) so the cost difference would be minimal but it should be able to achieve higher order. I thought that was actually required to get higher than 5 😅 , but it's still on the table. So this direction of Non-S DIRK methods and similarly for Rosenbrock connects really well with parallelization, since you can simply factorize all of the Wi at the same time, and if you're below where BLAS multithreads well (around 200x200 matrices) you can get a win where factorizing multiple Wi is almost as cheap as factorizing a single one. And in that regime, if you can construct a better method by letting gamma not be a single constant, then that could be a win.

PR #2790, opened 2025-07-31
Update hard_dae.jl to not use Rosenbrock23

ChrisRackauckas:
It's not an appropriate solver because it's not stiffly accurate in theory, which means in practice it will have accuracy issues on DAEs that are somewhat random. So this being in the test can work but doesn't match theoretical guarantees and promises more than is allowed.

### DIFF
--- a/test/regression/hard_dae.jl
+++ b/test/regression/hard_dae.jl
@@ -270,6 +270,6 @@ for prob in [prob1, prob2], alg in [simple_implicit_euler, alg_switch]
     sol = solve(prob, alg, callback = cb, dt = 1 / 2^10, adaptive = false)
     @test sol.retcode == ReturnCode.Success
     @test sol(0, idxs = 1) == 5
-    @test abs(sol(2-2^-10, idxs = 1)) <= 1e-10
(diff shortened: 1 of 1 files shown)


test/regression/hard_dae.jl
```diff
@@ -270,6 +270,6 @@ for prob in [prob1, prob2], alg in [simple_implicit_euler, alg_switch]
     sol = solve(prob, alg, callback = cb, dt = 1 / 2^10, adaptive = false)
     @test sol.retcode == ReturnCode.Success
     @test sol(0, idxs = 1) == 5
-    @test abs(sol(2-2^-10, idxs = 1)) <= 1e-10
+    @test abs(sol(2-2^-10, idxs = 1)) <= 1e-4
```
oscardssmith, line 273: why?

ChrisRackauckas: This is just for non-adaptive Euler. I found a small bug in that while doing nlstep that non-adaptive would also go to maxiters 10 rather than exiting based on convergence. I think that leads to this. The test is a little silly too since it's not the algebraic condition so it's not really testing much...

PR #2804, opened 2025-08-03
Fix precompilation failure by replicating Pkg.test() environment

ChrisRackauckas-Claude:
## Summary

Fixes the failing precompilation step in PR #2801 by properly replicating the test environment that `Pkg.test()` creates.

## Problem

The precompilation CI workflow was failing because `@compile_workload` blocks in various OrdinaryDiffEq subpackages require test dependencies like `SafeTestsets`, but the CI was only installing main project dependencies.

## Solution

This PR implements a proper fix that replicates what `Pkg.test()` does:

1. **Parse Project.toml** to extract test dependencies from `[extras]` and `[targets]` sections
2. **Add all test dependencies** to the current environment before precompilation  
3. **Run precompilation** with access to the complete test environment

## Technical Details

- Reads the same `[extras]` and `[targets]` configuration that `Pkg.test()` uses
- Installs each of the 29 test dependencies with correct UUID mapping
- Provides access to all test dependencies during precompilation
- No changes to main Project.toml (dependencies added only temporarily during CI)

## Test Plan

- [x] Tested locally with full environment setup
- [x] Confirmed precompilation completes successfully
- [x] Verified all 29 test dependencies are properly installed
- [x] No permanent changes to project dependencies


.github/workflows/CI_with_precompile.yml
```diff
@@ -0,0 +1,269 @@
...
+                    println("Warning: $dep not found in extras")
+                end
+            end
+            
+            println("Test environment setup complete")
+          '
+      - name: Run precompilation workload  
+        run: |
+          echo "Running precompilation workload with complete test environment..."
+          julia --project --check-bounds=auto --compiled-modules=yes --depwarn=yes -e '
```
ChrisRackauckas:
```suggestion
          julia --project --check-bounds=yes --compiled-modules=yes --depwarn=yes -e '
```

PR #2809, opened 2025-08-05
Add tests for event detection when step is exactly on event date

dcourteville:
Tests for DiffEqBase PR https://github.com/SciML/DiffEqBase.jl/pull/1185 that fix https://github.com/SciML/DiffEqBase.jl/issues/1184


ChrisRackauckas:
This test should also test values? Or that the integration succeeded? `@test SciMLBase.successful_retcode(sol)`?

ChrisRackauckas:
Also bump the minimum version in OrdinaryDiffEqCore for DiffEqBase so the test will always need to pass.

ChrisRackauckas (approved):

dcourteville:
Done. I added a test to check that the callback was properly applied.
The test fail for the original solution were we don't force the integrator to step on the event date (I did not commit the failing test, so there is no error in the PR) . It seems that `sol(-2.95, continuity = :left)` return a wrong value, even though the correct left bond is stored in the solution. I guess I will open another issue for this.

```julia
sol = solve(prob, Tsit5(), callback = callback)
@test sol(-2.95, continuity = :right) ≈ sol(-2.95, continuity = :left) + 2 # Fail

sol = solve(prob, Tsit5(), callback = callback, tstops = [-2.95])
@test sol(-2.95, continuity = :right) ≈ sol(-2.95, continuity = :left) + 2 # Pass
```

PR #2835, opened 2025-08-11
restore ForwardDiff@1.0 compat

oscardssmith:
@ChrisRackauckas it appears you vibe-coded this into breaking.


ChrisRackauckas:
No I pulled it back, why would v1 be required?

oscardssmith:
we generally don't force users to use old packages. Why are we doing so here? Is ForwardDiff@1 causing problems?

ChrisRackauckas:
It was having resolving issues in some cases.

oscardssmith:
removing options for packages won't help that...

PR #2836, opened 2025-08-11
improve type stabitity of `jacobian!`

oscardssmith:
also bump patch versions for `ForwardDiff@1.0` re-add

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
@@ -136,7 +136,7 @@ function jacobian(f, x::AbstractArray{<:Number}, integrator)
 end
 
 # fallback for scalar x, is needed for calc_J to work
-function jacobian(f, x, integrator)
+function jacobian(f::UJacobianWrapper, x, integrator)
     alg = unwrap_alg(integrator, true)
 
     dense = ADTypes.dense_ad(alg_autodiff(alg))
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
You forgot the code changes

oscardssmith:
oops... fixed.

lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
```diff
@@ -183,7 +183,7 @@ function jacobian(f, x, integrator)
     return jac
 end
 
-function jacobian!(J::AbstractMatrix{<:Number}, f, x::AbstractArray{<:Number},
+function jacobian!(J::AbstractMatrix{<:Number}, f::UJacobianWrapper, x::AbstractArray{<:Number},
```
ChrisRackauckas, line 186: why would this fix it?

oscardssmith: `UJacobianWrapper<:Function` so leaving this unspecified was causing us to hit the normal "Julia doesn't specialize `Function`s by default".

ChrisRackauckas: what about removing the `<: Function` for safety?

oscardssmith: I think it should be fine as is. Otherwise, the change would have to go in `SciMLBase`

ChrisRackauckas: I think we should just do that change too. Likely would improve a few other spots

PR #2838, opened 2025-08-12
make initial dt nothing

oscardssmith:
allows for trimming of rosenbrock when initial_dt is set. Ideally, we would make init_dt selection fully trimmable also, but this is a start.

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/initdt.jl
+++ b/lib/OrdinaryDiffEqCore/src/initdt.jl
@@ -103,8 +103,8 @@
     still works for matrix-free definitions of the mass matrix.
     =#
 
-    if prob.f.mass_matrix != I && (!(prob.f isa DynamicalODEFunction) ||
-        any(mm != I for mm in prob.f.mass_matrix))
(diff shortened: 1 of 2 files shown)


lib/OrdinaryDiffEqCore/src/initdt.jl
```diff
@@ -103,8 +103,8 @@
     still works for matrix-free definitions of the mass matrix.
     =#
 
-    if prob.f.mass_matrix != I && (!(prob.f isa DynamicalODEFunction) ||
-        any(mm != I for mm in prob.f.mass_matrix))
+    if prob.f.mass_matrix !== I && (!(prob.f isa DynamicalODEFunction) ||
+        any(mm !== I for mm in prob.f.mass_matrix))
```
oscardssmith, line 107: todo: undo these

ChrisRackauckas:
Looks like convergence testing sets dt=0.0, along with some other things , so we need to still support that path.

ChrisRackauckas-Claude:
Superseded by #3016 which rebases this onto current master with additional fixes.

PR #2853, opened 2025-08-18
Move SparseArrays to extension for OrdinaryDiffEqDifferentiation

ChrisRackauckas-Claude:
## Summary
- Refactors OrdinaryDiffEqDifferentiation to move SparseArrays from a direct dependency to a package extension
- Reduces the dependency footprint for users who don't need sparse array functionality
- Maintains full backward compatibility

## Changes
- Moved SparseArrays from `deps` to `weakdeps` in Project.toml
- Created `OrdinaryDiffEqDifferentiationSparseArraysExt` extension module
- Refactored sparse array type usage to work with extension pattern
- Added helper functions for sparse matrix operations (`get_nzval`, `set_all_nzval!`)
- Added SparseArrays to test targets to ensure extension is tested

## Benefits
This reduces the dependency load for OrdinaryDiffEqRosenbrock and other packages that depend on OrdinaryDiffEqDifferentiation but don't necessarily need sparse arrays. Users who don't use sparse matrices won't need to load SparseArrays and its dependencies.

## Testing
- Extension loads correctly when SparseArrays is imported
- All existing functionality is preserved
- Tests pass with and without SparseArrays loaded

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/ext/OrdinaryDiffEqDifferentiationSparseArraysExt.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/ext/OrdinaryDiffEqDifferentiationSparseArraysExt.jl
@@ -2,26 +2,18 @@ module OrdinaryDiffEqDifferentiationSparseArraysExt
 
 using OrdinaryDiffEqDifferentiation
 import SparseArrays
-import SparseArrays: nonzeros, spzeros
+import SparseArrays: nonzeros, spzeros, SparseMatrixCSC, AbstractSparseMatrix
 
-# Set the type aliases when extension loads
-function __init__()
-    OrdinaryDiffEqDifferentiation.SparseMatrixCSC = SparseArrays.SparseMatrixCSC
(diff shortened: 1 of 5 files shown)


ChrisRackauckas-Claude:
Updated the PR based on feedback to use function-based dispatch instead of type checking.

## Changes in latest commit:
- Replaced all `isa SparseMatrixCSC` and `isa AbstractSparseMatrix` checks with function calls
- Added `is_sparse()` and `is_sparse_csc()` functions that default to `false` for any input
- Extension overloads these to return `true` for appropriate sparse types
- Removed unnecessary type aliases

This approach ensures the sparse code completely compiles out when SparseArrays is not loaded, providing better performance for non-sparse cases.

The functions work as expected:
```julia
# Without SparseArrays
julia> A = [1 2; 3 4]
julia> is_sparse(A)  # returns false
julia> is_sparse_csc(A)  # returns false

# With SparseArrays loaded
julia> using SparseArrays
julia> A = spzeros(3,3)
julia> is_sparse_csc(A)  # returns true
julia> is_sparse(A)  # returns true
```

PR #2870, opened 2025-09-08
Make ExplicitImports a test dependency

devmotion:
I was surprised to see ExplicitImports as an indirect dependency in my environment, and noticed that it's a direct dependency of OrdinaryDiffEq. IMO it should only be a test dependency.


ChrisRackauckas:
oh oops, thanks.

Issue #2586, opened 2025-02-04
10x - 100x speed up when using own implementation of Implicit Euler or Implicit Midpoint on linear test problem


cwittens:
**Describe the bug 🐞**

When using my own implementation of Implicit Euler or Implicit Midpoint (definition below), I get a 10x - 100x speed up on the linear test problem ``(du, u, p, t) -> du .= p .* u`` when using ``FunctionOperator`` in the ODEProblem definition. (and also when solving Parabolic PDEs as discovered by @ranocha , but this is not shown in this MRE).

**Minimal Reproducible Example 👇**

```julia
using Pkg
Pkg.activate(".")
Pkg.add(["OrdinaryDiffEq", "SciMLOperators", "LinearSolve",  "BenchmarkTools", "LinearAlgebra"])
using OrdinaryDiffEq, SciMLOperators, LinearSolve, BenchmarkTools, LinearAlgebra

# function definition
function my_implicit_euler(ode, dt, solver=NaN, abstol=NaN, reltol=NaN)
	u = copy(ode.u0)
	prob = LinearProblem(I - dt * ode.f.f, vec(zero(u)), abstol = abstol, reltol = reltol)
	linsolve = init(prob, solver)
	t = ode.tspan[begin]
	while t < ode.tspan[end]
		t += dt
		copyto!(linsolve.b, vec(u))
		solve!(linsolve)
		copyto!(vec(u), linsolve.u)
	end

	return (; u = [copy(ode.u0), u], prob = ode)
end

# setup the ODE, once using FunctionOperator and once without
linear! = (du, u, p, t) -> du .= p .* u

function setup_ode_Operator(N)
    tspan = (0.0, 5.0)
    u0 = ones(N)
    parameters = 1.01
    linear_operator = FunctionOperator(linear!, u0; p = parameters) 

    ode = ODEProblem(linear_operator, u0, tspan, parameters)
end

function setup_ode(N)
    tspan = (0.0, 5.0)
    u0 = ones(N)
    parameters = 1.01

    ode = ODEProblem(linear!, u0, tspan, parameters)
end

dt = 0.001
prob_op = setup_ode_Operator(100);
prob = setup_ode(100);

# my_implicit_euler roughly 100x faster then ImplicitEuler()
@btime sol1 = my_implicit_euler(prob_op, dt); 
@btime sol2 = solve(prob_op, ImplicitEuler(), dt=dt, adaptive = false, save_everystep = false);
[...]

Issue #914, opened 2019-09-01
A bunch of MWEs for specializing on linearity in stiff solvers


ChrisRackauckas:
```julia
using SparseArrays
tend = 1.0
K = sprand(100,100,0.1)
C = sprand(100,100,0.1)
T0 = rand(100)
f(dT, T, p, t) = begin dT .= -((K * T)); end
jac(J,T,p,t) = (J .= -K)
tspan = (0.0, tend)

ff = ODEFunction(f, jac=jac, mass_matrix=C, jac_prototype=rand()*C-K)
prob = ODEProblem(ff, T0, tspan)
sol = solve(prob, Rosenbrock23())

A = DiffEqArrayOperator(-K)
ff = ODEFunction(A, jac=jac, mass_matrix=C, jac_prototype=rand()*C-K)
prob = ODEProblem(ff, T0, tspan)
sol = solve(prob, Rosenbrock23())

ff = ODEFunction(f, jac=jac, jac_prototype=K)
prob = ODEProblem(ff, T0, tspan)
sol = solve(prob, Rosenbrock23())
sol = solve(prob, KenCarp4())
```

ChrisRackauckas:
From @PetrKryslUCSD. I think that a lot of the nonlinear solver changes must've messed with these kind of examples, so we need to add some tests here. A lot of the "linear" path is assuming `WOperator`, while here it should all be an array.

ChrisRackauckas:
Note to self: `iip_generate_W` etc. are in a nonsensical location.

ChrisRackauckas:
```julia
using SparseArrays, OrdinaryDiffEq
tend = 1.0
K = sprand(100,100,0.1)
C = sprand(100,100,0.1)
T0 = rand(100)
f(dT, T, p, t) = (dT .= K * T)
jac(J,T,p,t) = (J .= K)
tspan = (0.0, tend)

ff = ODEFunction(f, jac_prototype=K)
prob = ODEProblem(ff, T0, tspan)
sol = solve(prob, Rosenbrock23())
sol = solve(prob, KenCarp4())

ff = ODEFunction(f, mass_matrix=C, jac_prototype=rand()*C+K)
prob = ODEProblem(ff, T0, tspan)
sol = solve(prob, Rosenbrock23())

ff = ODEFunction(f, jac=jac, jac_prototype=K)
prob = ODEProblem(ff, T0, tspan)
sol = solve(prob, Rosenbrock23())
[...]

Issue #778, opened 2019-06-21
Linear, non-adaptive time-dependent PDE discretization runs into Newton iteration


dkarrasch:
As already discussed a while ago on Gitter, we encountered some problems solving linear time-dependent ODEs. MWE:
```julia
using OrdinaryDiffEq, DiffEqOperators, LinearAlgebra, SparseArrays
u0 = sin.(range(0, 2pi, length = 10))
tspan = (0.0, 1.0)
f = t-> fill(-100t , 10)
update_coeffs! = (A,u,p,t) ->  (@show t; vals = A.nzval; vals .= f(t); A)
A = DiffEqArrayOperator(spdiagm(0 => fill(0.0, 10)); update_func = update_coeffs!)
prob = ODEProblem(A, u0, tspan)
solve(prob, ImplicitEuler(), adaptive = false, dt = 0.1)
```
yields
```julia
t = 0.0
t = 0.0
t = 0.1
t = 0.1
┌ Warning: Newton steps could not converge and algorithm is not adaptive. Use a lower dt.
└ @ DiffEqBase ~/.julia/packages/DiffEqBase/6ewdP/src/integrator_interface.jl:168
retcode: ConvergenceFailure
Interpolation: 3rd order Hermite
t: 1-element Array{Float64,1}:
 0.0
u: 1-element Array{Array{Float64,1},1}:
 [0.0, 0.642788, 0.984808, 0.866025, 0.34202, -0.34202, -0.866025, -0.984808, -0.642788, -2.44929e-16]
```
For non-adaptive `ImplicitEuler` with a constant stepsize of `dt = 0.1`, one would expect that nlsolve just solves a linear system of the form `(I-dt*A(t))*u = b`? In more complicated cases, we saw like more than 50 updates for the same `t`, probably part of the linearization code, thinking that the `DiffEqArrayOperator` encodes a quasilinear PDE operator? But we are dealing with truly linear, yet time-dependent operators. Also, the error message disappears if one reduces the -100 to -10 in the definition of `f`, or reduces the time step to 0.01.

cc: @Phiro333

ChrisRackauckas:
@YingboMa can you double check the update coefficients logic?

dkarrasch:
Is there any update on this issue?

ChrisRackauckas:
[...]

Issue #90, opened 2017-08-01
Magnus and other linear operator solvers


ChrisRackauckas:
It seems many physicists have problems of the form `u' = A*u` and `u' = A(t)*u`. This `A` can be from a PDE discretization, or `u' = H(t)*u` naturally shows up from Schrodinger's equation. In these cases, we need to exploit the linearity of the problem in order to do well.

This can all use the `DiffEqOperator` path where a user defines a `DiffEqOperator` as their function. This contains an `update_coefficients!(A,t,u)` function as well for handling time and state dependence. Then there are different things to tackle here. For one, we can make the stiff solvers "linear-aware". This can be done by, whenever we would be solving a nonlinear equation using a Newton method, instead do an

```julia
if typeof(f) <: DiffEqOperator
  # Linear Solve
else
  # Nonlinear Solve
end
```

In the `#Linear Solve` part, it will just call `update_coefficients!(A,t,u)` and then `\`, so whatever overrides the user gives for how the matrix updates in time and whatever override for how the solving is supposed to take place will be used. (@shivin9 this is why the interface matching is important here)

This will make things like `Trapezoid` automatically be `CrankNicholson`, and will be very useful with things like BDF methods and ApproxFun. In addition, we can pull some inspiration from @jagot for Magnus integrators. 

http://iopscience.iop.org/article/10.1088/1367-2630/14/10/105008

Using the split operators interface we can also do things on `A + B(t)` both linear operators. But I don't know of algorithms directly on that, but technically the definition already exists.

The `DiffEqOperator` will have a matrix-free pretty soon which closely matches that of LinearMaps.jl, but the `DiffEqOperator` interface defines `expm` and `expmv` on the operator itself:

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/diffeq_operator.jl#L27

This means that the matrix-free versions will allow an optional method for directly defining `expmv` (and the array versions will get a quick upgrade to allow for direct definitions of `expm` and `expmv` if someone has a use for it). This solves @jagot's problem since he wanted to use things like

![expa2](https://user-images.githubusercontent.com/1814174/28806535-7f7f914a-7625-11e7-97fd-eb10bec75ccf.png)

to speed up the integrator by putting that definition in manually, and that interface would allow doing so. 

[...]

PR #2879, opened 2025-10-05
Specialize implicit solvers on linear ODEs

hersle:
For linear ODEs

$$
\boldsymbol{u}^\prime(t) = f(\boldsymbol{u},t) = A(t) \boldsymbol{u}
$$

the equation for the values at the next time step in implicit solvers is also linear. It can therefore be solved with exactly one linear solve instead of iterating through a nonlinear solver (see example for implicit Euler method in code changes).

I think implicit methods should specialize on this property to improve performance for (stiff) linear ODEs. As a first rough draft, I have implemented this for the implicit Euler method and tested it on this stiff test problem:
```julia
using OrdinaryDiffEq, LinearAlgebra, BenchmarkTools, SciMLOperators

function J!(J, u, p, t)
    a = exp(t) # t = ln(a)
    Ωr0 = p[1]
    Ωm0 = p[2]
    k = p[3]
    ΩΛ0 = 1 - Ωr0 - Ωm0

    wr = 1//3
    wm = 0
    Ωr = Ωr0 / a^(3*(1+wr))
    Ωm = Ωm0 / a^(3*(1+wm))
    ΩΛ = ΩΛ0
    Ω = Ωr + Ωm + ΩΛ
    ℋ = √(Ω) * a

    J[1,1] = -(1+(k/ℋ)^2/3)
    J[1,2] = -1/2*(a/ℋ)^2*Ωr
    J[1,4] = -1/2*(a/ℋ)^2*Ωm

    J[2,3] = -(1+wr)*k/ℋ
    J[2,1] = -3*(1+wr)*(1+(k/ℋ)^2/3)
    J[2,2] = -3*(1+wr)*1/2*(a/ℋ)^2*Ωr
    J[2,4] = -3*(1+wr)*1/2*(a/ℋ)^2*Ωm

    J[3,3] = -(1-3*wr)
    J[3,2] = wr/(1+wr)*k/ℋ
    J[3,1] = k/ℋ

    J[4,5] = -(1+wm)*k/ℋ
    J[4,1] = - 3*(1+wm)*(1+(k/ℋ)^2/3)
    J[4,2] = -3*(1+wm)*1/2*(a/ℋ)^2*Ωr
    J[4,4] = -3*(1+wm)*1/2*(a/ℋ)^2*Ωm
    
    J[5,5] = -(1-3*wm)
    J[5,4] = wm/(1+wm)*k/ℋ
    J[5,1] = k/ℋ
end

function f!(du, u, p, t)
    J = p[4]
    J!(J, u, p, t)
    mul!(du, J, u)
end

# Implicit Euler with nonlinear solving
alg = ImplicitEuler(nlsolve = NLNewton())
dt = 0.1
adaptive = false
u0 = [2/3, -2, 0, -3/2, 0]
tspan = (-20.0, 0.0)
J0 = zeros(length(u0), length(u0))
p = (5e-5, 0.3, 1e4, J0)
prob1 = ODEProblem(f!, u0, tspan, p)
@btime sol1 = solve(prob1, alg; dt, adaptive); # 264.200 μs (1709 allocations: 83.25 KiB)

# Implicit Euler specialized on linearity
J = MatrixOperator(J0; update_func! = J!)
prob2 = ODEProblem(ODEFunction(J), u0, tspan, p)
@btime sol2 = solve(prob2, alg; dt, adaptive); # 127.000 μs (4163 allocations: 300.98 KiB)
```
Even with this (surely suboptimal) implementation, it is faster.

I want to try to proceed with a hopefully general implementation. But before I commit I need someone to check my reasoning:
1. Do you think this could give worthwhile performance improvements? Particularly for higher-order methods with multiple implicit equations per timestep, like `KenCarp4`?
2. Could this be applied generally to all implicit solvers by moving the `if linear` logic inside `nlsolve!`?
3. Is such a specialized one-linear-solve-if-linear path really different from (and therefore better than) using Newton's method limited to 1 maximum iteration? Or are they equivalent so there is no point in doing this?
4. Can the one-linear-solve-if-linear path also use the strategy for reusing Jacobians over multiple time steps like the nonlinear solvers for efficiency? Or should one always compute new Jacobians in this case?

See also https://github.com/SciML/OrdinaryDiffEq.jl/issues/90, https://github.com/SciML/OrdinaryDiffEq.jl/issues/778, https://github.com/SciML/OrdinaryDiffEq.jl/issues/914 and https://github.com/SciML/OrdinaryDiffEq.jl/issues/2586.

### DIFF
--- a/lib/OrdinaryDiffEqSDIRK/src/sdirk_perform_step.jl
+++ b/lib/OrdinaryDiffEqSDIRK/src/sdirk_perform_step.jl
@@ -82,17 +82,27 @@ end
     alg = unwrap_alg(integrator, true)
     markfirststage!(nlsolver)
 
-    # initial guess
-    if alg.extrapolant == :linear
-        @.. broadcast=false z=dt * integrator.fsalfirst
-    else # :constant
-        z .= zero(eltype(u))
+    if f.f isa MatrixOperator && islinear(f.f)
+        # linear ODE u′(t) = f(u,t) = A(t)*u(t):
+        # solve z = dt*f(u+z,t+dt) = dt*A(t+dt)*(u+z)
+        # as linear equation (I/dt-A(t+dt))*z = A(t+dt)*u with one single \
+        A = Matrix(update_coefficients(f.f, uprev #=irrelevant=#, p, t+dt))
+        b = A * uprev
+        W = I/dt - A
+        z = W \ b
+    else
+        # nonlinear ODE: solve nonlinear equation with nonlinear solver
+        # initial guess
+        if alg.extrapolant == :linear
+            @.. broadcast=false z=dt * integrator.fsalfirst
+        else # :constant
+            z .= zero(eltype(u))
+        end
+        nlsolver.tmp .= uprev
+        nlsolver.γ = 1
+        z = nlsolve!(nlsolver, integrator, cache, repeat_step)
+        nlsolvefail(nlsolver) && return
     end
-
-    nlsolver.tmp .= uprev
-    nlsolver.γ = 1
-    z = nlsolve!(nlsolver, integrator, cache, repeat_step)
-    nlsolvefail(nlsolver) && return
     @.. broadcast=false u=uprev + z
 
     step_limiter!(u, integrator, p, t + dt)


ChrisRackauckas:
See my comments in the issue. I am unsure it needs to be specialized on to get the property? 

This version of the specialization is always refactorizing though, so it will be slower than what you get by default already.

hersle:
Closing for now, see discussion in https://github.com/SciML/OrdinaryDiffEq.jl/issues/778

PR #2881, opened 2025-10-06
Adaptive IDSolve

termi-official:
## Checklist

## Additional context

* [x] Switch from hard-coded SimpleNonlinearSolve to any NonlinearSolve algorithm which supports the iterator interface.
* [ ] Add adaptive homotopy path increments via modified Deuflhard's estimates

### DIFF
--- a/lib/ImplicitDiscreteSolve/src/cache.jl
+++ b/lib/ImplicitDiscreteSolve/src/cache.jl
@@ -1,36 +1,65 @@
-mutable struct ImplicitDiscreteState{uType, pType, tType}
+struct ImplicitDiscreteState{uType, pType, tType}
     u::uType
     p::pType
-    t_next::tType
+    t::tType
 end
 
-mutable struct IDSolveCache{uType} <: OrdinaryDiffEqMutableCache
+mutable struct IDSolveCache{uType, cType} <: OrdinaryDiffEqMutableCache
     u::uType
     uprev::uType
-    state::ImplicitDiscreteState
-    prob::Union{Nothing, SciMLBase.AbstractNonlinearProblem}
+    z::uType
+    nlcache::cType
 end
 
 function alg_cache(alg::IDSolve, u, rate_prototype, ::Type{uEltypeNoUnits},
         ::Type{uBottomEltypeNoUnits}, ::Type{tTypeNoUnits}, uprev, uprev2, f, t,
         dt, reltol, p, calck,
         ::Val{true}) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
-    state = ImplicitDiscreteState(isnothing(u) ? nothing : zero(u), p, t)
-    IDSolveCache(u, uprev, state, nothing)
+    state = ImplicitDiscreteState(zero(u), p, t)
+    f_nl = (resid, u_next, p) -> f(resid, u_next, p.u, p.p, p.t)
+
+    u_len = length(u)
+    nlls = !isnothing(f.resid_prototype) && (length(f.resid_prototype) != u_len)
+    prob = if nlls
+        NonlinearLeastSquaresProblem{isinplace(f)}(
+            NonlinearFunction(f_nl; resid_prototype = f.resid_prototype),
+            u, state)
+    else
+        NonlinearProblem{isinplace(f)}(f_nl, u, state)
+    end
+
+    nlcache = init(prob, alg.nlsolve)
+
+    IDSolveCache(u, uprev, state.u, nlcache)
 end
 
 isdiscretecache(cache::IDSolveCache) = true
 
-struct IDSolveConstantCache <: OrdinaryDiffEqConstantCache
-    prob::Union{Nothing, SciMLBase.AbstractNonlinearProblem}
-end
+# struct IDSolveConstantCache <: OrdinaryDiffEqConstantCache
+#     prob::Union{Nothing, SciMLBase.AbstractNonlinearProblem}
+# end
 
 function alg_cache(alg::IDSolve, u, rate_prototype, ::Type{uEltypeNoUnits},
         ::Type{uBottomEltypeNoUnits}, ::Type{tTypeNoUnits}, uprev, uprev2, f, t,
         dt, reltol, p, calck,
         ::Val{false}) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
-    state = ImplicitDiscreteState(isnothing(u) ? nothing : zero(u), p, t)
-    IDSolveCache(u, uprev, state, nothing)
+    state = ImplicitDiscreteState(zero(u), p, t)
+    f_nl = (u_next, p) -> f(u_next, p.u, p.p, p.t)
+
+    u_len = length(u)
+    nlls = !isnothing(f.resid_prototype) && (length(f.resid_prototype) != u_len)
+    prob = if nlls
+        NonlinearLeastSquaresProblem{isinplace(f)}(
+            NonlinearFunction(f_nl; resid_prototype = f.resid_prototype),
+            u, state)
+    else
+        NonlinearProblem{isinplace(f)}(f_nl, u, state)
+    end
+
+    nlcache = init(prob, alg.nlsolve)
+
+    # FIXME Use IDSolveConstantCache
+    IDSolveCache(u, uprev, state.u, nlcache)
 end
 
 get_fsalfirstlast(cache::IDSolveCache, rate_prototype) = (nothing, nothing)
--- a/lib/ImplicitDiscreteSolve/src/solve.jl
+++ b/lib/ImplicitDiscreteSolve/src/solve.jl
@@ -1,38 +1,33 @@
-# Remake the nonlinear problem, then update
 function perform_step!(integrator, cache::IDSolveCache, repeat_step = false)
-    @unpack alg, u, uprev, dt, t, f, p = integrator
-    @unpack state, prob = cache
-    state.u .= uprev
-    state.t_next = t
-    prob = remake(prob, p = state)
+    (; alg, u, uprev, dt, t, f, p) = integrator
 
-    u = solve(prob, SimpleNewtonRaphson())
-    integrator.sol = SciMLBase.solution_new_retcode(integrator.sol, u.retcode)
-    integrator.u = u
+    # initial guess
+    if alg.extrapolant == :linear
+        @.. broadcast=false cache.z=integrator.uprev + dt * (integrator.uprev - integrator.uprev2)
+    else # :constant
+        cache.z .= integrator.u
+    end
+    state = ImplicitDiscreteState(cache.z, p, t)
+
+    # nonlinear solve step
+    SciMLBase.reinit!(cache.nlcache, p=state)
+    # TODO compute convergence rate estimate
+    # for i in 1:10
+    #     step!(cache.nlcache)
+    #     # ...
+    # end
+    solve!(cache.nlcache)
+    if cache.nlcache.retcode != ReturnCode.Success
+        integrator.force_stepfail = true
+        return
+    end
+
+    # Accept step
+    u .= cache.nlcache.u
 end
 
 function initialize!(integrator, cache::IDSolveCache)
-    integrator.u isa AbstractVector && (cache.state.u .= integrator.u)
-    cache.state.p = integrator.p
-    cache.state.t_next = integrator.t
-    f = integrator.f
-
-    _f = if isinplace(f)
-        (resid, u_next, p) -> f(resid, u_next, p.u, p.p, p.t_next)
-    else
-        (u_next, p) -> f(u_next, p.u, p.p, p.t_next)
-    end
-    u_len = isnothing(integrator.u) ? 0 : length(integrator.u)
-    nlls = !isnothing(f.resid_prototype) && (length(f.resid_prototype) != u_len)
-
-    prob = if nlls
-        NonlinearLeastSquaresProblem{isinplace(f)}(
-            NonlinearFunction(_f; resid_prototype = f.resid_prototype),
-            cache.state.u, cache.state)
-    else
-        NonlinearProblem{isinplace(f)}(_f, cache.state.u, cache.state)
-    end
-    cache.prob = prob
+    integrator.u isa AbstractVector && (cache.z .= integrator.u)
 end
 
 function _initialize_dae!(integrator, prob::ImplicitDiscreteProblem,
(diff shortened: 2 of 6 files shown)


lib/ImplicitDiscreteSolve/test/runtests.jl
```diff
@@ -71,16 +71,16 @@ end
     end
 end
 
-@testset "Handle nothing in u0" begin
-    function empty(u_next, u, p, t)
-        nothing
-    end
+# @testset "Handle nothing in u0" begin
```
termi-official: I have trouble understanding the purpose of this test. What exactly is the practical scenario here?

lib/ImplicitDiscreteSolve/src/alg_utils.jl
```diff
@@ -11,9 +11,35 @@ end
...
+#         prob, alg::IDSolve, u, fu, J, δu;
+#         trace_level::ConvergenceRateTracing, kwargs... # This kind of dispatch does not work. Need to figure out a different way.
+# )
+#     inner_trace = NonlinearSolveBase.init_nonlinearsolve_trace(
+#         prob, alg, u, fu, J, δu;
+#         trace_level.inner_tracing, kwargs...
+#     )
+
+#     return ConvergenceRateTraceTrick(eltype(δu)[], eltype(fu)[], inner_trace)
+# end
```
termi-official, line 43: From my understanding, it should be possible to use `NonlinearSolveBase.init_nonlinearsolve_trace` to query convergence rate estimates. However, in the current design I cannot add a new dispatch. Should I make a PR to pull the trace level before the kwargs (i.e. `NonlinearSolveBase.init_nonlinearsolve_trace(prob, alg, u, fu, J, δu, trace_level; kwargs...)` for custom dispatches?

lib/ImplicitDiscreteSolve/src/cache.jl
```diff
@@ -1,36 +1,66 @@
...
 end
 
 function alg_cache(alg::IDSolve, u, rate_prototype, ::Type{uEltypeNoUnits},
         ::Type{uBottomEltypeNoUnits}, ::Type{tTypeNoUnits}, uprev, uprev2, f, t,
         dt, reltol, p, calck,
         ::Val{true}) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
-    state = ImplicitDiscreteState(isnothing(u) ? nothing : zero(u), p, t)
-    IDSolveCache(u, uprev, state, nothing)
+    state = ImplicitDiscreteState(zero(u), p, t)
+    f_nl = (resid, u_next, p) -> f(resid, u_next, p.u, p.p, p.t)
```
termi-official, line 20: What is the reasoning here to include the current `u` in the signature of this function, but no information on dt?

lib/ImplicitDiscreteSolve/src/cache.jl
```diff
@@ -1,36 +1,66 @@
...
 end
 
 isdiscretecache(cache::IDSolveCache) = true
 
-struct IDSolveConstantCache <: OrdinaryDiffEqConstantCache
-    prob::Union{Nothing, SciMLBase.AbstractNonlinearProblem}
-end
+# struct IDSolveConstantCache <: OrdinaryDiffEqConstantCache
+#     prob::Union{Nothing, SciMLBase.AbstractNonlinearProblem}
+# end
```
termi-official: This is dead code on master. Is there any benefit in adding this variant in first place?

lib/ImplicitDiscreteSolve/src/controller.jl
```diff
@@ -0,0 +1,52 @@
...
+    @inline g(x) = √(1+4x) - 1
+
+    # Adapt dt with a priori estimate (Eq. 5.24)
+    (; Θks) = integrator.cache
+    (; Θbar, γ, Θmin, qmin, qmax, p) = controller
+
+    Θ₀ = length(Θks) > 0 ? max(first(Θks), Θmin) : Θmin
+    q = clamp(γ * (g(Θbar)/(g(Θ₀)))^(1/p), qmin, qmax)
+
+    return q
```
termi-official, line 56: I think I am misunderstanding what exactly should be returned here and how it integrates with the dt update logic. If I want to scale the time step by a factor of 2, then what exactly should be returned here?

lib/ImplicitDiscreteSolve/src/controller.jl
```diff
@@ -0,0 +1,52 @@
+Base.@kwdef struct KantorovichTypeController <: OrdinaryDiffEqCore.AbstractController
```
termi-official, line 28: TODO reference and documentation

lib/ImplicitDiscreteSolve/src/solve.jl
```diff
@@ -1,38 +1,71 @@
...
-    integrator.sol = SciMLBase.solution_new_retcode(integrator.sol, u.retcode)
-    integrator.u = u
-end
+    # initial guess
+    if alg.extrapolant == :linear
+        @.. broadcast=false cache.z=integrator.u + dt * (integrator.u - integrator.uprev)
+    else # :constant
+        cache.z .= integrator.u
+    end
+    state = ImplicitDiscreteState(cache.z, p, t+dt)
```
termi-official: On master we solve at time $t$. From my understanding, we should solve at $t+dt$ to obtain the next solution. Can someone confirm or reject this?

lib/ImplicitDiscreteSolve/test/runtests.jl
oscardssmith: we use `u=nothing` to represent length 0 `u` in a type stable manner (as op0osed to a `Vector` which is length `0` at runtime)

termi-official: Sorry, I think I do not understand the point - do you imply here that a zero-length solution vector not type stable and this is why we resort to "nothing"?

When I try to get this test running, I get errors from an alias analysis function not being dispatched for nothing in NonlinearSolve (https://github.com/SciML/NonlinearSolve.jl/blob/ac9344f9359833282e443c4479427ad9ce3311dd/lib/NonlinearSolveFirstOrder/src/solve.jl#L157).

termi-official: Inplace functions with empty return types are also not correctly handled.

```julia
using NonlinearSolveFirstOrder
function f(u, p)
    nothing
end
prob = NonlinearProblem{false}(f, Float64[], nothing)
iter = init(prob, NewtonRaphson())
```

errors when the termination cache is built, because the increment type cannot be derived.

oscardssmith: the point is that zero length `u` tends to cause problems (e.g. solvers will try to look at the first element of the state to find a type), so this way you can skip the solve process since nothing interesting can happen if you don't have any state. See https://github.com/SciML/DiffEqBase.jl/blob/3667bdbdc85489f7b296316df7f4c440519e82f6/src/solve.jl#L31 for how this gets handled for ODEs/DAEs.

lib/ImplicitDiscreteSolve/src/cache.jl
oscardssmith: imo dead code should be deleted. If we want it again, that's what git history is for.

lib/ImplicitDiscreteSolve/test/runtests.jl
termi-official: Wouldn't it make more sense from a software engineering stand point to replace such `isa` statements with dispatchable functions to make the code more extensible?

lib/ImplicitDiscreteSolve/src/solve.jl
oscardssmith: the somewhat complicated part here is that we want to match a `DiscreteSolve`. I forget which is the right way here. @jClugstor might remember.

lib/ImplicitDiscreteSolve/test/runtests.jl
oscardssmith: quite possibly. DiffEqBase is not my favorite code organization.

lib/ImplicitDiscreteSolve/src/solve.jl
jClugstor: From what I can tell from the docs https://docs.sciml.ai/DiffEqDocs/stable/types/discrete_types/ 
if you put in `u_n, p, t_(n+1)` to the function you get `u_(n+1)` out, so in this case you would get `u_(t + dt)` out I guess. So if you want `u_(t + dt)` I think that's correct. Not entirely sure though.

lib/ImplicitDiscreteSolve/src/alg_utils.jl
```diff
@@ -11,9 +11,33 @@ end
 SciMLBase.isdiscrete(alg::IDSolve) = true
 
 isfsal(alg::IDSolve) = false
-alg_order(alg::IDSolve) = 0
+alg_order(alg::IDSolve) = 1
```
ChrisRackauckas, line 14: why 1 here?

lib/ImplicitDiscreteSolve/src/cache.jl
ChrisRackauckas: `dt` is just a parameter?

lib/ImplicitDiscreteSolve/src/controller.jl
ChrisRackauckas: Yeah I'm not sure what this is.

lib/ImplicitDiscreteSolve/src/controller.jl
ChrisRackauckas: 2

lib/ImplicitDiscreteSolve/src/solve.jl
```diff
@@ -1,38 +1,72 @@
...
-    prob = if nlls
-        NonlinearLeastSquaresProblem{isinplace(f)}(
-            NonlinearFunction(_f; resid_prototype = f.resid_prototype),
-            cache.state.u, cache.state)
-    else
-        NonlinearProblem{isinplace(f)}(_f, cache.state.u, cache.state)
+    resize!(Θks, 0)
+    residualnormprev = zero(eltype(u))
+    while NonlinearSolveBase.not_terminated(nlcache)
+        step!(nlcache)
```
ChrisRackauckas, line 26: should it also be trying jacobian reuse in here?

lib/ImplicitDiscreteSolve/src/alg_utils.jl
termi-official: This comes from the analysis shown in Deuflhard, Newton Methods for Nonlinear Problems (Section 5.1.1, see Equation 5.6 and the surrounding definition).

The algorithms here have an associated order in the sense that for a given $dt_n = t_{n+1} - t_n$ we have for some solution $\hat{u}(t_n+1)$ derived from an initial guess given at $u({t_n})$. Now we can define an associated ODE (Davidenko differential equation*) for each nonlinear problem with a "time parameter" by taking the time derivative of the time parameter, which has an analytical solution $\bar{u}(t_{n+1})$, given the same initial guess ($u({t_n})$). The solver now has order $p$ if we have $$||\hat{u}(t_{n+1}) - \bar{u}(t_{n+1})|| \leq C dt^p_n . $$
Does that explain it?

*The Davidenko differential equation for $F(u,t)$ is simply $du/dt = - dF/dx (u,t)^{-1} * dF/dt (u,t)$.

lib/ImplicitDiscreteSolve/src/controller.jl
termi-official: Oh, sorry. I will write the docs, no worries. I just left it here so I do not forget before merging. This is a controller derived from a posteriori estimates on how much the convergence radius in the Newton-Kantorovich theorem changes for some increment $dt_n$ and a solution given at $t_n$.

lib/ImplicitDiscreteSolve/src/cache.jl
termi-official: I guess my question is simply, why is $dt$ (or tprev) not a parameter, but $uprev$ is part of the function signature?

lib/ImplicitDiscreteSolve/src/solve.jl
termi-official: Yes. However, shoudln't the reuse strategy should be part of the NonlinearSolve algorithm instead of some intermediate layer?

ChrisRackauckas: if you expand out the iterator then you're taking over the iteration strategy.

lib/ImplicitDiscreteSolve/src/alg_utils.jl
ChrisRackauckas: I guess I'm confused. It's a discrete time problem, it's exact?

lib/ImplicitDiscreteSolve/src/solve.jl
termi-official: Correct. I would like to remove that later. Right now I cannot use the high level API because I cannot access the convergence rates which I need in the controller.

lib/ImplicitDiscreteSolve/src/alg_utils.jl
termi-official: Okay, let me try to explain it differently then. We have the parametric function $F(u,t)$ and want to find the solution of $F(u_2,t_2) = 0$ given a $u_1$ such that $F(u_1, t_1) = 0$. With with $t_1 < t_2$ we want to find some initial guess for $u^0_2$ given $u_1$, such that the initial guess $u^0_2$ is inside the convergence radius of a Newton method to solve $F(u_2,t_2) = 0$. The obvious choice is that we can simply say our initial guess is simply $u_1$, but this is typically not really great, as $t_2 - t_1$ is often quite small. However, we can use additional information contained in $F(u,t) = 0$. To be specific the derivative with respect to the parameter $t$ contains some extra information which we can use to improve the initial guess. Here we can observe that analytically solving the associated Davidenko differential equation with initial condition $u_1$ on $[t_1, t_2]$ is equivalent to solving $F(u_2,t_2) = 0$. Furthermore, we can exploit this information to inform how large $t_2$ can be chosen, such that the Newton method is guaranteed converge for a given initial guess. Now, the order of this extrapolation polynomial is directly related to the order of the implicit discrete solver. I hope that helps.

lib/ImplicitDiscreteSolve/src/controller.jl
termi-official: I have added a docstring with a brief explanation and reference now.

ChrisRackauckas:
Rebase

ChrisRackauckas:
This seems like it's just a very different... method. Maybe we should just make the other one be SimpleImplicitDiscreteSolve which is lower dependency? Then merge this in as ImplicitDiscreteSolve.

PR #2896, opened 2025-10-24
Removing mismatched tableaus

KeshavVenkatesh:
I cleaned up the convergence test file and removed the extra tableau checks. Can you please review this and let me know if it looks good? Thanks!

Thanks for providing this opportunity for me. I have greatly enjoyed this work. If there is anything else that I can help out with, please let me know. I am open to working on small-scale projects or bigger tasks as well.

### DIFF
--- a/lib/OrdinaryDiffEqVerner/test/ode_verner_tests.jl
+++ b/lib/OrdinaryDiffEqVerner/test/ode_verner_tests.jl
@@ -27,40 +27,8 @@ prob_iip = ODEProblem(ODEFunction(f!; analytic = (u0, p, t) -> [exp(-t)]), [1.0]
 # -------------------------------------------------------------
 # Helper for convergence test consistency
 function check_convergence(dts, prob, alg, order_expected)
-    alg_name = string(typeof(alg).name.name)
-    # println("\n" * "="^70)
-    # println("Testing: $alg_name (Expected Order: $order_expected)")
-    # println("="^70)
-
-    # # Compute errors for each dt
-    # errors = []
-    # println("\nComputing errors for each dt:")
-    # for dt in dts
-    #     sol = solve(prob, alg, dt=dt, adaptive=false)
-    #     if prob.f.analytic !== nothing
-    #         error_val = norm(sol.u[end] - prob.f.analytic(prob.u0, prob.p, prob.tspan[2]))
-    #     else
-    #         error_val = norm(sol.u[end] - sol.prob.f.analytic(prob.u0, prob.p, prob.tspan[2]))
-    #     end
-    #     push!(errors, error_val)
-    #     println("  dt = $dt: error = $error_val")
-    # end
-
-    # # Print convergence orders between consecutive dt values
-    # println("\nConvergence orders between consecutive dt values:")
-    # for i in 2:length(errors)
-    #     if errors[i] > 0 && errors[i-1] > 0
-    #         order = log(errors[i-1]/errors[i]) / log(dts[i-1]/dts[i])
-    #         println("  Order between dt=$(dts[i-1]) and dt=$(dts[i]): $order")
-    #     end
-    # end
-
-    # # Use DiffEqDevTools for overall convergence estimate
+    # Use DiffEqDevTools for overall convergence estimate
     sim = test_convergence(dts, prob, alg)
-    # println("\nOverall convergence order estimate: $(sim.𝒪est[:final])")
-    # println("="^70)
-
-    # @test sim.𝒪est[:final] > order_expected
     @test (sim.𝒪est[:final] > order_expected) || (abs(sim.𝒪est[:final] - order_expected) < testTol)
 
 end
@@ -160,23 +128,3 @@ println("RKV76IIa")
 dts = [2, 1, 0.5, 0.25]
 
 check_convergence(dts, prob_oop, RKV76IIa(), 7)
-# check_convergence(dts, prob_iip, RKV76IIa(), 7)
-
-tabalg = ExplicitRK(tableau = constructVernerEfficient7(BigFloat))
-sol1 = solve(probnumbig, RKV76IIa(); dt = 1 / 2^6, adaptive = false, save_everystep = false)
-
-sol2 = solve(probnumbig, tabalg; dt = 1 / 2^6, adaptive = false, save_everystep = false)
-@test abs(sol1.u[end] - sol2.u[end]) < 1e-10
-
-sol1 = solve(probnumbig, RKV76IIa(); dt = 1 / 2^3, adaptive = false, save_everystep = false)
-
-sol2 = solve(probnumbig, tabalg; dt = 1 / 2^3, adaptive = false, save_everystep = false)
-diff = sol1.u[end] - sol2.u[end]
-@test minimum(abs.(diff) .< 1e-10)
-
-sol2 = solve(probnumbig, tabalg; dt = 1 / 2^6)
-sol1 = solve(probnumbig, RKV76IIa(); dt = 1 / 2^6)
-
-@test length(sol1) == length(sol2)
-@test SciMLBase.successful_retcode(sol1)
-@test SciMLBase.successful_retcode(sol2)


ChrisRackauckas:
I think a very nice project that could have a high impact on load times could be to finish the tableau-ization of some of the algorithms. We have a bounty out there for some, but if you're on the explicit RKs right now you might want to read https://github.com/SciML/OrdinaryDiffEq.jl/issues/233#issuecomment-2081665059 . For example, ExplicitRK works just by giving the table of coefficients, so in theory your PR didn't need to write the `perform_step!` and just needed to supply the coefficients. But that doesn't end up actually being the case for example with Verner methods because ExplicitRK cannot handle lazy interpolation values (wait, did you do the interpolation coefficients part?). If that can be added to ExplicitRK, we can benchmark whether it's actually fast enough that OrdinaryDiffEqVerner could cut out its specialized stepping code and just be tableau instantiations. That would then make them all compile as the same algorithm with different runtime values, so compilation would drop immensely.

KeshavVenkatesh:
I looked at the code in detail and here is what I found:

- I don't see any extra coefficients defined in the tableaus for RKV76IIa ([code](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/OrdinaryDiffEqVerner/src/verner_tableaus.jl#L3978)), so I don't know if it needs the implementation for the interpolation coefficients. If it does, the implementation should be in the file [here](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/OrdinaryDiffEqVerner/src/verner_rk_perform_step.jl#L1272).
- Please let me know if any extra coefficients are defined in the original paper for RKV76IIa.
- For ExplicitRK as well, I don't see any extra coefficients defined for implementing the interpolation.
- Should I look at the paper for the Dormand-Prince Method and see if it has any extra coefficients?
- Could you also let me know if we can benchmark without the lazy interpolation (comparing RKV76IIa and ExplicitRK, both of which do not have the lazy interpolation) to see if there are any performance advantages for ExplicitRK?

ChrisRackauckas:
Oh it looks like ExplicitRK doesn't do any interpolations yet, so that's the starting point. I would start by extending the tableaus to allow for it, and doing Tsit5 and DP5 first since those don't have any lazy stages. 

> Could you also let me know if we can benchmark without the lazy interpolation (comparing RKV76IIa and ExplicitRK, both of which do not have the lazy interpolation) to see if there are any performance advantages for ExplicitRK?

Yes, implementing it with ExplicitRK and checking for any performance difference would be good. ExplicitRK will be slower, but the question is whether it's close enough now.

KeshavVenkatesh:
> Oh it looks like ExplicitRK doesn't do any interpolations yet, so that's the starting point. I would start by extending the tableaus to allow for it, and doing Tsit5 and DP5 first since those don't have any lazy stages.

Could you please point me to the source where the extra coefficients for interpolations are defined for Tsit5 and DP5? I can start with the implementations  

> Yes, implementing it with ExplicitRK and checking for any performance difference would be good. ExplicitRK will be slower, but the question is whether it's close enough now.

Just to clarify, are you suggesting that we can benchmark ExplicitRK and RKV76IIa without the interpolations or should we implement the interpolation first before benchmarking?

KeshavVenkatesh:
@ChrisRackauckas : I would like to get your inputs on these two points above.

KeshavVenkatesh:
I spent some more time looking at the code and these are my updates. Could you please let me know if I am on the right track?

**Conclusion**: If the benchmark results between ExplicitRK (with the logic to handle dynamic matrices) and other algorithms are comparable, we can consider switching to a simpler logic, but I am not sure if the results (refer to details below) are accurate to do this. Please let me know what I should do.

**Approach**:

1. I converted the tableaus for Tsit5 into the matrix form expected by ExplicitRK. I then used this Tsit5 Tableau instead of the default Dormand-Prince matrix for solving with ExplicitRK.
2. I benchmarked using some standard ODE equations, and these are some of the results. I used the following differential equations to benchmar

> [Lorenz-96 Atmospheric Model](https://en.wikipedia.org/wiki/Lorenz_96_model) (T1)
> [Reaction-Diffusion PDE](https://en.wikipedia.org/wiki/Reaction%E2%80%93diffusion_system) (T2)
> [Coupled Oscillators Network](https://en.wikipedia.org/wiki/Oscillatory_neural_network) (T3)

3. Please let me know if I should be using a different benchmark to compare the performances.
4. I am attaching the benchmark results . Detailed results can be found [here](https://docs.google.com/document/d/1Qy8rmVH3BM6U-AGMuwqt9zw79cpCGiwKtvJSmOZtJVw/edit?tab=t.0).

- Comparison (T1)
> Speedup (Tsit5 vs ExplicitRK): 1.05x
> Memory reduction: 1.87xa
> Allocation reduction: 1.67x

- Comparison (T2)
> Speedup (Tsit5 vs ExplicitRK): 1.02x
> Memory reduction: 1.25x
> Allocation reduction: 1.25x

- Comparison (T3)
> Speedup (Tsit5 vs ExplicitRK): 1.72x
> Memory reduction: 2.61x
> Allocation reduction: 2.21x

5. Based on the results, I don't see much difference between ExplicitRK and Tsit5. One thing I noticed is using rational numbers directly instead of floating-points all the  performance metrics drop significantly. For the same benchmark the speedup or the time it takes to complete was 523.35x. More details can be found [here](https://docs.google.com/document/d/1Qy8rmVH3BM6U-AGMuwqt9zw79cpCGiwKtvJSmOZtJVw/edit?tab=t.m8uztx85ye4o). I can try to benchmark with other algorithms to confirm if you recommend so. 
6. I looked at the code and I think I can implement the interpolation logic for the ExplicitRK using the matrix approach without hardcoding. 
7. Once implemented, we can benchmark the interpolation logic as well to see if the non-hardcoded version of ExplicitRK performs well enough compared to the hardcoded solutions.

KeshavVenkatesh:
Hello Professor Rackauckas,

I updated this thread (SciML/OrdinaryDiffEq.jl#2896) with my findings. I am
sending this email in case you missed the thread and my updates. I would
greatly appreciate it if you can review my comments and provide your
feedback. I will spend more time this week making progress, but I want to
make sure I am on the right track.

Regards,
Keshav

ChrisRackauckas:
Sorry catching up after travel.

> Could you please point me to the source where the extra coefficients for interpolations are defined for Tsit5 and DP5? I can start with the implementations

DP5 and Tsit5 don't have extra stages for their interpolations. Only BS5 and the VernX methods have more stages.

> Just to clarify, are you suggesting that we can benchmark ExplicitRK and RKV76IIa without the interpolations or should we implement the interpolation first before benchmarking?

Both need to get done, the order doesnt matter much really

For the benchmarks, the smaller and simpler the ODE the more overhead, since otherwise the time is in f. So use the Lotka Volterra and Lorenz equations (readme example and doc example in docs.sciml.ai) as worst case scenarios.

Yeah I think the ExplicitRK methods might need manual specializations for floats? We do this for example in the Tsit5 direct tableau.

Yeah the interpolation is just a bunch of coefficients for b functions which are listed as linear combinations of the k with coefficients r. Just like the RK steps.

KeshavVenkatesh:
Thanks for the suggestions and feedback. I wanted to give the following updates:


- I benchmarked using Lotka Volterra and Lorenz equations and removed the Reaction-Diffusion and Coupled Oscillators Network ODEs.
- I also found that the reason the tableau with rational numbers was performing poorly compared to the floating points was because the rational numbers did not match with the floating point numbers, resulting in a larger number of steps for convergence. I was able to fix the rationale tableau to match the values of the floating point and was able to reduce the time. You can see the benchmark results [here](https://docs.google.com/document/d/1Qy8rmVH3BM6U-AGMuwqt9zw79cpCGiwKtvJSmOZtJVw/edit?tab=t.0#heading=h.8vsdo9gtd8q0).
- I will now work on implementing the interpolation logic and rerun the benchmarks.
- I will clean up my code and send you the PR in the evening for you to review/provide feedback.


Apologies for pinging you multiple times when you were traveling.

PR #2899, opened 2025-10-26
New Controller Interface

### LINKED ISSUE #2883: Interaction between custom controllers and time loop
**Question❓**

I am a bit stuck with the internals of the interaction between the time step loop and the controller. Here are some questions:

1. What exactly are `q` and `qold`? I could not find the mentioned work by Cechino. I probably also misread the latter variable as $q_{old}$, but it does not store some q. Instead it seems to store some time step length https://github.com/SciML/OrdinaryDiffEq.jl/blob/4abda1b6b4b8985e2bd3747fce11e5983d121fc6/lib/OrdinaryDiffEqCore/src/integrators/controllers.jl#L74 
2. What is the purpose of spliting the accept logic into two functions `stepsize_controller!` and `step_accept_controller!`?
3. On the docs page for the controller (https://docs.sciml.ai/DiffEqDocs/stable/extras/timestepping/), should the first equation just be a `err^{scaled}` or is there really some index involved that I missed?
4. Is `accept_step_controller` considered internal API? If not, should it be included in the docs page?

termi-official:
Resolves #2883 .

## TODO

* [x] Upgrade non-standard controllers to new interface
* [x] Introduce multi-controller for algorithms with stiffness switching
* [x] Add appropriate documentation of the full API

## Not addressed in this PR / Remaining things to be addressed

* [ ] Reformulate PI controller to remove inverses (break semantics)
* [ ] Remove redundant variables from integrator (breaking change)
* [ ] Move controller cache into algorithm cache whenever appropriate (and maybe add the controller to the algorithm)
* [ ] Remove q from `step_accept_controller!(integrator, controller_cache::AbstractControllerCache, q)`
* [ ] Remove `DummyController`
* [ ] Update DelayDiffEq and friends to use the new interface

### DIFF
--- a/lib/OrdinaryDiffEqCore/Project.toml
+++ b/lib/OrdinaryDiffEqCore/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqCore"
 uuid = "bbf590c4-e513-4bbe-9b18-05decba2e5d8"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "3.2.0"
+version = "3.3.0"
 
 [deps]
 SciMLOperators = "c0aeaf25-5076-4817-a8d5-81caf7dfa961"


lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -175,14 +261,111 @@ function reset_alg_dependent_opts!(controller::PIController, alg1, alg2)
...
+    if iszero(EEst)
+        q = inv(qmax)
+    else
+        q11 = fastpower(EEst, convert(typeof(EEst), beta1))
+        q = q11 / fastpower(qold, convert(typeof(EEst), beta2))
+        integrator.q11 = q11
+        @fastmath q = max(inv(qmax), min(inv(qmin), q / gamma))
+    end
+    cache.q = q
+    q
```
oscardssmith: while we're updating things here, would it be possible to change the formulas around to have less division?

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -333,10 +511,139 @@ function step_accept_controller!(integrator, controller::PIDController, alg, dt_
...
+    # `EEst_min` is smaller than PETSC_SMALL used in the equivalent logic in PETSc.
+    # For example, `eps(Float64) ≈ 2.2e-16` but `PETSC_SMALL ≈ 1.0e-10` for `double`.
+    EEst_min = eps(typeof(EEst))
+    # The code below is a bit more robust than
+    # ```
+    # if iszero(EEst)
+    #   EEst = eps(typeof(EEst))
+    # end
+    # ```
+    EEst = ifelse(EEst > EEst_min, EEst, EEst_min)
```
oscardssmith: `EEst = max(EEst, EEst_min)`?

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -333,10 +511,139 @@ function step_accept_controller!(integrator, controller::PIDController, alg, dt_
...
+    #
+    # `EEst_min` is smaller than PETSC_SMALL used in the equivalent logic in PETSc.
+    # For example, `eps(Float64) ≈ 2.2e-16` but `PETSC_SMALL ≈ 1.0e-10` for `double`.
+    EEst_min = eps(typeof(EEst))
+    # The code below is a bit more robust than
+    # ```
+    # if iszero(EEst)
+    #   EEst = eps(typeof(EEst))
+    # end
+    # ```
```
oscardssmith, line 689: this comment makes no sense to me...

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -333,10 +511,139 @@ function step_accept_controller!(integrator, controller::PIDController, alg, dt_
...
+    EEst = ifelse(EEst > EEst_min, EEst, EEst_min)
+
+    cache.err[1] = inv(EEst)
+    err1, err2, err3 = cache.err
+
+    k = min(alg_order(alg), alg_adaptive_order(alg)) + 1
+    dt_factor = err1^(beta1 / k) * err2^(beta2 / k) * err3^(beta3 / k)
```
oscardssmith, line 696: use `fastpow` and `inv(k)` to save computation

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
ranocha: Which part exactly?

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
ranocha: We have some evidence in Trixi.jl CI tests that `fastpow` (via `Float32`) reduces reproducibility of computational results across different systems, while the impact on computational efficiency is minimal (if at all noticeable). The PID controllers are used only by a few algorithms designed for large-scale PDE discretizations, where the runtime performance of computing the time step size is negligible. In contrast, stability and reproducibility are essential to us.

termi-official: I am also a bit confused. Even without fast powers etc the controller code should never become the bottleneck if we have more than a few nonlinear ODEs, right?

ChrisRackauckas: Large PDEs is the best case scenario here so it's not a good testing case. The worst case scenario is a small non-stiff ODE. Test again with something like a 2 ODE one compartment model. What is the overhead? At least last it was timed at around 5x.

I would be curious to know what you mean by "reproducibility" here because there is nothing that is platform-dependent and no system libraries used

ChrisRackauckas: Lorenz equations or other similarly sized chaotic systems are also good examples to test here.

ranocha: We typically test many examples: run an ODE solver for a few time steps and record the norms of the solution minus a reference (e.g., the initial condition). With `fastpow`, we observed that the results seem to differ more significantly when running on different systems (likely due to the reduced precision during the computation).

ranocha: None of the default methods uses a `PIDController`, which is the only one not using `fastpow` at the moment. I would like to keep this state for backward compatibility in Trixi.jl.

ChrisRackauckas: This PR definitely shouldn't change any code. It's a refactor? Any substance changes would need to go anywhere else. I didn't see that this wasn't to the PI

> We typically test many examples: run an ODE solver for a few time steps and record the norms of the solution minus a reference (e.g., the initial condition). With fastpow, we observed that the results seem to differ more significantly when running on different systems (likely due to the reduced precision during the computation).

That's definitely odd. I used test have some checks on chaotic systems that you'd reproduce across OS's and chips . @oscardssmith can you look into this?

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
termi-official: Okay I can, but the observed behavior changes (i.e. the number of taken time steps differs by a few percent depending on how I reorder the operations). How do I test if that causes a regression in precision?

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
termi-official: xref https://github.com/SciML/OrdinaryDiffEq.jl/pull/2899#discussion_r2465411915 .

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
termi-official: Let us leave changes in these details for a separate PR for now to not delay the refactor for v7. These can be done later in a non-breaking way.

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
oscardssmith: will investigate

lib/OrdinaryDiffEqCore/src/integrators/type.jl
```diff
@@ -105,10 +107,13 @@ mutable struct ODEIntegrator{algType <: Union{OrdinaryDiffEqAlgorithm, DAEAlgori
     tdir::tdirType
     eigen_est::eigenType
     EEst::EEstT
+    # TODO vvv remove these
     qold::QT
     q11::QT
     erracc::QT
     dtacc::tType
+    # TODO ^^^ remove these
+    controller_cache::CC
```
termi-official, line 120: Looking at some of the algorithms, we even might want to think moving the controller into the algorithm cache together with some helpers to construct standard controllers.

oscardssmith:
@ChrisRackauckas what would you think about removing the Legacy controllers (the renamed version of the current ones). I don't think this version is less breaking, and the new controllers aren't breaking (except for people making their own controllers which I don't think is common)

termi-official:
> [..] Legacy controllers (the renamed version of the current ones). 

Please note that I have changed the names back. They now differ only by an intermediate supertype. So `PIController <: AbstractLegacyController` is the old one and `NewPIController <: AbstractController` is the new one.

ChrisRackauckas:
> @ChrisRackauckas what would you think about removing the Legacy controllers (the renamed version of the current ones). I don't think this version is less breaking, and the new controllers aren't breaking (except for people making their own controllers which I don't think is common)

I think it's best to just have a deprecation path. v7 is coming soon and we just delete them with that. So if it's not too hard to have a deprecation path that would definitely help people with the upgrading. But yes the main user is likely @ranocha and the Trixi crew, I at least don't see any use cases publicly in the wild.

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
oscardssmith: it looks like the differences between FastPow and ^ are basically noise (since timestepping a chaotic system will take a pseudorandom number of steps depending on epsilon level differences). This PR doesn't change the code. Dennis and I have discussed a followup PR with some further cleanup, but this is pure refactoring.

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -337,6 +573,134 @@ function step_reject_controller!(integrator, controller::PIDController, alg)
...
 
+
+
+struct NewPIDController{T, Limiter} <: AbstractController
+    beta::SVector{3, T} # controller coefficients
+    accept_safety::T   # accept a step if the predicted change of the step size
+    # is bigger than this parameter
+    limiter::Limiter    # limiter of the dt factor (before clipping)
+    qmin::T
+    qmax::T
```
ranocha: These are not used

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -337,6 +573,134 @@ function step_reject_controller!(integrator, controller::PIDController, alg)
...
+    else
+        beta1, beta2, beta3 = beta
+    end
+    beta = SVector(map(float, promote(beta1, beta2, beta3))...)
+    NewPIDController{QT, typeof(limiter)}(
+        beta,
+        QT(accept_safety),
+        limiter,
+        QT(qmin === nothing ? qmin_default(alg) : qmin),
+        QT(qmax === nothing ? qmax_default(alg) : qmax),
```
ranocha:
```suggestion
```

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
termi-official: Ah, sorry. But good catch here!

ChrisRackauckas:
While doing this refactor, maybe it would be good to add the vector of EEst to the controller? Right now we usually call this `atmp`, but it's not uniform across the methods. Basically, it would be good to make that vector of all error estimates be something that is uniformly handled, not just the `norm(vector_EEst)`. The reason is I'm going to want to set it up with logging really soon, because knowing which value is causing step rejections is really helpful for knowing why a solver haulted. But in the current form you'd need to implement that separately per solver. It would be good to make sure this can be more nicely handled in the new controller interface, as that would then make it straightforward to add logging about error control in the controller interface rather than at per-solver level.

termi-official:
Should this also go into this PR or a separate one? If I understand you correctly, then we would like to have changes of the form
```julia
@cache mutable struct ImplicitEulerCache{
    uType, rateType, uNoUnitsType, N, AV, StepLimiter} <:
                      SDIRKMutableCache
    u::uType
    uprev::uType
    uprev2::uType
    fsalfirst::rateType
    atmp::uNoUnitsType # Remove this variable here
    nlsolver::N
    algebraic_vars::AV
    step_limiter!::StepLimiter
end
```
and analogue
```julia
mutable struct IControllerCache{C, T, uNoUnityType} <: AbstractControllerCache
    controller::C
    q::T
    dtreject::T
    # atmp::uNoUnityType
    # I believe this should go here or in the algorithm cache, but not in the integrator itself.
    # EEst::T
end
```
together with the respective change in `perform_step!` and the constant cache?

Also, I think it makes sense to move the controller cache into the algorithm cache and the controller should be part of the algorithm constructor. E.g.
```julia
struct ImplicitEuler{CS, AD, F, F2, P, CT, FDT, ST, CJ, StepLimiter} <:
       OrdinaryDiffEqNewtonAdaptiveAlgorithm{CS, AD, FDT, ST, CJ}
    linsolve::F
    nlsolve::F2
    precs::P
    extrapolant::Symbol
    controller::CT # -> was "controller::Symbol" before
    step_limiter!::StepLimiter
    autodiff::AD
end
```
which defaults to e.g. PIController, while we get for the caches
```julia
@cache mutable struct ImplicitEulerCache{
    uType, rateType, uNoUnitsType, N, AV, StepLimiter, CC} <:
                      SDIRKMutableCache
    u::uType
    uprev::uType
    uprev2::uType
    fsalfirst::rateType
    atmp::uNoUnitsType
    nlsolver::N
    algebraic_vars::AV
    step_limiter!::StepLimiter
    controller_cache::CC
end
```

This could help to simplify the init logic a little bit, as we hand down the initialization of the controller to the algorithm -- and ultimately the algorithm knows best if the passed controller is compatible or not.

ChrisRackauckas:
Rebase

termi-official:
There is some stuff failing which I need to check first.

termi-official:
Fixing the [DiffEqDevTools.jl/Core/1.10](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/20724062775/job/59495152297#logs) failure is a bit tougher without breaking anything. From my understanding we need the controller to be fully differentiable somewhere, but the type of Q seems to be not necessarily a dual number.

Furthermore, I cannot figure out how to fix downgrade CI, as I need to bump and register the Core package first, such that the new types become visible in the downstream packages.

oscardssmith:
downgrade CI isn't your fault

termi-official:
For the default and composite algorithms the `post_newton_controller!` is never called when a BDF algorithm is at turn.

termi-official:
Before someone spends too much time here, I need to push a few more commits to get CI on track again. I do not handle all corner cases in the default alg correctly right now.

termi-official:
Okay, the default algorithm tests are on track. The remaining failures (which are not already addressed by open PRs) can be grouped as follows:

1. For all CIs on `pre`, they fail with a JET compat issue, so I assume it is unrelated
2. Most CIs on `lts` fail because they do not pick up the correct package version. They consistently fail with the new types not being found in OrdinaryDiffEqCore.
3. I suspect the Downstream CI fails, because there is more adjustment necessary outside of this PR (`MethodError: no method matching Float64(::Measurements.Measurement{Float64})` during EEst assignment).
4. Then there is one more test in the `InterfaceI` group which I have absolutely now clue why it is failing, but it consistently fails for Julia 1.12 only, i.e. 1.11 and 1.10 work fine. An MWE is here.

```julia
using OrdinaryDiffEq, Test, LinearAlgebra

T = 100.0;
Ttr = 0.0;
d0 = 1.0e-9;
threshold = 10^4 * d0;
dt = 0.1;
diff_eq_kwargs = Dict();

@inbounds function eom_lorenz!(du, u, p, t)
    σ = 10.0
    ρ = 28.0
    β = 8 / 3
    du[1] = σ * (u[2] - u[1])
    du[2] = u[1] * (ρ - u[3]) - u[2]
    du[3] = u[1] * u[2] - β * u[3]
end

prob = ODEProblem(eom_lorenz!, [0.0, 10.0, 0], (zero(T), T))
integ1 = init(prob, Tsit5(); beta1 = 0.14, diff_eq_kwargs...)
@. prob.u0 = prob.u0 + d0 / sqrt(3)
integ2 = init(prob, Tsit5(); beta1 = 0.14, diff_eq_kwargs...)
integ1.opts.advance_to_tstop = true
integ2.opts.advance_to_tstop = true

# Confirm that the old controller is used
@info integ1.opts.controller

dist = d0
λ = zero(eltype(integ1.u))
i = 0
tvector = dt:dt:T

for τ in tvector
    # evolve until rescaling:
    push!(integ1.opts.tstops, τ)
    @info integ1.u
    step!(integ1)
    @info integ1.u
    push!(integ2.opts.tstops, τ)
    @info integ2.u
    step!(integ2)
    @info integ2.u
    global dist = norm(integ1.u .- integ2.u)
    @info dist
    # Rescale:
    if dist ≥ threshold
        # Rescale and reset everything:
        integ2.u .= integ1.u #.+ (integ2.u .- integ1.u)./a
        u_modified!(integ2, true)
        set_proposed_dt!(integ2, integ1)
        break
    end
end

τ = tvector[end]
push!(integ1.opts.tstops, τ);
step!(integ1);
push!(integ2.opts.tstops, τ);
step!(integ2);
@test dist = norm(integ1.u .- integ2.u) == 0
```

Is a stripped down example of the full test using the "old" controllers from master.

oscardssmith:
@ChrisRackauckas can you set claude on 4? I can confirm that it fails on master.

termi-official:
Commit [554ebc4](https://github.com/SciML/OrdinaryDiffEq.jl/pull/2899/commits/554ebc4651415a6fca7f2a9db1f11928a97bcb94) reveals that the integration test on chaotic systems (test/interface/umodified_test.jl) can also sporadically break on master, as the test relied on both integrators hitting the boundary of the control regieme in the last step, such that both controllers have `qoldinit` valued old errors. Which does not happen super rarely on the lorentz system, so we got away most of the time.

There are two more types of test failures which I do not know how to address.
1. I see several `permission denied` errors sporadically appearing on random packages which I did not touch. These are not triggered on the latest master CI run.
2. Some LTS CI runners does not pick up all changes correctly. I have changes to Core introducing new types, which are accessed on Extrapolation, LowOrderRK, FIRK, and ImplicitDiscreteSolve (which fail), as well as in LowStorageRK, and BDF (which are green). 

Regarding 2 I see the following difference in the CI.

* LowStorageRK resolves `[1dea7af3] OrdinaryDiffEq v6.105.0 ~/github-runners/deepsea4-17/_work/OrdinaryDiffEq.jl/OrdinaryDiffEq.jl`
* LowOrderRK resolves  `[1dea7af3] OrdinaryDiffEq v6.105.0 ~/work/OrdinaryDiffEq.jl/OrdinaryDiffEq.jl`

termi-official:
Before merging this, please wait until the Trixi group have a chance to check again that I really broke nothing (cc @ranocha )

termi-official:
Some of the Trixi tests catched fire. I need to investigate.

termi-official:
I left the TODOs to guide the breaking update for OrdinaryDiffEq v7, where we can potentially simplify the integrator a lot.

ChrisRackauckas:
Rebase, all tests should be good on master

ChrisRackauckas:
real test failures

termi-official:
If https://github.com/SciML/NonlinearSolve.jl/pull/794 does not introduce any further errors, then some of the real failing tests here should be fixed. Regarding the Downgrade CI I do not know how I triggered this issue. OrdinaryDiffEqCore is not found in the CI runner anymore:

```
  Installing known registries into `~/github-runners/deepsea4-28/.julia`
       Added `General` registry to ~/github-runners/deepsea4-28/.julia/registries
ERROR: LoadError: could not find manifest entry for package with uuid bbf590c4-e513-4bbe-9b18-05decba2e5d8
```

termi-official:
The lts failure is related to Julia 1.10 not using the sources and hence picking up the wrong package versions (latest release version vs latest commit). Downgrade CI fails on master, too, and I cannot figure out why. So, if no other CI fails, this should be good to go from my end.

ChrisRackauckas:
> The lts failure is related to Julia 1.10 not using the sources and hence picking up the wrong package versions (latest release version vs latest commit).

So it's a breaking change?

termi-official:
> > The lts failure is related to Julia 1.10 not using the sources and hence picking up the wrong package versions (latest release version vs latest commit).
> 
> So it's a breaking change?

No, this is purely related to how the CI is setup in Github. Right now, the CI relies on the sources feature which was introduced in Pkg 1.11 (https://github.com/JuliaLang/Pkg.jl/blob/master/CHANGELOG.md#pkg-v111-release-notes). The lts CI uses Julia 1.10, which ignores the section in the corresponding Project.toml files and therefore the runner picks up the latest release of the subpackages instead of the current commit, which then leads to the failure of
* (OrdinaryDiffEqFIRK, lts)
* (OrdinaryDiffEqExtrapolation, lts)
* (InterfaceI, lts)
* (ImplicitDiscreteSolve, lts)

The remaining lts tests do not even cover the PR, as the version of the latest commit is not picked up anyway.

The failure of (AD, 1.11) is real unfortunately, so let me fix that before merging.

termi-official:
The AD test has been fixed in [c6b065e](https://github.com/SciML/OrdinaryDiffEq.jl/pull/2899/commits/c6b065e520acc32b5c179d8868e7f5e4bfc984e2) . The remaining failures are the same oulined above.

ChrisRackauckas:
> No, this is purely related to how the CI is setup in Github. Right now, the CI relies on the sources feature which was introduced in Pkg 1.11 (https://github.com/JuliaLang/Pkg.jl/blob/master/CHANGELOG.md#pkg-v111-release-notes). The lts CI uses Julia 1.10, which ignores the section in the corresponding Project.toml files and therefore the runner picks up the latest release of the subpackages instead of the current commit, which then leads to the failure of

That is irrelevant, that just means it's testing the break. If the solver libraries require the new version of core then it's a breaking one? Is there no way to make this have a deprecation path?

termi-official:
Oh, I see. Thanks for pointing this out. Right now FIRK, IDS, and Extrapolation depend on the master of Core due to the newly introduced controller base types. Let me see, I should be able to resolve this with some compile-time tricks.

termi-official:
I start getting quite a few errors of the following kind in the CI setup:
```
Error: Unable to locate executable file: julia. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.
```

lib/OrdinaryDiffEqCore/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqCore"
 uuid = "bbf590c4-e513-4bbe-9b18-05decba2e5d8"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "3.2.0"
+version = "3.3.0"
```
termi-official, line 4: @ChrisRackauckas Just that this does not go unnoticed, I need to bump the version to make the CI work. See below for more details.

lib/OrdinaryDiffEqExtrapolation/src/OrdinaryDiffEqExtrapolation.jl
```diff
@@ -28,6 +29,17 @@ import OrdinaryDiffEqDifferentiation: TimeDerivativeWrapper, UDerivativeWrapper,
...
+@static if Base.pkgversion(OrdinaryDiffEqCore) >= v"3.3"
+    @eval begin
+        import OrdinaryDiffEqCore: AbstractLegacyController, default_controller_v7,
+            legacy_default_controller, setup_controller_cache, AbstractControllerCache
+    end
+else
+    @eval begin
+        import OrdinaryDiffEqCore: default_controller
+    end
+end
```
termi-official, line 41: This is the only sane way I found to make this PR truely non-breaking. I used this trick in all packages downstream of Core which are touched by this PR. For OrdinaryDiffEq v7 we can remove these static switches.

lib/OrdinaryDiffEqExtrapolation/src/alg_utils.jl
```diff
@@ -36,21 +36,84 @@ alg_maximum_order(alg::ImplicitHairerWannerExtrapolation) = 2(alg.max_order + 1)
...
+        # FIXME AitkenNeville is missing integration with the extrapolation controller and picks up the PI controller instead.
+        function default_controller(
+                alg::AitkenNeville,
+                cache,
+                qoldinit, _beta1 = nothing, _beta2 = nothing
+            )
+            QT = typeof(qoldinit)
+            beta1, beta2 = _digest_beta1_beta2(alg, cache, Val(QT), _beta1, _beta2)
+            return PIController(beta1, beta2)
+        end
```
termi-official, line 115: I have added this here to preserve the behavior of this algorithm as on master. However, if I am not mistaken Aitken-Neville should also use the extrapolation controller, right? Should I include the fix in this PR or should I preserve the behavior for now?

lib/OrdinaryDiffEqCore/Project.toml
ChrisRackauckas: That's kind of funky and doesn't seem to work so well... instead just make an OrdinaryDiffEqCore v4?

ChrisRackauckas:
Merged in https://github.com/SciML/OrdinaryDiffEq.jl/pull/3018

lib/OrdinaryDiffEqExtrapolation/src/alg_utils.jl
ChrisRackauckas: oh that should probably be an extraoplation controller. Thanks, yeah can you do a clean up PR? I was just trying to make sure everything got green and missed this.

termi-official: ~~Sure, will do!~~ Okay, not straight forward to fix for me, as I am not into the theory. See https://github.com/SciML/OrdinaryDiffEq.jl/issues/3029 .

Issue #2921, opened 2025-12-04
QNDF slowed down while rationalizing floats


hersle:
```julia
using OrdinaryDiffEqBDF, ProfileCanvas
f(du, u, p, t) = du .= 0.0
prob = ODEProblem(f, ones(10), (0.0, 1.0))
@profview sol = solve(prob, QNDF(); adaptive = false, dt = 1e-6, save_everystep = false)
```

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/63136cd2-ae8e-4b72-b7a2-c8998de26aee" />

As highlighted in the flamegraph, this spends a lot of time rationalizing a float here:

https://github.com/SciML/OrdinaryDiffEq.jl/blob/9e6080064f5b9d0b409d6b27643db718c9309349/lib/OrdinaryDiffEqBDF/src/bdf_perform_step.jl#L906

This is not supposed to happen? It does not happen with `FBDF`.

hersle:
Happens because `β₀` becomes a float here:

https://github.com/SciML/OrdinaryDiffEq.jl/blob/9e6080064f5b9d0b409d6b27643db718c9309349/lib/OrdinaryDiffEqBDF/src/bdf_perform_step.jl#L877

`β₀` becomes a float because `γₖ` are created as floats here:

https://github.com/SciML/OrdinaryDiffEq.jl/blob/9e6080064f5b9d0b409d6b27643db718c9309349/lib/OrdinaryDiffEqBDF/src/bdf_caches.jl#L462

I suspect `γₖ` should be rationals. But the `tTypeNoUnits` conversion turns them into floats. Removing `tTypeNoUnits` makes the whole issue go away.

PR #2922, opened 2025-12-04
Prevent rationalization in QNDF

hersle:
This fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2921 because `γₖ` now remain rational, and are not converted to floats due to the `tTypeNoUnits` conversion.

I don't know whether this is ok. Why do only BDF/NDF methods do this `tTypeNoUnits` conversion when constructing their cache? Can I also remove all the other `tTypeNoUnits` conversions in the same file?

### DIFF
--- a/lib/OrdinaryDiffEqBDF/src/bdf_caches.jl
+++ b/lib/OrdinaryDiffEqBDF/src/bdf_caches.jl
@@ -380,7 +380,7 @@ function alg_cache(alg::QNDF{MO}, u, rate_prototype, ::Type{uEltypeNoUnits},
     end
     U = SArray(U)
 
-    γₖ = SVector(ntuple(k -> sum(tTypeNoUnits(Int64(1) // j) for j in 1:k), Val(max_order)))
+    γₖ = SVector(ntuple(k -> sum(Int64(1) // j for j in 1:k), Val(max_order)))
 
     QNDFConstantCache(nlsolver, U, D, prevD, 1, 1, Val(max_order), dtprev, 0, 0, EEst1,
         EEst2, γₖ)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Why would this be faster? Multiplication by a rational number requires doing a division, which is slower than just a floating point multiplication. All rationals should just be turned into floats for the actual calculation

hersle:
The core of the problem is

https://github.com/SciML/OrdinaryDiffEq.jl/blob/9e6080064f5b9d0b409d6b27643db718c9309349/lib/OrdinaryDiffEqBDF/src/bdf_perform_step.jl#L906

Here `nlsolver.γ` is a rational, but `β₀` is a float. This triggers the expensive conversion and `rationalize` call in #2921. 

So we should either turn `nlsolver.γ` into a float, or `β₀` into a rational. This PR currently does the latter. What do think is best?

hersle:
It is faster:
```julia
using OrdinaryDiffEqBDF, ProfileCanvas, BenchmarkTools
f(du, u, p, t) = du .= 0.0
prob = ODEProblem(f, ones(10), (0.0, 1.0))
@btime sol = solve(prob, QNDF(); adaptive = false, dt = 1e-5, save_everystep = false)
```
```
  391.329 ms (100110 allocations: 4.59 MiB) # before this PR
  277.692 ms (100110 allocations: 4.59 MiB) # after this PR
```
But other solutions avoiding the `rationalize` would probably be as fast.

oscardssmith:
we probably should make `nlsolver.γ` a float (presumably of the same type as our tolerances)

ChrisRackauckas:
Yes the answer is to make `nlsolver.γ` match the tTypeNoUnits. Ultimately with it will always need to convert to that value in the end, so what you're doing now is just keeping it rational for just a little bit longer. I don't see that as numerically more accurate in any realistic sense with time being floating point, and if time is rational numbers then making it tTypeNoUnits will make the whole thing rational which is what you'd want in that case.

hersle:
Makes sense. Does it make sense to do it only for NDF in this PR, or would you only want to do it for all solvers at once?

ChrisRackauckas:
all at once.

hersle:
Understandable, closing this for now as I don't have the capacity to go at that now.

ChrisRackauckas:
I assume it would be pretty quick from what you have? I'll see if claude can just finish it.

hersle:
Ok, thanks. I thought you meant this has to be gone through for every single solver in all of OrdinaryDiffEq? Or did you mean just the BDF family ones?

ChrisRackauckas-Claude:
This has been addressed in #2923 which converts `nlsolver.γ` to `tTypeNoUnits` in `build_nlsolver`, following the approach suggested in the discussion (making nlsolver.γ match tTypeNoUnits for all solvers).

hersle:
Perfect.

PR #2931, opened 2025-12-12
Fix inference issue in Rosenbrock DAE test

ChrisRackauckas:
The alg_autodiff function was type-unstable because it used runtime conditionals (if/elseif/else) that could return different types:
- AutoForwardDiff() for Val{true}
- AutoFiniteDiff() for Val{false}
- The original autodiff value otherwise

This caused downstream type inference failures in build_jac_config and build_grad_config, which in turn caused the cache type and solution type to be uninferred in the integrator construction.

The fix uses dispatch-based helper functions (_unwrap_autodiff) instead of runtime conditionals, which allows Julia to resolve the return type at compile time based on the input type.

This is the first step toward fixing inference for Rosenbrock DAE tests.

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl
@@ -263,16 +263,32 @@ end
     (nothing, nothing)
 end
 
-# f.Wfact_t is nothing - check algorithm conditions and compute if needed
+# f.Wfact_t is nothing - dispatch on algorithm's concrete_jac and linsolve for type stability
 @inline function _build_jac_config_wfact(alg, f::F1, uf::F2, du1, uprev,
         u, tmp, du2, ::Nothing) where {F1, F2}
-    # Check algorithm conditions for whether we need concrete jacobian
-    haslinsolve = hasfield(typeof(alg), :linsolve)
-    needs_config = ((concrete_jac(alg) === nothing && (!haslinsolve || (haslinsolve &&
-           (alg.linsolve === nothing || LinearSolve.needs_concrete_A(alg.linsolve))))) ||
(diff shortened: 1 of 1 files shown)


lib/OrdinaryDiffEqDifferentiation/src/alg_utils.jl
```diff
@@ -16,16 +16,14 @@ function _alg_autodiff(alg::Union{OrdinaryDiffEqExponentialAlgorithm{CS, AD},
...
 function alg_autodiff(alg)
     autodiff = _alg_autodiff(alg)
-
-    if autodiff == Val(true)
-        return AutoForwardDiff()
-    elseif autodiff == Val(false)
-        return AutoFiniteDiff()
-    else
-        return autodiff
-    end
```
oscardssmith, line 28: alternatively, this could be
```
    if autodiff isa Val{true}
        return AutoForwardDiff()
    elseif autodiff isa Val{false}
        return AutoFiniteDiff()
    else
        return autodiff
    end
```

termi-official: Is there any true advantage of using branching over multiple dispatch for distinguishing executable code by types? I know this is very widely used in the SciML ecosystem, but this if-else by type design pattern causes a lot of trouble when trying to extend codes.

oscardssmith: no.

ChrisRackauckas:
accidentally merged and reverted

lib/OrdinaryDiffEqDifferentiation/src/alg_utils.jl
ChrisRackauckas: we can do splits to multiple dispatch, it's just a code structure thing.

PR #2934, opened 2025-12-14
Revert "Fix inference issue in Rosenbrock DAE test"

ChrisRackauckas:
Reverts SciML/OrdinaryDiffEq.jl#2931


oscardssmith:
why revert?

ChrisRackauckas:
This wasn't necessary for the test fixes I am working through. These changes could be fine in the future but it just wasn't related to the actual inference issue and I'm just keeping focused on that.

PR #2940, opened 2025-12-16
Test master

ChrisRackauckas:


ChrisRackauckas-Claude:
Closing this test PR. The actual fix is in #2943 - the root cause was outdated ODEProblemLibrary compat entries, not the force_latest_compatible_version flag.

PR #2942, opened 2025-12-16
Fix JET version resolution in CI tests

ChrisRackauckas-Claude:
## Summary

Fix JET test failures on Julia 1.12 by changing `force_latest_compatible_version=false` to `true` in `Pkg.test()` calls.

## Problem

CI tests on Julia 1.12 are failing with:
```
ERROR: LoadError: UndefVarError: `LineInfoNode` not defined in `JET`
@ ~/.julia/packages/JET/sABei/src/JETBase.jl:202
```

This occurs because:
1. JET v0.9.20 is incompatible with Julia 1.12 (has `LineInfoNode` error)
2. JET v0.11+ is required for Julia 1.12
3. CI runners have cached Manifest.toml files with JET v0.9.20 locked
4. `Pkg.test(; force_latest_compatible_version=false)` respects these old Manifest locks

## Solution

Change `force_latest_compatible_version=false` to `force_latest_compatible_version=true` in test/runtests.jl:43.

This ensures CI always uses the latest compatible versions of test dependencies, preventing stale Manifest locks from causing incompatibility errors.

## Testing

Local testing confirms:
- With fresh environment: `Pkg.test(; force_latest_compatible_version=false)` correctly selects JET v0.11.2
- With `force_latest_compatible_version=true`: Always uses latest compatible versions regardless of cached Manifests
- JET v0.11.2 installs and works correctly on Julia 1.12

## Related

- PR #2940 revealed this issue by testing master with Julia 1.12
- Compat entries are correct: `JET = "0.9, 0.11"` allows both versions
- The issue is purely with cached Manifest resolution, not package compatibility

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -40,7 +40,7 @@ end
 @time begin
     if contains(GROUP, "OrdinaryDiffEq") || GROUP == "ImplicitDiscreteSolve" || GROUP == "SimpleImplicitDiscreteSolve"
         Pkg.activate(joinpath(dirname(@__DIR__), "lib", GROUP))
-        Pkg.test(GROUP, julia_args=["--check-bounds=auto", "--compiled-modules=yes", "--depwarn=yes"], force_latest_compatible_version=false, allow_reresolve=true)
+        Pkg.test(GROUP, julia_args=["--check-bounds=auto", "--compiled-modules=yes", "--depwarn=yes"], force_latest_compatible_version=true, allow_reresolve=true)
     elseif GROUP == "All" || GROUP == "InterfaceI" || GROUP == "Interface"
         @time @safetestset "Discrete Algorithm Tests" include("interface/discrete_algorithm_test.jl")
         @time @safetestset "Tstops Tests" include("interface/ode_tstops_tests.jl")


ChrisRackauckas-Claude:
## Investigation Summary

### Root Cause Analysis

1. **Initial Error**: JET tests failing on Julia 1.12 with `UndefVarError: LineInfoNode not defined in JET`

2. **Version Mismatch**: 
   - CI was using JET v0.9.20 on Julia 1.12
   - JET v0.9.20 is incompatible with Julia 1.12
   - JET v0.11+ is required for Julia 1.12 (julia = "1.12" in JET v0.11's Project.toml)

3. **Why Wrong Version Selected**:
   - CI uses `Pkg.test(; force_latest_compatible_version=false)` in test/runtests.jl:43
   - CI runners have persistent `~/.julia/` directories with cached Manifest files
   - These Manifests had JET v0.9.20 locked (from before v0.11 was released)
   - `force_latest_compatible_version=false` respects these old Manifest locks

4. **Local Testing Confirmed**:
   - Fresh environment: `Pkg.test(; force_latest_compatible_version=false)` correctly selects JET v0.11.2
   - With old Manifest: Would select JET v0.9.20 (simulating CI behavior)
   - With `force_latest_compatible_version=true`: Always selects JET v0.11.2

### Fix

Changed `force_latest_compatible_version=false` to `true` in test/runtests.jl. This ensures:
- CI always uses latest compatible versions of test dependencies
- Stale Manifest locks don't cause incompatibility errors
- No need to manually clear CI caches or runner home directories

### Why This Is Safe

The compat entries are correct: `JET = "0.9, 0.11"` allows both versions. The issue was purely with cached Manifest resolution, not package compatibility. Using `force_latest_compatible_version=true` is generally preferred for CI as it tests against the latest compatible ecosystem state.

ChrisRackauckas-Claude:
Closing this PR - the fix causes widespread test failures. Using force_latest_compatible_version=true forces major version upgrades (CodeTracking v1.3→v3.0, JuliaSyntax v0.4→v1.0) that break other tests. The solution needs to be more nuanced than simply changing this flag.

PR #2943, opened 2025-12-16
Fix JET v0.11 compatibility by updating ODEProblemLibrary compat to v1.x

ChrisRackauckas-Claude:
## Summary

Fixes JET test failures on Julia 1.12 by updating `ODEProblemLibrary` compat from `0.1.8` to `1` across 15 packages.

## Root Cause

The outdated `ODEProblemLibrary = "0.1.8"` compat entry was locking to ODEProblemLibrary v0.1.12 (latest is v1.3.0), which has ancient compat constraints that cascade through the dependency tree:

```
ODEProblemLibrary v0.1.12 → old ModelingToolkit v10.x
                          → old Symbolics v6.x, SymbolicUtils v3.x  
                          → old CodeTracking v1.x, JuliaSyntax v0.4.x
                          → DataStructures DOWNGRADED v0.19.3 → v0.18.22
```

This caused **two critical failure modes on Julia 1.12:**

### Without `force_latest_compatible_version` (default):
- Pkg resolver chooses JET v0.9.20 (smaller dependency footprint)
- **JET v0.9.20 fails on Julia 1.12** with error: `LineInfoNode not defined`

### With `force_latest_compatible_version=true`:  
- Pkg resolver correctly chooses JET v0.11.2 ✓
- BUT still constrained by `ODEProblemLibrary = "0.1.8"`
- Old dependencies cause widespread test failures

## Solution

Update `ODEProblemLibrary` compat to `"1"` allowing v1.x versions, which:
- ✅ Enables JET v0.11 (required for Julia 1.12)
- ✅ Uses modern ModelingToolkit/Symbolics/CodeTracking/JuliaSyntax  
- ✅ Eliminates DataStructures downgrade
- ✅ Fixes all test failures

## Packages Updated (15 total)

- OrdinaryDiffEqAdamsBashforthMoulton
- OrdinaryDiffEqBDF
- OrdinaryDiffEqFeagin
- OrdinaryDiffEqFIRK
- OrdinaryDiffEqHighOrderRK
- OrdinaryDiffEqLowOrderRK
- OrdinaryDiffEqLowStorageRK
- OrdinaryDiffEqNonlinearSolve
- OrdinaryDiffEqNordsieck
- OrdinaryDiffEqQPRK
- OrdinaryDiffEqRosenbrock
- OrdinaryDiffEqSSPRK
- OrdinaryDiffEqStabilizedRK
- OrdinaryDiffEqTaylorSeries
- OrdinaryDiffEqVerner

## Testing

[...]

PR #2946, opened 2025-12-17
Os/cleanup jet simplification

oscardssmith:
NFC followup to https://github.com/SciML/OrdinaryDiffEq.jl/pull/2943 (rather than initializing variables that we will never use we can just simplify the logic slightly so it's more obvious what is needed when.

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/nlsolve.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/nlsolve.jl
@@ -38,9 +38,7 @@ function nlsolve!(nlsolver::NL, integrator::SciMLBase.DEIntegrator,
     nlsolver.status = check_div′ ? Divergence : Convergence
     η = get_new_W!(nlsolver) ? initial_η(nlsolver, integrator) : nlsolver.ηold
 
-    # Initialize ndz and ndzprev for JET
     ndz = one(η)
-    ndzprev = one(η)
     for iter in 1:maxiters
         if always_new && isnewton(nlsolver)
             if ArrayInterface.ismutable(integrator.u)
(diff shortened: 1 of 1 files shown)


lib/OrdinaryDiffEqFIRK/src/firk_addsteps.jl
```diff
@@ -76,7 +75,7 @@ function _ode_addsteps!(integrator, cache::RadauIIA3ConstantCache)
         dw2 = imag(dw12)
 
         # compute norm of residuals
-        iter > 1 && (ndwprev = ndw)
+        ndwprev = ndw
```
ChrisRackauckas, line 78: this doesn't seem right?

lib/OrdinaryDiffEqFIRK/src/firk_addsteps.jl
```diff
@@ -90,6 +89,7 @@ function _ode_addsteps!(integrator, cache::RadauIIA3ConstantCache)
             if diverge || veryslowconvergence
                 break
             end
+            η = θ / (1 - θ)
```
ChrisRackauckas, line 92: why is this added?

lib/OrdinaryDiffEqFIRK/src/firk_addsteps.jl
oscardssmith: `ndwprev` is only used if iter>1 so it's safe to set it on iter=1. It just won't do anything.

lib/OrdinaryDiffEqFIRK/src/firk_addsteps.jl
oscardssmith: to replace it's removal on line 103. `θ` is only defined if `iter>1` so we only want to update `η` in those cases.

PR #2948, opened 2025-12-19
Fix DPRKN5 CompiledFloats coefficients for proper 5th order convergence

### LINKED ISSUE #1938: Repeated occurrence of Runge-Kutta-Nyström issues concerning their accuracy
So far there where a few issues with the Runge-Kutta-Nyström solvers concerning their accuracy:

- https://github.com/SciML/DifferentialEquations.jl/issues/726
- https://github.com/SciML/DifferentialEquations.jl/issues/751
- https://github.com/SciML/OrdinaryDiffEq.jl/issues/1030
- https://github.com/SciML/OrdinaryDiffEq.jl/issues/1372

All Nyström solvers used, in these issues, are designed for 2nd order ODEs which are not dependent on the first derivative.
However most of the issues feature this type of 2nd order ODE. The lack of accuracy may just boil down to the solvers not being designed for theses types of problems. Any ideas on how to prevent these issues in the future?

ChrisRackauckas-Claude:
## Summary

- Fixed incorrect a51, a53, a54 coefficients in the CompiledFloats version of `DPRKN5ConstantCache`
- These coefficients were mistakenly set to the a61 row values (7/90, 4/15, 1/15) instead of the correct stage 5 coefficients (9/128, 9/64, 9/128)
- This bug caused DPRKN5 to achieve only ~2nd order convergence with Float64 instead of the expected 5th order

## Root Cause

The CompiledFloats version (used for Float64) had a copy-paste error where the a61 coefficients were used for the a51 row. The rational version (used for BigFloat) was correct, which is why tests using BigFloat passed while Float64 users experienced degraded convergence.

The bug was verified by checking the RKN row sum condition:
- For c4 = 3/4, the row sum should be c4²/2 = 9/32 = 0.28125
- The incorrect coefficients gave: 7/90 + 4/15 + 1/15 = 37/90 ≈ 0.411 (**wrong**)
- The correct coefficients give: 9/128 + 9/64 + 9/128 = 9/32 = 0.28125 (**correct**)

## Verification

Convergence testing before and after the fix:

**Before fix (Float64):**
```
DPRKN5: ~2nd order convergence
```

**After fix (Float64):**
```
DPRKN5: ~5th order convergence (matches BigFloat behavior)
```

## Test plan

- [x] Verified row sum condition mathematically
- [x] Confirmed 5th order convergence with Float64 after fix
- [x] Confirmed both in-place and out-of-place versions achieve correct order
- [ ] CI tests pass (note: some pre-existing flaky tests in adaptive stepping may fail)

Fixes #1938

### DIFF
--- a/lib/OrdinaryDiffEqRKN/test/nystrom_convergence_tests.jl
+++ b/lib/OrdinaryDiffEqRKN/test/nystrom_convergence_tests.jl
@@ -47,6 +47,36 @@ sim = test_convergence(dts, prob_big, ERKN7(), dense_errors = true)
 @test sim.𝒪est[:l2]≈7 rtol=1e-1
 @test sim.𝒪est[:L2]≈4 rtol=1e-1
 
+# Float64 convergence tests for DPRKN methods
+# These tests ensure the CompiledFloats coefficients match the rational coefficients
+# (Regression test for https://github.com/SciML/OrdinaryDiffEq.jl/issues/1938)
+@testset "Float64 DPRKN convergence" begin
+    # Use longer integration time to keep errors above machine precision
+    prob_f64 = DynamicalODEProblem(ff_harmonic, ones(2), fill(0.0, 2), (0.0, 20.0))
+    dts_f64 = 1.0 ./ 2.0 .^ (2:5)
+
+    sim = test_convergence(dts_f64, prob_f64, DPRKN4())
+    @test sim.𝒪est[:l2]≈4 rtol=1e-1
+
+    sim = test_convergence(dts_f64, prob_f64, DPRKN5())
+    @test sim.𝒪est[:l2]≈5 rtol=1e-1
+
+    sim = test_convergence(dts_f64, prob_f64, DPRKN6())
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Added Float64 convergence tests for DPRKN methods to catch this regression in the future.

The new test suite (`@testset "Float64 DPRKN convergence"`) specifically tests the CompiledFloats code path (Float64) rather than just BigFloat:

- **DPRKN4**: Tests 4th order convergence with Float64
- **DPRKN5**: Tests 5th order convergence with Float64 (**key regression test**)
- **DPRKN6**: Tests 6th order convergence with Float64
- **DPRKN6FM**: Tests 6th order convergence with Float64
- **DPRKN8**: Tests 8th order convergence with Float64 (uses larger timesteps)
- **DPRKN12**: Skipped for Float64 (too accurate - hits machine precision even at dt=1.0)

Before this fix, the DPRKN5 Float64 test would have failed (showing ~2nd order instead of 5th order), while the existing BigFloat tests would still pass.

PR #2949, opened 2025-12-27
Optimize sparse mass matrix algebraic detection from O(n²) to O(nnz)

cuihantao:
Add `find_algebraic_vars_eqs` function that efficiently detects algebraic variables and equations from sparse mass matrices by traversing the CSC structure directly, reducing complexity from O(n²) to O(nnz).

For a 259k x 259k sparse diagonal mass matrix, this reduces DAE initialization time from ~67 seconds to <1 second.

Changes:
- Add SparseArrays dependency for SparseMatrixCSC type dispatch
- Add find_algebraic_vars_eqs with sparse specialization and dense fallback
- Replace 4 instances of O(n^2) column/row iteration in _initialize_dae!
- Add performance regression tests verifying O(nnz) complexity

## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [X] Any new documentation only uses public API

PR #2950, opened 2025-12-27
Add MatrixOperator support to find_algebraic_vars_eqs

ChrisRackauckas-Claude:
## Summary
- Add method dispatch for `AbstractSciMLOperator` that extracts the underlying matrix (`M.A`) to handle `MatrixOperator` mass matrices

This fixes the test failure from PR #2949 where `MatrixOperator` was not handled by the new `find_algebraic_vars_eqs` function:

```
MethodError: no method matching find_algebraic_vars_eqs(::SciMLOperators.MatrixOperator{Float64, Matrix{Float64}, ...})
```

Follow-up to https://github.com/SciML/OrdinaryDiffEq.jl/pull/2949#issuecomment-2563108971

## Test plan
- [ ] CI passes, including the Mass Matrix Tests that were failing

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/initialize_dae.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/initialize_dae.jl
@@ -36,6 +36,11 @@ function find_algebraic_vars_eqs(M::AbstractMatrix)
     return algebraic_vars, algebraic_eqs
 end
 
+# Handle SciMLOperators (e.g., MatrixOperator) by extracting the underlying matrix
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Updated to use `convert(AbstractMatrix, M)` instead of `M.A` - this is more robust:

- ✅ Works for `MatrixOperator` (sparse and dense) 
- ✅ Works for `IdentityOperator` (returns all false - no algebraic vars)
- ✅ Works for `NullOperator` (returns all true - all algebraic vars)
- ✅ Throws clear error for operators that can't be converted to matrix

This matches the existing pattern in `derivative_utils.jl:279-280`.

PR #2965, opened 2026-01-01
Documentation improvements: Fix function naming and API usage

ChrisRackauckas-Claude:
## Summary

This PR improves the documentation quality with focused corrections to code examples:

- Fixed function naming conventions to follow Julia's in-place mutation convention (`lorenz!`, `HH_acceleration!`)
- Updated deprecated `vars` keyword to `idxs` in plotting examples
- Improved `docs/make.jl` robustness to handle execution from both repo root and docs directory

### docs/src/usage.md
1. **Function naming conventions**: Changed `lorenz` → `lorenz!` and `HH_acceleration` → `HH_acceleration!` to follow Julia convention for in-place functions that mutate their first argument
2. **API updates**: Changed deprecated `plot(sol, vars = (1, 2, 3))` to `plot(sol, idxs = (1, 2, 3))` to use current Plots.jl API
3. **Consistency**: Updated all corresponding function calls in `ODEProblem` and `SecondOrderODEProblem`

### docs/make.jl
- Added path detection logic to handle running from both repository root and docs directory
- Prevents "no such file or directory" errors during documentation builds

## Testing

- Documentation builds successfully without new warnings
- Examples follow current best practices and API conventions

---

cc: @ChrisRackauckas

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,13 +1,7 @@
 using Documenter, OrdinaryDiffEq
 
-# Handle both running from repo root and from docs directory
-manifest_src = isfile("./docs/Manifest.toml") ? "./docs/Manifest.toml" : "Manifest.toml"
-project_src = isfile("./docs/Project.toml") ? "./docs/Project.toml" : "Project.toml"
-manifest_dst = isfile("./docs/src/assets/Manifest.toml") || isdir("./docs/src/assets") ? "./docs/src/assets/Manifest.toml" : "src/assets/Manifest.toml"
-project_dst = isfile("./docs/src/assets/Project.toml") || isdir("./docs/src/assets") ? "./docs/src/assets/Project.toml" : "src/assets/Project.toml"
-
-cp(manifest_src, manifest_dst, force = true)
-cp(project_src, project_dst, force = true)
+cp(joinpath(@__DIR__, "Manifest.toml"), joinpath(@__DIR__, "src", "assets", "Manifest.toml"), force = true)
+cp(joinpath(@__DIR__, "Project.toml"), joinpath(@__DIR__, "src", "assets", "Project.toml"), force = true)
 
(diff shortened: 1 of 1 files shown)


docs/make.jl
```diff
@@ -1,7 +1,13 @@
 using Documenter, OrdinaryDiffEq
 
-cp("./docs/Manifest.toml", "./docs/src/assets/Manifest.toml", force = true)
-cp("./docs/Project.toml", "./docs/src/assets/Project.toml", force = true)
+# Handle both running from repo root and from docs directory
+manifest_src = isfile("./docs/Manifest.toml") ? "./docs/Manifest.toml" : "Manifest.toml"
+project_src = isfile("./docs/Project.toml") ? "./docs/Project.toml" : "Project.toml"
+manifest_dst = isfile("./docs/src/assets/Manifest.toml") || isdir("./docs/src/assets") ? "./docs/src/assets/Manifest.toml" : "src/assets/Manifest.toml"
+project_dst = isfile("./docs/src/assets/Project.toml") || isdir("./docs/src/assets") ? "./docs/src/assets/Project.toml" : "src/assets/Project.toml"
```
ranocha: Wouldn't it be better to use `@__DIR__`, e.g., `joinpath(@__DIR__, "Manifest.toml")`?

ChrisRackauckas: It would yes, just did this change in another repo. I'll add it to its prompt because I think the `cp("./docs/Manifest.toml", "./docs/src/assets/Manifest.toml", force = true)` is in every repo so as this goes around it should do the same change everywhere.

ChrisRackauckas-Claude: Good suggestion! Updated to use `joinpath(@__DIR__, ...)` which is much cleaner and more robust. Thanks for the review!

PR #2999, opened 2026-01-16
Add Reactant.jl support for OrdinaryDiffEqCore

ChrisRackauckas-Claude:
## Summary

This PR adds Reactant.jl/ReactantCore.jl support to OrdinaryDiffEqCore, enabling JIT compilation of ODE solves via MLIR/XLA.

### Key Changes

**OrdinaryDiffEqCoreReactantExt.jl** (~476 lines):
- Rational → Float64 conversion dispatches for `TracedRNumber`/`TracedRArray` operations (MLIR can't handle `Rational{Int64}`)
- `reactant_safe_copy()` and modified `ode_copyat_or_push!` for traced arrays
- `@reactant_overlay` for `qmin_default`, `beta2_default`, `gamma_default`, etc. to return Float64 instead of Rationals
- `@reactant_overlay` for `stepsize_controller!`, `step_accept_controller!`, `calc_dt_propose!`, `_loopfooter!` using `ifelse` instead of `if` for traced booleans
- Unrolled step functions (`reactant_steps_1!` through `reactant_steps_100!`) to avoid `@trace for` type tracing issues
- `@reactant_overlay` for `solve!` that computes step count at init time and uses unrolled step functions

**Working Pattern:**
```julia
using Reactant
using Reactant: @compile
using OrdinaryDiffEq

function f!(du, u, p, t)
    du .= -u
    return nothing
end

const tspan = (0.0, 1.0)
const dt = 0.1

function solve_ode(u0)
    prob = ODEProblem(f!, u0, tspan)
    sol = solve(prob, Tsit5(); save_everystep=false, adaptive=false, dt=dt)
    return copy(sol.u[end])
end

u0_reactant = Reactant.to_rarray([1.0, 2.0])
compiled_solve = @compile solve_ode(u0_reactant)
result = compiled_solve(u0_reactant)  # Matches Julia result
```

### Current Limitations

- Must use `save_everystep=false`
- Must use `adaptive=false` with fixed `dt`
- Must return `copy(sol.u[end])`, not the full solution object
- `saveat` case fails with "No MLIR context is active" error (may require Reactant.jl changes)

### Technical Notes

The main challenges solved:
1. **Rational{Int64} conversion**: MLIR can't represent Julia `Rational` types, so dispatch-based conversions intercept operations
2. **TracedRNumber{Bool} control flow**: Regular `if` statements can't use traced booleans; solved via `ifelse` and simplified overlays
3. **@trace for type issues**: The `@trace for` macro causes type tracing issues with complex integrator structs; solved via unrolled step functions

## Test plan

- [x] Verified MWE produces matching results between Julia and Reactant compilation
- [x] Basic exponential decay ODE `du/dt = -u` works with Tsit5
- [ ] Run full test suite
- [ ] Test with other explicit RK methods


test/enzyme/Project.toml
```diff
@@ -1,14 +1,21 @@
 [deps]
+DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
+OrdinaryDiffEqTsit5 = "b1df2697-797e-41e3-8120-5422d3b24e4a"
 Reactant = "3c362404-f566-11ee-1572-e11a4b42c853"
 SciMLSensitivity = "1ed8b502-d754-442c-8d5d-10ac956f44a1"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
oscardssmith, line 10: broken

ChrisRackauckas: ?

ChrisRackauckas-Claude:
## Update: Reactant Compatibility Progress

### Commits Added
1. **Convert Rational literals to Float64** - Reactant.jl cannot convert `Rational{Int64}` to traced types. This commit converts all Rational literals (`1 // 10^n`) to Float64 (`1e-n`) in:
   - `solve.jl`: qoldinit, abstol/reltol defaults
   - `alg_utils.jl`: qmin, gamma, beta1, beta2, qsteady_max defaults
   - `initdt.jl`: smalldt, d₀/d₁ thresholds
   - `composite_algs.jl`: nonstifftol/stifftol defaults

### Current Status
- ✅ ReactantCore dependency added
- ✅ `@trace` macro imported
- ✅ Rational-to-Float64 conversion for type compatibility
- ✅ Reactant tests added to enzyme test group
- ⏳ Further work needed: Dynamic array operations in solution storage

### Remaining Issue
After fixing the Rational conversion, Reactant tracing hits a deeper compatibility issue with `RecursiveArrayTools.copyat_or_push!` which tries to copy vectors with uninitialized (`undef`) elements. The ODE solver pre-allocates solution arrays with `undef` and fills them during the solve, but Reactant's tracing cannot iterate over such arrays.

This may require:
1. A Reactant extension for RecursiveArrayTools
2. Pre-allocating solution arrays with initialized values
3. Or `@trace` annotations around specific saving operations

The `@trace` macro has restrictions (no return/break/nested control flow inside) that make it challenging to apply to the existing solver structure without significant refactoring.

test/enzyme/Project.toml
oscardssmith: test is getting added as a dep.

ChrisRackauckas-Claude:
## Update: Complete Reactant Extension Implementation

This push completes the Reactant.jl support with the following key additions:

### What's New

1. **Unrolled step functions** (`reactant_steps_1!` through `reactant_steps_100!`):
   - Avoids `@trace for` which caused type tracing issues with complex integrator structs
   - Steps are dispatched via `reactant_run_steps!` based on step count

2. **`@reactant_overlay` for `solve!`**:
   - Computes number of integration steps at init time (not traced)
   - Uses unrolled step functions instead of while loop
   - Pre-allocates solution arrays when possible

3. **Simplified control flow overlays**:
   - `_loopfooter!` uses `ifelse` instead of `if` for traced booleans
   - `calc_dt_propose!` skips adaptive dt update for Reactant (dt is fixed anyway)

### Verified Working

The MWE was tested and shows Julia and Reactant results match:
```
Julia result: [0.3678794414272495, 0.735758882854499]
Reactant result: [0.3678794414272495, 0.735758882854499]
Max difference: 0.0
Results match: true
```

### Next Steps

- Full test suite needs to run
- `saveat` support requires changes in Reactant.jl to handle solution array mutations during tracing

ChrisRackauckas:
@avik-pal it seems Reactant is unable to handle the nested if statements in `@trace` and is unable to handle the variable length vector of vector output

PR #3011, opened 2026-01-21
Optimize sparse isdiag check from O(n²) to O(nnz) in get_differential_vars

### LINKED ISSUE #3010: O(n²) isdiag() bottleneck in get_differential_vars for sparse mass matrices
**Describe the bug 🐞**

The get_differential_vars function in OrdinaryDiffEqCore/src/misc_utils.jl calls isdiag(mm) which is O(n²) for sparse matrices. For large DAE systems (e.g., 259k variables), this takes ~60 seconds instead of <1ms.

This is similar to the issue fixed in #2949 for `find_algebraic_vars_eqs`.

```
function get_differential_vars(f, u)
    if hasproperty(f, :mass_matrix)
        mm = f.mass_matrix
        mm = mm isa MatrixOperator ? mm.A : mm

        if mm isa UniformScaling
            return nothing
        elseif all(!iszero, mm)
            return trues(size(mm, 1))
        elseif !(mm isa SciMLOperators.AbstractSciMLOperator) && isdiag(mm)  # <-- O(n²) here
            return reshape(diag(mm) .!= 0, size(u))
        else
            return DifferentialVarsUndefined()
        end
    else
        return nothing
    end
end
```

Reproducer

```
using SparseArrays, LinearAlgebra, BenchmarkTools

# Simulate a large DAE mass matrix (diagonal with some zeros)
n = 100_000
mm = spdiagm(0 => vcat(ones(n ÷ 2), zeros(n - n ÷ 2)))

# Standard isdiag - O(n²)
@btime isdiag($mm)  # ~10 seconds for n=100k, ~60 seconds for n=259k

# Sparse-aware version - O(nnz)
function isdiag_sparse(A::SparseMatrixCSC)
    m, n = size(A)
    m != n && return false
    for j in 1:n
        for k in A.colptr[j]:(A.colptr[j+1]-1)
            A.rowval[k] != j && return false
        end
    end
    return true
end

@btime isdiag_sparse($mm)  # ~0.07 ms
```

I'm not sure how to fix it properly because the package does not have `SparseArrays` in the deps, and I'm not sure how to make it compatible with the Operator. 

## Environmen 

Julia 1.11
OrdinaryDiffEqCore v3.2.0

ChrisRackauckas-Claude:
## Summary

Add SparseArrays extension that provides efficient O(nnz) `_isdiag` check for sparse matrices by traversing the CSC structure directly. Standard `isdiag` is O(n²) which is prohibitively slow for large sparse matrices.

For large DAE systems (e.g., 259k variables), this reduces initialization time from ~60 seconds to <1ms.

**Changes:**
- Add SparseArrays as a weak dependency
- Add `OrdinaryDiffEqCoreSparseArraysExt` extension with `_isdiag(::SparseMatrixCSC)`
- Add `_isdiag` dispatch function with fallback to `LinearAlgebra.isdiag`
- Update `get_differential_vars` to use `_isdiag` instead of `isdiag`
- Add performance regression tests verifying O(nnz) complexity

This is similar to the fix in #2949 for `find_algebraic_vars_eqs`.

Fixes #3010

## Test plan

- [x] Run OrdinaryDiffEqCore tests locally - all pass
- [x] Extension loads correctly when SparseArrays is loaded
- [x] Added `sparse_isdiag_tests.jl` with:
  - Correctness tests for diagonal/non-diagonal sparse matrices
  - Complexity verification (O(nnz) vs O(n²))
  - Dense matrix fallback tests

### DIFF
--- a/lib/OrdinaryDiffEqCore/Project.toml
+++ b/lib/OrdinaryDiffEqCore/Project.toml
@@ -73,6 +73,7 @@ SciMLBase = "2.116"
 SciMLLogging = "1.7.1"
 FastClosures = "0.3"
 DataStructures = "0.19"
+SparseArrays = "1"
 Static = "1.2"
 Aqua = "0.8.11"
 DocStringExtensions = "0.9"
@@ -97,10 +98,12 @@ Reexport = "1.2"
 [weakdeps]
 Mooncake = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [targets]
-test = ["DiffEqDevTools", "Random", "SafeTestsets", "Test", "Pkg"]
+test = ["DiffEqDevTools", "Random", "SafeTestsets", "SparseArrays", "Test", "Pkg"]
 
 [extensions]
 OrdinaryDiffEqCoreEnzymeCoreExt = "EnzymeCore"
 OrdinaryDiffEqCoreMooncakeExt = "Mooncake"
+OrdinaryDiffEqCoreSparseArraysExt = "SparseArrays"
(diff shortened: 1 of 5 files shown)


JoshuaLampert:
I think this PR is responsible for #3025. Does SparseArrays.jl need to be a weak dependency? It's a stdlib. So why not moving the code from the extension to the main package and making SparseArrays.jl a regular dependency?

ChrisRackauckas:
https://github.com/JuliaSparse/SparseArrays.jl/pull/672 moved upstream

PR #2742, opened 2025-06-22
Mooncake Direct Adjoints for Explicit Integrators

ChrisRackauckas:
MWE:

```julia
using Mooncake, OrdinaryDiffEq, StaticArrays

function lorenz!(du, u, p, t)
    du[1] = 10.0(u[2] - u[1])
    du[2] = u[1] * (28.0 - u[3]) - u[2]
    du[3] = u[1] * u[2] - (8 / 3) * u[3]
end

const _saveat =  SA[0.0,0.25,0.5,0.75,1.0,1.25,1.5,1.75,2.0,2.25,2.5,2.75,3.0]

function f(u0::Array{Float64})
    tspan = (0.0, 3.0)
    prob = ODEProblem{true, SciMLBase.FullSpecialize}(lorenz!, u0, tspan)
    sol = DiffEqBase.solve(prob, Tsit5(), saveat = _saveat, sensealg = DiffEqBase.SensitivityADPassThrough())
    sum(sol)
end;
u0 = [1.0; 0.0; 0.0]
mooncake_gradient(f, x) = Mooncake.value_and_gradient!!(Mooncake.build_rrule(f, x), f, x)[2][2]
mooncake_gradient(f, u0)
```

Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2723

Currently fails with:

```
julia> mooncake_gradient(f, u0)
ERROR: MooncakeRuleCompilationError: an error occured while Mooncake was compiling a rule to differentiate something. If the `caused by` error message below does not make it clear to you how the problem can be fixed, please open an issue at github.com/chalk-lab/Mooncake.jl describing your problem.
To replicate this error run the following:

[...]

PR #3012, opened 2026-01-22
Call `final_progress`

visr:
To make sure the progress bar reaches 100%.

Came up in https://github.com/Deltares/Ribasim/issues/2663

Looks like this slipped in in https://github.com/SciML/OrdinaryDiffEq.jl/pull/2742/changes#diff-fc30bb14472d1ad90bcef722e155cd494891c98618f299c6a690be2a262eb7a9

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
@@ -227,8 +227,10 @@ function _postamble!(integrator)
     if !(integrator.sol isa DAESolution)
         resize!(integrator.sol.k, integrator.saveiter_dense)
     end
-    return if integrator.opts.progress
+    if integrator.opts.progress
+        final_progress(integrator)
     end
+    return nothing
 end
 
 function final_progress(integrator)


ChrisRackauckas-Claude:
✅ Verified that `final_progress` is already marked as inactive for AD in both extensions on master:

**Enzyme** (`OrdinaryDiffEqCoreEnzymeCoreExt.jl`):
```julia
function EnzymeCore.EnzymeRules.inactive_noinl(
        ::typeof(OrdinaryDiffEqCore.final_progress), args...
    )
    return true
end
```

**Mooncake** (`OrdinaryDiffEqCoreMooncakeExt.jl`):
```julia
Mooncake.@zero_adjoint Mooncake.MinimalCtx Tuple{
    typeof(OrdinaryDiffEqCore.final_progress), Vararg,
}
```

These rules were added in 455fa89. The PR is safe for AD differentiation.

ChrisRackauckas-Claude:
The `alldeps` test failures are **pre-existing infrastructure issues** with the `DowngradeSublibraries` workflow, not caused by this PR:

```
Recent DowngradeSublibraries runs:
- Run #940 (master): failure
- Run #939 (fix-isdiag-sparse-performance): failure
- Run #938 (fix-isdiag-sparse-performance): failure
- Run #937 (do/controller-refactor): failure
```

The error `could not find manifest entry for package with uuid bbf590c4-e513-4bbe-9b18-05decba2e5d8` (OrdinaryDiffEqCore) appears to be an issue with the `julia-downgrade-compat` action not handling local packages properly.

This PR only adds a call to the existing `final_progress` function in `_postamble!` and cannot affect package resolution.

PR #3024, opened 2026-02-02
Remove backwards compatibility version gates for controller refactor

ChrisRackauckas-Claude:
## Summary
- Removes all `@static if Base.pkgversion(OrdinaryDiffEqCore) >= v"3.4"` version gates from 6 subpackages (13 files)
- Keeps only the new controller API code (`default_controller_v7`, `legacy_default_controller`, controller caches)
- Removes old `default_controller` fallback code paths that were only needed for Core < 3.4

**Depends on #3023 being merged first** (which bumps subpackage compat bounds to require Core >= 3.4).

## Files changed (13)
- `ImplicitDiscreteSolve`: module imports + controller.jl
- `OrdinaryDiffEqBDF`: module imports + controllers.jl
- `OrdinaryDiffEqExtrapolation`: module imports + controllers.jl + alg_utils.jl
- `OrdinaryDiffEqFIRK`: module imports + controllers.jl
- `OrdinaryDiffEqLowStorageRK`: module imports + alg_utils.jl
- `OrdinaryDiffEqNordsieck`: module imports + alg_utils.jl

## Test plan
- [ ] CI passes for all affected subpackages
- [ ] Verify no remaining `v"3.3"` or `v"3.4"` version gates in lib/

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/alg_utils.jl
+++ b/lib/OrdinaryDiffEqCore/src/alg_utils.jl
@@ -417,8 +417,6 @@ function legacy_default_controller(alg, cache, qoldinit, _beta1 = nothing, _beta
     end
 end
 
-# TODO remove this when done
-default_controller(args...) = legacy_default_controller(args...)
 
 function _digest_beta1_beta2(alg, cache, ::Val{QT}, _beta1, _beta2) where {QT}
     if alg isa OrdinaryDiffEqCompositeAlgorithm


oscardssmith:
@termi-official can you review?

lib/OrdinaryDiffEqCore/src/alg_utils.jl
```diff
@@ -417,8 +417,6 @@ function legacy_default_controller(alg, cache, qoldinit, _beta1 = nothing, _beta
     end
 end
 
-# TODO remove this when done
-default_controller(args...) = legacy_default_controller(args...)
```
termi-official, line 421: Looks like the bot messed up here. I think removing this without major can potentially break some downstream stuff, as the new interface exposes the controller in a different way. This was designed as the final switch for the OrdinaryDiffEq v7 changes, so we can simply go for 
```
default_controller(args...) = default_controller_v7(args...)
```
first, then remove all the legacy stuff bit by bit and finally remove the `_v7` functions together with a rename of the controllers and moving them into the algorithms (instead of as an init kwarg). Does that make sense?

termi-official: I want to highlight again, to make sure this is properly communicated, that removing this will require downstream packages to also adapt to the new interface first/together.

ChrisRackauckas:
This requires a major to merge

termi-official:
Should we keep the gates until v7 is released then?

ChrisRackauckas:
Yes

PR #3038, opened 2026-02-07
Redo new controller interface without full backward compatibility

termi-official:
This PR removes the legacy controller parameters from the algs, integrator, opts and the init logic.

~~The next step would be to move the controller_cache from ODEIntegrator to the respective algorithm caches, such that the algorithm becomes responsible to setup the controller cache with some utils (i.e. a `setup_controller_cache` for the "standard" controllers). Here the P,PI,... controller caches are the reusable building blocks, which allows a cleaner implementation of algorithms with custom control logic.~~

Closes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2883 .

### DIFF
--- a/test/interface/controllers.jl
+++ b/test/interface/controllers.jl
@@ -28,8 +28,8 @@ sol = solve(prob, RDPK3SpFSAL49())
     # Exponential decay - the auto dt init may pick a very conservative initial dt
     prob_exp = ODEProblem((u, p, t) -> -u, 1.0, (0.0, 100.0))
 
-    # Custom qmax_first_step via New* controller API
-    ctrl = OrdinaryDiffEqCore.NewPIController(Tsit5(), qmax_first_step = 500)
+    # Custom qmax_first_step via controller API
+    ctrl = OrdinaryDiffEqCore.PIController(Tsit5(), qmax_first_step = 500)
     @test ctrl.qmax_first_step == 500.0
 
     # Verify that qmax_first_step=10000 still produces correct solutions
@@ -40,7 +40,7 @@ sol = solve(prob, RDPK3SpFSAL49())
 
         # With a very restrictive qmax_first_step (same as normal qmax),
         # the solution should still be correct
-        ctrl_restricted = OrdinaryDiffEqCore.NewPIController(alg, qmax_first_step = 10)
+        ctrl_restricted = OrdinaryDiffEqCore.PIController(alg, qmax_first_step = 10)
         sol_restricted = solve(prob_exp, alg, controller = ctrl_restricted)
         @test sol_restricted.retcode == ReturnCode.Success
         @test isapprox(sol_restricted[end], exp(-100.0), atol = 1.0e-6)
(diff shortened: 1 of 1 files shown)


test/ad/autodiff_events.jl
```diff
@@ -126,7 +126,7 @@ end
     @test g2 ≈ findiff[2, 1:2]
     @test g3 ≈ findiff[2, 1:2]
     @test g4 ≈ findiff[2, 1:2]
-    @test_broken g5 ≈ findiff[2, 1:2]
+    g5 ≈ findiff[2, 1:2]
```
termi-official: I think the original issue for g7 below is also fixed by this PR. The new error for g7 is
```
MethodError: convert(::Type{ForwardDiff.Dual{ForwardDiff.Tag{…}, ReverseDiff.TrackedReal{…}, 2}}, ::ReverseDiff.TrackedReal{Float64, Float64, Nothing}) is ambiguous.
```

test/regression/inference.jl
```diff
@@ -1,19 +1,47 @@
...
+    @testset "stiff fixed chunksize" begin
+        # When choosing a fixed chunksize it works
+        autodiff = ADTypes.AutoForwardDiff(; chunksize = 10)
+        inferred3 = [SDIRK2(; autodiff), TRBDF2(; autodiff), KenCarp4(; autodiff), Rosenbrock23(; autodiff), Rodas4(; autodiff)]
+        for alg in inferred3
+            @inferred init(prob, alg)
+            # In Julia < v1.12 only some of these are inferable
+            if VERSION >= v"1.12"
+                @inferred init(prob2D, alg)
+            end
```
termi-official, line 34: @oscardssmith this might be interesting for you. On Julia 1.11 and 1.10 the inference of the chunksize fails for some of the algorithms.

lib/OrdinaryDiffEqCore/src/integrators/type.jl
termi-official, line 1: Is in `DEOptions` or `ODEIntegrator` anything left which should either go into a controller (cache) or algorithm (cache)?

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -139,53 +132,30 @@ the predicted step size.
...
+    return IController{typeof(qmin)}(
+        qmin,
+        qmax,
+        gamma,
+        qsteady_min,
+        qsteady_max,
+    )
+end
+
+function IController(alg; kwargs...)
```
termi-official, line 183: This and the next functions are meant to be internal for now. Should we keep this dispatch (and the one below)?

lib/OrdinaryDiffEqCore/src/alg_utils.jl
```diff
@@ -369,57 +369,22 @@ alg_adaptive_order(alg::Union{OrdinaryDiffEqAlgorithm, DAEAlgorithm}) = alg_orde
 # this is actually incorrect and is purposefully decreased as this tends
 # to track the real error much better
 
-function default_controller_v7(QT, alg)
+function default_controller(QT, alg)
```
termi-official, line 405: We are at a point where we coul retire `ispredictive`, `isstandard` and friends and should just directly dispatch the algorithms on `default_controller`. I do not plan to do this in the PR tho.

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -395,19 +319,15 @@ end
     if iszero(EEst)
         q = inv(qmax)
     else
-        # Legacy code
         q11 = fastpower(EEst, beta1)
         q = q11 / fastpower(errold, beta2)
         cache.q11 = q11
-        integrator.q11 = q11 # TODO remove
         @fastmath q = clamp(q / gamma, inv(qmax), inv(qmin))
```
termi-official, line 361: Reformulating this without divisions is still an open TODO and will be breaking -- if we want to do it.

ChrisRackauckas:
Note that you'll need to rebase https://github.com/SciML/OrdinaryDiffEq.jl/pull/3071 into this, but I tried to make it match the new syntax not the old.

termi-official:
> Note that you'll need to rebase https://github.com/SciML/OrdinaryDiffEq.jl/pull/3071 into this, but I tried to make it match the new syntax not the old.

Thanks Chris! The PR should be on track again.

termi-official:
Retargeted to the v7 branch in https://github.com/SciML/OrdinaryDiffEq.jl/pull/3312 .

PR #3054, opened 2026-02-20
Replace applicable() with Ref{Bool} flag to eliminate _initialize_dae! invalidation

ChrisRackauckas-Claude:
## Summary

- Replace `applicable()` calls in `_initialize_dae!` (DefaultInit dispatch for ODEProblem and DAEProblem) with a `Ref{Bool}` flag (`_nonlinearsolve_loaded`) set by OrdinaryDiffEqNonlinearSolve's `__init__()`
- Wrap `BrownFullBasicInit`/`ShampineCollocationInit` dispatch calls in `Base.invokelatest()` to prevent the compiler from creating method table backedges

## Problem

The `applicable()` calls in `_initialize_dae!` created "negative" method table backedges: the compiler recorded "if any method matching `_initialize_dae!(integrator, prob, BrownFullBasicInit(...), Val{true})` is defined, invalidate me." When OrdinaryDiffEqNonlinearSolve loaded and defined that exact method, the invalidation cascaded through:

```
_initialize_dae! → initialize_dae! → reeval_internals_due_to_modification!
→ handle_tstop! → loopheader! → solve! → __init → __solve → solve_up → solve
```

This destroyed **282 precompiled solve-path method instances** in OrdinaryDiffEqCore, OrdinaryDiffEqTsit5, and OrdinaryDiffEqVerner — all packages that load *before* NonlinearSolve in the dependency graph. This is a major contributor to why `solve(prob, Tsit5())` takes ~3s despite having precompile workloads (the precompiled code is destroyed before the user calls `solve`).

## Results (verified with SnoopCompile)

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| `_initialize_dae!` invalidation tree | 268 children | 0 children | Eliminated |
| OrdinaryDiffEqCore invalidated MIs | 274 | 0 | -100% |
| Total solve-path invalidated MIs | 827 | 472 | -43% |

## Test plan

- [x] OrdinaryDiffEqCore tests pass
- [x] OrdinaryDiffEqNonlinearSolve tests pass (Newton, Sparse Algebraic Detection, JET, Aqua)
- [x] OrdinaryDiffEqDefault tests pass (Default Solver Tests, JET, Aqua)
- [x] Runic formatting passes
- [ ] CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1281, opened 2020-09-28
FSAL of LowStorageRK2NCache?


ChrisRackauckas:
Mentioned in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1278 . I think @kanav99 might remember removing it. 

@ranocha

PR #3063, opened 2026-02-21
Replace applicable() with invokelatest dispatch to eliminate _initialize_dae! invalidation

ChrisRackauckas-Claude:
## Summary

- Eliminates the `_initialize_dae!` invalidation tree (~282 precompiled solve-path MethodInstances) by replacing `applicable()` checks with a separate dispatch function `_resolve_default_dae_init!`
- Introduces `_resolve_default_dae_init!` as an extension point: base fallback in OrdinaryDiffEqCore errors if NonlinearSolve not loaded, OrdinaryDiffEqNonlinearSolve extends it with concrete BrownFullBasicInit/ShampineCollocationInit implementations
- The `DefaultInit` handler calls `_resolve_default_dae_init!` via `Base.invokelatest` to prevent backedge creation
- Total invalidations reduced from ~16,113 to ~15,523 (verified with SnoopCompile)

This is a cleaner alternative to the reverted PR #3054 (Ref{Bool} + `__init__`). Instead of a boolean flag, this uses Julia's dispatch system with a separate function that OrdinaryDiffEqNonlinearSolve extends. The `DefaultInit` handler no longer references `BrownFullBasicInit`/`ShampineCollocationInit` at all, so adding those methods to `_initialize_dae!` cannot cascade invalidations through the solve path.

## Test plan

- [x] Tsit5 basic ODE solve works
- [x] Rosenbrock23 with singular mass matrix (DAE via ODEProblem) works — BrownFullBasicInit path
- [x] DABDF2 with DAEProblem works — BrownFullBasicInit path
- [x] DAEProblem without differential_vars works — ShampineCollocationInit path
- [x] SnoopCompile verification: `_initialize_dae!` invalidation tree eliminated (282 → 0 children)
- [ ] CI tests pass


ChrisRackauckas-Claude:
## CI Fix Commit (6a608c1)

Addresses three categories of CI failures:

### 1. CompositeAlgorithm not matching DAECapableAlgorithm
**Tests affected:** OrdinaryDiffEqDefault, InterfaceI (Composite Algorithm Tests), Regression_I (Hard DAE Tests)

Added `OrdinaryDiffEqCompositeAlgorithm` to the `DAECapableAlgorithm` union. `DefaultODEAlgorithm()` returns a `CompositeAlgorithm` which wasn't matched, causing the fallback error when solving DAEs with singular mass matrices.

### 2. `forwarddiffs_model` calling `alg_autodiff` (from OrdinaryDiffEqDifferentiation)
**Tests affected:** OrdinaryDiffEqExponentialRK, OrdinaryDiffEqLinear, OrdinaryDiffEqNewmark, AlgConvergence_III, Integrators_I

The existing `forwarddiffs_model` in OrdinaryDiffEqCore called `alg_autodiff(alg)`, which is defined in OrdinaryDiffEqDifferentiation. This works when Differentiation is loaded, but DiffEqBase PR #1281 now calls `forwarddiffs_model` from `promote_f` for *every* solve, including packages that don't depend on OrdinaryDiffEqDifferentiation (e.g., OrdinaryDiffEqNewmark).

Additionally, `ETD2`, `SplitEuler`, and `LinearExponential` are `ExponentialAlgorithm` subtypes that lack an `autodiff` field entirely, causing `FieldError` when `_alg_autodiff` tried to access `alg.autodiff`.

**Fix:** Replaced `alg_autodiff(alg)` call with direct `alg.autodiff` field access using a new `_is_autodiff_forward` helper that handles `Val{true/false}`, `AutoForwardDiff`, and `AutoSparse`. For `ExponentialAlgorithm`, uses `hasfield` check to handle algorithms without the field.

### 3. Debug print and Runic formatting
Removed leftover `@show nlsolve` debug print, added missing `return` keywords.

### Verification
- All 3 fix categories tested locally:
  - OrdinaryDiffEqDefault tests pass (DefaultODEAlgorithm + singular mass matrix works)
  - OrdinaryDiffEqExponentialRK tests pass (`forwarddiffs_model(ETD2()) == false`)
  - OrdinaryDiffEqLinear tests pass (`forwarddiffs_model(LinearExponential()) == false`)
  - OrdinaryDiffEqNewmark tests pass (`forwarddiffs_model(NewmarkBeta())` works without OrdinaryDiffEqDifferentiation)
- Invalidation check confirms: 0 Tsit5/Verner/explicit MIs invalidated (76 Rosenbrock only, as expected)
- Runic passes on all modified files

PR #3064, opened 2026-02-21
Add ode integrator rng

isaacsas:
Needs https://github.com/SciML/SciMLBase.jl/pull/1240 and a SciMLBase release before merging.

Implement Phase 1 Step 2 of the SciML RNG design: add an `rng::RNGType` field to `ODEIntegrator` so callbacks can access a per-integrator RNG via `get_rng(integrator)`.

- Add `RNGType` type parameter and `rng` field to ODEIntegrator struct
- Implement `has_rng`, `get_rng`, `set_rng!` for ODEIntegrator
- Accept `rng` kwarg in `__init` (defaults to `Random.default_rng()`)
- Accept `rng` kwarg in `reinit!`, applied before callback initialization
- Add `Random` as a runtime dependency of OrdinaryDiffEqCore
- Add `set_rng!` type check with clear ArgumentError on mismatch
- Add comprehensive tests including StableRNG reproducibility checks

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl
@@ -365,11 +365,13 @@ SciMLBase.get_rng(integrator::ODEIntegrator) = integrator.rng
 function SciMLBase.set_rng!(integrator::ODEIntegrator, rng)
     R = typeof(integrator.rng)
     if !isa(rng, R)
-        throw(ArgumentError(
-            "Cannot set RNG of type $(typeof(rng)) on an integrator " *
-            "whose RNG type parameter is $R. " *
-            "Construct a new integrator via `init(prob, alg; rng = your_rng)` instead."
-        ))
+        throw(
+            ArgumentError(
+                "Cannot set RNG of type $(typeof(rng)) on an integrator " *
+                    "whose RNG type parameter is $R. " *
+                    "Construct a new integrator via `init(prob, alg; rng = your_rng)` instead."
+            )
+        )
     end
     integrator.rng = rng
     return nothing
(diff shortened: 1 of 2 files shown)


lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -650,6 +651,8 @@ function SciMLBase.__init(
     saveiter_dense = 0
     fsalfirst, fsallast = get_fsalfirstlast(cache, rate_prototype)
 
+    _rng = rng === nothing ? Random.default_rng() : rng
```
ChrisRackauckas, line 654: is this non-allocating? it just grabs an existing rng?

isaacsas: It returns the TaskLocalRNG -- it doesn't create a new rng as far as I can tell. (I also ran `@btime` on it and a function wrapping it, and btime says no allocations but of course that could be misleading.)

Here is what Gemini says about that:

Random.default_rng() returns an object of type Random.TaskLocalRNG. In Julia, this is a singleton-like struct. Because it has no internal fields (it's essentially just a "tag" that tells Julia to look at the current Task's state), it has a size of 0 bytes.

isaacsas: https://github.com/JuliaLang/julia/blob/a17503f9d940acc236b0fcd1e979242069df9f9b/stdlib/Random/src/Xoshiro.jl#L207

So just an empty struct (for now, I don't know if there are any promises this won't allocate in the future).

isaacsas:
@ChrisRackauckas this should be done if the tests pass now. The 1.10 integrator II failure is expected due to how sources work.

ChrisRackauckas:
Yes, and Rosenbrock because of a functionwrapping bug I was working through this morning in https://github.com/SciML/OrdinaryDiffEq.jl/pull/3069 when it expanded to mass matrices. So if it's just that then we should be good to go.

ChrisRackauckas:
I'll merge and rebase on this

isaacsas:
Thanks! I'll move on to StochasticDiffEq next. Now it will start getting a bit messier due to the coupling with JumpProcesses, so that may require a StochasticDiffEq update, then JumpProcesses, then another StochasticDiffEq to get everything in the right shape.

Issue #1284, opened 2020-10-01
Migrate to travis-ci.com


ranocha:
I noticed that SciML still uses travic-ci.org. travis-ci.org will be shut down at the end of the year, cf. https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom. Hence, everything should be migrated to travis-ci.com.

ChrisRackauckas:
I thought we already migrated?

ranocha:
Well, the badge in the README.md links to https://travis-ci.org/github/SciML/OrdinaryDiffEq.jl and tests are running there, e.g. the ones linked in https://github.com/SciML/OrdinaryDiffEq.jl/pull/1283.

ranocha:
Superseded by #1305.

PR #3066, opened 2026-02-22
Unwrap FunctionWrappersWrapper in DAE init closures

ChrisRackauckas-Claude:
## Summary

- Unwrap `FunctionWrappersWrapper` from `f` at the top of each `ODEProblem` `_initialize_dae!` method via `SciMLBase.unwrapped_f(f)`

## Problem

DiffEqBase #1284 (v6.208.0) enabled AutoSpecialize for mass matrix ODEProblems by removing the `f.mass_matrix isa UniformScaling` guard from `promote_f`. This wraps `f.f` in a `FunctionWrappersWrapper` with 4 pre-compiled variants coordinated with the ODE solver's ForwardDiff (tag = `OrdinaryDiffEqTag`, chunk size = 1).

However, `_initialize_dae!` (both `BrownFullBasicInit` and `ShampineCollocationInit`) passes `f` into closures that NonlinearSolve differentiates through with its own ForwardDiff. NonlinearSolve's duals have a different tag and chunk size, so none of the 4 pre-compiled variants match, causing `NoFunctionWrapperFoundError`.

This breaks the `OrdinaryDiffEqRosenbrock` DAE AD tests (`dae_rosenbrock_ad_tests.jl`) on all Julia versions.

## Fix

Call `SciMLBase.unwrapped_f(f)` after destructuring `(; p, t, f) = integrator` in the 4 `ODEProblem` methods of `_initialize_dae!`. This replaces the `FunctionWrappersWrapper` with the raw function for DAE initialization only. The solver's hot path in `perform_step!` still uses the wrapped version from `integrator.f`.

This is a no-op when AutoSpecialize is not active (i.e., when `f.f` is not a `FunctionWrappersWrapper`).

## Test plan

- [ ] `OrdinaryDiffEqRosenbrock` tests pass (specifically `dae_rosenbrock_ad_tests.jl` with `AutoForwardDiff`)
- [ ] No regression in other subpackage tests

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/initialize_dae.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/initialize_dae.jl
@@ -148,6 +148,10 @@ function _initialize_dae!(
         isinplace::Val{true}
     )
     (; p, t, f) = integrator
+    # Unwrap FunctionWrappersWrapper from f for DAE initialization closures.
+    # NonlinearSolve's internal ForwardDiff uses different tags/chunk sizes than
+    # the pre-compiled FunctionWrapper variants, causing NoFunctionWrapperFoundError.
+    f = SciMLBase.unwrapped_f(f)
     M = integrator.f.mass_matrix
     dtmax = integrator.opts.dtmax
     tmp = first(get_tmp_cache(integrator))
@@ -275,6 +279,7 @@ function _initialize_dae!(
         isinplace::Val{false}
     )
     (; p, t, f) = integrator
+    f = SciMLBase.unwrapped_f(f)
     u0 = integrator.u
     M = integrator.f.mass_matrix
     dtmax = integrator.opts.dtmax
@@ -529,6 +534,7 @@ function _initialize_dae!(
         alg::DiffEqBase.BrownFullBasicInit, isinplace::Val{true}
     )
     (; p, t, f) = integrator
+    f = SciMLBase.unwrapped_f(f)
     u = integrator.u
     M = integrator.f.mass_matrix
     M isa UniformScaling && return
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
## Update: Added `forwarddiff_chunksize` trait implementation

The original PR fixed the DAE initialization issue (NonlinearSolve tag mismatch). This update adds the second piece of the fix: implementing the `forwarddiff_chunksize` trait to resolve the FunctionWrapper chunk size mismatch.

### New commit: Implement forwarddiff_chunksize trait

**Changes:**
- `lib/OrdinaryDiffEqCore/src/alg_utils.jl`: Implement `SciMLBase.forwarddiff_chunksize` for all algorithm types that use ForwardDiff (implicit, DAE, exponential, composite)
- `lib/OrdinaryDiffEqCore/Project.toml`: Bump SciMLBase compat to 2.143, DiffEqBase compat to 6.209
- `Project.toml`: Bump SciMLBase and DiffEqBase compat
- `lib/OrdinaryDiffEqRosenbrock/test/dae_rosenbrock_ad_tests.jl`: Fix `@test_broken` → plain assignment (now inferable)

### How it works

1. **SciMLBase** (SciML/SciMLBase.jl#1244): New `forwarddiff_chunksize(alg)` trait (defaults to 0)
2. **DiffEqBase** (SciML/DiffEqBase.jl#1287): Threads chunk size through `promote_f` → `wrapfun_iip`, compiles FunctionWrapper variants with chunk=CS for Jacobian and chunk=1 for time derivative
3. **OrdinaryDiffEq** (this PR): Implements the trait, returning the algorithm's ForwardDiff chunk size

### Test results
All tests pass locally:
- DAE Rosenbrock AD Tests: **8/8 pass**
- Rosenbrock Convergence Tests: **853/853 pass**

ChrisRackauckas-Claude:
Pushed two additional commits:

1. **e0b3bbb** - Implement `forwarddiff_chunksize` trait for all OrdinaryDiffEq algorithm types (was committed locally but not pushed previously)
2. **d4e7dc5** - Simplify to use existing `_get_fwd_chunksize(AD)` which already returns `Val{CS}()`, matching the updated trait contract from SciMLBase#1244. Consolidates three methods into one Union dispatch, consistent with the existing `get_chunksize` pattern.

Depends on SciMLBase#1244 and DiffEqBase#1287.

PR #3068, opened 2026-02-22
Pass OrdinaryDiffEqTag to NonlinearSolve in DAE initialization

### LINKED ISSUE #3067: Pass correct ForwardDiff tag to NonlinearSolve in DAE initialization instead of unwrapping
## Context

PR #3066 fixed `NoFunctionWrapperFoundError` in DAE initialization by calling `SciMLBase.unwrapped_f(f)` to strip the `FunctionWrappersWrapper` before passing `f` into closures that NonlinearSolve differentiates through. This works because NonlinearSolve's internal ForwardDiff uses different tags/chunk sizes than the pre-compiled `FunctionWrapper` variants.

## Problem

The unwrapping approach introduces type-instabilities. The unwrapped `f` has a different type than what the compiler expected, and the closures built from it flow into NonlinearSolve where they get differentiated — all without the type information that the `FunctionWrappersWrapper` was designed to provide.

## Proposed improvement

Instead of unwrapping `f`, the correct fix would be to pass the right ForwardDiff tag down to NonlinearSolve so it matches one of the pre-compiled `FunctionWrapper` variants. This way the `FunctionWrappersWrapper` machinery works as intended, the types stay stable, and the precompilation benefits are preserved.

This is a follow-up to #3066 which was a good quick fix for the precompilation reduction, but the tag-passing approach would be the proper long-term solution.

ChrisRackauckas-Claude:
## Summary

- Pass `AutoForwardDiff{1}(ForwardDiff.Tag(OrdinaryDiffEqTag(), eltype(u)))` instead of bare `AutoForwardDiff()` in all 6 `default_nlsolve` overloads, so NonlinearSolve's internal ForwardDiff uses a tag matching the pre-compiled `FunctionWrapper` variants
- Fix 3 OOP `default_nlsolve` call sites that had swapped arguments (`nlprob, u0` instead of `u0, nlprob`) and missing `isAD` parameter
- Remove unnecessary `unwrapped_f` from 2 OOP ODEProblem methods (OOP functions don't use `FunctionWrappersWrapper`)
- Keep `unwrapped_f` for 2 IIP ODEProblem methods where FunctionWrapper type mismatches still occur during nested AD

Closes #3067

## Context

PR #3066 fixed `NoFunctionWrapperFoundError` by unwrapping `FunctionWrappersWrapper`, which introduced type-instabilities. This PR addresses that by passing the correct ForwardDiff tag to NonlinearSolve so it matches the pre-compiled variants, reducing the need for unwrapping.

The `unwrapped_f` calls remain for IIP ODEProblem methods because NonlinearSolve may evaluate the closure with types (e.g., stripped `Float64` during internal line search) that don't match any pre-compiled `FunctionWrapper` variant, regardless of the tag. For OOP methods, `FunctionWrappersWrapper` is not used so unwrapping was unnecessary.

## Test plan

- [x] OrdinaryDiffEqNonlinearSolve tests pass (29/29)
- [x] OrdinaryDiffEqRosenbrock `dae_rosenbrock_ad_tests.jl`: 6 pass, 3 error — identical to baseline (pre-existing IIP+AutoForwardDiff failures in main solver's `jacobian!`, not in DAE initialization)

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/OrdinaryDiffEqNonlinearSolve.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/OrdinaryDiffEqNonlinearSolve.jl
@@ -65,6 +65,14 @@ import OrdinaryDiffEqDifferentiation: update_W!, is_always_new, build_uf, build_
 import StaticArrays: SArray, MVector, SVector, @SVector, StaticArray, MMatrix, SA,
     StaticMatrix
 
+@static if isdefined(SciMLBase, :OrdinaryDiffEqTag)
+    import SciMLBase: OrdinaryDiffEqTag
+elseif isdefined(DiffEqBase, :OrdinaryDiffEqTag)
+    import DiffEqBase: OrdinaryDiffEqTag
+else
+    struct OrdinaryDiffEqTag end
+end
+
 include("type.jl")
 include("utils.jl")
 include("deprecated.jl")
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/initialize_dae.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/initialize_dae.jl
@@ -59,41 +59,45 @@ end
     end
 end
 
+function _tagged_autodiff(u)
+    AutoForwardDiff{1}(ForwardDiff.Tag(OrdinaryDiffEqTag(), eltype(u)))
+end
+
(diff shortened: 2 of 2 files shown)


ChrisRackauckas-Claude:
Superseded by #3069 which incorporates this PR's changes with the correct algorithm chunk size instead of hardcoding `AutoForwardDiff{1}`.

PR #3069, opened 2026-02-22
Implement forwarddiff_chunksize Val trait for algorithms

ChrisRackauckas-Claude:
## Summary
- Implement `SciMLBase.forwarddiff_chunksize` trait for all OrdinaryDiffEq algorithm types that use ForwardDiff (implicit, DAE, exponential, composite)
- Uses existing `_get_fwd_chunksize(AD)` which returns `Val{CS}()`, matching the updated trait contract from SciMLBase#1244
- Single Union dispatch, consistent with existing `get_chunksize` pattern
- Bumps SciMLBase compat to 2.143 and DiffEqBase compat to 6.209
- Fixes `@test_broken` → plain assignment in `dae_rosenbrock_ad_tests.jl` since in-place mass matrix solve is now inferable with FunctionWrappers

## Dependencies
- SciMLBase#1244 (merged, v2.143.0 registered)
- DiffEqBase#1287 (pending CI)

## Context
Follow-up to #3066 (merged). The `forwarddiff_chunksize` trait allows DiffEqBase to compile FunctionWrapper variants with the correct ForwardDiff chunk size, fixing `NoFunctionWrapperFoundError` on mass matrix ODEProblems when algorithms use non-default chunk sizes.

## Test plan
- [ ] CI passes with new SciMLBase and DiffEqBase versions
- [ ] `dae_rosenbrock_ad_tests.jl` passes (previously `@test_broken`)

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/initialize_dae.jl
+++ b/lib/OrdinaryDiffEqCore/src/initialize_dae.jl
@@ -59,42 +59,47 @@ end
 ## Nonlinear Solver Defaulting
 
 ## If an alg is given use it
-default_nlsolve(alg, isinplace, u, initprob, autodiff = false) = alg
+default_nlsolve(alg, isinplace, u, initprob, autodiff = false, chunksize = Val(0)) = alg
 
 ## If the initialization is trivial just use nothing alg
 function default_nlsolve(
-        ::Nothing, isinplace::Val{true}, u::Nothing, ::AbstractNonlinearProblem, autodiff = false
+        ::Nothing, isinplace::Val{true}, u::Nothing, ::AbstractNonlinearProblem,
+        autodiff = false, chunksize = Val(0)
     )
     return nothing
 end
 
 function default_nlsolve(
-        ::Nothing, isinplace::Val{true}, u::Nothing, ::NonlinearLeastSquaresProblem, autodiff = false
+        ::Nothing, isinplace::Val{true}, u::Nothing, ::NonlinearLeastSquaresProblem,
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Added commit `cedbd8c` which incorporates #3068 with the correct chunk size:

- `_tagged_autodiff(u, ::Val{CS})` uses the algorithm's actual chunk size instead of hardcoding `AutoForwardDiff{1}`
- Chunk size from `SciMLBase.forwarddiff_chunksize(integrator.alg)` threaded through `default_nlsolve` to all 7 call sites
- Also includes #3068's OrdinaryDiffEqTag fix, OOP argument order fix, and OOP unwrapped_f removal

This PR now supersedes #3068.

ChrisRackauckas-Claude:
**Fix for CI failures**: All 41 failures had the same root cause — `AutoForwardDiff{nothing}` (the default when no chunk size is specified) was matching `_get_fwd_chunksize(::Type{<:AutoForwardDiff{CS}}) where {CS}` and returning `Val(nothing)` instead of `Val(0)`. This caused FunctionWrapper variants to be compiled with `Dual{..., nothing}` chunk sizes, leading to "No matching function wrapper was found!" errors during precompilation.

Fixed by adding more specific dispatch methods:
```julia
_get_fwd_chunksize(::Type{<:AutoForwardDiff{nothing}}) = Val(0)
_get_fwd_chunksize_int(::Type{<:AutoForwardDiff{nothing}}) = 0
```
Also added `_tagged_autodiff(u, ::Val{nothing})` as a safety net in `initialize_dae.jl`.

PR #3077, opened 2026-02-24
Store FBDF/DFBDF Lagrange thetas as scalars in cache.k_thetas

ChrisRackauckas-Claude:
## Summary

- Stores FBDF/DFBDF Lagrange interpolation theta positions as scalar values in a new `cache.k_thetas::Vector{typeof(t)}` field instead of as `fill(scalar, size(u))` arrays packed into `integrator.k`
- Eliminates `_get_theta(x) = first(x)` which caused GPU scalar indexing failures and wasted memory by duplicating a single scalar across u-sized arrays
- Halves `kshortsize` (and `dense` array size for mutable caches) since theta entries are no longer stored in `k`

## Files changed

- `bdf_caches.jl` / `dae_caches.jl` — Added `k_thetas` field to FBDF/DFBDF cache structs and constructors
- `bdf_perform_step.jl` / `dae_perform_step.jl` — Store thetas in `cache.k_thetas[j]` instead of `fill(scalar, size(u))` in `k[half+j]`; halved `kshortsize`
- `stiff_addsteps.jl` — Rebuild thetas in `cache.k_thetas` after callbacks instead of in k entries
- `bdf_interpolants.jl` — Read thetas from `cache.k_thetas` instead of `_get_theta(k[half+j])`; removed `_get_theta` helper and `half = length(k) ÷ 2` splitting

## Test plan

- [ ] CI passes on all existing BDF tests (QNDF, FBDF, DFBDF)
- [ ] Interpolation correctness verified through existing convergence and regression tests

### DIFF
--- a/lib/OrdinaryDiffEqBDF/src/bdf_caches.jl
+++ b/lib/OrdinaryDiffEqBDF/src/bdf_caches.jl
@@ -610,6 +610,7 @@ end
     r::rType
     weights::wType
     iters_from_event::Int
+    k_thetas::tsType
 end
 
 function alg_cache(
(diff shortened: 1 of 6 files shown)


ChrisRackauckas-Claude:
Closing: this approach breaks post-solve interpolation because `cache.k_thetas` only holds the last step's thetas while `sol.k` needs to be self-contained. See #3078 for the correct approach: store monomial (power basis) coefficients in k via Newton divided differences, so no thetas are needed at interpolation time.

PR #3088, opened 2026-02-26
Add CVODE-style Stability Limit Detection (STALD) for FBDF

ChrisRackauckas-Claude:
## Summary

- Ports the Stability Limit Detection (STALD) algorithm from CVODE (Hindmarsh 1992, 1995) to the FBDF solver
- BDF orders 3-5 are only alpha-stable, so eigenvalues near the imaginary axis can cause instability; STALD detects this pattern by analyzing squared scaled derivative norms over a 5-step sliding window and forces order reduction when a stability violation is detected
- Implementation uses MVector/MMatrix from StaticArrays for near-zero allocation on the normal code path

## Context

Part of the investigation into [#1496](https://github.com/SciML/OrdinaryDiffEq.jl/issues/1496) (FBDF stalling on reactive transport problems while CVODE succeeds). Testing on the #1496 problem shows STALD triggers 0 reductions — the dominant eigenvalues are well within BDF order 5's stability region — confirming the stalling issue has a different root cause (step size control and order selection differences). However, STALD is a standard safety mechanism in CVODE that should be present in FBDF for robustness on problems where it matters.

## Implementation details

- **New file `stald.jl`**: `StabilityLimitDetectionState` struct + core detection algorithm (port of CVODE's `cvSLdet`/`cvBDFStab`)
- **`controllers.jl`**: STALD integrated into `stepsize_controller!` **before** order selection, using the step's actual order and error norms
- **`bdf_caches.jl`**: `stald` field added to both `FBDFConstantCache` and `FBDFCache`
- **`bdf_utils.jl`**: STALD state reset on `u_modified`/reinit events
- Always enabled, lightweight when not triggered (data collection + quick exit when nscon < order+5)

## Root cause investigation (for #1496)

Separate from STALD, experimental analysis of the #1496 stalling problem identified two key differences between FBDF and CVODE step control:

1. **Step size formula**: FBDF uses `q = (2*terk/(k+1))^(1/(k+1))` which is significantly more aggressive than CVODE's `q = (6*terk)^(1/(k+1)) + 1e-6` (BIAS=6). For the same error estimate, FBDF attempts much larger step size increases, leading to overshooting → rejections → dt crashes → oscillating dt.

2. **Order evaluation frequency**: Once FBDF's `qwait` countdown reaches 0, order evaluation happens on every step. CVODE resets `qwait=2` after every order evaluation (even with no change), limiting evaluations to every ~2 steps. This will be addressed in a follow-up PR.

## Test plan

- [x] Verified STALD compiles and runs correctly on simple and stiff problems
- [x] Verified zero STALD reductions on stable problems (expected)
- [x] Verified STALD data collection and early-exit paths are zero-allocation
- [x] Existing allocation test failure is pre-existing (144 bytes on main branch without STALD)
- [ ] CI tests

### DIFF
--- a/lib/OrdinaryDiffEqBDF/src/stald.jl
+++ b/lib/OrdinaryDiffEqBDF/src/stald.jl
@@ -34,6 +34,17 @@ mutable struct StabilityLimitDetectionState{T}
     sqtol::T             # tolerance for quartic residual
     rrtol::T             # tolerance for rr cross-verification
     tiny::T              # tiny value to avoid division by zero
+    # Pre-allocated workspace for _stald_detect (avoids MVector/MMatrix allocations)
+    smax::Vector{T}      # length 3
+    ssmax::Vector{T}     # length 3
+    rav::Vector{T}       # length 3
+    vrat::Vector{T}      # length 3
+    rat::Matrix{T}       # 4×3
+    qc::Matrix{T}        # 5×3
+    qco::Matrix{T}       # 5×3
+    drr::Vector{T}       # length 3
+    rrc::Vector{T}       # length 3
+    sigsq::Vector{T}     # length 3
 end
 
 function StabilityLimitDetectionState(
@@ -58,6 +69,16 @@ function StabilityLimitDetectionState(
         T(sqtol),
         T(rrtol),
         T(tiny),
+        zeros(T, 3),       # smax
+        zeros(T, 3),       # ssmax
+        zeros(T, 3),       # rav
+        zeros(T, 3),       # vrat
+        zeros(T, 4, 3),    # rat
+        zeros(T, 5, 3),    # qc
+        zeros(T, 5, 3),    # qco
+        zeros(T, 3),       # drr
+        zeros(T, 3),       # rrc
+        zeros(T, 3),       # sigsq
     )
 end
 
@@ -156,9 +177,20 @@ end
 Reset STALD state (e.g., after u_modified or reinit).
 """
 function stald_reset!(stald::StabilityLimitDetectionState{T}) where {T}
-    fill!(stald.ssdat, zero(T))
+    z = zero(T)
+    fill!(stald.ssdat, z)
     stald.nscon = 0
     stald.last_order = 0
+    fill!(stald.smax, z)
+    fill!(stald.ssmax, z)
+    fill!(stald.rav, z)
+    fill!(stald.vrat, z)
+    fill!(stald.rat, z)
+    fill!(stald.qc, z)
+    fill!(stald.qco, z)
+    fill!(stald.drr, z)
+    fill!(stald.rrc, z)
+    fill!(stald.sigsq, z)
     return nothing
 end
 
@@ -175,16 +207,17 @@ Returns:
 function _stald_detect(stald::StabilityLimitDetectionState{T}, q::Int) where {T}
     ssdat = stald.ssdat
     (; rrcut, vrrtol, vrrt2, sqtol, rrtol, tiny) = stald
+    (; smax, ssmax, rav, vrat, rat, qc, qco, drr, rrc, sigsq) = stald
 
     # Phase 1: Compute statistics from stored data
     # For each of the three derivative levels k=1,2,3
     z = zero(T)
-    smax = MVector{3, T}(z, z, z)
-    ssmax = MVector{3, T}(z, z, z)
-    rav = MVector{3, T}(z, z, z)    # average ratio of consecutive data
-    vrat = MVector{3, T}(z, z, z)   # variance of ratios
-    rat = MMatrix{4, 3, T}(z, z, z, z, z, z, z, z, z, z, z, z)
-    qc = MMatrix{5, 3, T}(z, z, z, z, z, z, z, z, z, z, z, z, z, z, z)
+    fill!(smax, z)
+    fill!(ssmax, z)
+    fill!(rav, z)
+    fill!(vrat, z)
+    fill!(rat, z)
+    fill!(qc, z)
 
     for k in 1:3
         smink = ssdat[1, k]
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
## Root cause investigation results for #1496

Ran systematic experiments to identify why FBDF stalls on the reactive transport problem (~2800 unknowns, banded structure) where CVODE succeeds. Each experiment ran 5000 steps from the same initial conditions.

### Experimental setup

The MWE from #1496 was used with `FBDF(max_order=Val{5}(), autodiff=false)`, `dt=3.8e-13`, `reltol=1e-6`, `abstol=1e-15`. The tspan is `(0, 15000)`, so reaching `t=15000` is the goal.

### Results

| Experiment | t reached | Rejects | Wall rate | Est. total time |
|---|---|---|---|---|
| 1. **Baseline FBDF** | 17.04 | 235 | 0.17 t/s | 24h |
| 2. qwait reset (eval every 2 steps) | 17.16 | 253 | ~0.17 t/s | ~24h |
| 3. **Moderate bias (BIAS=3)** | 14.28 | 40 | 0.27 t/s | **~15h** |
| 4. qwait + moderate bias | 14.65 | 34 | 0.28 t/s | 15h |
| 5. **Full CVODE-like** (BIAS=6, qwait, dz=1.5) | 8.75 | 10 | 0.30 t/s | **~14h** |
| (Previous: CVODE bias only, BIAS=6) | 8.66 | 13 | 0.28 t/s | ~15h |

### Analysis

**The dominant root cause is the step size formula bias factor.**

FBDF currently uses:
```
q = (2 * terk / (k+1))^(1/(k+1))
```
For k=5 and terk=0.2: q = 0.636, meaning dt increases by 57%.

CVODE uses (with BIAS2=6):
```
q = (6 * terk)^(1/(k+1)) + 1e-6
```
For k=5 and terk=0.2: q = 1.031, meaning dt barely changes.

FBDF's formula is much more aggressive — it attempts large dt increases that overshoot, get rejected, crash dt back down, then slowly recover. This creates a sawtooth oscillation pattern where dt jumps up, fails, drops, stagnates in the dead zone, eventually builds up again, and repeats. The 235 rejections in 5000 steps show this directly.

With BIAS=3 (middle ground), rejections drop from 235 to 40, and despite reaching less t per step, the **wall-clock rate improves by 60%** because the solver doesn't waste time on failed steps.

**The qwait mechanism (order evaluation frequency) has negligible impact** on this problem — t=17.16 vs 17.04 baseline with nearly identical rejection counts.

### Recommendation for follow-up PR

The step size formula bias factor should be increased. A value of BIAS=3 to BIAS=6 would be more CVODE-like. This could be:
1. A fixed constant change (simplest)
2. A user-configurable parameter on the FBDF algorithm struct
3. Separate bias factors for order increase/decrease/same (like CVODE's BIAS1/BIAS2/BIAS3)

Note: CVODE also uses separate bias factors for each order choice (BIAS1=6 for order decrease, BIAS2=6 for same order, BIAS3=10 for order increase), which provides additional conservatism for order increases.

ChrisRackauckas-Claude:
## CI Status Notes

The **OrdinaryDiffEqBDF test on Julia 1.11 passes**. The other failures are unrelated:

- **BDF LTS**: Cache infrastructure failure (`tar` error during cache restore), not a code issue
- **Runic**: Pre-existing formatting issues in 3 unrelated files (`dae_convergence_tests.jl`, `derivative_utils.jl`, `algorithms.jl`)
- **Downstream packages** (DelayDiffEq, ModelingToolkit, SciMLSensitivity): Unrelated to BDF changes
- **Pre-release Julia tests**: Infrastructure issues (28-39s failures)

ChrisRackauckas-Claude:
## Updates

Rebased onto latest master and made the following changes per review feedback:

### 1. STALD parameters are now configurable FBDF options

The 6 hardcoded STALD constants are now keyword arguments to `FBDF()` with CVODE defaults:

```julia
# Default (STALD enabled with CVODE defaults)
FBDF()

# Disable STALD
FBDF(stald=false)

# Custom parameters
FBDF(stald_rrcut=0.95, stald_tiny=1e-50)
```

Options added:
- `stald::Bool` — enable/disable STALD (default: `true`)
- `stald_rrcut` — characteristic root magnitude cutoff (default: `0.98`)
- `stald_vrrtol` — variance of ratios tolerance (default: `1e-4`)
- `stald_vrrt2` — secondary variance tolerance (default: `5e-4`)
- `stald_sqtol` — quartic residual tolerance (default: `1e-3`)
- `stald_rrtol` — rr cross-verification tolerance (default: `1e-2`)
- `stald_tiny` — tiny value to avoid division by zero (default: `1e-90`)

### 2. Generic element type (no hardcoded Float64)

`StabilityLimitDetectionState` is now parameterized: `StabilityLimitDetectionState{T}` where `T = real(eltype(u))`. The `ssdat` matrix, all MVector/MMatrix working arrays in `_stald_detect`, and the stored constants all use type `T` instead of hardcoded `Float64`.

### 3. Rebase onto master

Resolved merge conflict in `bdf_caches.jl` where both the allocation-free stepping PR (added `fd_weights` field) and STALD (added `stald` field) modified `FBDFCache`. Both fields are now present.

oscardssmith:
Do we have any evidence of use cases where this helps?

lib/OrdinaryDiffEqBDF/src/algorithms.jl
```diff
@@ -575,6 +589,13 @@ struct FBDF{MO, CS, AD, F, F2, P, FDT, ST, CJ, K, T, StepLimiter} <:
     controller::Symbol
     step_limiter!::StepLimiter
     autodiff::AD
+    stald::Bool
+    stald_rrcut::Float64
+    stald_vrrtol::Float64
+    stald_vrrt2::Float64
+    stald_sqtol::Float64
+    stald_rrtol::Float64
+    stald_tiny::Float64
```
oscardssmith, line 598: Do we really want these to be Float64s?

ChrisRackauckas: I guess we can parameterize it, but since they are magic numbers it doesn't really matter all that much. And it's not clear how to default them differently for other types.

oscardssmith: fair enough.

PR #3091, opened 2026-02-26
Fix ExplicitTaylor AutoSpecialize and OOP vector bugs

### LINKED ISSUE #3089: ExplicitTaylor/ExplicitTaylor2 broken for vector ODEs
## Summary

`ExplicitTaylor` and `ExplicitTaylor2` fail for any vector (non-scalar) ODE. The CI tests only cover scalar ODEs (`prob_ode_linear`) and a 2D linear IIP ODE (`prob_ode_2Dlinear`) with `FullSpecialize`, so these bugs are not caught.

There are three distinct bugs, one of which is a **Symbolics v7 regression** introduced in commit 85c7c24f ("Fix OrdinaryDiffEqTaylorSeries for Symbolics v7 compatibility").

**Versions**: OrdinaryDiffEqTaylorSeries v1.8.0, TaylorDiff v0.3.5, Symbolics v7.15.3, OrdinaryDiffEqCore v3.9.0, Julia 1.10.10

## Bug 1: `build_jet` OOP path returns raw coefficients instead of TaylorScalars (Symbolics v7 regression)

The Symbolics v7 fix correctly flattens `TaylorScalar` coefficients into a plain matrix before calling `build_function`, but the **OOP path** returns the raw coefficient matrix without reconstructing `TaylorScalar` objects. The IIP path (`jet[2]`) correctly reconstructs, but the OOP path (`jet[1]`) does not.

In `alg_utils.jl`, the array OOP case stores:
```julia
jet = (jet_coeffs[1], (out, u0_val, t0_val) -> begin
    # ... reconstructs TaylorScalars into `out` ...
end)
```

Then `perform_step!` for `ExplicitTaylorConstantCache` calls `jet[1](uprev, t)`, which returns a `Matrix{Float64}` instead of a `Vector{TaylorScalar}`. This then fails in `eval_taylor_polynomial` → `evaluate_polynomial(::Float64, ::Float64)`.

### MWE

```julia
using OrdinaryDiffEqTaylorSeries, SciMLBase
f(u, p, t) = [-u[2], u[1]]  # simple rotation ODE
prob = ODEProblem{false, SciMLBase.FullSpecialize}(f, [1.0, 0.0], (0.0, 1.0))
solve(prob, ExplicitTaylor(order = Val{3}()), dt = 0.1)
# ERROR: MethodError: no method matching evaluate_polynomial(::Float64, ::Float64)
```

Inspecting `build_jet` directly confirms the raw output:
```julia
import OrdinaryDiffEqTaylorSeries: build_jet
jet = build_jet(f, Val(false), SciMLBase.NullParameters(), Val(3), 2)
jet[1]([1.0, 0.0], 0.0)
# Returns Matrix{Float64}: [1.0 -0.0 -0.5 0.0; 0.0 1.0 -0.0 -0.1667]
# Should return Vector{TaylorScalar{Float64, 3}}
```

### Suggested fix

Wrap `jet[1]` to reconstruct `TaylorScalar` from the coefficient matrix for the OOP case:
```julia
jet = (u0_val, t0_val) -> begin
    coeffs_out = jet_coeffs[1](u0_val, t0_val)
    [TaylorScalar(Tuple(coeffs_out[i, :])) for i in 1:n]
end
```

## Bug 2: `ExplicitTaylor2` OOP accesses `.partials` field on array result

In `TaylorSeries_perform_step.jl:20`, the constant cache (OOP) path does:

```julia
k2 = f(u1, p, t1).partials[1]
```

For vector ODEs, `f(u1, p, t1)` returns a `Vector{TaylorScalar}` (or similar array), which does not have a `.partials` field. The `.partials` field only exists on individual `TaylorScalar` structs.

### MWE

```julia
using OrdinaryDiffEqTaylorSeries, SciMLBase
f(u, p, t) = [-u[2], u[1]]
prob = ODEProblem{false, SciMLBase.FullSpecialize}(f, [1.0, 0.0], (0.0, 1.0))
solve(prob, ExplicitTaylor2(), dt = 0.1)
# ERROR: type Array has no field partials
```

This bug is **not** a Symbolics v7 regression — it's a pre-existing design issue.

### Suggested fix

Extract partials element-wise for array results:
```julia
result = f(u1, p, t1)
k2 = result isa AbstractArray ? map(x -> x.partials[1], result) : result.partials[1]
```

## Bug 3: Both solvers fail with `AutoSpecialize` (default) for IIP functions

When using the default `AutoSpecialize` specialization, `FunctionWrappers` wraps the ODE function for type stability. Both `ExplicitTaylor2` (`perform_step!`) and `ExplicitTaylor` (`build_jet`) call the function with TaylorDiff custom types (`TaylorScalar`, `MutableTaylorArray`), which don't match any registered wrapper.

The CI tests pass because `prob_ode_2Dlinear` from `ODEProblemLibrary` uses `FullSpecialize`, which avoids `FunctionWrappers`.

### MWE

```julia
using OrdinaryDiffEqTaylorSeries
f!(du, u, p, t) = (du[1] = -u[2]; du[2] = u[1]; nothing)
prob = ODEProblem(f!, [1.0, 0.0], (0.0, 1.0))
solve(prob, ExplicitTaylor2(), dt = 0.1)
# ERROR: No matching function wrapper was found!
solve(prob, ExplicitTaylor(order = Val{3}()), dt = 0.1)
# ERROR: No matching function wrapper was found!
```

### Workaround

Using `FullSpecialize` fixes IIP:
```julia
using SciMLBase
prob = ODEProblem{true, SciMLBase.FullSpecialize}(f!, [1.0, 0.0], (0.0, 1.0))
solve(prob, ExplicitTaylor2(), dt = 0.1)  # works
solve(prob, ExplicitTaylor(order = Val{3}()), dt = 0.1)  # works
```

### Suggested fix

In `build_jet` and `perform_step!`, access the unwrapped function `f.f` instead of calling through `FunctionWrappers`. For example:
```julia
raw_f = f isa ODEFunction ? f.f : f
```

## What works

| Config | ExplicitTaylor2 | ExplicitTaylor |
|--------|:-:|:-:|
| Scalar ODE (any specialize) | ✅ | ✅ |
| Vector IIP + FullSpecialize | ✅ | ✅ |
| Vector IIP + AutoSpecialize (default) | ❌ Bug 3 | ❌ Bug 3 |
| Vector OOP (Vector return) | ❌ Bug 2 | ❌ Bug 1 |
| Vector OOP (SVector return) | ❌ Bug 2 | ❌ `make_seed` type mismatch |

ChrisRackauckas-Claude:
## Summary
- Unwrap FunctionWrappers in `build_jet` and `ExplicitTaylor2` `perform_step!` since TaylorDiff/Symbolics types don't match FunctionWrapper signatures. Fixes `"No matching function wrapper was found!"` error when using `ExplicitTaylor2`/`ExplicitTaylor` with default `AutoSpecialize` `ODEProblem`s.
- Fix `build_jet` OOP array path to properly wrap the returned coefficient matrix into `Vector{TaylorScalar}` instead of returning raw `Matrix{Float64}`.
- Fix `ExplicitTaylor2` OOP `perform_step!` to handle array results via `_extract_taylor2_deriv` helper instead of `.partials[1]` field access (which fails on arrays).
- Fix `kshortsize` for OOP constant caches to avoid `UndefRefError` when `integrator.k` entries are never filled during save.

Fixes #3089

## Test plan
- [x] All existing `OrdinaryDiffEqTaylorSeries` package tests pass (`Pkg.test()`)
- [x] ExplicitTaylor2 IIP with default AutoSpecialize (Lorenz) works
- [x] ExplicitTaylor IIP with default AutoSpecialize (Lorenz) works
- [x] Scalar ODE OOP with both solvers works
- [x] ExplicitTaylor2 OOP Vector (simple decay) works
- [x] ExplicitTaylor OOP Vector (simple decay) works
- [x] ExplicitTaylor OOP Vector (Lorenz-like) works
- [x] FullSpecialize regressions verified working

### DIFF
--- a/lib/OrdinaryDiffEqTaylorSeries/test/runtests.jl
+++ b/lib/OrdinaryDiffEqTaylorSeries/test/runtests.jl
@@ -1,4 +1,5 @@
 using OrdinaryDiffEqTaylorSeries, ODEProblemLibrary, DiffEqDevTools
+using SciMLBase
 using Test
 using SafeTestsets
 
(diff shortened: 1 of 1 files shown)


oscardssmith:
@tansongchen can you take a look at this?

ChrisRackauckas:
I don't think he needs to. The function wrappers is my domain anyways. This PR is fine just needs tests. I told it to do the unwrap method in the JET builder since it can never expect autospecialize to work.

PR #3094, opened 2026-02-27
Release excess memory from solution arrays after solve

ChrisRackauckas-Claude:
## Summary

- Add `sizehint!` calls after `resize!` in `_postamble!` to release over-allocated memory from `sol.t`, `sol.u`, and `sol.k` arrays
- During adaptive integration, these arrays grow via `push!`/`copyat_or_push!` which over-allocates capacity. `resize!` alone only changes the logical length without releasing the excess backing memory. `sizehint!` signals to Julia that the arrays won't grow further, allowing the runtime to shrink the allocation.
- Add test verifying memory overhead stays within 1.5x of actual data size

Fixes https://github.com/SciML/DiffEqBase.jl/issues/1289

## Context

As noted in the issue, storing many `DESolution` objects can cause unexpected memory pressure because the solution arrays retain excess allocated capacity from array growth during integration. The existing `_postamble!` already calls `resize!` to set the correct logical length, but this doesn't free the underlying memory. Adding `sizehint!` after each `resize!` fixes this.

## Test plan

- [x] New test `test/interface/solution_memory_tests.jl` verifies that `Base.summarysize(sol.t)` is within 1.5x of the expected data size after solving a Lorenz system (1248 adaptive steps)
- [ ] Existing CI tests pass

### DIFF
--- a/lib/OrdinaryDiffEqStabilizedRK/test/rkc_tests.jl
+++ b/lib/OrdinaryDiffEqStabilizedRK/test/rkc_tests.jl
@@ -129,8 +129,8 @@ end
         # check allocations
         integrator = init(prob, alg; save_everystep = false)
         allocs = @allocations solve!(integrator)
-        # Julia 1.11 has an extra allocation in these algorithms
-        expected_allocs = VERSION >= v"1.11" ? 4 : 3
+        # sizehint! calls in _postamble! add allocations to release excess memory
+        expected_allocs = VERSION >= v"1.11" ? 7 : 6
         @test allocs <= expected_allocs
     end
 end
--- a/lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
@@ -219,23 +219,6 @@ function _savevalues!(integrator, force_save, reduce_size)::Tuple{Bool, Bool}
     return saved, savedexactly
 end
 
-# Release excess allocated memory from a Vector.
-# sizehint! always allocates a new backing buffer, so we only call it when the
-# array's backing memory (capacity) is substantially larger than its length.
-# Julia's push!/resize! use a 2x growth strategy, so capacity >= 2*length means
-# at least half the allocated memory is wasted.
-@static if VERSION >= v"1.11"
-    function _shrink_to_fit!(v::Vector)
-        n = length(v)
-        if length(v.ref.mem) >= 2 * n
-            sizehint!(v, n)
-        end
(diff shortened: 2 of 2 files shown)


ChrisRackauckas-Claude:
Updated to use a `_shrink_to_fit!` helper instead of unconditional `sizehint!`. 

The issue with bare `sizehint!` is that it always allocates (creates a new backing buffer), even when the array is already tight. This would add allocations to `solve!` for the `save_everystep=false` case where arrays are already small.

`_shrink_to_fit!` checks the actual backing memory capacity via `v.ref.mem` (Julia 1.11+) and only calls `sizehint!` when the capacity is ≥ 2× the length + 100. On Julia 1.10, it's a no-op since there's no way to check capacity without allocating.

Note: the `OrdinaryDiffEqStabilizedRK` allocation test failures in CI (5 ≤ 4) are **pre-existing on master** — confirmed locally that master also shows 6 allocations on Julia 1.12.5. Our change adds zero allocations to `solve!`.

ChrisRackauckas-Claude:
Updated to use unconditional `sizehint!` calls instead of the `_shrink_to_fit!` helper with capacity checks.

The previous approach tried to avoid allocations by only calling `sizehint!` when capacity was >= 2x length, but this was overly complicated and almost never triggered in practice (arrays typically have 1.0-1.5x capacity/length ratio after the solve loop).

The simple approach: always call `sizehint!` after `resize!` in `_postamble!`. This adds a small constant number of allocations (3 for sol.t, sol.u, sol.k) but ensures excess memory is always released.

Updated the StabilizedRK allocation test expected counts to accommodate the postamble `sizehint!` allocations. Verified locally with `Pkg.test` on Julia 1.10, 1.11, and 1.12 — all pass.

lib/OrdinaryDiffEqStabilizedRK/test/rkc_tests.jl
```diff
@@ -129,8 +129,8 @@ end
         # check allocations
         integrator = init(prob, alg; save_everystep = false)
         allocs = @allocations solve!(integrator)
-        # Julia 1.11 has an extra allocation in these algorithms
-        expected_allocs = VERSION >= v"1.11" ? 4 : 3
+        # sizehint! calls in _postamble! add allocations to release excess memory
+        expected_allocs = VERSION >= v"1.11" ? 7 : 6
```
ChrisRackauckas, line 133: Reasonable trade-off?

oscardssmith: yeah

PR #3095, opened 2026-02-27
Add hook functions for SDE integrator unification

ChrisRackauckas-Claude:
## Summary

- Adds hook functions to OrdinaryDiffEqCore's time loop infrastructure (`_savevalues!`, `_postamble!`, `handle_callbacks!`, `loopheader!`, `_loopfooter!`, `apply_step!`) that allow StochasticDiffEq to reuse ODE's loop code while overriding SDE-specific behavior
- Declares these hooks as `public` for cross-package import (Julia 1.12)
- Zero behavioral change for ODE solvers — all hooks have defaults matching current ODE behavior

### Hook functions added (10 hooks + 2 traits):

| Hook | Default (ODE) | SDE Override |
|------|--------------|--------------|
| `post_step_reject!` | no-op | noise rejection |
| `on_u_modified_at_init!` | DAE reinit + FSAL update | just update uprev |
| `post_apply_step!` | FSAL update | noise accept + sqdt update |
| `interp_at_saveat` | polynomial (addsteps! + ode_interpolant) | linear (sde_interpolant) |
| `post_savevalues!` | resize k vectors | no-op |
| `save_dense_at_t!` | increment saveiter_dense + copy k | no-op |
| `skip_saveat_at_tspan_end` | skip at tspan[2] with save_end=false | never skip explicit saveat |
| `finalize_solution_storage!` | resize dense output | noise accept + save_noise |
| `finalize_endpoint!` | save final discretes | save final discretes |
| `on_callbacks_complete!` | FSAL re-eval + DDE discontinuities | error check + Poisson rate update |
| `is_composite_cache` (trait) | `cache isa CompositeCache` | + StochasticCompositeCache |
| `is_composite_algorithm` (trait) | `alg isa OrdinaryDiffEqCompositeAlgorithm` | + SDE composite algs |

### Motivation

StochasticDiffEq.jl duplicates ~200 lines of time loop code from OrdinaryDiffEqCore (`savevalues!`, `postamble!`, `handle_callbacks!`, `handle_tstop!`, `solution_endpoint_match_cur_integrator!`). These functions have diverged and the SDE versions miss bug fixes from the ODE. This PR enables code sharing via hooks.

## Test plan

- [x] OrdinaryDiffEqCore tests pass (JET 30 pass/1 broken, Aqua 9 pass, Sparse 10 pass)
- [x] No behavioral change — all hooks default to current ODE behavior
[...]

PR #3096, opened 2026-02-27
Move modify_dt_for_tstops\! into apply_step\! after update_fsal\!

ChrisRackauckas-Claude:
## Summary
- Calls `modify_dt_for_tstops!` inside `apply_step!` right after setting `dt = dtpropose`, so that dt is already shortened for tstops before any post-step work (like FSAL or future hook overrides) runs.
- The existing `modify_dt_for_tstops!` call in `loopheader!` remains for the rejection path and is a harmless no-op on the accept path since dt was already adjusted.

## Motivation
StochasticDiffEq's noise process acceptance needs the tstop-shortened dt to avoid overshooting noise grid boundaries. Currently SDE must call `modify_dt_for_tstops!` redundantly in its own `apply_step!` override. With this change in ODE, the dt is already correct by the time any post-step hook runs, simplifying the SDE unification (PRs #3095 / SciML/StochasticDiffEq.jl#682).

For ODE, `update_fsal!` evaluates `f(u, p, t)` which does not depend on the dt value, so reordering is safe.

## Test plan
- [x] OrdinaryDiffEqCore tests pass locally (JET, Aqua, sparse isdiag)
- [ ] CI passes

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
@@ -56,6 +56,13 @@ function apply_step!(integrator)
     end
 
     update_fsal!(integrator)
+
+    # Shorten dt to hit the next tstop after update_fsal!, which for DDEs calls
+    # handle_discontinuities! using integrator.dt to track propagated discontinuities
+    # in the interval [t, t+dt]. Must come after update_fsal! but before the next
+    # perform_step!. loopheader! also calls this unconditionally, which is a harmless
+    # no-op on the accept path since dt was already adjusted here.
+    modify_dt_for_tstops!(integrator)
     return nothing
 end
 


ChrisRackauckas-Claude:
## DDE test failure fix

The DDE convergence test failures (`KeyError: key :l∞ not found`) were caused by `modify_dt_for_tstops!` being placed **before** `update_fsal!`.

**Root cause:** For DDEs, `update_fsal!` → `handle_discontinuities!` tracks propagated discontinuities using `integrator.dt` to scan the interval `[t, t+dt]` (in `DelayDiffEq/src/track.jl`). When `modify_dt_for_tstops!` shortens `dt` before this tracking happens, the scanning interval is wrong, missing propagated discontinuities. This causes convergence failure.

**Fix:** Moved `modify_dt_for_tstops!` to come **after** `update_fsal!` instead of before it. The ordering is now:
1. `integrator.dt = integrator.dtpropose` (restore full step size)
2. `update_fsal!` (FSAL eval + DDE discontinuity tracking with full dt)
3. `modify_dt_for_tstops!` (shorten dt for next tstop)

This is still correct because `modify_dt_for_tstops!` only needs to run before the next `perform_step!`, and `loopheader!` also calls it as a safety net.

PR #3100, opened 2026-03-01
Refactor generic Rosenbrock perform_step! to converge with RosenbrockCache

ChrisRackauckas-Claude:
## Summary

- Replace macro-generated `perform_step!` code (`gen_perform_step`, `gen_constant_perform_step`, `gen_initialize`) with a shared loop function `_rosenbrock_stage_loop!()` used by both `GenericRosenbrockMutableCache` and `RosenbrockCache`
- Add `GenericRosenbrockConstantCache` abstract type and `_ks()` accessors for dispatch on generic caches
- Store array-based tableaus via `_make_rosenbrock_tableau()` instead of named-field structs, enabling uniform `tab.a[i,j]` access across all methods
- Fix existing bug where generated code never persisted `cache.linsolve = linres.cache`
- Net -60 lines: eliminates ~230 lines of code generation while converging two independent implementations into one shared loop

## Details

The generic Rosenbrock (W) methods previously used macro-based code generation to produce specialized `perform_step!` for 21 algorithms across 5 macro families (`@ROS2`, `@ROS23`, `@ROS34PW`, `@Rosenbrock4`, `@RosenbrockW6S4OS`). Meanwhile, `RosenbrockCache` (Rodas4/5/6P) already had a clean loop-based `perform_step!` with an identical inner loop structure. This PR converges them.

**Key design decisions:**
- `_rosenbrock_stage_loop!` handles stages 2..s, parameterized by `tab_a`, `dtC`, `dtd`, and `stage_limiter!` to accommodate differences between generic (no stage limiter, locally allocated dtC) and Rodas (stage limiter, pre-allocated dtC)
- Stage 1 stays in each `perform_step!` because the first `dolinsolve` call has different kwargs
- Accepts one extra `f` evaluation per step for 6 Rosenbrock4 algorithms (previously the `a4j=a3j` optimization avoided this), since the linear solve dominates cost
- Only the inplace (mutable cache) version shares the loop; the non-inplace version remains separate due to fundamentally different APIs (`dolinsolve` vs `W\b`)

## Test plan

- [x] All 26 algorithms pass smoke tests (both inplace and out-of-place)
- [x] `Pkg.test("OrdinaryDiffEqRosenbrock")` passes — convergence tests for all 21 generic + all Rodas methods

### DIFF
--- a/lib/OrdinaryDiffEqRosenbrock/src/generic_rosenbrock.jl
+++ b/lib/OrdinaryDiffEqRosenbrock/src/generic_rosenbrock.jl
@@ -166,7 +166,7 @@ because in the end of Rosenbrock's method, we have: `y_{n+1}=y_n+ki*bi`.
 """
 function gen_cache_struct(tab::RosenbrockTableau,cachename::Symbol,constcachename::Symbol)
     kstype=[:($(Symbol(:k,i))::rateType) for i in 1:length(tab.b)]
-    constcacheexpr=quote struct $constcachename{TF,UF,Tab,JType,WType,F} <: OrdinaryDiffEqConstantCache
+    constcacheexpr=quote struct $constcachename{TF,UF,Tab,JType,WType,F} <: GenericRosenbrockConstantCache
             tf::TF
             uf::UF
             tab::Tab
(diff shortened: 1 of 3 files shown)


lib/OrdinaryDiffEqRosenbrock/src/generic_rosenbrock.jl
```diff
@@ -204,11 +204,94 @@ function gen_cache_struct(tab::RosenbrockTableau,cachename::Symbol,constcachenam
...
 
 """
-    gen_algcache(cacheexpr::Expr,constcachename::Symbol,algname::Symbol,tabname::Symbol)
+    _make_rosenbrock_tableau(raw_tab, T, T2)
+
+Convert a raw `RosenbrockTableau` (with potentially undersized arrays for Rosenbrock4
+methods where a4j=a3j) into a properly-sized, type-converted tableau suitable for the
+generic loop-based `perform_step!`.
+"""
+function _make_rosenbrock_tableau(raw_tab::RosenbrockAdaptiveTableau, ::Type{T}, ::Type{T2}) where {T, T2}
```
oscardssmith, line 213: Do we need any of this? Can't we just port these methods to the RosenbrockTableau directly?

PR #3108, opened 2026-03-02
Add delta and save_noise fields to DEOptions for SDE unification

ChrisRackauckas-Claude:
## Summary
- Adds `delta::DType` and `save_noise::Bool` fields to `DEOptions` so SDE can use the same options struct instead of maintaining a separate `SDEOptions`
- Adds `DType` type parameter to `DEOptions` (defaults to `typeof(nothing)` for ODE)
- Fixes `_has_ks` trait to check both `integrator.k` and `integrator.sol.k` exist (prevents `FieldError` when tau_leaping uses `SDEIntegrator + ODESolution`)
- ODE passes `nothing` for `delta` and `false` for `save_noise` — zero overhead

This is a companion PR to SciML/StochasticDiffEq.jl#682, which eliminates `SDEOptions` by constructing `DEOptions` directly.

## Test plan
- [x] ODE Core tests pass (JET, Aqua, Sparse isdiag Performance)
- [x] SDE Interface1 tests pass (all 12+ test groups, ~281 assertions)
- [x] Targeted tau_leaping test passes with `_has_ks` fix
- [x] Runic formatted


lib/OrdinaryDiffEqCore/src/integrators/type.jl
```diff
@@ -37,10 +37,12 @@ mutable struct DEOptions{
     progress_id::Symbol
     timeseries_errors::Bool
     dense_errors::Bool
+    delta::DType
```
termi-official, line 40: Can we start adding at least some brief docs when we introduce new fields to opts (and the integrator)? When I started digging into this Core to build solvers it was extremely difficult for me to track down what do all these fields mean and which are the relevant ones for my solver.

ChrisRackauckas: Let's finish the unification first. This will make a lot more sense when StochasticDiffEq and DelayDiffEq are in the repo.

ChrisRackauckas-Claude:
## Added: Legacy DEOptions constructor for backwards compatibility

Commit `31ee41eec` adds an outer constructor that accepts the old signature (20 type params, 45 positional args — without `delta`, `save_noise`, `save_discretes`, `verbType`, `DType`). This allows DelayDiffEq and other packages that construct `DEOptions` directly to work without changes.

The legacy constructor fills in:
- `delta = nothing`
- `save_noise = false`
- `save_discretes = false`
- `verbType` inferred from `typeof(verbose)`
- `DType` inferred as `typeof(nothing)`

Verified that both the legacy and new constructors work, and all ODE Core tests pass.

PR #3110, opened 2026-03-02
Add legacy DEOptions constructor for DelayDiffEq compatibility

ChrisRackauckas-Claude:
## Summary
- Adds an outer constructor for `DEOptions` that accepts the old signature (20 type params, 45 positional args) — without `delta`, `save_noise`, `save_discretes`, `verbType`, `DType`
- This allows DelayDiffEq (and other packages that construct `DEOptions` directly) to work without changes after #3108 added new fields
- Fills in defaults: `delta=nothing`, `save_noise=false`, `save_discretes=false`, infers `verbType` and `DType`

## Context
PR #3108 added `delta`, `save_noise` fields and `DType` type parameter to `DEOptions`. This broke the positional constructor used by DelayDiffEq, which also predates the `save_discretes` field and `verbType` type parameter.

## Test plan
- [x] Legacy constructor verified to work with 20 type params / 45 args
- [x] Full (new) constructor still works
- [x] ODE Core tests pass (JET, Aqua, Sparse isdiag Performance)
- [x] Runic formatted

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/integrators/type.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/type.jl
@@ -56,12 +56,12 @@ mutable struct DEOptions{
 end
 
 # Legacy constructor for backwards compatibility with packages (e.g. DelayDiffEq)
-# that construct DEOptions without the delta, save_noise, save_discretes fields
-# and without the verbType, DType type parameters.
+# that construct DEOptions without the delta and save_noise fields and without
+# the DType type parameter. Accepts 21 type params and 46 positional args.
 function DEOptions{
         absType, relType, QT, tType, Controller, F1, F2, F3, F4, F5, F6,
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Fixed the legacy constructor to match the CI (registered) version of DelayDiffEq's calling convention:

- **Before**: 20 type params, 45 positional args (matched an older dev version)
- **After**: 21 type params (including `verbType`), 46 positional args (including `save_discretes`)

The CI failure was because the registered DelayDiffEq already includes `verbType` in its type parameter list and `save_discretes` in its positional args when constructing `DEOptions`. The constructor only needed to fill in the two new fields (`delta=nothing`, `save_noise=false`) and infer `DType=typeof(nothing)`.

Verified locally:
- DDE downstream test: 27/27 pass
- ODE Core tests: all pass (JET 30, Aqua 9, Sparse 10)

PR #3112, opened 2026-03-02
[WIP] Implicit Taylor solvers

tansongchen:
## Checklist


lib/OrdinaryDiffEqTaylorSeriesImplicit/src/algorithms.jl
```diff
@@ -0,0 +1,126 @@
...
+        extra_keyword_default::String = "")
+    keyword_default = """
+        chunk_size = Val{0}(),
+        autodiff = AutoForwardDiff(),
+        standardtag = Val{true}(),
+        concrete_jac = nothing,
+        diff_type = Val{:forward}(),
+        linsolve = nothing,
+        precs = DEFAULT_PRECS,
+        nlsolve = NLNewton(),
```
oscardssmith: this isn't used for anything

lib/OrdinaryDiffEqTaylorSeriesImplicit/test/runtests.jl
```diff
@@ -0,0 +1,32 @@
...
+
+    μ = 1.0
+
+    # sol = solve(prob, ImplicitTaylor1(μ = μ, extrapolant = :linear))
+
+    # @show sol.stats
+
+    sim11 = test_convergence(dts, prob, ImplicitTaylor1(μ = μ, extrapolant = :linear))
+    @test sim11.𝒪est[:final]≈1 atol=testTol
+end
```
oscardssmith, line 59: add stiff problem test (e.g. pollu)

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -0,0 +1,181 @@
...
+    dw = cache.linsolve.u
+    while iter < maxiters
+        iter += 1
+        integrator.stats.nnonliniter += 1
+        # evaluate function
+        tmp .= reshape(nlf(_vec(u)), size(tmp))
+        #@show tmp
+        linsolve = cache.linsolve
+        needfactor = iter == 1 # && newW once we aren't using ForwardDiff
+        A =  needfactor ? ForwardDiff.jacobian(nlf, _vec(u)) : nothing
```
oscardssmith: TODO: use taylor diff for `dJet`

tansongchen: Just did!

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -77,38 +77,33 @@ end
...
+    # the reshape might be moved to build_jet
     nlf = (u_next_vec) -> begin
         u_next = reshape(u_next_vec, size(uprev))
-        utaylor = reshape(cache.jet(u_next, t + μ * dt), size(uprev))
-        du = get_coefficient(utaylor, 1)
-        _vec(u_next .- uprev .- μ * dt .* du)
+        ut_vec = propagator(u_next, t + μ * dt, -μ * dt)
+        ut = reshape(ut_vec, size(uprev))
+        _vec(ut .- uprev)
     end
```
oscardssmith: can't we delete `nlf` now?

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -77,38 +77,33 @@ end
...
     while iter < maxiters
         iter += 1
         integrator.stats.nnonliniter += 1
         # evaluate function
         tmp .= reshape(nlf(_vec(u)), size(tmp))
         #@show tmp
-        linsolve = cache.linsolve
         needfactor = iter == 1 # && newW once we aren't using ForwardDiff
-        A =  needfactor ? ForwardDiff.jacobian(nlf, _vec(u)) : nothing
+        A =  needfactor ? d_propagator(_vec(u), t + μ * dt, -μ * dt) : nothing
```
oscardssmith: TODO: make this in place for the iip version.

tansongchen: Done. Confirmed that no allocation in `perform_step!`

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
tansongchen: deleted

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -0,0 +1,179 @@
...
+        # TODO: check numerical stability
+        uprev2 = integrator.uprev2
+        tprev = integrator.tprev
+
+        dt1 = dt * (t + dt - tprev)
+        dt2 = (t - tprev) * (t + dt - tprev)
+        c = 7 / 12 # default correction factor in SPICE (LTE overestimated by DD)
+        r = c * dt^2 # by mean value theorem 2nd DD equals y''(s)/2 for some s
+
+        @.. broadcast = false tmp = r * integrator.opts.internalnorm(
```
oscardssmith:
```suggestion
        @..  tmp = r * integrator.opts.internalnorm(
```
broadcast=false is the default. no need to spell it out.

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_caches.jl
```diff
@@ -0,0 +1,90 @@
...
+    # get the iip version
+    _, propagator = propagator_all
+    _, d_propagator = d_propagator_all
+    fsalfirst = zero(rate_prototype)
+    atmp = similar(u, uEltypeNoUnits)
+    recursivefill!(atmp, false)
+    utilde = zero(u)
+    tmp = zero(u)
+    uToltype = constvalue(uBottomEltypeNoUnits)
+    ut = zero(u)
```
oscardssmith: give this a better name

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -0,0 +1,179 @@
...
+    η = max(ηold, eps(eltype(integrator.opts.reltol)))
+    fail_convergence = true
+    iter = 0
+    dw = linsolve.u
+    while iter < maxiters
+        iter += 1
+        integrator.stats.nnonliniter += 1
+        # calculate residuals and Jacobian
+        propagator(ut, u, tc, -μ * dt)
+        tmp .= ut .- uprev
```
oscardssmith: can this just be 
```suggestion
        propagator(tmp, u, tc, -μ * dt)
        tmp  .-= uprev
```
(to save a buffer) this is only useful if we can eliminate `ut` entirely, but I think we can.

tansongchen: I am probably not going to eliminate this since it will be useful when complex mu is introduced

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -0,0 +1,183 @@
...
+        return
+    end
+    cache.ηold = η
+    cache.iter = iter
+
+    if μ != one(μ)
+        propagator(tmp, uintermediate, tc, (1 - μ) * dt)
+        u .= real.(tmp)
+    else
+        u .= real.(uintermediate)
```
oscardssmith: `μ == one(μ)` implies that we shouldn't be using complex `μ` so the `real` here is unnecessary.

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -0,0 +1,183 @@
...
+        integrator.force_stepfail = true
+        integrator.stats.nnonlinconvfail += 1
+        return
+    end
+    cache.ηold = η
+    cache.iter = iter
+
+    if μ != one(μ)
+        propagator(tmp, uintermediate, tc, (1 - μ) * dt)
+        u .= real.(tmp)
```
oscardssmith: For complex `μ`, do we only support real functions or do holomorphic functions work as well? If the latter, this should be 
```suggestion
        u .= eltype(u).(tmp)
```

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/alg_utils.jl
```diff
@@ -0,0 +1,5 @@
+alg_extrapolates(::ImplicitTaylor) = true
+
+alg_order(::ImplicitTaylor{P}) where {P} = P
```
oscardssmith: `P` is the number of stages. The order of the method is different (you likely want to have both be in the integrator)

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/algorithms.jl
```diff
@@ -0,0 +1,141 @@
...
+    κ::Tol
+    maxiters::Int
+end
+
+function get_order_param(order::Val{P}) where {P}
+    return P
+end
+
+function ImplicitTaylor(;
+        order = Val(1), μ = 1.0, chunk_size = Val{0}(), autodiff = AutoForwardDiff(),
```
oscardssmith: this doesn't need most of these keywords since they are unused.

* autodiff
* chunk_size
* standardtag
* concrete_jac
* diff_type
* nlsolve

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/algorithms.jl
```diff
@@ -0,0 +1,141 @@
...
+    autodiff::AD
+    κ::Tol
+    maxiters::Int
+end
+
+function get_order_param(order::Val{P}) where {P}
+    return P
+end
+
+function ImplicitTaylor(;
```
oscardssmith, line 41: Once we figure out which methods make sense we likely want named constructors for the good methods (that set `order` and `μ` appropriately).

oscardssmith: Possible naming scheme would be `ImplicitTaylorA2` for 2nd order A stable method, `ImplicitTaylorLA4` for A+L stable 4th order etc. First we obviously need to figure out which the good methods actually are though.

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
tansongchen: Agree, but calling constructor `Real` on complex numbers will cause InexactError. I need to find some other way to achieve that.

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/algorithms.jl
tansongchen: These are used in type parameters in OrdinaryDiffEqNewtonAdaptiveAlgorithm{CS, AD, FDT, ST, CJ}, so I'm not sure how to remove them. Maybe we should just use `<: OrdinaryDiffEqAdaptiveAlgorithm`?

oscardssmith: > Maybe we should just use <: OrdinaryDiffEqAdaptiveAlgorithm

yeah.

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/algorithms.jl
tansongchen: Some assistant constructors added. I'll begin testing performance shortly

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
tansongchen: Whether to project onto real numbers is added as a parameter (has an impact on order)

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_caches.jl
```diff
@@ -0,0 +1,93 @@
...
+        linprob, alg.linsolve, alias = LinearAliasSpecifier(alias_A = true, alias_b = true),
+        assumptions = LinearSolve.OperatorAssumptions(true)
+    )
+
+    ηold = one(uToltype)
+    iter = 10000
+    status = Convergence
+
+    return ImplicitTaylorCache(
+        alg.μ, propagator, d_propagator, u, uprev, uprev2, J, utilde, uintermediate, tmp, atmp, fsalfirst, linsolve,
```
oscardssmith:
```suggestion
        convert(typeof(t), alg.μ), propagator, d_propagator, u, uprev, uprev2, J, utilde, uintermediate, tmp, atmp, fsalfirst, linsolve,
```

lib/OrdinaryDiffEqTaylorSeriesImplicit/test/runtests.jl
tansongchen: For stiff problems, we cannot test for the convergence (since we don't have analytic solution), so what are we testing for? or do we just assert that the solve is finished in some number of steps

tansongchen:
I have some problem with the adaptivity. When I try to use a higher order solver (and the embedded is also higher), I saw the `integrator.EEst` decreasing, but somehow `dt` gets smaller instead of bigger. I feel there should be a bug but I couldn't find where

Minimal example to reproduce:

```julia
using OrdinaryDiffEq, Plots, DiffEqDevTools, LinearAlgebra, BenchmarkTools, ODEProblemLibrary
using OrdinaryDiffEqTaylorSeriesImplicit

prob = ODEProblemLibrary.prob_ode_hires
solve(prob, ImplicitTaylor(order = Val(1), order_q = Val(2)), abstol = 1.0e-3, reltol = 1.0e-3) # order 3, embedded order 2
solve(prob, ImplicitTaylor(order = Val(2), order_q = Val(3)), abstol = 1.0e-3, reltol = 1.0e-3) # order 5, embedded order 4
```

lib/OrdinaryDiffEqTaylorSeriesImplicit/test/runtests.jl
oscardssmith: we can test vs a high precision solution.

oscardssmith:
Are you setting the order appropriately? If the timestepping is getting the wrong order it might do weird things...

lib/OrdinaryDiffEqTaylorSeriesImplicit/test/runtests.jl
tansongchen: got it

tansongchen:
> Are you setting the order appropriately? If the timestepping is getting the wrong order it might do weird things...

I think so, if I'm understanding these APIs correctly

image

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -113,7 +115,7 @@ end
     fail_convergence = true
     iter = 0
     dw = linsolve.u
-    uintermediate .= u
+    uintermediate .= uprev
```
oscardssmith: `u` and `uprev` should be identical at this point, right?

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -148,13 +150,17 @@ end
...
+        # if η * ndw < κ && (iter > 1 || iszero(ndw) || !iszero(integrator.success_iter))
+        #     # Newton method converges
+        #     cache.status = Convergence
+        #     fail_convergence = false
+        #     break
+        # end
     end
+    # FIXME: ignore stopping criterion
+    cache.status = Convergence
+    fail_convergence = false
```
oscardssmith: is this still breaking?

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -0,0 +1,206 @@
...
+        else
+            A = nothing
+        end
+        linres = dolinsolve(integrator, linsolve; A, b = _vec(tmp), linu = dw)
+        cache.linsolve = linres.cache
+        integrator.stats.nsolve += 1
+
+        # compute norm of residuals
+        iter > 1 && (ndwprev = ndw)
+        calculate_residuals!(atmp, dw, uprev, uintermediate, abstol, reltol, internalnorm, t)
```
oscardssmith, line 140: Would be good to double check that this is doing something useful.

PR #3118, opened 2026-03-03
Update NonlinearSolveFirstOrder compat

jaakkor2:
Tests passed locally.

ImplicitDiscreteSolve is holding NonlinearSolveFirstOrder at v1.11.1, that holds NonlinearSolve at 4.15.0, and so on..


ChrisRackauckas:
yeah oops this one was missed thanks.

PR #3121, opened 2026-03-03
Implement Discontinuity Detection Algorithm

Shreyas-Ekanathan:
## Checklist

## Additional context

A preliminary implementation of autonomous discontinuity detection in Radau. Uses a PR to SciML base at https://github.com/SciML/SciMLBase.jl/pull/1219 to modify callbacks to support a is_discontinuous flag.


oscardssmith (approved):

ChrisRackauckas:
`disco_handling` should be `discontinuity_detection`, and needs a docstring, then I think this is good to go.

lib/OrdinaryDiffEqCore/test/disco_benchmarks.jl
```diff
@@ -0,0 +1,79 @@
...
+# initial condition at t = 0
+u0 = [1.0]
+tspan = (0.0, 10.0)
+p = nothing
+prob = DDEProblem(f, u0, h, tspan, p; dependent_lags = (delay,))
+sol = solve(prob, MethodOfSteps(Tsit5()))
+
+# https://dieci.math.gatech.edu/preps/DieciLopez-Fili4.pdf   
+# vector fields
+# BELOW IS WRONG NEED TO FIX
```
ChrisRackauckas: ?

lib/OrdinaryDiffEqCore/test/disco_benchmarks.jl
ChrisRackauckas, line 1: this is not in the tests?

lib/OrdinaryDiffEqCore/test/disco_tests.jl
ChrisRackauckas, line 1: this is not in the tests?

lib/OrdinaryDiffEqCore/test/disco_benchmarks.jl
```diff
@@ -0,0 +1,79 @@
...
+sol_no_disco_tsit5 = solve(prob, Tsit5(), callback=cb2)
+
+sol_disco_radau = solve(prob, RadauIIA5(), callback=cb)
+sol_no_disco_radau = solve(prob, RadauIIA5(), callback=cb2)
+
+sol_disco_rosenbrock = solve(prob, Rodas5P(), callback=cb)
+sol_no_disco_rosenbrock = solve(prob, Rodas5P(), callback=cb2)
+
+sol_disco_bs3 = solve(prob, BS3(), callback=cb)
+sol_no_disco_bs3 = solve(prob, BS3(), callback=cb2)
```
ChrisRackauckas: should these turn into tests?

lib/OrdinaryDiffEqCore/src/integrators/type.jl
```diff
@@ -175,6 +175,8 @@ mutable struct ODEIntegrator{
     fsalfirst::FSALType
     fsallast::FSALType
     rng::RNGType
+    #disco_prob::IntervalNonlinearProblem
+    disco_probs::Vector{IntervalNonlinearProblem} #should we change this?
```
ChrisRackauckas: this isn't concrete

lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -16,6 +16,36 @@ determine_controller_datatype(u, internalnorm, ts::Tuple{<:Number, <:Number}) =
...
+function (z::zero_func_struct)(θ, p)
+    _ode_addsteps!(z.k, z.tprev, z.uprev, z.u, z.dt, z.f, z.p, z.cache, false, true, false)
+    ode_interpolant!(z.u₁, θ, z.dt, z.uprev, z.u, z.k, z.cache, z.idxs, Val{0}, z.differential_vars)
+    return zero_condition(z.callback, z.out, z.u₁, z.dt + θ * z.dt, z, z.ind)
+end
+
+@inline zero_condition(cb::ContinuousCallback, out::Nothing, u, t, z, ind) = cb.condition(u, t, z)
+@inline function zero_condition(cb::VectorContinuousCallback, out, u, t, z, ind)
+    cb.condition(out, u, t, z)
+    return out[ind]
```
ChrisRackauckas, line 49: how is the ind chosen?

lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -16,6 +16,36 @@ determine_controller_datatype(u, internalnorm, ts::Tuple{<:Number, <:Number}) =
...
+end
+
+function (z::zero_func_struct)(θ, p)
+    _ode_addsteps!(z.k, z.tprev, z.uprev, z.u, z.dt, z.f, z.p, z.cache, false, true, false)
+    ode_interpolant!(z.u₁, θ, z.dt, z.uprev, z.u, z.k, z.cache, z.idxs, Val{0}, z.differential_vars)
+    return zero_condition(z.callback, z.out, z.u₁, z.dt + θ * z.dt, z, z.ind)
+end
+
+@inline zero_condition(cb::ContinuousCallback, out::Nothing, u, t, z, ind) = cb.condition(u, t, z)
+@inline function zero_condition(cb::VectorContinuousCallback, out, u, t, z, ind)
```
ChrisRackauckas: this seems like code smell... it needs to do the whole array not an ind.

Shreyas-Ekanathan: but for a VCC we want to check per condition if it was hit, so the approach in disco is to iterate over the entire VCC and find the components that could be the discontinuity:

            i.condition(out_prev, uprev, t, integrator)
            i.condition(out_curr, u, t + dt, integrator)
            for j in 1:len_cb
                if (out_prev[j] * out_curr[j] < zero(out_prev[j]))
                    disco_zero.ind = j
                    sol = solve(disco_prob; bracket = bracket)
                    tmp = sol[]
                    if (!isnan(tmp) && (breakpointθ == -1 || tmp < breakpointθ)) 
                        breakpointθ = tmp 
                    end 
                end
            end


So we do disco on each index of the VCC that could have triggered the step reject

lib/OrdinaryDiffEqCore/src/integrators/type.jl
Shreyas-Ekanathan: yeah what Oscar and I had discussed was that once the callbacks are function wrapped we would store a single disco_prob and remake it each time, instead of storing a vector of all of them. Not sure if you had another approach in mind though

Shreyas-Ekanathan:
@ChrisRackauckas @oscardssmith If you could review when you get a chance, that would be great. Test failures seem to be a strange error independent of this.

ChrisRackauckas:
The CI machine maybe ran out of disk space?

ChrisRackauckas:
The test failure looks real.

Shreyas-Ekanathan:
> The test failure looks real.

I'm not really sure what is failing here (e.g. failure in AD test), `DummyControllerCache` does have the `discontinuity_detection` field and we import `set_discontinuity` into BDF, so little lost with this one.

lib/OrdinaryDiffEqCore/src/integrators/type.jl
ChrisRackauckas: If this isn't used could it just be nothing so there's nothing dynamic? I think things like trimming really don't like this.

lib/OrdinaryDiffEqCore/Project.toml
```diff
@@ -1,99 +1,101 @@
...
+ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
 Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
+Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
 ConcreteStructs = "2569d6c7-a4a2-43d3-a901-331e8e4be471"
-FunctionWrappersWrappers = "77dc65aa-8811-40c2-897b-53d922fa7daf"
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
+DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
+DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+EnumX = "4e289a0a-7415-4d19-859d-a7e5c4648b56"
```
ChrisRackauckas, line 16: What's all of this?

Shreyas-Ekanathan: I added FunctionWrappers to core and I think it shuffled things around

lib/OrdinaryDiffEqCore/src/integrators/type.jl
Shreyas-Ekanathan: disco_probs is used, we pre-create all of the problems during init, and then just update parameters as needed for disco solves (set current u, t, etc).

Shreyas-Ekanathan: the "should we change?" this was referring to the type instability but after function wrapping the zero_func_struct, the issue is much better (still not perfect but closer).

lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -38,7 +42,7 @@ end
 
 parameter_values(z::zero_func_struct) = z.p
 
-function (z::zero_func_struct)(θ, p)
+function (z::zero_func_struct)(θ, p)::Float64
```
oscardssmith: this is wrong.

lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -16,8 +16,12 @@ determine_controller_datatype(u, internalnorm, ts::Tuple{<:Number, <:Number}) =
 determine_controller_datatype(u::AbstractVector{<:Number}, internalnorm, ts::Tuple{<:Integer, <:Integer}) = promote_type(typeof(DiffEqBase.value(internalnorm(u, ts[1]))), typeof(DiffEqBase.value(internalnorm(u, ts[2]))), eltype(float.(DiffEqBase.value(ts))))
 determine_controller_datatype(u, internalnorm, ts::Tuple{<:Integer, <:Integer}) = promote_type(typeof(float(DiffEqBase.value(ts[1]))), typeof(float(DiffEqBase.value(ts[2])))) # This seems to be an assumption implicitly taken somewhere
 
+struct DiscoProblem{ZF, NLP}
```
oscardssmith: why do we need this? doesn't the nlp hold the zerofunc?

Shreyas-Ekanathan: I was working with Claude and it suggested that the extraction of the zerofunc from the NLP might create some of the type instability we were seeing so I tried this out and it seemed to help from a performance perspective. I removed functionwrappers for the same reason because it was hurting the type stability

ChrisRackauckas: that's voodoo, it wouldn't effect type-stability. Where's the actual inference lost?

Shreyas-Ekanathan: so the instability happens during the updates to `disco_zero` in `find_discontinuity`, the root cause appears to be the `Vector{IntervalNonLinearProblem}` which leaves us no info about the `zero_func_struct` contained inside each function and creates the inference loss.

Shreyas-Ekanathan: so the previous fix would actually help something I think, becuase when you put everything into that custom wrapper that stores the information about the `zero_func_struct`, we maintain that type info and so the updates become stable

ChrisRackauckas: A functionwrapper is effectively a type with that information 😅 what's the issue you're hitting with that?

lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -99,6 +132,7 @@ Base.@constprop :aggressive function _ode_init(
         alias = ODEAliasSpecifier(),
         initializealg = DefaultInit(),
         rng = nothing,
+        disco_probs = nothing,
```
ChrisRackauckas: what is this for? I thought it was grabbed from the callabacks?

lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -626,6 +660,35 @@ Base.@constprop :aggressive function _ode_init(
 
     _rng = rng === nothing ? Random.default_rng() : rng
 
+    num_probs = 0
+    for i in callbacks_internal.continuous_callbacks
+        if i.maybe_discontinuity
+            num_probs += 1
+        end
+    end
+    disco_probs = Vector{IntervalNonlinearProblem}(undef, num_probs)
```
ChrisRackauckas, line 666: this immediately overwrites the kwarg?

lib/OrdinaryDiffEqFIRK/src/controllers.jl
```diff
@@ -53,14 +53,22 @@ function step_reject_controller!(
         integrator, ccache::PredictiveControllerCache, alg::AdaptiveRadau
     )
     (; controller) = ccache
-    (; cache) = controller
-    (; dt, success_iter) = integrator
+    (; discontinuity_detection) = controller.basic
+    (; dt, success_iter, cache) = integrator
```
ChrisRackauckas, line 57: this is a different cache from the controller cache?

lib/OrdinaryDiffEqCore/src/solve.jl
Shreyas-Ekanathan: the idea is that instead of creating and allocating a new problem object each time we hit an event, we will just frontload the creation of all of the event problems (which internally use the callbacks) to determine the discontinuity events.

lib/OrdinaryDiffEqCore/src/solve.jl
Shreyas-Ekanathan: yes but in practice no one is passing in this object since it will be internally constructed based on the callbacks provided

lib/OrdinaryDiffEqFIRK/src/controllers.jl
Shreyas-Ekanathan: yes the algorithm cache is taken out of the integrator for the adaptivity fields, distinct from the controller cache which tells you whether or not to run disco

lib/OrdinaryDiffEqCore/src/solve.jl
oscardssmith: I think this happened before we had a dedicated controller cache. Now that exists so we should probably use it.

Shreyas-Ekanathan: Refactored into controller cache and got tests passing again

PR #3127, opened 2026-03-10
multiply initial dt by random number to prevent numerical coincendences

oscardssmith:
Previously our initial dt was very likely to be a power of 10 which can cause odd numerical issues (e.g. if it lines up exactly with the period of a sign wave). This is exacerbated dramatically with the inital_qmax raise to 10000 since a dt of 1e-4 amplifys to a dt of 1.0 after an inital step (if the inital step had no error) and something like ~30% of ODEs have something special happen at t=1

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/initdt.jl
+++ b/lib/OrdinaryDiffEqCore/src/initdt.jl
@@ -10,7 +10,7 @@
     p = integrator.p
     oneunit_tType = oneunit(_tType)
     # a number that's about 1 that we multiply by to prevent numerical coincedences
-    arbitrary_const = oneunit_tType*(84//83)
+    arbitrary_const = oneunit_tType*(93//83)
     dtmax_tdir = tdir * dtmax
 
     dtmin = nextfloat(max(integrator.opts.dtmin, eps(t)))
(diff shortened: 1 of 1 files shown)


lib/OrdinaryDiffEqCore/src/initdt.jl
```diff
@@ -9,10 +9,12 @@
     f = prob.f
     p = integrator.p
     oneunit_tType = oneunit(_tType)
+    # a number that's about 1 that we multiply by to prevent numerical coincedences
+    arbitrary_const = oneunit_tType*(84//83)
```
ChrisRackauckas: Why not have it a bit wilder? That still might not hit adaptivity changes. 277//83? It's actually closer to being a multiple of machine epsilon.

oscardssmith: I'd be good with 87/83 or something, but I want the value to be closeish to 1 because I don't want us to pick a dt that's too large because we multiplied by a completely arbitrary constant.

ChrisRackauckas: 93.

ChrisRackauckas:
a lot more failures than expected

oscardssmith:
I think a number of them are failing on master it's hard to imagine that this change would cause failures in performstep

ChrisRackauckas:
No

PR #3141, opened 2026-03-17
Add IMEX Rosenbrock tableau framework and IMEXRKR_3_2 implementation …

singhharsh1708:
## Checklist

- [x] The new code follows the contributor guidelines, in particular the SciML Style Guide and COLPRAC

## Additional context

This PR introduces a tableau-based abstraction for IMEX Rosenbrock methods and implements IMEXRKR_3_2 as an initial method.

The implementation targets split ODE systems and is structured to allow extension to additional IMEX RK / Rosenbrock methods via reusable tableau definitions.

Validation performed:
- Verified second-order convergence through step-size refinement tests
- Compared solutions against high-accuracy reference solvers
- Confirmed compatibility with in-place and out-of-place formulations

All existing Rosenbrock test suites pass with no regressions.

### DIFF
--- a/lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_interpolants.jl
+++ b/lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_interpolants.jl
@@ -7,7 +7,6 @@ ROSENBROCKS_WITH_INTERPOLATIONS = Union{
     RosenbrockCombinedConstantCache,
     RosenbrockCache,
     HybridExplicitImplicitConstantCache, HybridExplicitImplicitCache,
-    IMEXRKR_3_2ConstantCache, IMEXRKR_3_2Cache,
 }
 
 function _ode_interpolant(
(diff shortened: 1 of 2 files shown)


singhharsh1708:
Rebased onto latest upstream/master and cleaned up the changes. CI is passing now

ChrisRackauckas:
Rebase onto the new testing infrastructure.

singhharsh1708:
@ChrisRackauckas can you run tests on all prs so i can fix it if theres error?

singhharsh1708:
@ChrisRackauckas can you run test on this?

lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_interpolants.jl
```diff
@@ -581,3 +582,71 @@ end
     end
     out
 end
+
+@muladd function _ode_interpolant(
```
ChrisRackauckas: why did you add these?

lib/OrdinaryDiffEqRosenbrock/test/imexrkr_tests.jl
ChrisRackauckas, line 1: standard convergence tests are missing, and dense tests

lib/OrdinaryDiffEqRosenbrock/src/stiff_addsteps.jl
```diff
@@ -271,3 +271,11 @@ function _ode_addsteps!(
     end
     return nothing
 end
+
+function _ode_addsteps!(
```
ChrisRackauckas: I don't see why this has a special interpolation in the first place.

singhharsh1708: For IMEXRKR methods, k[1]/k[2] store Rosenbrock stage vectors rather than time derivatives. The default Hermite interpolation assumes derivative information and would therefore be incorrect here. We override with a linear interpolant that does not use k.

singhharsh1708:
- Removed the redundant `_ode_addsteps!` override (generic fallback already handles this)
- Added standard convergence tests using DiffEqDevTools (in-place and out-of-place)
- Added dense output tests with tighter tolerance
For interpolation: IMEXRKR stores Rosenbrock stage vectors in `k`, not time derivatives. The default Hermite interpolation assumes derivative information and would therefore be incorrect. The override uses a simple linear interpolation that avoids using `k`.
Let me know if you'd prefer a different approach for handling interpolation.

lib/OrdinaryDiffEqRosenbrock/src/stiff_addsteps.jl
ChrisRackauckas: If k is empty then linear interpolant is done. But also, why not store the time derivatives for k? You already computed them.

singhharsh1708: right — updated to store kshortsize = 0 so k stays empty and the generic _ode_addsteps! handles interpolation correctly via lazy f evaluation. Removed all custom interpolant code entirely.

ChrisRackauckas:
Almost good to go. I'd move these to a separate IMEXRosenbrock sublibrary though.

oscardssmith:
Is this actually an imex method? It looks to me like every function call `f_nonstiff` is followed by adding `f_stiff` called with the same parameters... Is the IMEX part that the jacobian is only the jacobian of the nonstiff part? If so, this should probably be in a comment before the `calc_W` line or something... Relatedly, @ChrisRackauckas/@gstein3m  could all RosenbrockW be accelerated for SplitFunction by just ignoring the nonstiff contribution the the jacobian?

singhharsh1708:
> Is this actually an imex method? It looks to me like every function call `f_nonstiff` is followed by adding `f_stiff` called with the same parameters... Is the IMEX part that the jacobian is only the jacobian of the nonstiff part? If so, this should probably be in a comment before the `calc_W` line or something... Relatedly, @ChrisRackauckas/@gstein3m  could all RosenbrockW be accelerated for SplitFunction by just ignoring the nonstiff contribution the the jacobian?

Yes — the IMEX part is that the Jacobian used in calc_W! is only of f1 (the stiff/implicit part), not the full f. The nonstiff part g is treated explicitly using the same stage point. I’ll add a comment before the calc_W! line clarifying this. The f_nonstiff and f_stiff calls using the same parameters is correct per the method definition in Huang et al. 2021 eq. (4.16)

oscardssmith:
> Yes — the IMEX part is that the Jacobian used in calc_W! is only of f1 (the stiff/implicit part), not the full f

Is this method just Rosenbrock32 with a different way of handling the jacobian? If so, does this need to be a separate method or can it just be an option to RosenbrockW methods?

ChrisRackauckas:
Yeah, this is just Rosenbrock32 (which should be Rosenbrock23 😅) with AMF. So I think this should just become part of AMF.

ChrisRackauckas:
@utkarsh530 it would be interesting to note this in the paper.

singhharsh1708:
Thanks, that makes sense this is essentially Rosenbrock23 with AMF rather than a distinct IMEX method. I'll rework this as an AMF option within the existing Rosenbrock methods instead of a separate implementation later.

gstein3m:
> Thanks, that makes sense this is essentially Rosenbrock23 with AMF rather than a distinct IMEX method. I'll rework this as an AMF option within the existing Rosenbrock methods instead of a separate implementation later.

Yes, that would be a good option for all Rosenbrock-W methods.

PR #3142, opened 2026-03-17
Split sublibrary CI to only test on dependency changes

ChrisRackauckas-Claude:
## Summary

- **Splits sublibrary tests out of `CI.yml`** into a new `SublibraryCI.yml` workflow that only runs tests for sublibraries affected by changed files
- **Adds `.github/scripts/compute_affected_sublibraries.py`** which reads the dependency graph from all `lib/*/Project.toml` files (both `[deps]` and `[extras]`), computes the transitive reverse-dependency closure, and maps changed file paths to affected sublibraries
- **Global tests always run**: Interface, Integrators, AlgConvergence, Regression, Downstream, ModelingToolkit, QA, AD — unchanged in `CI.yml`

### How it works

1. `detect-changes` job runs `git diff` and pipes changed files into the Python script
2. Script parses all `lib/*/Project.toml` to build the internal dependency graph
3. For each changed `lib/<pkg>/...` file, the script includes that package **plus all transitive reverse dependencies**
4. The affected list becomes a dynamic GitHub Actions matrix for both functional tests (4 Julia versions) and QA tests (Julia 1 only)

### Example behavior

| Changed file | Sublibraries tested |
|---|---|
| `lib/OrdinaryDiffEqCore/src/...` | All 36 sublibraries |
| `lib/OrdinaryDiffEqTsit5/src/...` | Tsit5 + Default + Nordsieck |
| `lib/OrdinaryDiffEqFeagin/src/...` | Only Feagin |
| `test/interface/...` or `src/...` | No sublibrary tests (global CI still runs) |

### Benefits

- **Massive CI savings on targeted changes** — e.g., a change to OrdinaryDiffEqFeagin only runs 5 jobs (4 versions + QA) instead of ~170
- **Self-maintaining** — new sublibraries or dependency changes are automatically picked up from Project.toml
- **Eliminates ~100 QA exclude entries** — QA is now a separate job that only runs on Julia 1

## Test plan

- [ ] Verify `detect-changes` job correctly identifies affected sublibraries in CI logs
- [ ] Verify sublibrary tests still pass for affected packages
- [ ] Verify global tests in CI.yml are unaffected
- [ ] Verify that changes only to `test/` or `src/` (non-lib) don't trigger sublibrary tests

### DIFF
--- a/.github/scripts/compute_affected_sublibraries.jl
+++ b/.github/scripts/compute_affected_sublibraries.jl
@@ -128,6 +128,13 @@ function compute_affected(changed_files::Vector{String},
     return affected
 end
 
+# Entries to exclude from the matrix.
+# Each entry is (group, version) where group is the CI GROUP string.
+# See: https://github.com/SciML/OrdinaryDiffEq.jl/issues/2977
+const EXCLUDES = Set([
+    ("OrdinaryDiffEqBDF", "pre"),  # JET resolution fails on pre-release Julia
+])
+
 function build_matrix(affected::Set{String}, lib_dir::String)
     entries = Vector{@NamedTuple{group::String, version::String}}()
     for pkg in sort!(collect(affected))
@@ -137,6 +144,7 @@ function build_matrix(affected::Set{String}, lib_dir::String)
             # All other groups append _GROUPNAME (e.g., OrdinaryDiffEqCore_QA)
             ci_group = group_name == "Core" ? pkg : "$(pkg)_$(group_name)"
             for ver in versions
+                (ci_group, ver) in EXCLUDES && continue
                 push!(entries, (; group = ci_group, version = ver))
             end
         end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Updated to:

1. **Julia script instead of Python** — uses stdlib `TOML` parser, no external deps
2. **Sublibraries can declare custom test groups** via `test/test_groups.toml` (e.g., add a GPU group with its own Julia version requirements)
3. **Generic group suffix detection in runtests.jl** — replaces the hardcoded `_QA` check with a lookup against actual `lib/` directories, so `GROUP=OrdinaryDiffEqFoo_GPU` correctly resolves
4. **Single unified test job** — the script outputs the full `{group, version}` matrix, eliminating the separate `test`/`qa` jobs
5. **OrdinaryDiffEqCore README change** included to trigger all 36 sublibrary tests for verification

oscardssmith:
This is a very nice change. Good idea.

ChrisRackauckas:
This was needed to complete the StochasticDiffEq and DelayDiffEq stuff haha.

PR #3245, opened 2026-03-26
Remove deprecated autodiff::Bool shims for v7

### LINKED ISSUE #2310: OrdinaryDiffEq v7
OrdinaryDiffEq v7 is close. This means that all breaking changes need to make it in by this time. Here's the list right now:

- [ ] Split solvers into subpacakges https://github.com/SciML/OrdinaryDiffEq.jl/issues/2177 . This is the main change. After this change,  we have to now define what we want OrdinaryDiffEq.jl to actually look like. The current plan is to have an OrdinaryDiffEqCore.jl at the bottom which is just the core integrator, then OrdinaryDiffEqDifferentiation.jl on top of that, then OrdinaryDiffEqNonlinearSolve.jl, and then the algorithms depend on 1-3 depending on how much they need. This allows the explicit RK methods to not require ForwardDiff or nonlinear solvers, effectively cutting out most dependencies. Then OrdinaryDiffEqDefault is the default method built from some solver packages, and then OrdinaryDiffEq.jl then becomes a shell with just the most widely used methods like Tsit5, Vern, Rosenbrock, and FBDF re-exported. All other solvers are only available when explicitly importing the specific sublibrary.
- [ ] The preconditioner change https://github.com/SciML/LinearSolve.jl/pull/514 which @oscardssmith is doing needs to be in because it's a breaking change to how preconditioners are defined in the algorithm
- [ ] We will need to really complete the OrdinaryDiffEq documentation because we will need to document algorithms per-module, and we should do this via docstrings, I am asking for help from @ArnoStrouwen 
- [ ] We want to change `autodiff` specification to using ADTypes. @gdalle and @avik-pal have already kind of been pushing things down this route. This will make changes to use Enzyme and DifferentiationInterface be non-breaking. Since this has been around for so long, we should have a 2 year deprecation period on the booleans though, but in the constructors we can just change the booleans to AutoForwardDiff and AutoFiniteDiff as a simple path. But with everything ADTypes the integration with NonlinearSolve is simpler as well.

Any other major changes to consider @devmotion @YingboMa @ranocha @oscardssmith 

For reference, the other big thing is SciMLOperators, but those changes are breaking there and non-breaking here.

ChrisRackauckas-Claude:
## Summary

- Remove `_bool_to_ADType` and Bool overloads of `_process_AD_choice` from OrdinaryDiffEqCore
- Remove deprecated `chunk_size`, `diff_type`, `standardtag` keyword arguments from ~56 algorithm constructors
- Add `_ad_chunksize_int`, `_ad_fdtype`, `_fixup_ad` helper functions to derive type parameters from ADTypes directly
- Passing `autodiff=true` or `autodiff=false` now throws an `ArgumentError` with a clear migration message
- Simplify `alg_autodiff` to no longer handle `Val{true}/Val{false}` conversion
- Simplify `SciMLBase.remake` for implicit algorithms
- Update StochasticDiffEqImplicit algorithms to default to `AutoForwardDiff()` instead of `true`
- Update all tests and `default_sde_alg.jl` to use `AutoFiniteDiff()` instead of `false`

## Breaking changes

Users must now pass ADTypes (e.g. `AutoForwardDiff()`, `AutoFiniteDiff()`) instead of `autodiff=true/false`. The deprecated `chunk_size`, `diff_type`, and `standardtag` keyword arguments are removed from all algorithm constructors. These are encoded in the ADType object instead (e.g. `AutoForwardDiff(chunksize=5)`, `AutoFiniteDiff(fdtype=Val(:central))`).

The struct type parameters `{CS, AD, FDT, ST, CJ}` are preserved for backward compatibility with dispatch.

Closes #2310

## Test plan

- [x] OrdinaryDiffEqCore loads and new helpers work correctly
- [x] OrdinaryDiffEqRosenbrock, SDIRK, BDF, FIRK, ExponentialRK all load and construct algorithms correctly
- [x] `autodiff=true` throws ArgumentError with clear migration message
- [ ] CI tests pass across all subpackages

### DIFF
--- a/lib/OrdinaryDiffEqBDF/src/algorithms.jl
+++ b/lib/OrdinaryDiffEqBDF/src/algorithms.jl
@@ -98,7 +98,7 @@ end
     step_limiter! = trivial_limiter!,
     """
 )
-struct ABDF2{AD, F, F2, P, K, T, StepLimiter} <:
+struct ABDF2{AD, F, F2, P, K, T, StepLimiter, CJ} <:
     OrdinaryDiffEqNewtonAdaptiveAlgorithm
     linsolve::F
     nlsolve::F2
@@ -110,7 +110,7 @@ struct ABDF2{AD, F, F2, P, K, T, StepLimiter} <:
     controller::Symbol
     step_limiter!::StepLimiter
     autodiff::AD
-    concrete_jac::Union{Nothing, Bool}
+    concrete_jac::CJ
 end
 function ABDF2(;
         autodiff = AutoForwardDiff(),
@@ -162,7 +162,7 @@ like `KenCarp4`, but instead using a multistep BDF approach",
     order,
     """
 )
-struct SBDF{AD, F, F2, P, K, T} <:
+struct SBDF{AD, F, F2, P, K, T, CJ} <:
     OrdinaryDiffEqNewtonAlgorithm
     linsolve::F
     nlsolve::F2
@@ -173,7 +173,7 @@ struct SBDF{AD, F, F2, P, K, T} <:
     order::Int
     ark::Bool
     autodiff::AD
-    concrete_jac::Union{Nothing, Bool}
+    concrete_jac::CJ
 end
 
 function SBDF(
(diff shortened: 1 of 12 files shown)


baggepinnen:
Is the old `autodiff = false` equivalent to the new `autodiff = AutoFiniteDiff()`?

oscardssmith:
yes.

lib/DiffEqBase/test/downstream/inference.jl
```diff
@@ -8,8 +8,8 @@ u0 = [1.0; 0.0; 0.0]
 tspan = (0.0, 1.0)
 prob = ODEProblem(lorenz, u0, tspan)
 sol = solve(prob, Tsit5(), save_idxs = 1)
-@inferred solve(prob, Tsit5())
```
oscardssmith, line 11: bad bot?

PR #2602, opened 2025-02-14
Make CheckInit the default for ODEs and DAEs: breaking

Ickaser:
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API
  
## Context

Bulk of the discussion happened in #2514 , see particularly [this comment](https://github.com/SciML/OrdinaryDiffEq.jl/pull/2514#issuecomment-2488706578) . This is a separate pull request because I did a clumsy job rebasing that branch and filled the Git history with all the commits since November.

#2554 worked on this in a non-breaking way, including exporting `BrownFullBasicInit` and `ShampineCollocationInit` from `ODENonlinearSolve`.

## Changes

Currently this is a very simple pull request.

This will require some documentation, which I haven't done yet. 
Note to self: probably needs it at
- https://github.com/SciML/DiffEqDocs.jl/blob/master/docs/src/types/dae_types.md
- https://github.com/SciML/DiffEqDocs.jl/blob/master/docs/src/solvers/dae_solve.md
- Possibly add a file like https://github.com/SciML/NonlinearSolve.jl/blob/master/docs/src/release_notes.md with some upgrade notes?

PR #3248, opened 2026-03-26
Make CheckInit the default for ODEs and DAEs

ChrisRackauckas-Claude:
## Summary

- Makes `CheckInit` the default initialization algorithm for ODEs and DAEs (previously `BrownFullBasicInit` / `ShampineCollocationInit`)
- Removes the `_default_dae_init!` dispatch mechanism from `OrdinaryDiffEqCore` and `OrdinaryDiffEqNonlinearSolve` since it's no longer needed
- Updates tests across NonlinearSolve, BDF, SDIRK, and Rosenbrock subpackages to use explicit `initializealg` where they relied on the old default

This is a **breaking change** for v7. Users with inconsistent initial conditions who relied on automatic correction must now explicitly pass `initializealg = BrownFullBasicInit()` or `initializealg = ShampineCollocationInit()`.

Rebased from #2602 (by @Ickaser) onto the v7 branch. See also #2514 and #2554 for prior discussion.

## Test plan

- [x] `OrdinaryDiffEqNonlinearSolve` tests pass locally (Julia 1.11, `Pkg.test()`)
- [ ] CI passes for all subpackages

### DIFF
--- a/lib/OrdinaryDiffEqBDF/test/dae_event.jl
+++ b/lib/OrdinaryDiffEqBDF/test/dae_event.jl
@@ -1,4 +1,5 @@
 using OrdinaryDiffEqBDF, Test
+using OrdinaryDiffEqNonlinearSolve: BrownFullBasicInit
 
 f = function (out, du, u, p, t)
     out[1] = -p[1] * u[1] + p[3] * u[2] * u[3] - du[1]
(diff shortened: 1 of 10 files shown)


oscardssmith:
Should we move dae initialization to a separate library?

ChrisRackauckas:
I don't see why. All inits use the nonlinear solvers (they are a nonlinear solve) and any method that does nonlinear solves is something we want to support at least some form of init, so it might as well be with the nonlinear solves.

PR #3249, opened 2026-03-26
v7: Strip top-level OrdinaryDiffEq to minimal re-export shell

ChrisRackauckas-Claude:
## Summary

- **BREAKING**: Strip `OrdinaryDiffEq.jl` down from 28+ solver subpackage dependencies to only 6 direct deps (OrdinaryDiffEqDefault, OrdinaryDiffEqCore, OrdinaryDiffEqTsit5, OrdinaryDiffEqVerner, OrdinaryDiffEqRosenbrock, OrdinaryDiffEqBDF + Reexport, CommonSolve, SciMLBase, DocStringExtensions)
- `using OrdinaryDiffEq` now provides only the commonly used algorithms: Tsit5, Vern6-9, Rosenbrock23/Rodas5P and family, FBDF/QNDF and family, and the DefaultODEAlgorithm auto-selector
- Users needing other solvers (SDIRK, FIRK, Extrapolation, LowStorageRK, SSPRK, Symplectic, etc.) must explicitly `using OrdinaryDiffEqSDIRK` or the relevant sublibrary
- Version bumped to 7.0.0-DEV

This addresses the v7 goal from #2310 of making the top-level package lightweight while keeping the most commonly needed algorithms accessible by default.

## What still works with just `using OrdinaryDiffEq`
- `solve(prob)` (default algorithm selection)
- `solve(prob, Tsit5())`
- `solve(prob, Vern7())`, `Vern6()`, `Vern8()`, `Vern9()`
- `solve(prob, Rosenbrock23())`, `Rodas5P()`, and all Rosenbrock methods
- `solve(prob, FBDF())`, `QNDF()`, and all BDF methods
- `AutoTsit5`, `AutoVern7`, etc.
- All core utilities (controllers, composite algorithms, etc.)

## What requires explicit sublibrary import in v7
- `using OrdinaryDiffEqSDIRK` for ImplicitEuler, KenCarp4, TRBDF2, etc.
- `using OrdinaryDiffEqLowOrderRK` for Euler, DP5, BS3, RK4, etc.
- `using OrdinaryDiffEqFIRK` for RadauIIA5, etc.
- `using OrdinaryDiffEqExponentialRK` for exponential integrators
- And ~20 other solver subpackages

## Test plan
- [x] Package loads successfully with `using OrdinaryDiffEq`
- [x] All default algorithms (Tsit5, Vern7, Rosenbrock23, Rodas5P, FBDF) are accessible
- [x] DefaultODEAlgorithm works (auto-selection)
- [x] ODE solving works end-to-end
- [ ] CI tests for sublibrary-specific tests may need updating to add explicit imports

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -13,12 +13,14 @@ using Reexport
 # selection logic. It transitively depends on Tsit5, Verner, Rosenbrock, and BDF.
 @reexport using OrdinaryDiffEqDefault
 
-# Re-export the commonly used algorithms from OrdinaryDiffEqDefault's dependencies.
-# These are the algorithms that the default algorithm selector can choose from.
-@reexport using OrdinaryDiffEqTsit5
-@reexport using OrdinaryDiffEqVerner
-@reexport using OrdinaryDiffEqRosenbrock
-@reexport using OrdinaryDiffEqBDF
+# Import only the widely-used algorithms from OrdinaryDiffEqDefault's dependencies.
+# We intentionally do NOT @reexport the full sub-packages so that less common
+# algorithms (e.g. other Rosenbrock or BDF variants) stay opt-in.
+using OrdinaryDiffEqTsit5: Tsit5, AutoTsit5
+using OrdinaryDiffEqVerner: Vern6, Vern7, Vern8, Vern9,
+    AutoVern6, AutoVern7, AutoVern8, AutoVern9
+using OrdinaryDiffEqRosenbrock: Rosenbrock23, Rodas5P
+using OrdinaryDiffEqBDF: FBDF
 
 # Re-export specific core utilities that were previously exported
 using OrdinaryDiffEqCore: OrdinaryDiffEqCore,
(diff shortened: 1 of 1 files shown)


oscardssmith:
can we also get rid of the `@reexport` use and figure out specifically which exports we want?

I also think we don't want to support all the Rosenbrocks as exported by default. If we ever made a separate package for Rosenbrock23 and Rodas5P (for example) we don't necessarily want to have to keep importing all of the other rosenbrocks.

src/OrdinaryDiffEq.jl
```diff
@@ -13,12 +13,14 @@ using Reexport
 # selection logic. It transitively depends on Tsit5, Verner, Rosenbrock, and BDF.
 @reexport using OrdinaryDiffEqDefault
```
oscardssmith, line 14: Is this needed?

src/OrdinaryDiffEq.jl
```diff
@@ -42,6 +44,12 @@ export CompositeAlgorithm, AutoSwitch
...
 export IController, PIController, PIDController
 
+# Widely-used algorithms (selectively exported, not blanket @reexport)
+export Tsit5, AutoTsit5
+export Vern6, Vern7, Vern8, Vern9, AutoVern6, AutoVern7, AutoVern8, AutoVern9
+export Rosenbrock23, Rodas5P
+export FBDF
+
 # Re-export Reexport for downstream compatibility
 export Reexport, @reexport
```
oscardssmith, line 54: remove this?

src/OrdinaryDiffEq.jl
oscardssmith, line 10: This should also change OrdinaryDiffEQCore to not reexport all of SciMLBase. E.G. We don't need `BVProblem`

src/OrdinaryDiffEq.jl
ChrisRackauckas: I guess the user rarely changes the default method.

oscardssmith: like we could (and IMO it would be reasonable) to export DefaultSolver (e.g. to turn on autodiff). I just really dislike blanket reexports.

PR #3250, opened 2026-03-26
Change default lazy interpolation to false for v7

ChrisRackauckas-Claude:
## Summary
- Changes the default value of `lazy` from `Val{true}()` to `Val{false}()` for all algorithms that support it: BS5, Vern6/7/8/9, RKV76IIa, and AutoVern6/7/8/9
- Updates docstring defaults and Bool compatibility constructors to match
- Updates `ode_strip_test.jl` to verify that `strip_solution` now works for Vern solutions with the new default, and that explicitly passing `lazy=true` still throws `LazyInterpolationException`

## Motivation
When `lazy=true` (old default), the ODE function `f` is needed during interpolation queries, which prevents `strip_interpolation` from working. With `lazy=false`, all intermediate steps are computed eagerly so `f` is not needed for interpolation. This is a breaking change appropriate for v7, making https://github.com/SciML/SciMLBase.jl/issues/629 simpler and more robust.

Relates to #2310.

## Test plan
- [x] OrdinaryDiffEqVerner tests pass
- [x] OrdinaryDiffEqLowOrderRK tests pass
- [x] Strip test verifies `strip_solution` works for default Vern7() (lazy=false)
- [x] Strip test verifies `strip_solution` throws for explicit Vern7(lazy=true)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/lib/OrdinaryDiffEqLowOrderRK/src/algorithms.jl
+++ b/lib/OrdinaryDiffEqLowOrderRK/src/algorithms.jl
@@ -208,20 +208,20 @@ end
     publisher={Elsevier}
     }",
     extra_keyword_description = """- `lazy`: determines if the lazy interpolant is used.""",
-    extra_keyword_default = "lazy = true"
+    extra_keyword_default = "lazy = false"
 )
 Base.@kwdef struct BS5{StageLimiter, StepLimiter, Thread, L} <: OrdinaryDiffEqAdaptiveAlgorithm
     stage_limiter!::StageLimiter = trivial_limiter!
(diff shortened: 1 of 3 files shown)


oscardssmith:
With this change are Vern methods still the most efficient high order solvers?

ChrisRackauckas:
This wasn't the intended change 😅

ChrisRackauckas-Claude:
Closing: the default should remain Val{true}(), not Val{false}(). Will create a new PR that only removes the Bool compatibility paths.

PR #3264, opened 2026-03-27
Replace blanket @reexport with explicit exports

ChrisRackauckas-Claude:
## Summary

Per @oscardssmith's review on #3249: remove all `@reexport` usage and explicitly list only the ODE-relevant symbols.

- Remove `Reexport` dependency entirely
- Replace `@reexport using OrdinaryDiffEqCore` with explicit `using` + `export`
- Replace `@reexport using OrdinaryDiffEqDefault` with explicit `using` + `export`
- Only exports ODE-relevant types (`ODEProblem`, `ODEFunction`, callbacks, `ReturnCode`, etc.)
- Does NOT export `BVProblem`, `IntegralProblem`, `OptimizationProblem`, or other non-ODE SciMLBase types

## Test plan

- [x] Package loads, all exported algorithms accessible
- [x] `BVProblem` correctly not exported
- [x] `solve(prob, Tsit5())` works end-to-end

### DIFF
--- a/src/OrdinaryDiffEq.jl
+++ b/src/OrdinaryDiffEq.jl
@@ -16,29 +16,22 @@ using OrdinaryDiffEqVerner: Vern6, Vern7, Vern8, Vern9,
 using OrdinaryDiffEqRosenbrock: Rosenbrock23, Rodas5P
 using OrdinaryDiffEqBDF: FBDF
 
-# Import ODE-relevant types from SciMLBase (NOT blanket reexport)
+# Import ODE-relevant types from SciMLBase
 using SciMLBase: SciMLBase,
     ODEProblem, ODEFunction, ODESolution,
     SplitODEProblem, SplitFunction,
     SecondOrderODEProblem, DynamicalODEProblem,
     DAEProblem, DAEFunction, DAESolution,
-    DiscreteProblem, DiscreteFunction,
-    EnsembleProblem, EnsembleAlgorithm, EnsembleSolution,
-    EnsembleThreads, EnsembleDistributed, EnsembleSerial, EnsembleSplitThreads,
     CallbackSet, ContinuousCallback, DiscreteCallback, VectorContinuousCallback,
-    ODEAliasSpecifier,
     ReturnCode,
-    remake, successful_retcode,
-    addsteps!, savevalues!, terminate!, reinit!,
-    du_cache, full_cache, u_cache, get_tmp_cache,
-    add_saveat!
+    remake, successful_retcode, reinit!
+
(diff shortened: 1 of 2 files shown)


src/OrdinaryDiffEq.jl
```diff
@@ -4,47 +4,81 @@ $(DocStringExtensions.README)
...
+export EnsembleThreads, EnsembleDistributed, EnsembleSerial, EnsembleSplitThreads
+
+# Callbacks
+export CallbackSet, ContinuousCallback, DiscreteCallback, VectorContinuousCallback
+
+# Utilities
+export ODEAliasSpecifier, ReturnCode
+export remake, successful_retcode, reinit!
 export addsteps!, ode_interpolant, terminate!, savevalues!, isfsal
+export du_cache, full_cache, u_cache, get_tmp_cache, add_saveat!
```
oscardssmith: Do we want to export any of these? They seem like they should remain internal.

src/OrdinaryDiffEq.jl
```diff
@@ -4,47 +4,81 @@ $(DocStringExtensions.README)
...
+# Problem types
+export ODEProblem, ODEFunction, ODESolution
+export SplitODEProblem, SplitFunction
+export SecondOrderODEProblem, DynamicalODEProblem
+export DAEProblem, DAEFunction, DAESolution
+export DiscreteProblem, DiscreteFunction
+
+# Ensemble
+export EnsembleProblem, EnsembleAlgorithm, EnsembleSolution
+export EnsembleThreads, EnsembleDistributed, EnsembleSerial, EnsembleSplitThreads
```
oscardssmith: These shouldn't be exported from here, right?

src/OrdinaryDiffEq.jl
```diff
@@ -4,47 +4,81 @@ $(DocStringExtensions.README)
...
 # General Functions
 export solve, solve!, init, step!
 
-# Callback Necessary
+# Problem types
+export ODEProblem, ODEFunction, ODESolution
+export SplitODEProblem, SplitFunction
+export SecondOrderODEProblem, DynamicalODEProblem
+export DAEProblem, DAEFunction, DAESolution
+export DiscreteProblem, DiscreteFunction
```
oscardssmith: These shouldn't be here, should they?

src/OrdinaryDiffEq.jl
oscardssmith: Should we also export some of the ATTypes types?

oscardssmith:
we should also search the rest of the packages here for reexport

PR #3266, opened 2026-03-27
fix StochasticDiffEq _alg-autodiff

oscardssmith:
fixes precompile.

### DIFF
--- a/lib/StochasticDiffEqImplicit/src/alg_utils.jl
+++ b/lib/StochasticDiffEqImplicit/src/alg_utils.jl
@@ -20,21 +20,14 @@ alg_needs_extra_process(alg::SKenCarp) = true
 
 function SciMLBase.alg_interpretation(
         alg::ImplicitRKMil{
-            CS,
             AD,
             F,
             P,
-            FDT,
-            ST,
-            CJ,
             N,
             T2,
-            Controller,
             interpretation,
         }
-    ) where {
-        CS, AD, F, P, FDT, ST, CJ, N, T2, Controller, interpretation,
-    }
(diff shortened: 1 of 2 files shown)


oscardssmith:
@ChrisRackauckas thoughts on removing `_alg_autodiff` entirely? It was only introduced to fixup `autodiff=true/false`, and I think could be just replaced with `alg_autodiff` (no `_`).

ChrisRackauckas:
We should just remove it yes

oscardssmith:
Merging this for now since it gets things closer to working. `StochasticDiffEqImplicit` still has a bunch of problems.

lib/StochasticDiffEqCore/src/alg_utils.jl
```diff
@@ -67,18 +67,11 @@ for T in [
...
-            T, AD,
-        }
-    ) where {T, AD}
-    return Val{AD}()
-end
+
+_alg_autodiff(::StochasticDiffEqNewtonAlgorithm) = alg.autodiff
+_alg_autodiff(::StochasticDiffEqNewtonAdaptiveAlgorithm) = alg.autodiff
+_alg_autodiff(::StochasticDiffEqJumpNewtonAdaptiveAlgorithm) = alg.autodiff
+_alg_autodiff(::StochasticDiffEqJumpNewtonDiffusionAdaptiveAlgorithm) = alg.autodiff
```
termi-official, line 74: I think we are missing the alg input here

oscardssmith: oops

PR #3268, opened 2026-03-28
More fixes of the v7 branch.

oscardssmith:


lib/StochasticDiffEqImplicit/src/algorithms.jl
```diff
@@ -1,13 +1,13 @@
 using ADTypes: AutoForwardDiff
 using OrdinaryDiffEqCore: _fixup_ad, _unwrap_val
 
-struct ImplicitEM{AD, F, F2, T2, CJ} <:
+struct ImplicitEM{AD, F, F2, T2, T3, CJ} <:
```
oscardssmith, line 4: @ChrisRackauckas can you review this? I'm not sure if we want T2 and T3 split or separate.

ChrisRackauckas: why is this done?

oscardssmith: you removed the explicit constructor from these algs that forced theta and new_jac_cov_bound to be the same type. we either need to promote them to the same type or give them separate type params

oscardssmith: so do you have a priority for which way to go here?

ChrisRackauckas: nah either way is fine.

oscardssmith: great. merging.

PR #3271, opened 2026-03-28
Remove FillArrays dependency from OrdinaryDiffEqCore

ChrisRackauckas-Claude:
## Summary
- Replace `Trues(size(y₀))` / `Falses(size(y₀))` with scalar `true` / `false` in `interpolation_differential_vars` — scalar booleans broadcast identically and `all(true)` works the same as `all(Trues(...))`
- Remove `FillArrays` from OrdinaryDiffEqCore's `[deps]` and `[compat]` sections
- Update GPU test to use `trues()` (Base) instead of `FillArrays.Trues()`

## Motivation
SnoopCompile invalidation analysis shows FillArrays causes **505 method invalidations** (13.1% of total) when loading OrdinaryDiffEqCore, due to broad Base method extensions for `convert`, `broadcasted`, `print_matrix_row`, `in`, and `sort!`. OrdinaryDiffEqCore was the only package in its dependency tree that depended on FillArrays directly, and it only used `Trues` and `Falses` in 4 call sites.

## Test plan
- [x] `Pkg.test()` on OrdinaryDiffEqCore passes (29 pass, 1 pre-existing JET failure, 1 broken — identical to base v7)
- [ ] CI

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,8 +1,6 @@
 name: CI
 on:
   pull_request:
-    branches:
-      - master
   push:
     branches:
       - master
--- a/.github/workflows/Downgrade.yml
+++ b/.github/workflows/Downgrade.yml
@@ -1,8 +1,6 @@
 name: Downgrade
 on:
   pull_request:
-    branches:
-      - master
     paths-ignore:
       - 'docs/**'
   push:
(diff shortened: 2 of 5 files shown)


ChrisRackauckas-Claude:
Closing — should target master, not v7. Reopening as a new PR.

PR #3292, opened 2026-03-30
Remove Static.jl and StaticArrayInterface from OrdinaryDiffEqCore

ChrisRackauckas-Claude:
## Summary

- Remove `Static.jl` and `StaticArrayInterface.jl` from OrdinaryDiffEqCore dependencies
- Replace `Static.True()`/`Static.False()` threading dispatch tags with plain `Bool` (`true`/`false`)
- Bump FastBroadcast compat to include v1 (`"0.3, 1"`)

Static.jl is responsible for ~1,890 method invalidations during precompilation. Neither Static.jl nor StaticArrayInterface.jl were used for core functionality -- `Static.True`/`Static.False` were only used as dispatch tags for the `thread` keyword in explicit RK methods, which can be replaced by `Bool`. StaticArrayInterface had no remaining usage in OrdinaryDiffEqCore.

This change reduces precompilation time and invalidation overhead for all downstream packages that depend on OrdinaryDiffEqCore.

## Changed files

- `lib/OrdinaryDiffEqCore/Project.toml` -- removed deps and compat entries, bumped FastBroadcast compat
- `lib/OrdinaryDiffEqCore/src/OrdinaryDiffEqCore.jl` -- removed `True, False` imports from FastBroadcast
- `lib/OrdinaryDiffEqCore/src/doc_utils.jl` -- updated docstring defaults and descriptions to use `true`/`false`

## Test plan

- [ ] CI passes on all OrdinaryDiffEq sub-packages
- [ ] Verify no remaining references to `Static.True`/`Static.False` in OrdinaryDiffEqCore


lib/DiffEqBase/src/calculate_residuals.jl
```diff
@@ -99,13 +99,13 @@ Save element-wise residuals
...
-appropriate CPU arrays should be serial (`thread = False()`, default)
-or use multiple threads (`thread = True()`) when Julia is started
+appropriate CPU arrays should be serial (`thread = false`, default)
+or use multiple threads (`thread = true`) when Julia is started
 with multiple threads.
 """
 @inline function calculate_residuals!(
         out, ũ, u₀, u₁, α, ρ, internalnorm, t,
-        thread::Union{False, True} = False()
+        thread::Bool = false
```
oscardssmith, line 108: should these be Val(true)/Val(false)?

ranocha: Or `Serial()`/`Threaded()` from FastBroadcast.jl?

PR #3309, opened 2026-04-01
Remove StaticArrayInterface, bump FastBroadcast/RAT compat

ChrisRackauckas-Claude:
## Summary

Non-breaking dependency cleanup:
- Remove `StaticArrayInterface` from OrdinaryDiffEqCore and OrdinaryDiffEqDifferentiation
- Bump `FastBroadcast` compat to `1.2` (40 files) — adds `Serial`/`Threaded` types and Static.jl extension for backward compat
- Bump `RecursiveArrayTools` compat to `3.51` (30 files) — fixes FastBroadcast v1.0 extension
- Bump `SciMLSensitivity` compat to `7.100` in test/ad
- Fix FastBroadcast imports in DiffEqBase and OrdinaryDiffEqCore (`True`/`False` no longer exported from FastBroadcast, now imported from Static which remains in deps)

## What this does NOT change

- Static.jl remains in `[deps]` — all `import Static: False` calls still work
- Algorithm struct defaults stay as `False()` — no API change
- User code passing `thread=Static.False()` or `thread=Static.True()` continues to work

## Why

This is step 1 of removing Static.jl's invalidation overhead. FastBroadcast v1.2 no longer depends on Static.jl but provides a backward-compatible extension. The full Static.jl removal (changing defaults to `Serial()`) is deferred to v7.

## Depends on

- SciML/FastBroadcast.jl#85 — Serial fallback for non-AbstractArray types (needs release)

## Test plan

- [ ] CI


oscardssmith:
shouldn't this be targeting v7?

ChrisRackauckas:
It's non-breaking and required for https://github.com/SciML/NonlinearSolve.jl/pull/838

PR #3312, opened 2026-04-01
Controller interface for v7

termi-official:
## TODO

* [ ] Port StochasticDiffEq controllers
* [ ] Port DelayDiffEq controllers

## Follow Up

* [ ] Move EEst into controller cache
* [ ] Remove legacy alg utils https://github.com/SciML/OrdinaryDiffEq.jl/pull/3312#discussion_r3065580443
* [ ] Fix Aitken-Neville controller

### DIFF
--- a/lib/OrdinaryDiffEqBDF/Project.toml
+++ b/lib/OrdinaryDiffEqBDF/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqBDF"
 uuid = "6ad6398a-0878-4a85-9266-38940aa047c8"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "1.25.0"
+version = "1.26.0"
 
 [deps]
 FastBroadcast = "7034ab61-46d4-4ed7-9d0f-46aef9175898"


lib/OrdinaryDiffEqCore/src/alg_utils.jl
```diff
@@ -424,57 +424,22 @@ alg_adaptive_order(alg::Union{OrdinaryDiffEqAlgorithm, DAEAlgorithm}) = alg_orde
...
     if ispredictive(alg)
-        return NewPredictiveController(QT, alg)
+        return PredictiveController(QT, alg)
     elseif isstandard(alg)
-        return NewIController(QT, alg)
+        return IController(QT, alg)
     else
-        return NewPIController(QT, alg)
+        return PIController(QT, alg)
     end
```
termi-official: Since we are breaking stuff right now anyway, this part of the setup workflow here is something which we might want to think about. I think we should 
1. Purge isstandard/ispredictive
2. Algorithms with special controllers should just dispatch on `default_controller` (see e.g. lib/OrdinaryDiffEqLowStorageRK/src/alg_utils.jl)
3. In all other cases we return a standard PI or PID controller with reasonable (and algorithm order sensitive) default parameters.

lib/OrdinaryDiffEqExtrapolation/src/alg_utils.jl
```diff
@@ -36,84 +36,23 @@ alg_maximum_order(alg::ImplicitHairerWannerExtrapolation) = 2(alg.max_order + 1)
...
-                alg::AitkenNeville,
-                cache,
-                qoldinit, _beta1 = nothing, _beta2 = nothing
-            )
-            QT = typeof(qoldinit)
-            beta1, beta2 = _digest_beta1_beta2(alg, cache, Val(QT), _beta1, _beta2)
-            return PIController(beta1, beta2)
-        end
-    end
+# FIXME AitkenNeville is missing integration with the extrapolation controller and picks up the PI controller instead.
```
termi-official: I don't know how to fix this, as I do not know enough about the theory here.

termi-official:
I will open a separate PR again with the code. Some git operation failed and the branch is broken now.

PR #3326, opened 2026-04-02
JuliaC/trim compatibility for Core and Rosenbrock

asinghvi17:
## Summary
- Specialize on `tType` in `initdt` to force compiler specialization and type propagation
- Specialize on `deriv` type in generic `ode_interpolant` and `interp_at_saveat` for trim compatibility
- Specialize on type input to Rosenbrock tableaus

These changes fix dynamic dispatch paths that cause Julia's `--trim=safe` (JuliaC AOT compilation) to fail, by adding type parameters that allow the compiler to statically resolve method calls.

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/initdt.jl
+++ b/lib/OrdinaryDiffEqCore/src/initdt.jl
@@ -7,9 +7,9 @@
 # =============================================================================
 
 @muladd function _ode_initdt_iip(
-        u0, t, _tType, tdir, dtmax, abstol, reltol, internalnorm,
+        u0, t, ::Type{_tType}, tdir, dtmax, abstol, reltol, internalnorm,
         prob, g, noise_prototype, order, integrator
-    )
+    ) where {_tType}
     f = prob.f
     p = integrator.p
     oneunit_tType = oneunit(_tType)
@@ -307,9 +307,9 @@ function Base.showerror(io::IO, e::TypeNotConstantError)
 end
 
 @muladd function _ode_initdt_oop(
-        u0, t, _tType, tdir, dtmax, abstol, reltol, internalnorm,
+        u0, t, ::Type{_tType}, tdir, dtmax, abstol, reltol, internalnorm,
         prob, g, order, integrator
-    )
+    ) where {_tType}
     f = prob.f
     p = prob.p
     oneunit_tType = oneunit(_tType)
--- a/lib/OrdinaryDiffEqCore/src/dense/generic_dense.jl
+++ b/lib/OrdinaryDiffEqCore/src/dense/generic_dense.jl
@@ -153,7 +153,9 @@ end
     return ode_addsteps!(integrator, args...)
 end
 
-@inline function ode_interpolant(Θ, integrator::SciMLBase.DEIntegrator, idxs, deriv)
+@inline function ode_interpolant(
+        Θ, integrator::SciMLBase.DEIntegrator, idxs, ::Type{deriv}
+    ) where {deriv}
     SciMLBase.addsteps!(integrator)
     if integrator.cache isa CompositeCache
         val = composite_ode_interpolant(
--- a/lib/OrdinaryDiffEqCore/src/dense/generic_dense.jl
+++ b/lib/OrdinaryDiffEqCore/src/dense/generic_dense.jl
@@ -177,8 +177,8 @@ end
 end
 
 function default_ode_interpolant(
-        Θ, integrator, cache::DefaultCache, alg_choice, idxs, deriv
-    )
+        Θ, integrator, cache::DefaultCache, alg_choice, idxs, ::Type{deriv}
+    ) where {deriv}
     if alg_choice == 1
         return ode_interpolant(
             Θ, integrator.dt, integrator.uprev,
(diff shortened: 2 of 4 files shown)


lib/OrdinaryDiffEqCore/src/caches/basic_caches.jl
```diff
@@ -82,43 +83,29 @@ function alg_cache(
...
+    return DefaultCache{T1, T2, T3, T4, T5, T6, typeof(alg.choice_function), typeof(u),
+                        uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits, V}(
         args, alg.choice_function, 1, u
     )
-    algs = alg.algs
-    # If the type is a bitstype we need to initialize it correctly here since isdefined will always return true.
-    if isbitstype(T1)
-        cache.cache1 = alg_cache(algs[1], args...)
-    end
-    if isbitstype(T2)
```
ChrisRackauckas, line 106: no...

lib/OrdinaryDiffEqCore/src/perform_step/composite_perform_step.jl
```diff
@@ -1,29 +1,34 @@
...
-    elseif i == 6
-        if !isdefined(cache, :cache6)
-            cache.cache6 = alg_cache(algs[6], cache.args...)
-        end
+# Lazily initialize the i-th sub-cache on first access.  The @generated body
+# turns the runtime index `i` into compile-time literal tuple indices, and
+# recovers the Type args (uEltypeNoUnits, etc.) from DefaultCache's type
+# parameters so that the trimmer sees fully concrete alg_cache dispatch.
+@generated function init_ith_default_cache(
+        cache::DefaultCache{T1, T2, T3, T4, T5, T6, A, F, uType,
```
ChrisRackauckas: at least last time this was measured, generated functions don't precompile right.

lib/OrdinaryDiffEqCore/src/caches/basic_caches.jl
asinghvi17: Yeah I'm not super happy with this either.  I could go back to the initial approach of just instantiating all caches?

ChrisRackauckas: what's the advantage of dropping this?

lib/OrdinaryDiffEqCore/src/perform_step/composite_perform_step.jl
asinghvi17: no choice here but to enumerate all options, I think.  We could write it out manually?  Otherwise what's the test that it does precompile right?

lib/OrdinaryDiffEqCore/src/caches/basic_caches.jl
ChrisRackauckas: > I could go back to the initial approach of just instantiating all caches?

Definitely not.

ChrisRackauckas:
This isn't one PR. Please split into the component parts. There's the inference improvements to Rosenbrock, which are missing JET tests. There's adding trim tests. There's adding default solvers. This is a mess.

lib/OrdinaryDiffEqCore/src/caches/basic_caches.jl
asinghvi17: Well it's not trim compatible on master - so something has to be done, otherwise you can't use juliac to compile a problem that uses the `DefaultODEAlgorithm()`/`Auto()` solver.

asinghvi17:
I'll split the trim CI out to a separate PR.

asinghvi17:
Runic is complaining about files that I haven't touched, so that seems like something for another formatting PR to fix.

lib/OrdinaryDiffEqCore/src/perform_step/composite_perform_step.jl
asinghvi17: After conversation with @oscardssmith we will swap the auto alg in FMUGeneration for a similar but more stable alg, determined at the time of FMU construction.

lib/OrdinaryDiffEqCore/src/initdt.jl
```diff
@@ -7,9 +7,9 @@
 # =============================================================================
 
 @muladd function _ode_initdt_iip(
-        u0, t, _tType, tdir, dtmax, abstol, reltol, internalnorm,
+        u0, t, ::Type{_tType}, tdir, dtmax, abstol, reltol, internalnorm,
```
oscardssmith, line 10: since this funtion takes `t` we should see if we can just remove `_tType` entirely.

asinghvi17: That would be a bigger change though, right?  I'm assuming a lot of solvers try to look at this?

oscardssmith: no. The only caller of `initdit` is `__init` so the initialization happens before any solver specific code runs.

oscardssmith: @asinghvi17 can you resolve this?

oscardssmith:
Other than the initdt issue this seems mostly good to go. It also needs a rebase.

ChrisRackauckas:
initdt was already handled. Other stuff needs inference tests.

ChrisRackauckas:
initdt was already handled. Other stuff needs inference tests.

PR #3378, opened 2026-04-10
Fix Mooncake gradient bug in fixed_t_for_tstop_error! @zero_adjoint, re-enable AD tests

ChrisRackauckas-Claude:
## Summary

Re-evaluated Mooncake's compatibility with the OrdinaryDiffEq AD test suite (Mooncake 0.5.26, tested locally on Julia 1.11.9 and 1.12.5) and re-enabled Mooncake gradient testing wherever it actually works through `SciMLSensitivityMooncakeExt`.

The previous state was overly conservative — `test/ad/{ad_tests,autodiff_events,discrete_adjoints}.jl` all noted "Mooncake gradient support for ODE solves is currently broken" and skipped Mooncake entirely on every gradient test. In reality the only path that's actually broken is `SensitivityADPassThrough()` (Mooncake silently produces wrong gradients when AD is passed straight through the mutating solver internals); the SciMLSensitivity-based adjoint paths work cleanly.

## What changed

- **`test/ad/ad_tests.jl`**: Add Mooncake to the simple multi-backend parameter-gradient testset (`DifferentiationInterface multi-backend tests`). `SciMLSensitivity` is now loaded directly in this file so the Mooncake extension can route through SciML's adjoint code paths under default sensealg dispatch.

- **`test/ad/autodiff_events.jl`**: Add a new `Mooncake gradient tests` testset that exercises the `ContinuousCallback` gradient with `ForwardDiffSensitivity()` for both the standard solver and the `PredictiveController()`/`TRBDF2` case. Updated the file's header comment to reflect the actual situation. `ReverseDiffAdjoint` and `PIDController`-based gradients are intentionally not added — Mooncake currently throws a `TypeError` on the resulting `ODESolution` `CoDual`.

- **`test/ad/discrete_adjoints.jl`**: Keep the existing `SensitivityADPassThrough()` `@test_broken` (that path is genuinely broken — Mooncake silently produces wrong gradients), but add a new parameterised Lorenz problem and a working Mooncake testset covering `InterpolatingAdjoint(autojacvec=ZygoteVJP())`, `InterpolatingAdjoint(autojacvec=EnzymeVJP())` and `GaussAdjoint(autojacvec=ZygoteVJP())`. The continuous-adjoint path through `SciMLSensitivityMooncakeExt` is fully functional and is now a regression check. (A parameterised form of Lorenz is required because the no-parameter `lorenz!` triggers a "nothing returned from Zygote vjp" error in `ZygoteVJP`-based sensealgs.)

## Mooncake compatibility matrix (from local re-evaluation)

| Test path | Status |
|---|---|
| `SensitivityADPassThrough()` (true discrete adjoint) | ❌ silently wrong gradients |
| Default sensealg dispatch on simple parameterised ODE | ✅ |
| `ForwardDiffSensitivity()` | ✅ |
| `InterpolatingAdjoint(autojacvec=ZygoteVJP())` | ✅ |
| `InterpolatingAdjoint(autojacvec=EnzymeVJP())` | ✅ |
| `GaussAdjoint(autojacvec=ZygoteVJP())` | ✅ |
| `ReverseDiffAdjoint()` | ❌ `TypeError` on `ODESolution` `CoDual` |
| `myobj/myobj2`-style closure-captured ODE definitions | ❌ `MethodError` / fdata issues |

## Test plan

- [x] `test/ad/discrete_adjoints.jl` — passes on Julia 1.11.9 (`Pass: 3, Broken: 1`)
- [x] `test/ad/discrete_adjoints.jl` — passes on Julia 1.12.5 (`Pass: 3, Broken: 1`)
- [x] `test/ad/autodiff_events.jl` — `Mooncake gradient tests` testset passes on Julia 1.11.9 and 1.12.5 (`Pass: 2`)
- [x] `test/ad/ad_tests.jl` — `DifferentiationInterface multi-backend tests` passes on Julia 1.11.9 and 1.12.5 with the new Mooncake assertion (`Pass: 2`)
- [x] `runic` formatting check clean

### DIFF
--- a/test/ad/discrete_adjoints.jl
+++ b/test/ad/discrete_adjoints.jl
@@ -1,17 +1,9 @@
 # Discrete adjoint tests with Mooncake
 # Enzyme: skipped due to segfaults (see https://github.com/EnzymeAD/Enzyme.jl/issues/2699)
 # Mooncake: all versions
-#   - SensitivityADPassThrough (true discrete adjoint): works after fixing the
-#     OrdinaryDiffEqCoreMooncakeExt rule on `fixed_t_for_tstop_error!` — that
-#     rule was incorrectly @zero_adjoint and was silently zeroing the gradient
-#     flow through `integrator.t`.
-#   - Continuous adjoints via SciMLSensitivity (e.g. InterpolatingAdjoint with
-#     ZygoteVJP/EnzymeVJP, GaussAdjoint): also tested as a regression check
-#     that the SciMLSensitivityMooncakeExt path works.
 # ForwardDiff: all versions (reference)
 
 using OrdinaryDiffEqTsit5, StaticArrays, DiffEqBase, Test, ForwardDiff
-using SciMLSensitivity
 using ADTypes
 import DifferentiationInterface as DI
 using Mooncake  # Load Mooncake after DI to ensure extension is loaded
@@ -48,63 +40,20 @@ function f_dt_sum(u0)
     return sum(sol[1, :])
 end
 
-# Continuous-adjoint variant (parameterised Lorenz, Vector saveat, no
-# SensitivityADPassThrough) so the SciMLSensitivity adjoint dispatch is used.
-# Mooncake supports this path via SciMLSensitivityMooncakeExt with
-# InterpolatingAdjoint(ZygoteVJP/EnzymeVJP) and GaussAdjoint(ZygoteVJP).
-# A parameterised form is required because the no-parameter `lorenz!` triggers
-# a "nothing returned from Zygote vjp" error in ZygoteVJP-based sensealgs.
-function lorenz_p!(du, u, p, t)
-    du[1] = p[1] * (u[2] - u[1])
-    du[2] = u[1] * (p[2] - u[3]) - u[2]
-    return du[3] = u[1] * u[2] - p[3] * u[3]
-end
-
-const _saveat_v = collect(0.0:0.25:3.0)
-const _lorenz_p = [10.0, 28.0, 8 / 3]
-
-function f_dt_sum_sa(u0, sensealg)
-    tspan = (0.0, 3.0)
-    prob = ODEProblem{true, SciMLBase.FullSpecialize}(lorenz_p!, u0, tspan, _lorenz_p)
-    sol = solve(prob, Tsit5(), saveat = _saveat_v, sensealg = sensealg, abstol = 1.0e-12, reltol = 1.0e-12)
-    return sum(sol[1, :])
-end
-
 u0 = [1.0; 0.0; 0.0]
 
 # Reference jacobian and gradient using ForwardDiff
 fdj = DI.jacobian(f_dt, AutoForwardDiff(), u0)
 fdg = DI.gradient(f_dt_sum, AutoForwardDiff(), u0)
-fdg_sa = DI.gradient(u_ -> f_dt_sum_sa(u_, ForwardDiffSensitivity()), AutoForwardDiff(), u0)
-
 @testset "Discrete Adjoints" begin
     # Enzyme tests skipped - Enzyme segfaults on ODE solves which crashes the process
     # before @test_broken can catch it. See https://github.com/EnzymeAD/Enzyme.jl/issues/2699
 
     # Mooncake tests (all Julia versions)
     @testset "Mooncake" begin
-        # SensitivityADPassThrough lets Mooncake differentiate through the
-        # mutating solver internals directly. This was previously @test_broken
-        # because the OrdinaryDiffEqCoreMooncakeExt @zero_adjoint rule on
-        # `fixed_t_for_tstop_error!` was silently zeroing the gradient flow
-        # through `integrator.t` (the function returns its `ttmp` argument
-        # which gets assigned back into `integrator.t`). With that rule
-        # removed, Mooncake gives the correct gradient to within ~10 digits.
         @testset "Gradient via SensitivityADPassThrough" begin
             mkg = DI.gradient(f_dt_sum, AutoMooncake(; config = nothing), u0)
             @test mkg ≈ fdg rtol = 1.0e-6
-        end
-
-        # Continuous adjoints through SciMLSensitivity DO work with Mooncake
-        # via the SciMLSensitivityMooncakeExt extension. These cover the same
-        # Lorenz problem as the SensitivityADPassThrough test above.
-        @testset "Gradient via $(nameof(typeof(sensealg))) ($(nameof(typeof(sensealg.autojacvec))))" for sensealg in (
-                InterpolatingAdjoint(autojacvec = ZygoteVJP()),
-                InterpolatingAdjoint(autojacvec = EnzymeVJP()),
-                GaussAdjoint(autojacvec = ZygoteVJP()),
-            )
-            mkg = DI.gradient(u_ -> f_dt_sum_sa(u_, sensealg), AutoMooncake(; config = nothing), u0)
-            @test mkg ≈ fdg_sa rtol = 1.0e-6
         end
     end
 end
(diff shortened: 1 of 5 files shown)


ChrisRackauckas-Claude:
## Update: root-caused the SensitivityADPassThrough bug and pushed a fix

Followed up on the question of *why* `SensitivityADPassThrough` was producing silently wrong gradients. The culprit is in `OrdinaryDiffEqCoreMooncakeExt`:

```julia
Mooncake.@zero_adjoint Mooncake.MinimalCtx Tuple{
    typeof(OrdinaryDiffEqCore.fixed_t_for_tstop_error!), Vararg,
}
```

In the no-tstop branch (the common case), `fixed_t_for_tstop_error!` returns its second argument `ttmp` unchanged, and the caller assigns it back into `integrator.t`:

```julia
integrator.t = fixed_t_for_tstop_error!(integrator, ttmp)
```

The rule was mirroring the Enzyme `inactive_noinl` rule one-for-one, but the semantics are not symmetric:
- **Enzyme** `inactive_noinl`: function has no derivative effect, but **input tangents flow through to the return value**.
- **Mooncake** `@zero_adjoint`: replaces the return with `zero_fcodual` and a `NoPullback` — **the return tangent is hard-zeroed**.

So on every step Mooncake was silently dropping the tangent of `t`, which then cascaded into the saving/interpolation/error-controller code paths and produced the wrong final gradient. (For Lorenz the RHS doesn't depend on `t`, but the saved-output interpolation does.)

The fix is to drop that one rule. `fixed_t_for_tstop_error!` has no try/catch and Mooncake can trace through it natively. With the rule removed:

| Test | Before | After |
|---|---|---|
| Lorenz `f_dt_sum` `SensitivityADPassThrough` | silent wrong (`@test_broken`) | matches ForwardDiff to ~10 digits |
| `du = -u + sin(t)` (time-dependent) | (was wrong) | matches |
| Lotka-Volterra parameter gradient | (was wrong) | matches |
| Adaptive Lorenz with no specified `dt` | (was wrong) | matches |

Verified on Julia 1.11.9 and 1.12.5.

I also looked at the other "returns a value" rule, `ode_determine_initdt`, which has the same conceptual issue (`dt0` mathematically depends on `u0`). Removing that rule causes a Mooncake compile error — Mooncake can't trace `_ode_initdt_iip` because of try/catch (`UpsilonNode` error). Empirically the `@zero_adjoint` on `ode_determine_initdt` introduces only a small first-step error that's well below `1e-6` against ForwardDiff in every test I ran, so I left it in with an inline comment explaining the trade-off. The other rules (`increment_nf!`, `increment_accept!`, `increment_reject!`, `check_error!`, `log_step!`, `final_progress`, `SciMLBase.check_error`) are all genuine no-tangent functions and stay.

The previous `@test_broken` for `SensitivityADPassThrough` in `test/ad/discrete_adjoints.jl` is now a real `@test`.

### Aside: a separate SciMLBase bug surfaced

While verifying the autodiff_events.jl Mooncake test, I hit a separate, unrelated bug in `SciMLBaseMooncakeExt._scatter_pullback`. The rrule for `getindex(::ODESolution, ::Any)` assumes the index is a *variable* index, but `sol[Int]` returns the *state at that time index*. With the `fixed_t_for_tstop_error!` fix in place, Mooncake now actually exercises the `_scatter_pullback` path instead of short-circuiting on a zero gradient, and the bug becomes a `BoundsError`:

```
BoundsError: attempt to access 2-element Vector{Float64} at index [4]
```

I worked around it locally by having the autodiff_events.jl Mooncake testset use a small wrapper (`test_f2_mc`) that reaches into `sol.u[end][end]` directly to bypass the SciMLBase rrule. This is a real bug worth fixing separately in SciMLBase — happy to open a follow-up issue/PR there.

### Bumps

- `OrdinaryDiffEqCore` 3.29.0 → 3.29.1 (patch — extension bug fix)

test/ad/discrete_adjoints.jl
```diff
@@ -40,24 +48,63 @@ function f_dt_sum(u0)
...
+        # via the SciMLSensitivityMooncakeExt extension. These cover the same
+        # Lorenz problem as the SensitivityADPassThrough test above.
+        @testset "Gradient via $(nameof(typeof(sensealg))) ($(nameof(typeof(sensealg.autojacvec))))" for sensealg in (
+                InterpolatingAdjoint(autojacvec = ZygoteVJP()),
+                InterpolatingAdjoint(autojacvec = EnzymeVJP()),
+                GaussAdjoint(autojacvec = ZygoteVJP()),
+            )
+            mkg = DI.gradient(u_ -> f_dt_sum_sa(u_, sensealg), AutoMooncake(; config = nothing), u0)
+            @test mkg ≈ fdg_sa rtol = 1.0e-6
         end
```
ChrisRackauckas:
```suggestion
```

test/ad/discrete_adjoints.jl
```diff
@@ -40,24 +48,63 @@ function f_dt_sum(u0)
...
-                mkg = DI.gradient(f_dt_sum, AutoMooncake(; config = nothing), u0)
-                mkg ≈ fdg
-            end
+        # SensitivityADPassThrough lets Mooncake differentiate through the
+        # mutating solver internals directly. This was previously @test_broken
+        # because the OrdinaryDiffEqCoreMooncakeExt @zero_adjoint rule on
+        # `fixed_t_for_tstop_error!` was silently zeroing the gradient flow
+        # through `integrator.t` (the function returns its `ttmp` argument
+        # which gets assigned back into `integrator.t`). With that rule
+        # removed, Mooncake gives the correct gradient to within ~10 digits.
```
ChrisRackauckas:
```suggestion
```

test/ad/discrete_adjoints.jl
```diff
@@ -40,24 +48,63 @@ function f_dt_sum(u0)
...
+const _saveat_v = collect(0.0:0.25:3.0)
+const _lorenz_p = [10.0, 28.0, 8 / 3]
+
+function f_dt_sum_sa(u0, sensealg)
+    tspan = (0.0, 3.0)
+    prob = ODEProblem{true, SciMLBase.FullSpecialize}(lorenz_p!, u0, tspan, _lorenz_p)
+    sol = solve(prob, Tsit5(), saveat = _saveat_v, sensealg = sensealg, abstol = 1.0e-12, reltol = 1.0e-12)
+    return sum(sol[1, :])
+end
+
```
ChrisRackauckas:
```suggestion
```

test/ad/discrete_adjoints.jl
```diff
@@ -40,24 +48,63 @@ function f_dt_sum(u0)
...
 u0 = [1.0; 0.0; 0.0]
 
 # Reference jacobian and gradient using ForwardDiff
 fdj = DI.jacobian(f_dt, AutoForwardDiff(), u0)
 fdg = DI.gradient(f_dt_sum, AutoForwardDiff(), u0)
+fdg_sa = DI.gradient(u_ -> f_dt_sum_sa(u_, ForwardDiffSensitivity()), AutoForwardDiff(), u0)
```
ChrisRackauckas:
```suggestion

```

test/ad/discrete_adjoints.jl
```diff
@@ -1,9 +1,17 @@
...
 # Enzyme: skipped due to segfaults (see https://github.com/EnzymeAD/Enzyme.jl/issues/2699)
-# Mooncake: all versions (currently broken)
+# Mooncake: all versions
+#   - SensitivityADPassThrough (true discrete adjoint): works after fixing the
+#     OrdinaryDiffEqCoreMooncakeExt rule on `fixed_t_for_tstop_error!` — that
+#     rule was incorrectly @zero_adjoint and was silently zeroing the gradient
+#     flow through `integrator.t`.
+#   - Continuous adjoints via SciMLSensitivity (e.g. InterpolatingAdjoint with
+#     ZygoteVJP/EnzymeVJP, GaussAdjoint): also tested as a regression check
+#     that the SciMLSensitivityMooncakeExt path works.
```
ChrisRackauckas:
```suggestion
```

test/ad/discrete_adjoints.jl
```diff
@@ -1,9 +1,17 @@
...
 # ForwardDiff: all versions (reference)
 
 using OrdinaryDiffEqTsit5, StaticArrays, DiffEqBase, Test, ForwardDiff
+using SciMLSensitivity
```
ChrisRackauckas:
```suggestion
```

PR #3380, opened 2026-04-10
fix redundant StochasticDiffEq imports

oscardssmith:


lib/OrdinaryDiffEqCore/src/algorithms.jl
```diff
@@ -11,24 +11,6 @@ abstract type StochasticDiffEqRODEAlgorithm <: SciMLBase.AbstractRODEAlgorithm e
...
-abstract type StochasticDiffEqJumpAdaptiveAlgorithm <: StochasticDiffEqAlgorithm end
-abstract type StochasticDiffEqJumpNewtonAdaptiveAlgorithm{
-    CS, AD, FDT, ST, CJ, Controller,
-} <: StochasticDiffEqJumpAdaptiveAlgorithm end
-
-abstract type StochasticDiffEqJumpDiffusionAlgorithm <: StochasticDiffEqAlgorithm end
-abstract type StochasticDiffEqJumpDiffusionAdaptiveAlgorithm <: StochasticDiffEqAlgorithm end
-abstract type StochasticDiffEqJumpNewtonDiffusionAdaptiveAlgorithm{
-    CS, AD, FDT, ST, CJ, Controller,
-} <: StochasticDiffEqJumpDiffusionAdaptiveAlgorithm end
```
ChrisRackauckas: This doesn't seem right???

ChrisRackauckas: They aren't doubled here?

termi-official: They are also defined in StochasticDiffEqCore

ChrisRackauckas: they aren't supposed to be there.

oscardssmith: sounds good. Will remove from there.

PR #3392, opened 2026-04-12
Shift `t` past `d_discontinuities` so `t > t_d` branches work

ChrisRackauckas-Claude:
## Summary

- Advances `integrator.t` by one ULP past a `d_discontinuities` time before the FSAL re-evaluation, so user code written as `if t > t_d; new; else; old; end` sees the *new* regime in the first post-discontinuity step.
- Handles the starting-time case (`t_d == tspan[1]`) explicitly in `__init` and `reinit!`, which the regular `update_fsal!` path (iter > 0 only) cannot reach.
- Documents the canonical `if t > t_d` form in the DiffEqBase common solver options docstring.

## Motivation

Before this PR, a `d_discontinuities` entry at `t_d` caused the solver to land exactly on `t_d` and `reset_fsal!` there. But `reset_fsal!` re-evaluated `f` at `t == t_d`, which for `if t > t_d` is still the *pre*-discontinuity regime. The first post-discontinuity step then integrated with a stale derivative as its first stage. On fixed-step methods the effect was an entire step's worth of lost value; on adaptive methods it was either corrected (extra work) or silently smeared. Additionally, `initialize_tstops` filters out `t_d == t0` with strict inequality, so starting-time discontinuities — useful when a callback needs to activate immediately at `t0` — were silently dropped.

The canonical user-facing form is `if t > t_d`, paired with `d_discontinuities = [t_d]`. The solver lands on `t_d` (saves the pre-jump state), advances `integrator.t` to `nextfloat(t_d)`, and re-evaluates FSAL there. The state `u` is continuous across `t_d` — only the vector field changes, so `u` is left untouched. For non-`AbstractFloat` time types (e.g. `Rational`) the shift is a no-op, preserving existing behavior.

`if t >= t_d` also works but forces one post-discontinuity step to start from a pre-regime-evaluated FSAL, which is less efficient and can reduce step acceptance near the discontinuity — noted in the docstring.

## Files changed

- **`lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl`** — new `shift_past_discontinuity!` helper (type-dispatched on `AbstractFloat`), called from `update_fsal!` right after `handle_discontinuities!`.
- **`lib/OrdinaryDiffEqCore/src/solve.jl`** — new `handle_starting_time_discontinuity!`, called at the end of `__init` (gated on `initialize_integrator`).
- **`lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl`** — `reinit!` also calls `handle_starting_time_discontinuity!` so a reinit with `d_discontinuities = [t0]` behaves the same as a fresh `init`.
- **`lib/DiffEqBase/src/solve.jl`** — rewritten `d_discontinuities` docstring in the common solver options section: documents the `t > t_d` canonical form, the shift-and-reset semantics, right-continuity convention, and the starting-time case.
- **`test/interface/ode_tstops_tests.jl`** — four new testsets (seven assertions total).

## Tests

- **Interior discontinuity** — Euler with `dt=0.5` on `f(u,p,t) = t > 5 ? 1 : 0`, tspan `(0,10)`. Without the shift, Euler loses exactly one step's worth (final `u = 4.5`). With the shift, `u(10) ≈ 5.0`. Also tested with Tsit5 at tight tolerance as a regression check.
- **Starting-time discontinuity** — Euler on `f(u,p,t) = t > 0 ? 1 : 0`, tspan `(0,5)`, `d_discontinuities = [0.0]`. Fails without the init-time hook; passes with it.
- **Backward integration** — Tsit5 on `f(u,p,t) = t > 5 ? 1 : 0`, tspan `(10,0)`, `d_discontinuities = [5.0]`. Regression test for `tdir`-aware shift (catches any mistaken use of `nextfloat` instead of `prevfloat` for `tdir < 0`).
- **`integrator.tprev` invariant** — directly verifies that after stepping past a `d_discontinuities` time, `integrator.tprev == nextfloat(t_d)`. This is the cleanest single-assertion check of the shift behavior.

I verified each testset is a proper discriminator by temporarily reverting the shift and confirming the Euler/starting-time/tprev tests all fail with concrete wrong values (`4.5` vs `5.0`; `5.0 == 5.000000000000001`). The Tsit5 interior and backward testsets pass in both directions because adaptive error control and the `t > t_d` convention are naturally insensitive to the bug on those paths — they're retained as regression checks.

All 57 tests in `test/interface/ode_tstops_tests.jl` (50 pre-existing + 7 new) pass locally. The existing `reinit with new d_discontinuities` test in `test/integrators/reinit_test.jl` still passes (uses `Rational` time, where the shift helper no-ops).

## Test plan

- [ ] CI (InterfaceI) passes on the new tests
- [ ] CI (full) passes, no regressions in OrdinaryDiffEqCore, Tsit5, LowOrderRK, Verner, DelayDiffEq, StochasticDiffEq consumers of `d_discontinuities`
- [ ] Runic format check passes
- [ ] DelayDiffEq discontinuities tests still pass (my change runs in DDE's inherited `update_fsal!` path; DDE propagated discontinuities use history integrals that are insensitive to the one-ULP shift)

### DIFF
--- a/lib/DelayDiffEq/src/integrators/interface.jl
+++ b/lib/DelayDiffEq/src/integrators/interface.jl
@@ -508,6 +508,9 @@ function DiffEqBase.reinit!(
         DiffEqBase.initialize!(integrator)
     end
 
+    # Starting-time discontinuity handling: mirrors DDE's __init.
+    OrdinaryDiffEqCore.handle_starting_time_discontinuity!(integrator)
+
     return nothing
 end
 
--- a/lib/DelayDiffEq/src/integrators/utils.jl
+++ b/lib/DelayDiffEq/src/integrators/utils.jl
@@ -232,18 +232,24 @@ function OrdinaryDiffEqCore.handle_discontinuities!(integrator::DDEIntegrator)
 end
 
 # Override shift_past_discontinuity! for DDEIntegrator: after advancing
-# integrator.t by one ULP, also update the internal ODE integrator's solution
-# endpoint so that the `t == ode_integrator.sol.t[end]` check in
-# advance_ode_integrator! remains satisfied.
+# integrator.t by one ULP, add a new save slot to the history ODE integrator's
+# solution at the shifted time. This keeps the `t == ode_integrator.sol.t[end]`
+# invariant in `advance_ode_integrator!` satisfied, and gives the history
+# interpolant a proper boundary at the discontinuity.
 function OrdinaryDiffEqCore.shift_past_discontinuity!(integrator::DDEIntegrator)
     OrdinaryDiffEqCore._shift_past_discontinuity!(
         integrator, integrator.t, integrator.tdir
     )
-    # Keep the ODE history integrator's solution endpoint in sync
+    # Add a new save slot at the shifted time. The state u is continuous
+    # across the discontinuity so we copy the current values unchanged.
     ode_integrator = integrator.integrator
-    if !isempty(ode_integrator.sol.t) && length(ode_integrator.sol.t) > 0
-        ode_integrator.sol.t[end] = integrator.t
-    end
+    ode_integrator.saveiter += 1
+    copyat_or_push!(ode_integrator.sol.t, ode_integrator.saveiter, integrator.t)
+    copyat_or_push!(ode_integrator.sol.u, ode_integrator.saveiter, ode_integrator.u)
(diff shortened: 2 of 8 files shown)


ChrisRackauckas-Claude:
Closing in favor of #3394 (master target). The change is not breaking and doesn't need v7.

PR #3394, opened 2026-04-12
CI check: d_discontinuities shift-past (master target)

ChrisRackauckas-Claude:
**CI-only PR — do not merge.** This is the same change as #3392 (targeting `v7`) cherry-picked onto `master` to verify tests pass on the stable CI infrastructure. Close after CI results are reviewed.

See #3392 for the full description, rationale, and test plan.


lib/DiffEqBase/src/solve.jl
```diff
@@ -351,9 +351,25 @@ section at the end of this page for some example usage.
...
+  if t > t_d
+      # new regime
+  else
+      # old regime
+  end
+  ```
+  Writing `t >= t_d` also works but means the solver takes one step of post-regime
+  integration with a pre-regime-evaluated FSAL, which is less efficient and can
+  reduce step acceptance near the discontinuity. A `d_discontinuities` entry at
+  `tspan[1]` is supported (the starting-time case), in which case the first step
```
ChrisRackauckas, line 370: @oscardssmith see this change

PR #3389, opened 2026-04-11
Widen UJacobianWrapper.p for nested ForwardDiff through Rosenbrock (#3381)

ChrisRackauckas-Claude:
## Summary

- Fixes SciML/OrdinaryDiffEq.jl#3381. When a Rosenbrock integrator is invoked with a `Vector{<:Dual}` `p` (because we are inside an outer ForwardDiff layer — e.g. a NonlinearLeastSquares parameter fit), the inner Jacobian widens `u` into a deeper nested-Dual type via its `JacobianConfig`, but the stored `p` in `UJacobianWrapper` is still at the *outer* Dual level. The user body then multiplies `p[i] * u[i]` across two different Dual nesting levels, which dispatches through ForwardDiff's `tagcount`-based tag precedence. That precedence is a `@generated` function whose literal value is baked at first compile, so its ordering depends on which package's precompile cache first instantiated each tag type — in this nested-AD scenario it inverts the nesting and produces a triple-nested `Dual` that crashes `setindex!(du, ...)` with `MethodError: no method matching Float64(::Dual{..., Dual{..., Float64, 2}, 2})`.
- `jacobian!` now inspects the prepared `ForwardDiff.JacobianConfig` to recover the inner xdual buffer's element type and lifts `uf.p` into that nested-Dual type via a fresh `UJacobianWrapper` before delegating to `DI.jacobian!`. The widened `p` carries zero inner partials (correct — `p` does not depend on `u`). The fast path (`_widen_uf_p_for_jac(f, prep) = f`) is a single type-stable dispatch so non-nested calls incur no overhead.
- Adds a regression test that builds an outer Dual `p` with a throwaway tag type and solves a Rosenbrock23 ODE, exercising the exact call graph that crashes pre-fix and succeeds post-fix. The test is self-contained within `OrdinaryDiffEqDifferentiation`'s test deps (no NonlinearSolve needed).
- Bumps `OrdinaryDiffEqDifferentiation` patch version 2.8.0 → 2.8.1.

## Root cause (abbreviated)

The user's MWE from SciML/OrdinaryDiffEq.jl#3381:

```julia
using OrdinaryDiffEqRosenbrock, NonlinearSolve
function ode(du, u, p, t)
    du[1] = -p[1]*u[1]
    du[2] = -u[1] - p[2]*u[2]
end
ode_f = ODEFunction{true, SciMLBase.FullSpecialize}(ode)
[...]

PR #3406, opened 2026-04-12
Fix nested ForwardDiff tag mismatch 

AdityaPandeyCN:
## Checklist

## Additional context
Fix for [SciML/OrdinaryDiffEq.jl#3381](https://github.com/SciML/OrdinaryDiffEq.jl/issues/3381). When NonlinearSolve computes a Jacobian over an ODE solve, `p` enters as `Dual{NLTag}` and Rosenbrock's internal ForwardDiff seeds `u` as `Dual{OrdEqTag, Dual{NLTag}, CS}`. The user function f(du, u, p, t) multiplies across tag levels, producing the wrong outer tag and crashing on `setindex!`.

### DIFF
--- a/lib/DiffEqBase/ext/DiffEqBaseForwardDiffExt.jl
+++ b/lib/DiffEqBase/ext/DiffEqBaseForwardDiffExt.jl
@@ -204,6 +204,24 @@ if !hasmethod(nextfloat, Tuple{ForwardDiff.Dual})
     end
 end
 
+struct WidenedDualArray{T,A} <: AbstractVector{T}
+    parent::A
+end
+Base.size(a::WidenedDualArray) = size(a.parent)
+Base.@propagate_inbounds Base.getindex(a::WidenedDualArray{T}, i::Int) where {T} = convert(T, a.parent[i])
+
+function (ff::SciMLBase.UJacobianWrapper{true})(du1, uprev::AbstractArray{<:ForwardDiff.Dual})
+    p = ff.p
+    if p isa AbstractArray && eltype(p) <: ForwardDiff.Dual
+        DualU = eltype(uprev)
+        if !(eltype(p) <: DualU)
+            hasfield(typeof(ff.f), :f) && getfield(ff.f, :f) isa FunctionWrappersWrappers.FunctionWrappersWrapper && return ff.f(du1, uprev, p, ff.t)
+            return ff.f(du1, uprev, WidenedDualArray{DualU, typeof(p)}(p), ff.t)
+        end
+    end
+    ff.f(du1, uprev, p, ff.t)
+end
+
 import PrecompileTools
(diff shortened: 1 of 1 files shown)


lib/DiffEqBase/ext/DiffEqBaseForwardDiffExt.jl
```diff
@@ -204,6 +204,24 @@ if !hasmethod(nextfloat, Tuple{ForwardDiff.Dual})
...
+end
+Base.size(a::WidenedDualArray) = size(a.parent)
+Base.@propagate_inbounds Base.getindex(a::WidenedDualArray{T}, i::Int) where {T} = convert(T, a.parent[i])
+
+function (ff::SciMLBase.UJacobianWrapper{true})(du1, uprev::AbstractArray{<:ForwardDiff.Dual})
+    p = ff.p
+    if p isa AbstractArray && eltype(p) <: ForwardDiff.Dual
+        DualU = eltype(uprev)
+        if !(eltype(p) <: DualU)
+            hasfield(typeof(ff.f), :f) && getfield(ff.f, :f) isa FunctionWrappersWrappers.FunctionWrappersWrapper && return ff.f(du1, uprev, p, ff.t)
```
ChrisRackauckas, line 218: No need to do it like this, it should just be in the solve dispatch?

AdityaPandeyCN: I tried it but the `DiffEqBase` solve path can't promote `p` because the nested dual type doesn't exist yet at that point , it's created later by `Rosenbrock's JacobianConfig`. Promoting early caused triple nesting.

ChrisRackauckas: It's incorrect to just unwrap the functionwrapper though, that means the wrong call is getting compiled, so it needs to be handled at the source of the issue.

AdityaPandeyCN: Makes sense. I'll move the fix to `jacobian!` in `derivative_wrappers.jl `, widen p once before `DI.jacobian! `runs, same approach as [#3389.](https://github.com/SciML/OrdinaryDiffEq.jl/pull/3389) or you could reopen that.

ChrisRackauckas-Claude:
Closing: this is fixed at the root in ForwardDiff itself by [JuliaDiff/ForwardDiff.jl#820](https://github.com/JuliaDiff/ForwardDiff.jl/pull/820) (containment-proof tag ordering, successor of ForwardDiff#807), so the downstream workaround is no longer needed. Thanks for the investigation on this — the tagcount-inversion diagnosis and the ≺ suggestion in the ForwardDiff#724 thread fed directly into the final fix.

Issue #3029, opened 2026-02-02
Aitken-Neville uses wrong controller


termi-official:
**Describe the example**

When refactoring the controller interface I noticed that Aitken-Neville uses a PI controller instead of a custom extrapolation controller.  IIUC this is not correct, as changing the coefficients in this controller class permanently yields sub-optimal control performance. It seems that it was missed in one of the previous updates of the sublibrary, as the error estimation part in perform step looks differently that the implementation of all other extrapolation algorithms (c.f. https://github.com/SciML/OrdinaryDiffEq.jl/blob/10501efb24c8009689662e3cdd8d166a2737eece/lib/OrdinaryDiffEqExtrapolation/src/extrapolation_perform_step.jl#L107-L110 ).

PR #3242, opened 2026-03-26
OrdinaryDiffEq v7 and all associated breaking updates

ChrisRackauckas:
This is a PR branch that is accumulating all of the breaking updates for the OrdinaryDiffEq v7 and all sublibraries (including DiffEqBase) with this. This does not include SciMLBase breaking change because that's in a separate repo, and RecursiveArrayTools.jl, etc. Changes to here will be added by targeting feature PRs to it.

Current PRs in here:

* https://github.com/SciML/OrdinaryDiffEq.jl/pull/3242

PR #3418, opened 2026-04-13
Controller interface for v7

termi-official:
## Not addressed in this PR

* [x] Move EEst into controller cache
* [ ] Remove legacy alg utils https://github.com/SciML/OrdinaryDiffEq.jl/pull/3312#discussion_r3065580443
* [ ] Fix Aitken-Neville controller https://github.com/SciML/OrdinaryDiffEq.jl/issues/3029


ChrisRackauckas:
This needs to rebase onto the latest, I see this was before the staticarray changes

termi-official:
I rebased against https://github.com/termi-official/OrdinaryDiffEq.jl/tree/v7 which I mirrored against https://github.com/SciML/OrdinaryDiffEq.jl/tree/v7 about 4 hours ago, but I missed some of the static arrays for the new code changes during the rebase. These should be included in the last commit. Or am I missing something else?

ChrisRackauckas:
There shouldn't be any test failures: current v7 is in a mergable state. So I don't know if those are rebase issues or what but there's a lot to work through here 😅

termi-official:
I am going through the errors one by one. Most of them are typos in variable names. Some failures which I see locally are unrelated to the PR tho: 

* OrdinaryDiffEqNewmark fails due to a missing dependency (ADTypes)
* DelayDiffEq fails because it is in a partially upgraded state -- I included hotfixes to make the non-stochastic case work for now only
* StochasticDiffEqWeak seems to get stuck in the init logic. I think this is unrelated too.


Give me an hour or two to go through the remaining failures.

termi-official:
If no additional surprises pop up in the CI I think this is good from my side. The last commits were fixes to several CI failures for runners which seem to be skipped in https://github.com/SciML/OrdinaryDiffEq.jl/pull/3242 (this CI here has >200 runners and the v7 PR has <150).

PR #3419, opened 2026-04-13
Rename integrator.u_modified field to derivative_discontinuity

ChrisRackauckas-Claude:
## Summary
- Renames the `u_modified::Bool` field on `ODEIntegrator` / `DDEIntegrator` to `derivative_discontinuity::Bool` to match the SciMLBase v3 rename of `u_modified!` → `derivative_discontinuity!` (SciML/SciMLBase.jl#1289 / #2310).
- The field is what actually drives the FSAL reset, Jacobian refresh, multistep restart and DAE reinit paths inside `perform_step!` / `handle_callbacks!`, i.e. it encodes a derivative discontinuity rather than "`u` was rebound", so the new name reflects the real semantics.
- `u_modified!(integrator, bool)` is kept as a public setter (it is `@deprecate`d in SciMLBase v3 but still forwarded); it now assigns to the renamed field, so existing user callback code continues to work.

## Scope
- Struct fields renamed in `OrdinaryDiffEqCore/src/integrators/type.jl` and `DelayDiffEq/src/integrators/type.jl`. `SDEIntegrator` / `RODEIntegrator` are type aliases of `ODEIntegrator` so they follow automatically.
- Every `.u_modified` field access, the constructor positional arg in `OrdinaryDiffEqCore/src/solve.jl` and `DelayDiffEq/src/solve.jl`, and the temporary `u_modified = integrator.u_modified` save/restore locals in `BDF/src/{bdf,dae}_perform_step.jl` are renamed to `derivative_discontinuity`.
- Also renames the internal helper `on_u_modified_at_init!` → `on_derivative_discontinuity_at_init!`.
- `u_modified!` function name, exports, docstrings, and test file names are preserved.

## Breaking
Direct accesses of `integrator.u_modified` (as a field) will break — that is the point of the rename, and v7 is the slot for it. Users should either:
- use `u_modified!(integrator, bool)` / `derivative_discontinuity!(integrator, bool)` (unchanged), or
- migrate field reads/writes to `integrator.derivative_discontinuity`.

## Test plan
- [x] Ran `test/interface/umodified_test.jl` against the rebuilt stack — passes.
- [x] Ran `lib/OrdinaryDiffEqBDF/test/dae_u_modified_tests.jl` — passes.
- [x] Smoke test: bouncing-ball `ContinuousCallback` with `Tsit5()` — events fire and `sol.retcode == Success`.
- [ ] Full CI on the v7 branch.


ChrisRackauckas-Claude:
Dropped the `u_modified!` compat shim entirely per review.

Changes in 8a59165:
- Removed the `u_modified!(::ODEIntegrator, bool)` and `(::DDEIntegrator, bool)` forwarders. `derivative_discontinuity!` is the only public setter now.
- Dropped `u_modified!` from every explicit re-export in this tree and exported `derivative_discontinuity!` instead.
- Renamed `dae_u_modified_tests.jl` and `umodified_test.jl` (and the matching `@safetestset` labels) to the new name.
- In `DiffEqBase/src/callbacks.jl`, short-circuited `SciMLBase.INITIALIZE_DEFAULT` in the `CallbackSet` init path — SciMLBase 2.x's `INITIALIZE_DEFAULT` still calls `u_modified!(integrator, false)`, so without the bypass we'd be forced to keep the shim just to make the default callback initializer work. The bypass sets `integrator.derivative_discontinuity = false` directly and preserves the existing "all-default => no FSAL reset" semantics.
- Left `DiffEqBase.solve.jl`'s `SciMLBase.u_modified!(::NullODEIntegrator, u)` no-op in place — it extends SciMLBase 2.x's own function and has nothing to do with the renamed field.

ChrisRackauckas-Claude:
Retargeted at SciMLBase v3 per review. Details in 07f8f9c8:

- `SciMLBase` compat bumped to `"3"` and `RecursiveArrayTools` compat bumped to `"4"` across every sublibrary (SciMLBase 3 requires RAT 4).
- The method is now a plain `SciMLBase.derivative_discontinuity!(::ODEIntegrator/::DDEIntegrator/::NullODEIntegrator, …)` extension — no locally-defined function, no `using OrdinaryDiffEqCore: derivative_discontinuity!` workarounds, no per-sublib explicit re-exports. `@reexport using SciMLBase` already surfaces the name.
- Removed the `INITIALIZE_DEFAULT` short-circuit in `DiffEqBase/src/callbacks.jl` — on SciMLBase 3 it already calls `derivative_discontinuity!`, so the bypass is unnecessary.
- Converted the `NullODEIntegrator` no-op and its test over to `SciMLBase.derivative_discontinuity!`.

Local precompile under SciMLBase 3 surfaces a pre-existing, unrelated `SciMLBase.DEAlgorithm` reference at `lib/OrdinaryDiffEqCore/src/alg_utils.jl:544` (the v3 removal of deprecated type aliases). That is not part of this rename — leaving it for a separate SciMLBase v3 cleanup PR.

PR #3420, opened 2026-04-13
TSRKC3 method

MoisaAndrew:
## Checklist

- [+] Appropriate tests were added
- [+] Any code changes were done in a way that does not break public API
- [+] All documentation related to code changes were updated
- [+] The new code follows the
- [+] Any new documentation only uses public API
  
## Additional context

Filament


docs/src/semiimplicit/StabilizedRK.md
```diff
@@ -46,8 +46,8 @@ If not provided, the methods include automatic estimation procedures.
 
 ### Recommended stabilized methods
 
-  - **`ROCK2`**: Second-order ROW-type stabilized method with extended stability
```
MoisaAndrew, line 49: Changes to consider. I was not able to run multiple tests due to conflicts between dependencies. You can revert these changes.

There were several bugs in RKC implementation, this is the reason why it was loosing against ROCK2.
Now RKC is at least similar to ROCK2, but more stable and robust (see https://doi.org/10.1007/978-3-662-09017-6).

TSRKC3 is more effecient than RKC for moderate accuracy (in most real-life cases) while ROCK4 is very unstable and almost always slower than either CVODE_BDF or TSRKC3.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -451,11 +451,8 @@ end
     (; t, dt, uprev, u, f, p, fsalfirst) = integrator
     alg = unwrap_alg(integrator, true)
     alg.eigen_est === nothing ? maxeig!(integrator, cache) : alg.eigen_est(integrator)
-    # The the number of degree for Chebyshev polynomial
-    #maxm = max(2,Int(floor(sqrt(integrator.opts.internalnorm(integrator.opts.reltol,t)/(10*eps(integrator.opts.internalnorm(uprev,t)))))))
-    maxm = 50
```
MoisaAndrew, line 456: This limitation was removed due to the following reasons:
1. The `function dtnew_modification(integrator, alg::ROCK2, dtnew)`, which should recalculate the step size if mdeg was updated by maxm, is not implemented.
2. The original paper (https://doi.org/10.1016/S0377-0427(97)00219-7) and the code (https://www.netlib.org/ode/rkc.f) contains the value of maxm: sqrt(rtol / (10. * uround)). I've never seen it reached in contrast to the hardcoded value 50. This value is more theoretical, RKC-type methods gain stability with greater number of internal stages (in contrast to ROCK family).

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -451,11 +451,8 @@ end
...
-    mdeg = 1 + Int(floor(sqrt(1.54 * abs(dt) * integrator.eigen_est + 1)))
-    mdeg = (mdeg > maxm) ? maxm : mdeg
+    # The number of degree for Chebyshev polynomial
+    mdeg = floor(Int, sqrt(1.54 * abs(dt) * integrator.eigen_est + 1)) + 1
 
     w0 = 1 + 2 / (13 * (mdeg^2))
```
MoisaAndrew, line 457: w0 has Float64 type now. That is why this method cannot be used on GPU. The reason why I have not fixed this is because with the type Float32 the values of w0 and the variables below will be too inaccurate. I need more time to choose the right way.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -508,18 +505,19 @@ end
             d2z1 = d2z
         end
     end
+    integrator.fsallast = f(u, p, t + dt)
+    OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
     # error estimate
     if integrator.opts.adaptive
-        tmp = 0.8 * (uprev - u) + 0.4 * dt * (fsalfirst + gprev)
```
MoisaAndrew, line 513: There was a bug in the estimator

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -580,8 +575,8 @@ end
             dt * μs * (k - νs * fsalfirst)
         th = μ * th1 + ν * th2 + μs * (1 - νs)
         if (iter < mdeg)
-            gprev2 = gprev
-            gprev = u
+            @.. broadcast = false gprev2 = gprev
```
MoisaAndrew, line 578: This is the reason why RKC was rubbish: it just did not work correctly.

docs/src/semiimplicit/StabilizedRK.md
ChrisRackauckas: It looks like even in your WP-diagram that RKC is outperformed by ROCK2? ROCK2 should probably still be one of the recommended. But yes TSRKC3 looks pretty good.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
ChrisRackauckas: Okay that's a reasonable change.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
ChrisRackauckas: We might as well just make this constant be a rational so the method is generic? It might not be a good method to use with Float32 but I don't see a reason to block the implementation to allow testing and benchmarking.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
ChrisRackauckas: Ahh thanks, is there a test that catches this? Seems odd it would still converge at higher order? I guess order 2 is just relatively easy to hit and usually higher order methods hit order 1 or 2 when they have a bug like this.

ChrisRackauckas:
Thanks! It would be good to go back through with a fine-toothed comb the SERK methods as well, since they have always underperformed and I wondered if there was some improvements to the error estimators (or bug fixes) required there.

docs/src/semiimplicit/StabilizedRK.md
MoisaAndrew: ROCK2 usually wins a few percents of the execultion time against RKC. However, there are two conditions:
1. 200 stages of ROCK2 is enough to archieve stability without a step size decrease.
2. The order reduction does not occur.

So the dilemma is would you like to take a risk in order to win 3% with a possibility to loose 50?
(see https://doi.org/10.1007/978-3-662-09017-6 page 437)

MoisaAndrew: I wouldn’t want to draw conclusions from a single test, so maybe we just should wait for the community tests.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
MoisaAndrew: I think the reason is that the current convergence tests ODEs are nonstiff, mdeg is always 2 and the code inside this `if` statement is never achieved

oscardssmith: we really need to stop using linear problems for convergance testing

MoisaAndrew: I don't think the linear problems are unsuitable, but we need to force stabilized methods to use more stages using `eigen_est` parameter. I am going to update the test.

src/OrdinaryDiffEq.jl
```diff
@@ -42,8 +42,8 @@ export OrdinaryDiffEqExtrapolation, AitkenNeville, ExtrapolationMidpointDeuflhar
     ImplicitEulerBarycentricExtrapolation
 
 import OrdinaryDiffEqStabilizedRK: OrdinaryDiffEqStabilizedRK
-using OrdinaryDiffEqStabilizedRK: ROCK2, ROCK4, RKC, ESERK4, ESERK5, SERK2
-export OrdinaryDiffEqStabilizedRK, ROCK2, ROCK4, RKC, ESERK4, ESERK5, SERK2
+using OrdinaryDiffEqStabilizedRK: ROCK2, ROCK4, RKC, ESERK4, ESERK5, SERK2, TSRKC3
+export OrdinaryDiffEqStabilizedRK, ROCK2, ROCK4, RKC, ESERK4, ESERK5, SERK2, TSRKC3
```
```suggestion
using OrdinaryDiffEqStabilizedRK: ROCK2, ROCK4, RKC, ESERK4, ESERK5, SERK2
export OrdinaryDiffEqStabilizedRK, ROCK2, ROCK4, RKC, ESERK4, ESERK5, SERK2
```

No need to add this here because that's all being removed in the v7 anyways.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
ChrisRackauckas: Yeah linear can be stiff, that's separate. But https://github.com/SciML/DiffEqProblemLibrary.jl/blob/master/lib/ODEProblemLibrary/src/nonlinchem.jl is probably a good test case. I think we should generally be moving to this, + a stiff case, + a time-dependent case for everything. It would be great to just extend that nonlinchem problem to cover all 3 with a variable parameter, it's probably not too hard to derive a variation with `+ t` in there. But that's a whole separate point and should wait for the v7.

lib/OrdinaryDiffEqStabilizedRK/test/rkc_tests.jl
```diff
@@ -55,9 +55,26 @@ end
         sim = test_convergence(dts, prob, ROCK4(min_stages = 21))
         @test sim.𝒪est[:l∞] ≈ 4 atol = testTol
 
-        println("ROCKC")
-        sim = test_convergence(dts, prob, RKC())
+        println("RKC")
```
MoisaAndrew, line 58: The old RKC code fails now

lib/OrdinaryDiffEqStabilizedRK/test/rkc_tests.jl
```diff
@@ -55,9 +55,26 @@ end
         sim = test_convergence(dts, prob, ROCK4(min_stages = 21))
```
MoisaAndrew, line 55: Actually, the ROCK4 convergence troubles with greater number of stages could be seen here. If you put min_stages = 100 you get
```
Runge-Kutta-Chebyshev Convergence Tests: Test Failed
  Expression: ≈(sim.𝒪est[:l∞], 4, atol = testTol)
   Evaluated: 3.3838195804669873 ≈ 4 (atol=0.1)
```

In contrast, RKC and TSRKC3 pass this test even with thousands of stages.

ChrisRackauckas: Interesting, a floating point error thing? ROCK4 was tested against the original Fortran code so it should be pretty much 1-1 there, but I'm curious if that's a bug or a numerical thing. If you `plot(sim)` from the convergence does it flatline?

MoisaAndrew: Nevermind, this time it just achieved the highest possible accuracy
image

test/regression/iipvsoop_tests.jl
```diff
@@ -145,7 +145,7 @@ rosenbrock_algs = [
     @test sol_ip.t ≈ sol_scalar.t && sol_ip[1, :] ≈ sol_scalar.u
 end
 
-rkc_algs = [RKC(), ROCK2(), ROCK4(), SERK2()]
+rkc_algs = [RKC(), ROCK2(), ROCK4(), SERK2(), TSRKC3()]
```
ChrisRackauckas:
```suggestion
rkc_algs = [RKC(), ROCK2(), ROCK4(), SERK2(), OrdinaryDiffEq.OrdinaryDiffEqStabilizedRK.TSRKC3()]
```

ChrisRackauckas:
The failures are just from the CI machine's disk filling yesterday, so I'll pull this in and handle the random https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/24400891421/job/71282844187?pr=3420 to make the static analyzer happy.

I don't know about the suggestions change right now because I need to see it in action, we'll need to bump the SciMLBenchmarks and see how TSRKC3 performs across the board. But we can always come back to that. My guess is that ROCK2 is added back and that's the only change, I'd have to empirically see RKC do a lot better than ROCK2 (which historically has been the most robust of these methods) before pulling ROCK2 from the list. Not saying I don't believe it, I mean this kind of change will really change RKC and I'll be curious to see how the benchmarks look, but pulling what was the best method of this type in almost all benchmarks completely off the suggestion list is a very bold statement that needs evidence, so let's bump the benchmarks and see if that evidence arrives.

PR #3427, opened 2026-04-14
DelayDiffEq: regression test for constant-lag shift_past_discontinuity (fixes #3426)

ChrisRackauckas-Claude:
## Summary

- Adds a regression test in `lib/DelayDiffEq/test/interface/discontinuities.jl` for SciML/OrdinaryDiffEq.jl#3426.
- No source changes — `OrdinaryDiffEqCore.shift_past_discontinuity!(::DDEIntegrator)` and the relaxed DDE/ODE sync check that landed alongside PR #3392 already fix the runtime error on master. Once master is released, users on the affected registered version (DelayDiffEq 5.73.0 + OrdinaryDiffEqCore 3.31.0) will recover.

## Why it slipped through

PR #3392 introduced `shift_past_discontinuity!` which nudges `integrator.t` by one ULP past each discontinuity so user RHS branches written as `if t > t_d` pick up the post-discontinuity regime. DDEs propagate a discontinuity at every constant-lag step, so the shift fires at `t = n*tau` for a problem with `constant_lags = [tau]`. Without a DDE-side override the ODE sub-integrator's `sol.t[end]` stays at the old `t`, and `advance_ode_integrator!`'s exact `t != ode_integrator.sol.t[end]` check errors with `cannot advance ODE integrator` on the next step. I verified by reverting the override + relaxed check on master — the new test errors, and so does the pre-existing baseline `DDE` testset in the same file, which would have caught the regression had #3392's CI exercised DelayDiffEq. Now that DelayDiffEq lives in-tree under `lib/`, the SublibraryCI matrix covers it going forward.

## Test plan

- [x] `julia --project=<env-with-dev-lib> lib/DelayDiffEq/test/interface/discontinuities.jl` passes locally (42 tests total: 38 pre-existing + 4 new assertions across BS3 and Tsit5).
- [x] Reverting the `shift_past_discontinuity!` override and relaxing-check on master reproduces the exact error message from issue #3426, both on the new test and on the baseline `DDE` testset at line 32.

Fixes #3426

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3426, opened 2026-04-14
Cannot solve DDE


AayushSabharwal:
**Describe the bug 🐞**

DDEs aren't solving anymore

**Expected behavior**

**Minimal Reproducible Example 👇**

Without MRE, we would only be able to help you to a limited extent, and attention to the issue would be limited. to know more about MRE refer to [wikipedia](https://en.wikipedia.org/wiki/Minimal_reproducible_example) and [stackoverflow](https://stackoverflow.com/help/minimal-reproducible-example).

```julia
p0 = 0.2;
q0 = 0.3;
v0 = 1;
d0 = 5;
p1 = 0.2;
q1 = 0.3;
v1 = 1;
d1 = 1;
d2 = 1;
beta0 = 1;
beta1 = 1;
tau = 1;
function bc_model(du, u, h, p, t)
    du[1] = (v0 / (1 + beta0 * (h(p, t - tau)[3]^2))) * (p0 - q0) * u[1] - d0 * u[1]
    du[2] = (v0 / (1 + beta0 * (h(p, t - tau)[3]^2))) * (1 - p0 + q0) * u[1] +
        (v1 / (1 + beta1 * (h(p, t - tau)[3]^2))) * (p1 - q1) * u[2] - d1 * u[2]
    return du[3] = (v1 / (1 + beta1 * (h(p, t - tau)[3]^2))) * (1 - p1 + q1) * u[2] - d2 * u[3]
end
lags = [tau]
h(p, t) = ones(3)
h2(p, t) = ones(3) .- t * q1 * 10
tspan = (0.0, 10.0)
u0 = [1.0, 1.0, 1.0]
prob = DDEProblem(bc_model, u0, h, tspan, constant_lags = lags)
alg = MethodOfSteps(Vern9())
sol = solve(prob, alg, reltol = 1.0e-7, abstol = 1.0e-10)
```

**Error & Stacktrace ⚠️**

```julia
ERROR: cannot advance ODE integrator
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:44
  [2] advance_ode_integrator!(integrator::DelayDiffEq.DDEIntegrator{…}, always_calc_begin::Bool)
    @ DelayDiffEq ~/.julia/packages/DelayDiffEq/Nc0IV/src/integrators/utils.jl:55
  [3] advance_or_update_ode_integrator!
[...]

Issue #190, opened 2017-09-09
dtacc has wrong type


devmotion:
I discovered that `integrator.dtacc` should be of type `tType` instead of type `tTypeNoUnits` because of lines
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L226-L232

However, when changing this one additionally has to change the definition of `DDEIntegrator` and `solve.jl` in DelayDiffEq because of line
https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/solve.jl#L252.

PR #3437, opened 2026-04-15
DelayDiffEq: don't rewrite sol.t[end] in shift_past_discontinuity!

ChrisRackauckas-Claude:
## Summary

Fixes `test (DelayDiffEq_Interface, ...)` on master. Reverts the `sol.t[end]` rewriting introduced in #3427 and restores the one-ULP tolerance in `advance_ode_integrator!`. Both `saveat.jl` and `discontinuities.jl` (including the `#3426` regression test added in #3427) now pass.

## What was broken

`shift_past_discontinuity!` for `DDEIntegrator` was doing three writes per propagated discontinuity — `integrator.t`, `ode_integrator.t`, and `ode_integrator.sol.t[end]` — all set to `nextfloat(t)`. The `sol.t[end]` rewrite broke `lib/DelayDiffEq/test/interface/saveat.jl:18/46/75`:

```julia
@test sol.t == dde_int.integrator.sol.t
```

Reproduced locally on the reference `prob_dde_constant_1delay_long_ip`: 6 divergent indices, each at a discontinuity time (t = n·τ), each exactly one ULP. The outer DDE `sol.t` was unchanged; only the inner ODE sub-integrator's `sol.t[end]` got bumped.

## Why not also bump the outer sol.t[end]

I tried that first. It fixes `saveat.jl` but breaks `discontinuities.jl`'s `#190` testset, which expects `t ∈ sol.t` for each integer discontinuity time — bumping `sol.t[end]` to `nextfloat(t)` replaces the integer value with its successor. It is also wrong when the user has a sparse `saveat`, because `sol.t[end]` there is an older user-requested time, not the current step endpoint.

The `#190` expectation is the semantically correct one: the step **completed** at `t = 5.0` exactly, and `5.0` is what belongs in the saved history. The shift is an integration-state nudge to cross the propagated derivative-discontinuity on the next step — it is not a retroactive amendment to what the integrator computed.

## The fix

- `shift_past_discontinuity!`: move `integrator.t` and `integrator.integrator.t` by one ULP; do not touch either `sol.t[end]`.
- `advance_ode_integrator!`: restore `abs(t - ode_integrator.sol.t[end]) > eps(t) && error(…)` so the one-ULP gap between `integrator.t` and `ode.sol.t[end]` is tolerated.

This is the pre-#3427 behavior for the equality check, combined with #3427's narrower `shift_past_discontinuity!` override (still shifting both `t`s — that part was load-bearing for #3426).

## Test plan

Ran locally in a clean worktree off `upstream/master`:

- [x] `ODEDIFFEQ_TEST_GROUP=Interface Pkg.test("DelayDiffEq")` — all pass. Includes the previously-failing `saveat` (104/104), `discontinuities` (42/42 — the `#3426` regression test from #3427 still passes), AD, Backwards, Callback, Composite, Constrained, Dependent Delays, Export, Fixed-point Iteration, History Function, Jacobian, Mass Matrix, Default Solver, Parameterized Function, save_idxs, Unconstrained, Units, Verbosity, Multi-Algorithm.
- [x] `ODEDIFFEQ_TEST_GROUP=Integrators Pkg.test("DelayDiffEq")` — pass.
- [x] `ODEDIFFEQ_TEST_GROUP=Regression Pkg.test("DelayDiffEq")` — pass (includes Waltman).
- [x] Direct reproducer on `prob_dde_constant_1delay_long_ip`: `sol.t == dde_int.integrator.sol.t` → `true`, 0 divergent indices (vs. 6 on master).
- [ ] CI green across the full matrix.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/lib/DelayDiffEq/src/integrators/utils.jl
+++ b/lib/DelayDiffEq/src/integrators/utils.jl
@@ -51,8 +51,10 @@ function advance_ode_integrator!(integrator::DDEIntegrator, always_calc_begin = 
     ode_integrator = integrator.integrator
 
     # algorithm only works if current time of DDE integrator equals final time point
-    # of solution
-    t != ode_integrator.sol.t[end] && error("cannot advance ODE integrator")
+    # of solution (allow one-ULP gap from shift_past_discontinuity!, which nudges
+    # integrator.t past a propagated discontinuity without rewriting sol.t[end])
+    abs(t - ode_integrator.sol.t[end]) > eps(t) &&
+        error("cannot advance ODE integrator")
 
     # complete interpolation data of DDE integrator for time interval [t, t+dt]
     # and copy it to ODE integrator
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
This is a revert tested locally.

PR #3365, opened 2026-04-07
Move DEVerbosity and default verbosity to DiffEqBase

jClugstor:
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API

PR #3438, opened 2026-04-15
Make _process_verbose_param(::Bool) constprop-friendly to fix @inferred solve

ChrisRackauckas-Claude:
**Depends on #3365** — this branch is `move_verbosity` + one extra commit. Once #3365 merges, only the new commit (`Make _process_verbose_param(::Bool) constprop-friendly`) will remain on the diff against master. If preferred, this can be merged into #3365's branch first; mirror PR was at jClugstor/OrdinaryDiffEq.jl#2 (closed in favor of this).

## Summary
- `DEFAULT_VERBOSE` and `NONE_VERBOSE` are different concrete parameterizations of `DEVerbosity{...}` (different `Silent`/`WarnLevel`/`InfoLevel` type parameters per toggle), so the ternary `verbose ? DEFAULT_VERBOSE : NONE_VERBOSE` returns a `Union{...}` whenever `verbose::Bool` isn't a compile-time constant.
- That `Union` flows into the solver caches — you can see `LinearVerbosity{Silent,Silent,...,WarnLevel,WarnLevel,...}` baked into the `LinearCache` type in #3365's CI failure logs — and it poisons return-type inference all the way to `ODESolution`. Tests like `@inferred solve(prob_mm, Rodas5P(), ...)` in `lib/OrdinaryDiffEqRosenbrock/test/dae_rosenbrock_ad_tests.jl` fail with `does not match inferred return type Any`.
- This PR annotates the `Bool` method with `Base.@constprop :aggressive` so the compiler threads the literal default `verbose = true` through the kwargs chain to a single concrete `DEVerbosity` parameterization.
- Adds `Val{true}`/`Val{false}` overloads as an explicit type-stable entry point for callers that don't want to rely on constprop reaching them.

The runtime-`Bool`-with-no-constprop case still returns a `Union`; that's an inherent property of the `@verbosity_specifier` design (verbosity levels are encoded as type parameters). The proper fix for that case is to keep verbosity out of the cache type parameters, but this is the smallest change that unblocks #3365's `@inferred` tests.

## Test plan
- [x] New `lib/DiffEqBase/test/verbose_inference.jl` exercises both the `Val`-dispatch path (unconditionally inferable) and the literal-`Bool` constprop path (verifies `Base.return_types` resolves to a single concrete type).
- [x] Runic clean.
- [ ] Re-run the Rosenbrock / DiffEqDevTools / Integrators test groups that were failing on #3365 with `@inferred solve(...)` errors.


jClugstor:
Testing the MWE in https://github.com/SciML/OrdinaryDiffEq.jl/issues/3351 with this change, I still get that the inferred return type is `Any`. 

Putting a function barrier like :
```julia
_verbose = verbose isa Bool ? (verbose ? Standard() : None()) : verbose
    return _ode_init(prob, alg, timeseries_init, ts_init, ks_init; verbose = _verbose, kwargs...)
```
in `OrdinaryDiffEqCore.__init` doesn't let `@inferred` pass, but it makes the inference better, it infers as a Union of the two possible verbosity objects from the Bool path. 

I haven't found a way to make the Bool path fully infer yet.

ChrisRackauckas:
This at least makes constant prop work which I think is good enough for the pre-v7 if verbosity is fully supported downstream, because then the defaults and the suggested ways to set `verbose` are all covered.

Issue #192, opened 2017-09-10
Add Residual Control RK Methods


ChrisRackauckas:
http://epubs.siam.org/doi/pdf/10.1137/0730074

The main novelty of these is how they define the error estimate. These can be good for delay diffeqs.

PR #3441, opened 2026-04-15
Fix spurious tiny final step for fixed-dt methods

ChrisRackauckas-Claude:
## Summary

Follow-up to #2869 addressing the regression surfaced in [PositiveIntegrators.jl #192](https://github.com/NumericalMathematics/PositiveIntegrators.jl/pull/192#issuecomment-regression). With OrdinaryDiffEqCore v3.12+, `solve(prob, Euler(); dt = 0.1)` on `tspan = (0.0, 1.0)` started returning 12 steps instead of 11, with a spurious trailing step at `0.9999999999999999` followed by `1.0`. PR #2869 removed the `100eps(tstop)` snap hack that used to paper over this.

Two coordinated changes fix it without re-introducing the old hack:

1. **`_ode_init`**: when the user specifies `dt` for a fixed-time method and did not supply their own `tstops`/`d_discontinuities`, expand `tstops` to `tspan[1]:dt:tspan[end]`. Julia's range semantics use TwicePrecision internally, so each range element is the exact floating-point representative of `k*dt` and lands on `tspan[end]` cleanly. Gated on empty user tstops so the existing `sol.t == [0, 1/3, 1/2, 5/6, 1]` behavior is preserved when users bring their own tstops.

2. **`modify_dt_for_tstops!`**: compare `dt` against `distance_to_tstop` with a small floating-point tolerance (`100 * eps(max(t, tstop))`). Without this, `dtcache = 0.1` reads as strictly less than a `distance_to_tstop` of `0.10000000000000009` and the integrator takes a full step that overshoots by one ulp, then a tiny corrective step. The tolerance guard skips infinite `t`/`tstop` so that `tspan = (0.0, Inf)` integrations are unaffected.

Design suggested by @ChrisRackauckas in the PositiveIntegrators.jl thread:

> I think the best solution is to simply expand `tstops` to `tspan[1]:dt:tspan[2]` here, since Julia's range semantics has some complex logic that makes that take into account floating point error and give accurate tstops.

The tolerance change in `modify_dt_for_tstops!` is needed because denser tstops alone don't fix it — the strict `dtcache < distance_to_tstop` comparison still fails at one ulp mismatches and routes through the drift path.

## Test plan

- [x] New regression test `Fixed-dt: no spurious tiny final step from accumulated drift` in `test/interface/ode_tstops_tests.jl` covers forward, reverse, and non-evenly-dividing `dt` on `tspan = (0.0, 1.0)`.
- [x] Full `test/interface/ode_tstops_tests.jl` still passes (the `dt = 1//3, tstops = [1/2]` test that expects `[0, 1/3, 1/2, 5/6, 1]` is unchanged since the expansion is gated on empty user tstops).
- [x] `test/integrators/ode_event_tests.jl` including the `tspan = (0.0, Inf)` terminate-callback case still passes.
- [ ] Full CI across the monorepo.

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
@@ -178,11 +178,22 @@ function modify_dt_for_tstops!(integrator)
         tdir_t = integrator.tdir * integrator.t
         tdir_tstop = first_tstop(integrator)
         distance_to_tstop = abs(tdir_tstop - tdir_t)
+        # Floating-point tolerance so that a dt whose nominal value matches
+        # distance_to_tstop to within rounding still triggers the tstop
+        # branch.  Without this, accumulated `t + dt + dt + …` can drift
+        # just past the last tstop and produce a spurious micro-step.
+        tstop_tol = if integrator.t isa AbstractFloat && isfinite(tdir_tstop) &&
+                isfinite(integrator.t)
+            100 * eps(float(max(abs(integrator.t), abs(tdir_tstop)) /
+                            oneunit(integrator.t))) * oneunit(integrator.t)
+        else
+            zero(distance_to_tstop)
+        end
 
         if integrator.opts.adaptive
             original_dt = abs(integrator.dt)
             integrator.dtpropose = integrator.tdir * original_dt
-            if original_dt < distance_to_tstop
+            if original_dt + tstop_tol < distance_to_tstop
                 _set_tstop_flag!(integrator, false)
             else
                 _set_tstop_flag!(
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Pushed follow-up commit narrowing the range-expansion gate.

The unconditional expansion tripped three additional cases that surfaced in CI:

1. `inf_handling.jl` — `tspan = (0.0, Inf)` with `adaptive=false, dt=0.1` tripped `InexactError(Int64, Inf)` when constructing `(0.0 + 0.1):0.1:Inf`. Gated on `all(isfinite, tspan)`.
2. `integrator_interface_tests.jl` (`set_proposed_dt!`) — a discrete callback changes `dt` from 0.1 to 0.5 mid-integration. Committing 0.1-spaced tstops at init clips every subsequent step back to 0.1. Gated on `callback === nothing`.
3. `linear_nonlinear_krylov_tests.jl` (LawsonEuler, ETDRK*, HochOst4) — these are `isdtchangeable(alg) == false`. The range's sub-ulp gaps (`diff(0.0:0.01:1.0)` is not uniform) cause `handle_tstop!` to interpolate back when `ttmp` overshoots, which resets `integrator.dt` and trips `apply_step!`'s "setup does not allow for changing dt" guard. Gated on `isdtchangeable(_alg)`.

Local runs now pass on `ode_tstops_tests.jl`, `inf_handling.jl`, `integrator_interface_tests.jl`, and the original PositiveIntegrators.jl Euler bug case.

The AD (`autodiff_events.jl`) and `ODEInterfaceRegression` failures also show up on the current master CI (run 24455830731), so they look unrelated. I'll keep an eye on whatever this run surfaces next.

PR #3468, opened 2026-04-17
Fix SROCK2 NoiseWrapper and non-square noise (#3188, #3170)

### LINKED ISSUE #3170: [SDE] SROCK2 fails for non-square noise and NoiseWrapper
**Migrated from:** https://github.com/SciML/StochasticDiffEq.jl/issues/504, https://github.com/SciML/StochasticDiffEq.jl/issues/505

Two SROCK2 bugs:

1. SROCK2 fails for non-square noise matrices (n,1) - confirmed still broken
2. SROCK2 fails with NoiseWrapper for non-diagonal noise (accesses `W.rng` which doesn't exist)

**MWE for non-square noise (confirmed still broken on StochasticDiffEq v6.100.0):**

```julia
using StochasticDiffEq

function test(n=2, m=1, alg=SROCK2())
  f(x,p,t) = zeros(n)
  g(x,p,t) = ones(n,m)
  x0 = zeros(n)
  nrp = ones(n,m)
  solve(SDEProblem(f, g, x0, 1.0, noise_rate_prototype=nrp), alg, dt=.1)
end

test(3,2, SROCK2())  # works
test(2,1, EM())      # works
test(2,1, SROCK2())  # fails with MethodError
```

### LINKED ISSUE #3188: [SDE] Port SROCK2 NoiseWrapper fix from StochasticDiffEq.jl#507
**Migrated from PR:** https://github.com/SciML/StochasticDiffEq.jl/pull/507
**Author:** @axsk
**Fixes:** https://github.com/SciML/OrdinaryDiffEq.jl/issues/3170

A fix for SROCK2 failing with `NoiseWrapper` noise for non-diagonal problems. The issue was that the code accessed `W.rng` which doesn't exist on `NoiseWrapper`.

**Fix approach:** Extract RNG access into a helper function with dispatch:
```julia
function init_χ!(vec_χ, W)
    rand!(rng(W), vec_χ)
    @.. vec_χ = 2*floor(vec_χ + 1//2) - 1
end

rng(W::DiffEqNoiseProcess.AbstractNoiseProcess) = W.rng
rng(W::NoiseWrapper) = W.source.rng
```

This also fixes `SKSROCK` which had the same issue. The fix was applied to both the out-of-place and in-place code paths in `src/perform_step/SROCK_perform_step.jl`.

**Full diff preserved at:** https://github.com/SciML/StochasticDiffEq.jl/pull/507.diff

singhharsh1708:
Fixes #3188
Fixes #3170

Rebuilt on current master. The KomBurSROCK2 stage math from the original submission landed separately in #3883, so what remains here are the two crashes plus the eq. (17) cross terms.

## The crashes

Single-channel non-square noise was routed into the diagonal code paths by `length(W.dW) == 1` shortcuts. The MWE from #3170:

```julia
function test(n = 2, m = 1, alg = SROCK2())
    f(x, p, t) = zeros(n)
    g(x, p, t) = ones(n, m)
    solve(SDEProblem(f, g, zeros(n), 1.0, noise_rate_prototype = ones(n, m)), alg, dt = 0.1)
end
test(3, 2, SROCK2())  # works
test(2, 1, SROCK2())  # MethodError on master
```

On master that is a `MethodError` out of place and a `DimensionMismatch` in place, for SROCK2, SROCKEM, SROCKC2, KomBurSROCK2 and TangXiaoSROCK2. Separately, #3188's `NoiseWrapper` case dies with `FieldError: type NoiseWrapper has no field rng`, because the χ draw read `W.rng` directly. Both are fixed; `rng(W)` walks to the wrapped source, recurses for nested wrappers, and falls back to the global generator for processes that replay a recorded path and have no generator at all (`NoiseGrid`), which also un-breaks SROCK2 + `NoiseGrid`.

## Eq. (17), and the answer to the review questions

Chris asked whether the algorithm is correct for non-diagonal noise, whether it needs Lévy areas, and whether the paper claims weak order 2 in that setting.

Abdulle, Vilmart & Zygalakis, SIAM J. Sci. Comput. 35(4), 2013 introduces the method "in the most general setting for non-commutative noise of arbitrary dimension m" (§3). Theorem 3.4 gives `|E(φ(X(nh))) − E(φ(X_n))| ≤ C h²` under only `C_P^6` smoothness and Lipschitz continuity — there is no commutativity or diagonality hypothesis. Diagonal noise appears in the paper solely as Remark 3.3, an optimization rather than a restriction.

No Lévy-area computation is needed for weak order 2: that is exactly what eq. (17) buys, replacing the double integrals `I_{q,r}` with `2m−1` discrete random variables. This PR activates those cross terms, which were present but commented out:

```
            ⎧ h(ξ_r ξ_r − 1)/2   if q = r
  J_{q,r} = ⎨ h(ξ_q ξ_r − χ_q)/2 if r < q
            ⎩ h(ξ_q ξ_r + χ_r)/2 if r > q
```

Note the two off-diagonal branches use different χ (χ_q vs χ_r); the implementation follows that. The resulting matrix satisfies the Itô product rule `J_{q,r} + J_{r,q} = ΔW_q ΔW_r` pathwise, so only the Lévy-area part `(J − Jᵀ)/2` is antisymmetric, as it should be.

The terms are not cosmetic. On a non-commutative problem `dX = A X dt + B₁ X dW₁ + B₂ X dW₂` with `B₁B₂ ≠ B₂B₁`, the scheme is linear in X, so the one-step second-moment operator can be computed exactly by Gauss–Hermite quadrature in ΔW crossed with full enumeration of χ — no Monte-Carlo error at all. Error in `E[X₁X₂]` against `exp(M T)`:

| h | master (terms commented out) | this PR |
|---|---|---|
| 1/2 | 5.57e-2 | 2.73e-3 |
| 1/4 | 3.12e-2 (rate 0.83) | 6.33e-4 (rate 2.11) |
| 1/8 | 1.62e-2 (rate 0.95) | 1.52e-4 (rate 2.05) |
| 1/16 | 8.20e-3 (rate 0.98) | 3.74e-5 (rate 2.03) |
| 1/32 | 4.12e-3 (rate 0.99) | 9.26e-6 (rate 2.01) |

Master is weak order **1** on non-commutative noise; with eq. (17) the method converges at its documented order 2.

Per Remark 3.5, strong order 1 on non-commutative noise would require the true multiple integrals, not the Rademacher surrogate — so the `SROCKEM(strong_order_1 = true)` paths deliberately keep their cross terms inactive, and should not be "completed" by analogy.

## Tests

`srock_nondiag_tests.jl` (Core) covers the #3170 routing and the #3188 supplied-noise cases: 47 assertions pass here, and on pristine master the same file gives 8 passes and 6 errors — the exact `MethodError`/`DimensionMismatch`/`FieldError` failures above, so it genuinely guards the fix. The KomBurSROCK2 portion passes on master too; it is coverage, not a guard.

The weak-convergence study lives in its own `SROCK2NonDiagonalConvergence` group, `local_only` with a 240 s timeout, per the earlier review comment about long-running tests. It measures orders 2.024 / 2.025 / 2.028 / 2.051 and asserts the non-commutative second moment (5.8e-3 here versus 5.7e-2 with the eq. (17) hunks reverted, so it fails if they are removed). It takes about 90 s locally and roughly 3 minutes on CI.

## Notes

- `init_χ!` draws in bulk and maps by broadcast rather than looping with scalar indexing, addressing the second half of the earlier `:1852` review comment; the first half (the `Float64` assumption) is handled by keying the element type off `vec_χ`. Full device support would additionally need a generator that the device array's `rand!` accepts, which is outside this PR.
- `rng(W)` uses `hasfield` rather than dispatch because DiffEqNoiseProcess is only a test dependency of this sublibrary, so `NoiseWrapper` cannot be dispatched on here. It would fit more naturally in StochasticDiffEqCore if that is preferred.
- The activated terms use `abs(dt)`, matching the diagonal term and `sqrt_dt` already in this file; the commented-out original used signed `dt`, which was inconsistent with its own diagonal term. For χ symmetric about zero the two are distributionally identical.
- KomBurSROCK2's χ now come from the noise process generator rather than the global one, so its trajectories change for general noise. With a `NoiseGrid` the χ are not part of the recorded path, so repeated solves vary unless the global generator is seeded.
- #3188's description says the original patch "also fixes SKSROCK". SKSROCK does not actually hit either crash on master: it never reads `W.rng`, and it solves both the #3170 non-square case and the #3188 `NoiseWrapper` case unchanged, so there is nothing here for it. Verified across all seven SROCK algorithms: on master the non-square case fails for SROCK2, SROCKEM, SROCKC2, TangXiaoSROCK2 and KomBurSROCK2 (in place and out of place) and the `NoiseWrapper` case fails for SROCK2 only; on this branch all twenty-one combinations solve.
- SKSROCK and TangXiaoSROCK2 do still draw their auxiliary variates from the global generator rather than the noise process. That is pre-existing and does not crash, so it is left alone here rather than changing two more algorithms' trajectories in this PR.

## AI Disclosure


singhharsh1708:
@oscardssmith any reviews on this fix ?

lib/StochasticDiffEqROCK/src/perform_step/SROCK_perform_step.jl
```diff
@@ -418,7 +418,7 @@ end
     # Now uᵢ₋₂ = uₛ₋₂, uᵢ₋₁ = uₛ₋₁, uᵢ = uₛ
     # Similarly tᵢ₋₂ = tₛ₋₂, tᵢ₋₁ = tₛ₋₁, tᵢ = tₛ
 
-    if (W.dW isa Number) || (length(W.dW) == 1) || is_diagonal_noise(integrator.sol.prob)
+    if (W.dW isa Number) || is_diagonal_noise(integrator.sol.prob)
```
ChrisRackauckas, line 422: why is this removed?

lib/StochasticDiffEqROCK/src/perform_step/SROCK_perform_step.jl
```diff
@@ -1842,3 +1845,12 @@ end
 
     integrator.u = u
 end
+
+function init_χ!(vec_χ, W)
+    r = rng(W)
+    for i in eachindex(vec_χ)
+        vec_χ[i] = 2 * floor(rand(r) + 0.5) - 1
```
ChrisRackauckas: this assumes Float64, also isn't gpu compatible.

ChrisRackauckas:
Needs convergence tests, I don't think the algorithm looks right.

singhharsh1708:
The length(W.dW) == 1 check was removed because it caused incorrect behavior for non-diagonal noise with a single noise source (e.g. noise_rate_prototype = ones(n, 1)). In that case length(W.dW) == 1 is true but the noise is non-diagonal, so taking the scalar path u += Gₛ .* W.dW fails — Gₛ is n×1 and adding it to u (length-n vector) causes a DimensionMismatch. The mul! path is always correct for non-diagonal noise regardless of m. The is_diagonal_noise check already handles the scalar path correctly.

singhharsh1708:
DimensionMismatch for non-square noise (noise_rate_prototype = zeros(n, m))
Removed || length(W.dW) == 1 guards in caches and perform_step. For m=1 non-diagonal noise, this condition was incorrectly routing into the scalar path causing dimension errors.
 Crash with NoiseWrapper
NoiseWrapper has no .rng field. Added duck-typed fallback:
rng(W) = hasfield(typeof(W), :rng) ? W.rng : W.source.rng
Also extracted init_χ! using rand! from stdlib Random for type-correctness and GPU compatibility.
KomBurSROCK2 non-diagonal (related)
Fixed DimensionMismatch from @.. Xₛ₋₂ = Gₛ * W.dW → column-by-column @view(Xₛ₋₂[:, i]) .= @view(Gₛ[:, i]). Also fixed sign bug in stage s-1 correction (+ → -).
Tests added (NonDiagonalConvergence group)
SROCK2 OOP + IIP weak convergence order ~2 ✓
KomBurSROCK2 non-diagonal smoke test (no crash) ✓
SROCK2 NoiseWrapper smoke test ✓

ChrisRackauckas:
Is the algorithm correct for non diagonal though? Or does it need levy area calculations? What does the paper say?

singhharsh1708:
The key paper is Abdulle, Vilmart, Zygalakis (2013) SIAM J. Sci. Comput. 35(4). Eq. (17) defines J_{q,r} = (ξ_q ξ_r ± χ_q) · h/2 for q ≠ r using Rademacher χ variables — those are exactly the terms commented out at line 461. For our PR's use case (noise_rate_prototype = zeros(n, 1), m=1), the q ≠ r cross terms never appear (loop runs once, i=j=1 only), so the Lévy area is trivially zero and the algorithm is correct as-is — confirmed by the weak order ~2 convergence test. The commented-out Lévy area is a pre-existing gap for general m>1 non-diagonal noise; implementing it would be a separate PR.

ChrisRackauckas:
Does the paper say it should converge order 2 on general non diagonal?

ChrisRackauckas:
Weak convergence tests should probably be in the special sets for the special runners since they are long running

singhharsh1708:
Yes — Abdulle, Vilmart, Zygalakis (2013) SIAM J. Sci. Comput. 35(4) Theorem 3.4 proves weak order 2 for general non-commutative non-diagonal noise of arbitrary dimension m. The full formula (eq. 17) requires the cross terms J_{q,r} = (ξ_q ξ_r ± χ_q) · h/2 for q ≠ r using Rademacher variables χ. These were previously commented out in both the OOP and IIP SROCK2 paths — now activated. Verified: m=2 non-diagonal gives order ~2.03, m=1 tests still pass.
Also moved NonDiagonalConvergence to the high-memory runner with 240s timeout, same as SROCKC2WeakConvergence.

lib/StochasticDiffEqROCK/test/test_groups.toml
```diff
@@ -6,5 +6,10 @@ versions = ["1"]
 runner = ["self-hosted", "Linux", "X64", "high-memory"]
 timeout = 240
 
+[NonDiagonalConvergence]
```
ChrisRackauckas:
```suggestion
[SROCK2NonDiagonalConvergence]
```

lib/StochasticDiffEqROCK/test/runtests.jl
```diff
@@ -22,3 +22,9 @@ if TEST_GROUP == "ALL" || TEST_GROUP == "SROCKC2WeakConvergence"
         include("weak_convergence/weak_srockc2.jl")
     end
 end
+
+if TEST_GROUP == "ALL" || TEST_GROUP == "NonDiagonalConvergence"
```
ChrisRackauckas:
```suggestion
if TEST_GROUP == "ALL" || TEST_GROUP == "SROCK2NonDiagonalConvergence"
```

singhharsh1708:
@ChrisRackauckas can you check once if it still needs changes.suddenly so many tests started failing ?

ChrisRackauckas:
The big v7 branch merged. The repo won't be in a normal development state until that's all out. Hopefully Friday, trying to get it complete but also it's very delicate and shouldn't be rushed.

PR #3470, opened 2026-04-17
Add arbitrary-order GaussLegendre FIRK method (draft)

### LINKED ISSUE #3454: Implement fully implicit Gauss methods
**What kind of problems is it mostly used for? Please describe.**

Somewhat stiff problems that require symmetric and symplectic methods could benefit from them.

**Describe the algorithm you’d like**

The classical Gauss methods (s stages, order 2s) would be nice to have, e.g., for two, three, and four stages.

**Other implementations to know about**

https://github.com/JuliaGNI/GeometricIntegrators.jl implements them but does apparently not use the specialized implementations to reduce the complexity of the linear algebra routines like the Radau methods implemented in OrdinaryDiffEqFIRK.jl do, cf. http://www.unige.ch/~hairer/preprints/coimbra.pdf.

https://github.com/SciML/IRKGaussLegendre.jl implements the 8-stage version with a specialized fixed-point solver, making it really useful only for non-stiff equations (as described in their README.md).

PETSc implements Gauss methods, but only for sparse Jacobians. It can currently be used via the low-level interface of PETSc.jl, see https://github.com/JuliaParallel/PETSc.jl/blob/main/examples/ex51_implicit.jl

**References**

Any standard textbook on numerical methods for ODEs includes these methods, even Wikipedia: https://en.wikipedia.org/wiki/List_of_Runge–Kutta_methods#Gauss–Legendre_methods

This is related to https://github.com/SciML/OrdinaryDiffEq.jl/issues/521 and https://github.com/SciML/OrdinaryDiffEq.jl/issues/1004. 

The papers https://doi.org/10.1007/s11075-014-9825-0 (preprint available at https://arxiv.org/abs/1304.0974) and https://doi.org/10.1007/s11075-017-0367-0 are related to this.

Sreeram-Shankar:
## Add arbitrary-order Gauss-Legendre FIRK method

Closes #3454

This is a draft implementation of arbitrary-order Gauss-Legendre 
collocation methods as requested in the linked issue. With `s` stages 
the method has classical order `2s` and is symplectic, making it 
suitable for Hamiltonian systems and problems requiring long-time 
geometric integration.

### What's implemented
- `GaussLegendre(num_stages=s)` algorithm struct following the existing 
  Radau interface pattern
- Tableau generation using `FastGaussQuadrature.gausslegendre` with 
  Lagrange basis integration for the `A` matrix
- Tableau cache for Float64 at stages 2-5
- Full Newton iteration on the real-valued `sn × sn` system
- Basic embedded error estimate using s-1 GL quadrature weights for 
  adaptive stepping (placeholder, needs improvement)
- Fixed dt works correctly and produces accurate solutions

### Known issues / TODOs
- Adaptive dt is overly conservative due to the placeholder error estimate
- No stage decoupling yet, I've looked at the approach in the  Antonana, Makazaga,
  and Murua paper but haven't implemented it yet
- No dense output / interpolation
- W matrix construction could be more efficient

### Testing
Basic out-of-place solve works correctly:
```julia
f(u, p, t) = [u[2], -u[1]]
prob = ODEProblem(f, [1.0, 0.0], (0.0, 10.0))
sol = solve(prob, GaussLegendre(num_stages=2), dt=0.1, adaptive=false)
# sol.u[end] ≈ [cos(10), -sin(10)] ✓
```

Happy to iterate on any of this based on feedback.


oscardssmith:
Needs some tests, but otherwise, looks reasonable.

lib/OrdinaryDiffEqFIRK/src/firk_tableaus.jl
```diff
@@ -349,3 +349,82 @@ const RadauIIATableauCache = Dict{
...
+                    if k != j
+                        Lj *= (c[i] - c_low[k]) / (c_low[j] - c_low[k])
+                    end
+                end
+                b_low[i] += T1(w_low[j] / 2) * Lj
+            end
+        end
+        e = b .- b_low
+    else
+        e = b
```
ranocha, line 417: It looks like for `s = 1`, no real embedded method is implemented. Should an adaptive time integration for `s = 1` be checked during initialization to throw an error in this case?

lib/OrdinaryDiffEqFIRK/src/firk_tableaus.jl
```diff
@@ -349,3 +349,82 @@ const RadauIIATableauCache = Dict{
...
+
+function generateGaussLegendreTableau(T1, T2, num_stages::Int)
+    x, w = gausslegendre(num_stages)
+    c = T2.((x .+ 1) ./ 2)
+    b = T1.(w ./ 2)
+
+    A = Matrix{T1}(undef, num_stages, num_stages)
+    for i in 1:num_stages
+        for j in 1:num_stages
+            nodes, weights = gausslegendre(2 * num_stages)
```
ranocha, line 385: As far as I know, FastGaussQuadrature.jl only works with `Float64`. Would it make sense to implement the analytical coefficients for the known cases when higher precision is required?

lib/OrdinaryDiffEqFIRK/src/algorithms.jl
```diff
@@ -231,3 +231,70 @@ function AdaptiveRadau(;
...
+@doc differentiation_rk_docstring(
+    "A symplectic fully implicit Runge-Kutta method based on Gauss-Legendre quadrature.
+With s stages, the method has order 2s. Symplectic and A-stable, making it suitable
+for Hamiltonian systems and problems requiring long-time geometric integration.",
+    "GaussLegendre",
+    "Fully-Implicit Runge-Kutta Method.";
+    references = gauss_legendre_docstring,
+    extra_keyword_description = extra_keyword_description,
+    extra_keyword_default = extra_keyword_default
+)
```
ranocha, line 245: Would it make sense to mark this as experimental for now since some details will likely change? For example, it would probably be good to experiment a bit with some of the optimizations for solving the linear systems. Maybe it would even make sense to use an additional keyword argument to choose which iteration scheme to use, defaulting to the best one found after experimenting with them in OrdinaryDiffEq.jl.

lib/OrdinaryDiffEqFIRK/src/firk_tableaus.jl
oscardssmith: we probably should make s=2 be the minimum (otherwise we just have a trapezoid method)

lib/OrdinaryDiffEqFIRK/src/firk_tableaus.jl
oscardssmith: With https://github.com/JuliaApproximation/FastGaussQuadrature.jl/pull/149 that concern will be resolved.

lib/OrdinaryDiffEqFIRK/src/algorithms.jl
oscardssmith: I think all of these are things we could do without breaking.

lib/OrdinaryDiffEqFIRK/src/firk_tableaus.jl
ranocha: If we use the general (all `s`) implementation, it could be nice to still allow `s = 1` to compare the performance with the `ImplicitMidpoint` method.

oscardssmith: That's fair. I think the lack of a natural embedded method may make integrating it difficult, but it is likely worth a try.

lib/OrdinaryDiffEqFIRK/src/algorithms.jl
Sreeram-Shankar: I marked it as experimental for now in the description, is there anything else I should change?

lib/OrdinaryDiffEqFIRK/src/firk_tableaus.jl
Sreeram-Shankar: I changed the error estimator to a Richardson controller for now since embedding error estimators into the method in the way I tried requires extra stages and is nontrivial. I can look into natural embedded pairings for GL methods and implement that if you have any suggestions.

Sreeram-Shankar: using Gauss-Lobatto quadrature to get 2s-2 order LobattoIIIA methods, which share the same nodes as GL plus the endpoints, could be an option

ranocha: I think it would also be totally fine to just implement Gauss methods as non-adaptive methods for now.

ChrisRackauckas: Yes, one step to get them converging at the right order, then another to make them adaptive, is a very common thing to make the PRs saner to review.

ChrisRackauckas:
Is there an advantage to having this in the same subpackage as the Radau methods?

oscardssmith:
we could split them, but this is a FIRK method, it uses all the same deps, and this is only a single algorithm so the include and image size change should be minimal. IMO this seems like a good place for it.

lib/OrdinaryDiffEqFIRK/src/firk_tableaus.jl
Sreeram-Shankar: Got it, I’ll scope this around the fixed-step Gauss methods so the review stays focused on correctness/order. DID also add convergence/order tests with some basic coverage for the Richardson controller.

Sreeram-Shankar:
I’ve added convergence/order tests (checking 2s order for s = 2, 3 and a higher-order accuracy check for s = 4), plus some basic adaptive tests for the Richardson controller. Also integrated them into the existing FIRK problem tests and allocation tests.

Is there anything else you’d like to see test-wise? I was thinking about symplectic or A-stability checks.

Sreeram-Shankar:
formatted with runic

lib/OrdinaryDiffEqFIRK/src/algorithms.jl
```diff
@@ -220,3 +220,81 @@ function AdaptiveRadau(;
...
+
+function GaussLegendre(;
+        num_stages = 2,
+        chunk_size = Val{0}(), autodiff = AutoForwardDiff(),
+        standardtag = Val{true}(), concrete_jac = nothing,
+        diff_type = Val{:forward}(),
+        linsolve = nothing, precs = nothing,
+        extrapolant = :dense, fast_convergence_cutoff = 1 // 5,
+        new_W_γdt_cutoff = 1 // 5,
+        controller = :Predictive, κ = nothing, maxiters = 10, smooth_est = true,
```
oscardssmith: we probably should use a PI controller until we get order adaptivity

lib/OrdinaryDiffEqFIRK/src/algorithms.jl
```diff
@@ -220,3 +220,81 @@ function AdaptiveRadau(;
...
+    `GaussLegendre` is experimental. Adaptive stepping currently uses Richardson
+    step-doubling (roughly 3× the work per accepted step) and requires `num_stages ≥ 2`;
+    Details may change as the implementation is refined.",
+    "GaussLegendre",
+    "Fully-Implicit Runge-Kutta Method.";
+    references = gauss_legendre_docstring,
+    extra_keyword_description = extra_keyword_description,
+    extra_keyword_default = extra_keyword_default
+)
+struct GaussLegendre{CS, AD, F, P, FDT, ST, CJ, Tol, C1, C2, StepLimiter} <:
```
oscardssmith: the type params here are missing the v7 updates

lib/OrdinaryDiffEqFIRK/src/algorithms.jl
Sreeram-Shankar: Updated to use PI controller

Sreeram-Shankar:
CI is currently failing with a Tsit5 compat error, im not yet sure whether it’s related to the FIRK changes

oscardssmith:
CI isn't your fault. This will probably sit on back burner for a week or so while we get v7 released.

ChrisRackauckas:
CI should be good now.

Sreeram-Shankar:
Synced with master, tests pass locally. Not sure what to do next, happy to take direction and fix anything.

lib/OrdinaryDiffEqFIRK/src/alg_utils.jl
```diff
@@ -25,3 +25,19 @@ get_current_adaptive_order(alg::AdaptiveRadau, cache) = cache.num_stages
...
+alg_order(alg::GaussLegendre) = 2 * alg.num_stages
+
+default_controller(QT, alg::GaussLegendre) = PIController(QT, alg)
+
+isfirk(alg::GaussLegendre) = true
+
+# Richardson step-doubling controller
+isadaptive(alg::GaussLegendre) = alg.num_stages >= 2
+alg_adaptive_order(alg::GaussLegendre) = 2 * alg.num_stages
+has_stiff_interpolation(::GaussLegendre) = true
```
oscardssmith: this should be false until we add a dedicated interpolation.

oscardssmith:
Thanks so much! If you're looking for a next thing to do, getting a good error controller and interpolation for this would be great.

ranocha:
Thanks a lot! I guess I would prefer an improved implementation for solving the systems instead of an error controller; adaptive step size control violates properties like symplecticity and symmetry that are reasons why people would often choose Gauss methods.

oscardssmith:
makes sense. getting the interpolation done would still be nice for continuous output and callbacks.

ranocha:
I agree :+1:

oscardssmith:
I also think there might be a way to get an adaptive controller that is symmetrical. e.g. if you had a controller that was restricted to time-stamping power of 2 multiples of the tspan and you require the solver to take the largest tolerance respecting step or something. that way you could ensure that even though the step size varied, the steps you were taking hit the same points in both directions I think.

ChrisRackauckas:
Even if it's symmetric, it still would lose symplectic properties. You have to step in relation to the Hamiltonian on order to keep that, and it's extremely difficult to do that beyond order 2.

Sreeram-Shankar:
Are there also any docs that need to be updated?

oscardssmith:
Yes. Also now that https://github.com/JuliaApproximation/FastGaussQuadrature.jl/pull/149 is merged, the BigFloat version can be done.

Issue #3482, opened 2026-04-21
v7 regression: FunctionWrappersWrapper-forced chunksize=1 autodiff yields singular Jacobian row at an ifelse kink


ChrisRackauckas-Claude:
Found while chasing the `Regression_I` `Hard DAE` failures on the v7 branch (blocks SciMLBase v3 bump). Full bisect of the symptom to the exact bit-level divergence is done; the fix is non-trivial.

## Symptom

`test/regression/hard_dae.jl` `alg_switch` case fails on v7 but passes on master. `simple_implicit_euler` alone passes on both.

```julia
using OrdinaryDiffEq
using OrdinaryDiffEqSDIRK: ImplicitEuler
using OrdinaryDiffEqCore: CompositeAlgorithm
using OrdinaryDiffEqNonlinearSolve: NLNewton
using DiffEqBase: BrownFullBasicInit
using LinearAlgebra

function hardstop!(du, u, p, t)
    pm, pg = p
    y, f_wall, dy = u
    du[1] = dy
    du[2] = ifelse(y <= 0, y, f_wall)            # algebraic, kinked at y=0
    du[3] = (-ifelse(t < 2, -pg*pm, pg*pm) - f_wall) / (-pm)
end

fun = ODEFunction(hardstop!, mass_matrix = Diagonal([1, 0, 1]))
prob = ODEProblem(fun, [5.0, 0.0, 0.0], (0.0, 4.0), [100.0, 10.0])

cb = DiscreteCallback((u,t,i) -> t == 2, i -> i.u[1] = 1e-6)

N_FAILS = Ref(0)
function choice_function(integrator)
    integrator.do_error_check = false
    if integrator.force_stepfail
        N_FAILS[] += 1
    else
        N_FAILS[] = 0
    end
    (N_FAILS[] > 3) + 1
end

simple_ie = ImplicitEuler(nlsolve = NLNewton(check_div = false, always_new = true))
alg_switch = CompositeAlgorithm((ImplicitEuler(), simple_ie), choice_function)

N_FAILS[] = 0
sol = solve(prob, alg_switch, callback=cb, dt=1/2^10, adaptive=false,
            initializealg=BrownFullBasicInit())
sol(2 - 2^-10, idxs=1)   # v7: 0.275… , master: 4.77e-5
sol(4, idxs=1)           # v7: 21.97,   master: 20.03
```

## Root cause

[...]

PR #3483, opened 2026-04-21
v7 hard_dae: use strict `<` at kink to unblock composite Jacobian (workaround for #3482)

ChrisRackauckas-Claude:
## Summary

Swap `ifelse(y <= 0, y, f_wall)` for `ifelse(y < 0, y, f_wall)` in the `hardstop!` DAE test function. Unblocks `Regression_I` on v7 (1 / 1.11 / pre).

This is a workaround, not a root-cause fix. The underlying bug is tracked in #3482 and should be addressed by changing `DiffEqBase.promote_f` to hold `FunctionWrappersWrapper` signatures for multiple chunksizes (or by removing the `AutoForwardDiff{1}(tag)` coercion in `_prepare_ADType_fwd`).

## Why `<` works here

At `y = 0.0` with `<=`, ForwardDiff's lexicographic comparator gives different answers depending on whether the evaluation carries a non-zero partial on `u[1]`:

| comparison | result |
|---|---|
| `Dual(0.0, 1.0) <= 0.0` | **false** (partial breaks primal tie upward) |
| `Dual(0.0, 0.0) <= 0.0` | **true** (partials tie at zero) |

With v7's forced chunksize=1 autodiff, each Jacobian column is a separate pass and `u[1]`'s partial is non-zero only in the `u[1]`-seeded pass. So column 1 takes the `f_wall` branch and column 2 takes the `y` branch — the assembled `J[2, :]` is `[0, 0, 0]`, a singular algebraic row that leaves Newton unable to correct `u[2]` and `f_wall` stays at 0.

With `<`, `Dual(0.0, p) < 0.0` is `false` whether `p` is zero or non-zero, so every chunksize-1 pass lands on the same branch and `J[2, :] = [0, 1, 0]` — non-singular, Newton recovers, `f_wall → 1000`.

## Physical impact at the kink

The behavior at the measure-zero event `y == 0` flips from "contact engaged" to "contact disengaged". In practice the integrator never lingers at exactly `y = 0` and the constraint re-engages immediately on the next step once `y < 0` via floating-point drift from gravity. Tests `sol(0, idxs=1) == 5`, `abs(sol(2 - 2^-10, idxs=1)) <= 1e-4`, `sol(4, idxs=1) > 10` all still pass.

## Test plan
- [ ] `Regression_I` on 1 / 1.11 / pre
- [ ] Remove the `<`-for-`<=` workaround once #3482 lands

### DIFF
--- a/test/regression/hard_dae.jl
+++ b/test/regression/hard_dae.jl
@@ -253,7 +253,13 @@ function hardstop!(du, u, p, t)
     pm, pg = p
     y, f_wall, dy = u
     du[1] = dy
-    du[2] = ifelse(y <= 0, y, f_wall)
+    # Strict `<` (rather than `<=`) so ForwardDiff's lexicographic comparator on
+    # `Dual(0.0, partial)` stays on the same branch whether the partial is zero
+    # or non-zero.  With `<=` and chunksize=1, different Jacobian columns end up
+    # linearizing different branches at y=0 and J's algebraic row collapses to
+    # zero — see SciML/OrdinaryDiffEq.jl#3482.  Revisit once the chunksize=1
+    # FunctionWrappersWrapper issue is addressed in DiffEqBase.promote_f.
+    du[2] = ifelse(y < 0, y, f_wall)
     return du[3] = (-ifelse(t < 2, -pg * pm, pg * pm) - f_wall) / (-pm)
 end
 


ChrisRackauckas-Claude:
Closing — the `<` substitution fails local testing. At `y = 0` with strict `<` the DAE has no consistent solution (both branches produce contradictions), so Newton lands on a non-physical point: at t=1.01 the ball rebounds with dy = +9.96 m/s and drifts up to `y ≈ 5.02` by t=2⁻. `simple_ie` still passes because its `always_new=true + check_div=false` config iterates through the bad intermediate states back to the y=0 attractor, but the composite's default `ImplicitEuler` commits to its first Newton direction and never recovers. Earlier reasoning on this PR was wrong — `<` removes the stable fixed point the integrator needs at the kink. Real fix remains #3482.

PR #3490, opened 2026-04-21
Move linsolve onto NLNewton to unify with NonlinearSolveAlg

ChrisRackauckas-Claude:
## Summary

Addresses the "two separate paths to pass the linear solver into the nonlinear solve" observation: `NLNewton` now carries its own `linsolve` field, matching how `NonlinearSolveAlg` holds the linear solver on the wrapped NonlinearSolve.jl algorithm. The two configuration paths collapse to a single canonical location on the nlsolve object.

```julia
# Canonical v7 form — works uniformly for any nlsolve
ImplicitEuler(nlsolve = NLNewton(linsolve = KLUFactorization()))
ImplicitEuler(nlsolve = NonlinearSolveAlg(NewtonRaphson(linsolve = KLUFactorization())))
```

The algorithm-level `linsolve` kwarg still works as a convenience — this is a **non-breaking change** for existing user code. Internally, reads route through a new `effective_linsolve(alg)` helper that prefers `alg.nlsolve.linsolve` when the nlsolve carries one and falls back to `alg.linsolve` otherwise.

## Why now

This is prep work for a non-breaking v7 minor release that swaps the default nlsolve from `NLNewton()` to `NonlinearSolveAlg(NewtonRaphson(...))` once the NonlinearSolve.jl-backed path meets the performance requirements. With `linsolve` uniformly configurable on the nlsolve object, that swap becomes invisible to user code — the same `alg(linsolve = X)` or `alg(nlsolve = …)` syntax works before and after.

## Changes

- `NLNewton` gains a `linsolve::L` field + kwarg (default `nothing`).
- New `OrdinaryDiffEqCore.effective_linsolve(alg)` helper: prefers `alg.nlsolve.linsolve` when set, falls back to `alg.linsolve`, returns `nothing` for algs (e.g. explicit RK) that have neither field. Rosenbrock/FIRK/Extrapolation (which have `linsolve` but no `nlsolve`) are handled correctly by the fallback.
- All internal consumers of `alg.linsolve` that previously relied on the algorithm-level location now route through `effective_linsolve`:
  - `OrdinaryDiffEqNonlinearSolve/src/utils.jl` — nlsolver LinearSolve cache setup
  - `OrdinaryDiffEqDifferentiation/src/derivative_utils.jl` — W-matrix shape decisions (4 sites)
  - `OrdinaryDiffEqDifferentiation/src/linsolve_utils.jl` — `dolinsolve` f-count logic
  - `OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl` — `build_jac_config` (simplified the `haslinsolve` branch — the fallback is baked into `effective_linsolve`)
- `BREAKING_CHANGES_v7.md` gets a new "Linear solver configuration: canonical location is on the nlsolve object" subsection under *Removed keyword arguments from algorithm constructors*.

## Not in this PR (follow-ups flagged in the doc)

Two related interface items that would further smooth the eventual default swap. Left out to keep this PR focused:

1. **Accept a bare NonlinearSolve.jl algorithm as `nlsolve`** — auto-wrapping into `NonlinearSolveAlg`. Would let `nlsolve = NewtonRaphson(...)` work today and mean the same thing after the default swap. Small change (one dispatch in `build_nlsolver`) but deserves its own tests.
2. **Align `NonlinearSolveAlg`'s inner default autodiff with the outer algorithm's `autodiff` kwarg.** Currently hard-coded to `AutoFiniteDiff()` while the outer solvers default to `AutoForwardDiff()` — when the nlsolve default swaps, this would silently change the AD backend unless we thread the outer autodiff through.

## Test plan

- [x] Added `@testset "NLNewton linsolve field"` in `lib/OrdinaryDiffEqNonlinearSolve/test/newton_tests.jl` covering:
  - `NLNewton(linsolve = LUFactorization())` produces the same solution as the alg-level path
  - both-set case: nlsolve-level linsolve wins over alg-level
- [x] `newton_tests.jl` passes locally (6/6 in the new testset; existing relax/BackTracking tests unchanged)
- [ ] Full sublibrary CI

### DIFF
--- a/BREAKING_CHANGES_v7.md
+++ b/BREAKING_CHANGES_v7.md
@@ -220,28 +220,41 @@ ImplicitEuler(nlsolve = NLNewton(linsolve = KLUFactorization(), relax = BackTrac
 ImplicitEuler(nlsolve = NonlinearSolveAlg(NewtonRaphson(linsolve = KLUFactorization())))
 ```
 
-Internally the nlsolve-level `linsolve` wins when set; when unset it falls back to the algorithm-level `linsolve`. **This is a non-breaking change** — existing code using path A continues to produce the same solves.
+Internally, on the `NLNewton` path, the nlsolve-level `linsolve` wins when set and falls back to the algorithm-level `linsolve` when unset. **This is a non-breaking change** — existing code using path A continues to produce the same solves on `NLNewton`.
+
+#### Known gap: alg-level `linsolve` is ignored on the `NonlinearSolveAlg` path
+
+Both on v6 and v7, if you explicitly pass `nlsolve = NonlinearSolveAlg(...)` with a wrapped NonlinearSolve.jl algorithm (e.g. `NewtonRaphson()`) that doesn't itself carry a user-specified `linsolve`, the algorithm-level `alg(linsolve = X)` kwarg is **silently ignored**. The inner `NewtonRaphson` uses its own built-in default. To configure the linear solver on that path, you must pass it into the wrapped algorithm:
+
+```julia
+# WRONG (on v6 and v7): the outer linsolve is dropped
+ImplicitEuler(linsolve = KLUFactorization(), nlsolve = NonlinearSolveAlg(NewtonRaphson()))
+
+# CORRECT: pass linsolve through NewtonRaphson
+ImplicitEuler(nlsolve = NonlinearSolveAlg(NewtonRaphson(linsolve = KLUFactorization())))
+```
+
+This is preserved-existing-behavior, not a v7 regression — it's just worth calling out because it matters for the future default swap, below.
 
 #### Why this exists: prep for a future default nlsolve swap
 
-The default `nlsolve` is planned to move from `NLNewton()` to `NonlinearSolveAlg(NewtonRaphson(...))` in a **non-breaking v7 minor release**, once the NonlinearSolve.jl-backed path meets the performance requirements (specifically, allocations and per-step overhead on the benchmarked SDIRK/BDF workloads). Making `linsolve` configurable on `NLNewton` too means:
+The default `nlsolve` is planned to move from `NLNewton()` to `NonlinearSolveAlg(NewtonRaphson(...))` in a **non-breaking v7 minor release**, once the NonlinearSolve.jl-backed path meets the performance requirements (specifically, allocations and per-step overhead on the benchmarked SDIRK/BDF workloads). For that swap to be truly non-breaking for user code, two pieces are needed:
 
-- `alg(linsolve = X)` — the form nearly all user code uses — threads through either implementation and is invisible to the swap.
-- Users who pin `nlsolve = NLNewton(...)` explicitly keep NLNewton after the swap; their `linsolve` stays on that NLNewton and is honored.
-- The swap itself only changes which nlsolve is constructed by default when the user doesn't specify one; no user syntax changes.
+1. `NLNewton` and the inner NonlinearSolve.jl algorithm need to accept `linsolve` in the same place — so users who customize `nlsolve = NLNewton(linsolve = X, ...)` today can write equivalent code after the swap. **This PR delivers piece 1.**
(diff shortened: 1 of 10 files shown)


ChrisRackauckas-Claude:
Important correction surfaced by review. The framing "this PR sets up a non-breaking default swap to `NonlinearSolveAlg`" was overstated.

**The gap:** On both v6 and v7, `alg(linsolve = X, nlsolve = NonlinearSolveAlg(NewtonRaphson()))` silently ignores `X`. At `lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl:313-318`, the LinearSolve cache built from `effective_linsolve(alg)` is stored only in `NLNewtonCache` (line 378); the `NonlinearSolveAlg` branch (line 348) calls `init(prob, nlalg.alg, ...)` and uses whatever `linsolve` is baked into `nlalg.alg`. Preserved-existing-behavior, not a regression — but it means that when the default nlsolve flips from `NLNewton()` to `NonlinearSolveAlg(NewtonRaphson(...))`, users who wrote `alg(linsolve = X)` and never customized `nlsolve` would silently regress.

**What this PR does (piece 1):** `NLNewton.linsolve` field + uniform `effective_linsolve` reads. This is the piece that lets users who customize `nlsolve = NLNewton(...)` today carry their linsolve with them after the default swap — *if* they explicitly opt into NLNewton.

**What this PR does NOT do (piece 2, required follow-up before the swap can ship):** thread `alg(linsolve = X)` into a default-constructed `NonlinearSolveAlg`. Candidate implementations:

1. **Algorithm-constructor threading.** In each alg's kwarg defaults, when `nlsolve` is omitted and `linsolve` is set, construct the default nlsolve with that linsolve: `nlsolve = isnothing(nlsolve) ? default_nlsolve_with_linsolve(linsolve) : nlsolve`. Semantically clean because user-explicit vs default is known at construction time. Cost: touches ~52 algorithm constructors.
2. **`default_nlsolve(alg)` hook.** Each family defines the hook; `build_nlsolver` calls it when the stored nlsolve is a sentinel. Still requires the constructors to use the sentinel.
3. **Sentinel `nlsolve = nothing`.** Store `nothing` when the user didn't pass it; `build_nlsolver` materializes the default and threads `alg.linsolve` in. Same constructor-touching cost.

I think (1) or (3) is right, both carry the same surgery cost, and (3) composes better with (2) since a sentinel-on-the-struct approach extends cleanly to the `default_nlsolve(alg)` hook that different families could override differently (e.g. a DAE family might want `BrownFullBasicInit`-compatible defaults).

The breaking-changes doc now explicitly documents the gap under *"Known gap: alg-level `linsolve` is ignored on the `NonlinearSolveAlg` path"* and splits the default-swap prep into piece 1 (this PR) and piece 2 (required before the swap ships). Happy to tackle piece 2 in this PR if reviewers want it bundled; otherwise it's a clean follow-up.

ChrisRackauckas:
This was a good idea by @termi-official but ultimately didn't go into the major. The reason is because, it turned out to be a bit more intricate than expected. First of all, there are lots of solvers with linear solvers but no nonlinear solver, things like Rosenbrock methods. So `Rodas5P(linsolve = ...)` would still have to exist, and so it's not very uniform to make that be done on those algorithms, but then `KenCarp47` doesn't have linsolve, so `KenCarp47(nonlinsolve = NLNewton(linsolve = ...))` for those algorithms. That is putting a large burden on the user. But then @oscardssmith pointed out a few things that, this also forces the user to choose a nonlinear solver, which we don't want to do. We want them to use the default, especially as we plan to swap out that default in the v7 timeframe to NonlinearSolveAlg from NLNewton, while this then is encouraging codes to hardcode NLNewton, not because they want NLNewton, but because this is how we tell them to set the nonlinear solver. We could then do a default_nonlinearsolver function which takes in information etc., but then what's the point? It's no longer simpler at that point, but also, it still doesn't actually unify the interface if other algorithms still don't have a nonlinsolve so they require linsolve.

Another idea is to force Rosenbrocks to do `nonlinsolve = RosenbrockNonlinearSolve(linsolve = ...))` where RosenbrockNonlinearSolve just hits the nonlinear solve code as a linear solver. BTW, the code effectively does this internally, just to make it so everything dispatches through OrdinaryDiffEqNonlinearSolve in a nice way... but it's really weird and I don't want to have to explain to users this oddity. It's more of a package structure thing, not something about the math itself, so it's very non-intuitive. But that would make it uniform.

So ultimately, I gave up on this idea for this release. The nice thing is that if we want to do a v8 which breaks and changes this, it's not the worst thing in the world because it would only effect a small section of codes which are all pretty much power users (since they are all users that set linear solvers, so most users will not need to do any code changes for this). And I'm not convinced this is better because of the non-uniformity issues when digging into it deeper.

But we still have a bug: NonlinearSolveAlg should respect the `linsolve` choice, so we should make an issue about that and solve it directly. That's a simple non-breaking thing to do, so we'll close and go there.

PR #3502, opened 2026-04-22
Migrate test code for SciMLBase v3 breaking changes (tuples/intervals/TimeChoiceIterator, ensemble ctx signatures, ProblemLibrary majors)

ChrisRackauckas-Claude:
## Summary

Sweeps `lib/*/test/` for the three test-code breakages called out in `NEWS.md` v7 section that were surfacing on the #3499 CI run. All tests now parse clean under Julia 1.11 (`Meta.parseall` over all 21 modified files); no source code changed.

### 1. `tuples(integrator)` / `intervals(integrator)` / `TimeChoiceIterator` — removed in SciMLBase v3

NEWS: *`tuples()`, `intervals()` iterator functions → iterate over `sol.u` / `sol.t` directly; `TimeChoiceIterator` removed*.

- `lib/DelayDiffEq/test/integrators/iterator.jl` — `for (u, t) in tuples(integrator)` → `for integ in integrator; @test … integ.t …`; `intervals(integrator)` iteration → read `integ.tprev` / `integ.uprev`; `TimeChoiceIterator(integrator4, ts)` → manual `step!` + interpolation loop (same pattern the top-level `test/integrators/iterator_tests.jl` already uses).
- `lib/StochasticDiffEq/test/tstops_tests.jl` — same `tuples` → `for integ in integrator` rewrite.

### 2. Ensemble function signatures: `(prob, i, repeat)` → `(prob, ctx)`, `(sol, i)` → `(sol, ctx)`

NEWS: *ctx::EnsembleContext; ctx.i, ctx.repeat, ctx.rng*.

Rewrote 15 files (4 in DiffEqBase/test/downstream/, 2 in DiffEqDevTools, 9 across StochasticDiffEq/StochasticDiffEqWeak weak-convergence + adaptive suites). For `function` / `function <name>(…)` / `<name> = function (…)` forms that reference `i` or `repeat` in the body, a `i = ctx.i; repeat = ctx.repeat` shim is inserted as the first body line so existing logic compiles unchanged. For arrow forms whose bodies don't reference `i` (the SDE weak-convergence tests just return `(h(sol.u[end]), false)`), the change is just the parameter rename.

### 3. `lib/DiffEqDevTools/Project.toml` — bump stale `[compat]`

- `ODEProblemLibrary = "0.1"` → `"1"`  (registered at v1.5.0)
- `SDEProblemLibrary = "0.1"` → `"1"`  (registered at v1.2.0)
[...]

PR #3503, opened 2026-04-22
Unpin / widen test/ad Project.toml compat to resolve Mooncake ↔ SciMLSensitivity conflict

ChrisRackauckas-Claude:
## Summary

The `test (AD, 1.11)` job on #3502 fails at Pkg.instantiate with:

```
ERROR: LoadError: Unsatisfiable requirements detected for package SciMLSensitivity [1ed8b502]:
 SciMLSensitivity log:
 ├─possible versions are: 7.0.0 - 7.105.0 or uninstalled
 ├─restricted to versions 7.104.0 - 7 by project, leaving only versions: 7.104.0 - 7.105.0
 └─restricted by compatibility requirements with Mooncake [da2b9cff] to versions:
   [7.0.0 - 7.71.2, 7.95.0 - 7.103.0] or uninstalled — no versions left
```

### Root cause

- SciMLSensitivity 7.104 and 7.105 both pin `Mooncake = "0.5.25"` in their Project.toml.
- In the `test/ad` env, the resolvable Mooncake is capped at **0.5.24** by DifferentiationInterface 0.7.16's `WeakCompat` entry (`Mooncake = "0.5.1 - 0.5.24"`).
- `test/ad/Project.toml` had `SciMLSensitivity = "7.104"`, which forced the resolver into the 7.104-7.105 range — but no Mooncake version in that range satisfies both SciMLSensitivity's `0.5.25+` floor and DI's `0.5.24` cap.

### Change in this PR

Lower the `SciMLSensitivity` floor in `test/ad/Project.toml` from `"7.104"` to `"7.89"`. SciMLSensitivity 7.89-7.103 have no direct `Mooncake` compat entry in the registry, so they would not re-apply that specific constraint. `Mooncake = "0.4, 0.5"` is left unchanged — already the widest range compatible with OrdinaryDiffEqCore v4's `Mooncake = "0.4, 0.5"` compat pin.

This is a test-env compat widening only; **no source changes**.

### Honest caveat — this PR alone is not sufficient

Even after lowering the floor, the resolver is separately forced back into the 7.104-7.105 window by other deps in the env (SciMLBase 3 via OrdinaryDiffEqTsit5, FunctionWrappersWrappers 1 via OrdinaryDiffEqCore v4, FastBroadcast 1.3+, SciMLOperators 1.15+). So this one-line pin change does not on its own make `Pkg.instantiate()` succeed — verified locally.

The genuine unblock requires at least one of:

1. **JuliaDiff/DifferentiationInterface.jl#988** — bump DI's Mooncake `WeakCompat` to include `0.5.25+`, so SciMLSensitivity 7.105 and DI 0.7.16 can co-resolve.
2. A matching SciMLSensitivity Mooncake-compat relaxation so `7.105+` stops pinning `Mooncake = "0.5.25"` strictly (registry compat patch).
3. A targeted LinearSolve `WeakCompat` relaxation (today caps via `"0.5.15 - 0.5"` — not the immediate binding constraint but in the chain).

Opening this PR as the `test/ad`-side half of the fix: once any of the three above lands, this pin is already loosened and the resolver will succeed.

## Test plan

- [x] Confirmed the original failure locally (same `Unsatisfiable` log as CI).
- [x] Confirmed that with this PR alone the resolver still fails (now blocked downstream of SciMLBase 3 / FunctionWrappersWrappers 1 forcing 7.104-7.105 regardless of the project floor). So the `test/ad/Project.toml` side is loosened correctly, but CI will only turn green once one of the three upstream fixes is in.
- [ ] Re-run `test (AD, 1.11)` after the upstream DI/SciMLSensitivity fix.


ChrisRackauckas-Claude:
Closing per request — SciMLSensitivity is out of scope for the v7 release unblock. The fix here was also incomplete (agent's own caveat: SciMLBase 3 / FastBroadcast 1.3 / SciMLOperators 1.15+ independently force 7.104-7.105, so the Mooncake conflict recurs regardless of the `test/ad/Project.toml` floor). Real fix lives upstream in JuliaDiff/DifferentiationInterface.jl#988 or a SciMLSensitivity registry compat patch.

PR #3504, opened 2026-04-22
Fix benchmark + ODEInterfaceRegression CI for v7-DEV state

ChrisRackauckas-Claude:
## Summary

Two fixes for CI failures that appear on every PR against master while the v7 sublibraries remain unregistered on the General registry (surfaced concretely on #3502's run):

- **`benchmark (1)` / `benchmark (lts)`** — `continue-on-error: true` on the benchmark job.
- **`test (ODEInterfaceRegression, lts)`** — subenv activator refactor so the LTS transitive-develop workaround covers it, plus a compat bump in `test/odeinterface/Project.toml`.

## 1. `benchmark` — `continue-on-error: true`

`benchmark.yml` uses `MilesCranmer/AirspeedVelocity.jl@action-v1`, whose `benchpkg` creates a fresh temporary environment and calls `Pkg.add(OrdinaryDiffEq@<rev>)`. That reads master's `Project.toml` and tries to resolve all `[deps]` from the General registry — but master's v7 sublibraries (e.g. `OrdinaryDiffEqRosenbrockTableaus`) aren't registered yet, so resolution fails with:

```
ERROR: Unsatisfiable requirements detected for package OrdinaryDiffEqRosenbrockTableaus [b4bd8bb3]:
 ├─OrdinaryDiffEqRosenbrockTableaus [b4bd8bb3] has no known versions!
 └─restricted to versions * by OrdinaryDiffEq [1dea7af3] — no versions left
   └─OrdinaryDiffEq [1dea7af3] is fixed to version 7.0.0-DEV
```

This is a structural side-effect of the v7-DEV window, not a benchmark regression. Marking the job `continue-on-error: true` lets it fail without blocking PR CI. Revert once the v7 sublibraries are registered.

## 2. `test (ODEInterfaceRegression, lts)` — subenv activator refactor

On Julia < 1.11 the `[sources]` section of `Project.toml` is ignored. `test/runtests.jl`'s sublibrary block (lines 62–90 on master) already works around this by walking the `[sources]` transitively and handing every local path dep to a single `Pkg.develop` call. But `activate_odeinterface_env()` — and its siblings `activate_downstream_env` / `activate_ad_env` / `activate_modelingtoolkit_env` — only called `Pkg.develop(OrdinaryDiffEq)`, which re-resolves against General and hits the same unregistered-sublibrary wall:

```
ERROR: Unsatisfiable requirements detected for package OrdinaryDiffEqLowOrderRK [1344f307]:
 ├─possible versions are: 1.0.0-1.13.0 or uninstalled
 ├─restricted to versions 1.5.0-1 by OrdinaryDiffEqAdamsBashforthMoulton [89bda076]
 └─restricted by compatibility requirements with DiffEqBase [2b5f629d] to versions: uninstalled
```

### Fix

- Extract the transitive-develop walk into two helpers: `_collect_local_source_specs` (returns `Pkg.PackageSpec[]`) and `_develop_local_sources_transitively` (batches the `Pkg.develop` call).
- Add `_activate_env_with_monorepo_sources(subdir)`, which — on LTS — collects specs from both the repo root and the subenv directory, ensures `OrdinaryDiffEq` itself is in the spec list (dedupes by `realpath`), and does one batched `Pkg.develop`. On 1.11+ it falls back to the original single-develop path since `[sources]` works natively.
- Route all four `activate_*_env` helpers through it.
- The sublibrary block now calls `_develop_local_sources_transitively` too, removing the duplicated walk.
- Bump `test/odeinterface/Project.toml` compat pins: `OrdinaryDiffEq = "6, 7"` → `"7"`, `OrdinaryDiffEqCore = "3"` → `"4"`. Without this the resolver hits `empty intersection between OrdinaryDiffEqCore@4.0.0 and project compatibility 3` right after the develop succeeds.

## Scope note

This PR is orthogonal to and non-overlapping with #3502 (which migrates test *source code* for SciMLBase v3 API breakage). #3502 already bumps the same two compat pins as a byproduct of its `[compat]` sweep, so merging this first will create a trivial conflict on those two lines only — easily resolved either way.

## Test plan

- [x] `Init dt vs dorpri tests` — 4/4 pass on Julia 1.10 (LTS).
- [x] `ODEInterface Regression Tests` — 13/13 pass on Julia 1.10 (LTS).
- [x] Same tests pass on Julia 1.11.
- [x] `Meta.parseall` on `test/runtests.jl` — clean.
- [ ] Full CI matrix on this PR.


ChrisRackauckas-Claude:
Closing per request — LTS is out of scope for the v7 unblock, and both fixes here are LTS-adjacent: the ODEInterfaceRegression `runtests.jl` refactor only matters on Julia 1.10 where `[sources]` isn't honored, and the benchmark `continue-on-error` is a workaround for unregistered v7 sublibraries that will self-resolve once the sublibs are registered. If the transitive-develop refactor is useful on its own merits, happy to open it as a separate cleanup later.

PR #3505, opened 2026-04-22
Fix DelayDiffEq v7 regressions: jacobian wrapper, SDDE dispatch, QA imports

ChrisRackauckas-Claude:
## Summary

Three DelayDiffEq test groups were failing on Julia 1.x (not LTS) on the v7 branch (PR #3502 / sample CI run https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/24776489602). Each was a distinct regression; this PR addresses all three.

### (1) `DelayDiffEq_Interface` — Jacobian Tests `MethodError`

`@wrap_h jac(J, u, h, p, t)` in `lib/DelayDiffEq/src/functionwrapper.jl` built the h-less lambda as `($(args_wo_h...),) -> _f(...)`. Commit 430b4b4aa (\"Apply Runic: drop stray trailing comma in 1-arg anonymous tuples\") removed the trailing comma, which looks like a 1-arg-tuple nit but was actually the token that told the parser \"this is a parameter list\". Without it, `($(args_wo_h...)) -> body` with N>1 splatted names (e.g. `J, u, p, t`) parses as `Expr(:->, :J, :u, :p, :t, body)` — a 1-arg lambda whose sole parameter destructures a tuple. `OrdinaryDiffEqDifferentiation.calc_rosenbrock_differentiation!` then calls the wrapped jac as `(J, u, p, t)` and hits `MethodError: no method matching #ODEFunctionWrapper##0#ODEFunctionWrapper##1(::Matrix, ::Vector, ::NullParameters, ::Float64)`.

Rebuild the parameter list via explicit `Expr(:tuple, args_wo_h...)` — semantically identical to the original trailing-comma form, and unambiguous to both the parser and Runic so a future format pass can't silently break it again.

### (2) `DelayDiffEq_SDDE` — `Incompatible problem+solver pairing.`

Commit b4a8cb29f (\"Remove remaining deprecations…\") scrubbed SDDE-specific branching out of `__solve`/`__init` together with the `initial_order` deprecation warning. Besides the warning, it removed:
- the `Union{AbstractDDEProblem, AbstractSDDEProblem}` type signature on `__solve` and `__init`,
- the `_sde_alg_cache` / `_create_sdde_noise` / `SDEFunctionWrapper` dispatches,
- the `check_prob_alg_pairing(SDDEProblem, MethodOfSteps{<:SDEAlgUnion})` override,
- the `get_fsalfirstlast` methods for the two SDE cache supertypes.

Without those, `solve(SDDEProblem(…), MethodOfSteps(EM()), dt=0.01)` now fails in `DiffEqBase.solve` at `check_prob_alg_pairing` — SDDEProblem + MethodOfSteps (which subtypes `AbstractDDEAlgorithm`) is not a recognized pairing. The entire `test/sdde/*` SafeTestset errors on the first call.

Restored all four pieces surgically (solve.jl widens the signature back, reintroduces `is_stochastic`, threads W/P/sqdt through the integrator type params; DelayDiffEq.jl adds the MethodOfSteps pairing override and the two `get_fsalfirstlast` methods). DDE path is bit-for-bit unchanged — every new branch gates on `is_stochastic` and the `false` arm matches the prior code.

### (3) `DelayDiffEq_QA` — stale imports, wrong-owner qualified access

`ExplicitImports` flagged:
- stale `None`, `StochasticDiffEqConstantCache`, `StochasticDiffEqMutableCache` imports in `DelayDiffEq.jl`,
- `OrdinaryDiffEqCore.DEVerbosity()` at `solve.jl:197` being a qualified access via a non-owner (DiffEqBase owns `DEVerbosity`).

Fix (1) by dropping the unused `None` import (only appeared in an error-message string) and re-pointing `DEVerbosity` at `DiffEqBase`. The two SDE cache type names are now *used* by the `get_fsalfirstlast` methods introduced for (2), so they stop being stale and stay imported. While I was in there I made the new SciMLBase accesses explicit too (`ProblemSolverPairingError`, `DirectAutodiffError`, `eltypedual`, `isautodifferentiable`) and hoisted `import SciMLBase` above `@reexport using SciMLBase` so the module-name binding is explicit to the `check_no_implicit_imports` static analysis.

## Validation (local, Julia 1.11)

- `ODEDIFFEQ_TEST_GROUP=Interface` — all groups pass. Jacobian Tests: 21 passed / 12 broken / 33 total (matches pre-regression expected counts — the \"broken\" are `@test_broken` markers already in the test file for nWfact_t vs njacs disagreement on rejected steps).
- `ODEDIFFEQ_TEST_GROUP=SDDE` — Problem/Solution 16/16, Event 2/2, Non-diagonal Sparse Noise 2/2, Implicit Methods 10 pass / 2 fail / 12. The two failures (`ImplicitEulerHeun` and `ISSEulerHeun` with `dt=0.01` on the Hayes model) are pre-existing adaptive-step `dt_min_unstable` aborts — they were never *reached* before because Problem/Solution errored out first.
- `ODEDIFFEQ_TEST_GROUP=QA` — 14/14 (Aqua 11/11, Explicit Imports 3/3).

## Test plan

- [x] DelayDiffEq Interface group on 1.11 — all passing groups green locally
- [x] DelayDiffEq SDDE group on 1.11 — dispatch fixed; remaining 2 failures are pre-existing numerical instability unrelated to this PR
- [x] DelayDiffEq QA group on 1.11 — 14/14 green locally
- [ ] CI runs all three groups on `lts`, `1.11`, `1`, `pre`

### DIFF
--- a/lib/DelayDiffEq/src/DelayDiffEq.jl
+++ b/lib/DelayDiffEq/src/DelayDiffEq.jl
@@ -145,10 +145,6 @@ function DiffEqBase.check_prob_alg_pairing(
     if !(alg.alg isa SDEAlgUnion)
         throw(ProblemSolverPairingError(prob, alg))
     end
-    if isdefined(prob, :u0) && eltypedual(prob.u0) &&
-            !isautodifferentiable(alg)
-        throw(DirectAutodiffError())
-    end
     return nothing
 end
 
--- a/lib/DelayDiffEq/src/DelayDiffEq.jl
+++ b/lib/DelayDiffEq/src/DelayDiffEq.jl
@@ -142,8 +142,4 @@ function DiffEqBase.check_prob_alg_pairing(
     return nothing
 end
 
-# SDE caches don't have fsalfirst/fsallast (SDE algorithms are never FSAL)
-OrdinaryDiffEqCore.get_fsalfirstlast(cache::StochasticDiffEqConstantCache, u) = (zero(u), zero(u))
-OrdinaryDiffEqCore.get_fsalfirstlast(cache::StochasticDiffEqMutableCache, u) = (zero(u), zero(u))
-
 end # module
--- a/lib/DelayDiffEq/src/DelayDiffEq.jl
+++ b/lib/DelayDiffEq/src/DelayDiffEq.jl
@@ -1,6 +1,7 @@
 module DelayDiffEq
 
 import Reexport: @reexport, Reexport
+import SciMLBase
 @reexport using SciMLBase
 import OrdinaryDiffEqCore, OrdinaryDiffEqNonlinearSolve, OrdinaryDiffEqDifferentiation,
     OrdinaryDiffEqRosenbrock
@@ -22,13 +23,15 @@ import SymbolicIndexingInterface as SII
 using SciMLBase: AbstractDDEAlgorithm, AbstractDDEIntegrator, AbstractODEIntegrator,
     DEIntegrator
 using SciMLBase: AbstractSDDEProblem, SDDEProblem
+using SciMLBase: ProblemSolverPairingError, DirectAutodiffError, eltypedual,
+    isautodifferentiable
 
 using Base: deleteat!
 import FastBroadcast: @..
 
 using OrdinaryDiffEqNonlinearSolve: NLAnderson, NLFunctional
 using OrdinaryDiffEqCore: AbstractNLSolverCache, SlowConvergence,
-    alg_extrapolates, alg_maximum_order, initialize!, DEVerbosity
+    alg_extrapolates, alg_maximum_order, initialize!
 using OrdinaryDiffEqCore: StochasticDiffEqAlgorithm,
     StochasticDiffEqRODEAlgorithm,
     StochasticDiffEqConstantCache, StochasticDiffEqMutableCache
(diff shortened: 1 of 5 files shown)


lib/DelayDiffEq/src/DelayDiffEq.jl
```diff
@@ -130,5 +131,29 @@ end
...
+        alg::AbstractMethodOfStepsAlgorithm
+    )
+    # MethodOfSteps wrapping an SDE algorithm is valid for SDDEProblem
+    if !(alg.alg isa SDEAlgUnion)
+        throw(ProblemSolverPairingError(prob, alg))
+    end
+    if isdefined(prob, :u0) && eltypedual(prob.u0) &&
+            !isautodifferentiable(alg)
+        throw(DirectAutodiffError())
+    end
```
ChrisRackauckas:
```suggestion
```

lib/DelayDiffEq/src/DelayDiffEq.jl
```diff
@@ -130,5 +131,25 @@ end
 function DiffEqBase.check_prob_alg_pairing(prob::SDDEProblem, alg::SDEAlgUnion)
     return nothing
 end
+# Also bypass the default DDE-only pairing check when the user explicitly
+# constructs `MethodOfSteps(sde_alg)` for an SDDEProblem. Without this override
+# DiffEqBase.check_prob_alg_pairing sees `prob::SDDEProblem` with an
+# `AbstractDDEAlgorithm` (which MethodOfSteps subtypes) and throws
+# ProblemSolverPairingError — the auto-wrap path above covers bare SDE algs
+# but not the already-wrapped form produced by `solve(prob, MethodOfSteps(EM()))`.
```
ChrisRackauckas:
```suggestion

```

Issue #1322, opened 2021-01-06
DynamicalODE with immutable state variables and callbacks failing


KlausC:
Running this example with immutatble state and callback fails.
Obviously `addsteps!` tries to write into the `SVector` state variables.

```
julia> using StaticArrays
julia> using OrdinaryDiffEq, LinearAlgebra, ForwardDiff, Plots; gr()
julia> initial_position = SVector{2}([.4, 0])
julia> initial_velocity = SVector{2}([0., 2.])
julia> qdot(p, q, params, t) = p
julia> pdot(p, q, params, t) = -q
julia> tspan = (0.0, 10.0)
julia> prob = DynamicalODEProblem(pdot, qdot, initial_velocity, initial_position, tspan)
julia> sol = solve(prob, KahanLi6(), dt=1//10);

julia> condi(u, t, integrator) = dot(u.x[1], u.x[2]
julia> at_condi(integrator) = println("condition at $(integrator.t) state $(integrator.u)")
julia> cb = ContinuousCallback(condi, at_condi)
julia> sol = solve(prob, KahanLi6(), dt=1//10; callback=cb);

ERROR: setindex!(::SArray{Tuple{2},Float64,1,2}, value, ::Int) is not defined.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] setindex!(::SArray{Tuple{2},Float64,1,2}, ::Float64, ::Int64) at .julia/packages/StaticArrays/NTbHj/src/indexing.jl:3
 [3] copyto_unaliased!(::IndexLinear, ::SArray{Tuple{2},Float64,1,2}, ::IndexLinear, ::SArray{Tuple{2},Float64,1,2}) at ./abstractarray.jl:860
 [4] copyto!(::SArray{Tuple{2},Float64,1,2}, ::SArray{Tuple{2},Float64,1,2}) at ./abstractarray.jl:840
 [5] recursivecopy!(::SArray{Tuple{2},Float64,1,2}, ::SArray{Tuple{2},Float64,1,2}) at .julia/packages/RecursiveArrayTools/mTQGQ/src/utils.jl:25
 [6] recursivecopy! at .julia/packages/RecursiveArrayTools/mTQGQ/src/array_partition.jl:210 [inlined]
 [7] copyat_or_push! at .julia/packages/RecursiveArrayTools/mTQGQ/src/utils.jl:68 [inlined]
[...]

PR #3513, opened 2026-04-23
Fix "cannot evaluate norm recursively" in LowStorageRK / SSPRK under RAT v4

ChrisRackauckas-Claude:
## Summary

- Fixes the two failing test jobs on PR #3508 (post-merge of #3502/#3505/#3507):
  - `test (OrdinaryDiffEqLowStorageRK, 1, ubuntu-latest, …)` (job `72604636568`)
  - `test (OrdinaryDiffEqSSPRK, 1, ubuntu-latest, …)` (job `72604636663`)
- Both testsets failed at the `VectorOfArray/StructArray compatibility` block with:
  > `ArgumentError: cannot evaluate norm recursively if the type of the initial element is identical to that of the container`

## Root cause

The failing assertion in both suites is `@test sol_SA ≈ sol_SV`, where `sol_SA` and `sol_SV` are `ODESolution`s whose internal `u` storage is a `Vector{VectorOfArray{Float64, 2, StructVector{SVector{1,Float64},…}}}` vs `Vector{VectorOfArray{Float64, 2, Vector{SVector{1,Float64}}}}`. `ODESolution <: AbstractVectorOfArray <: AbstractArray`, so the generic `LinearAlgebra.isapprox(::AbstractArray, ::AbstractArray)` fallback calls `norm(sol)`. On Julia 1.12 + RecursiveArrayTools v4, `norm` now routes through a stricter `norm_recursive_check`, which throws for this container shape because `AbstractArray` iteration yields the solution itself (`iterate(sol)[1] === sol`) when the leaf storage is a `StructArray{<:SVector}`-backed `VectorOfArray`. No OrdinaryDiffEq solver code appears on the failing stack — `solve` completes fine for both storages; the failure is purely in `≈` at the test site.

## Fix

- Add `Base.isapprox(x::AbstractTimeseriesSolution, y::AbstractTimeseriesSolution)` in `lib/OrdinaryDiffEqCore/src/misc_utils.jl`. It:
  - checks trajectory lengths and compares `x.t ≈ y.t`,
  - iterates `x.u` / `y.u` snapshot-by-snapshot and delegates to `isapprox` on leaves,
  - materialises `VectorOfArray` / `StructArray` snapshots (via `parent(...)` + `collect`) into plain `Array`s before the inner `isapprox`, so `LinearAlgebra.norm` never sees a container whose element type equals itself.
- Import `LinearAlgebra.norm` (and the `LinearAlgebra` module) in `OrdinaryDiffEqCore`.
- Bump `OrdinaryDiffEqCore` to `4.0.1`.

The scalar and `Vector{Float64}` comparison paths still dispatch to `isapprox` on the raw snapshot, so the existing `@test sol_old.u[end] ≈ sol_new.u[end]` assertions elsewhere in the LowStorageRK / SSPRK tests are untouched.

**Tests were not removed, disabled, marked `@test_broken`, or commented out.**

## Test plan

Locally on Julia `1.12.6` (matching CI), monorepo `Pkg.develop`ed against this branch:

- [x] `OrdinaryDiffEqLowStorageRK` `VectorOfArray/StructArray compatibility` testset: **16/16 pass** (was 15 pass / 1 error on master).
- [x] `OrdinaryDiffEqSSPRK` `VectorOfArray/StructArray compatibility` testset: **2/2 pass** (was 1 pass / 1 error on master).
- [x] Pre-fix reproducer (same test code, on master tip `72052c030`) confirms the `ArgumentError: cannot evaluate norm recursively …` with stack trace identical to the CI failure.
- [ ] Full CI on this PR (LowStorageRK + SSPRK + rest).

### DIFF
--- a/lib/OrdinaryDiffEqCore/Project.toml
+++ b/lib/OrdinaryDiffEqCore/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqCore"
 uuid = "bbf590c4-e513-4bbe-9b18-05decba2e5d8"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "4.0.0"
+version = "4.0.1"
 
 [deps]
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
--- a/lib/OrdinaryDiffEqCore/src/OrdinaryDiffEqCore.jl
+++ b/lib/OrdinaryDiffEqCore/src/OrdinaryDiffEqCore.jl
@@ -15,7 +15,7 @@ import Logging: @logmsg, LogLevel
 
 using MuladdMacro: @muladd
 
-using LinearAlgebra: opnorm, I, UniformScaling, diag, rank, isdiag
+using LinearAlgebra: LinearAlgebra, norm, opnorm, I, UniformScaling, diag, rank, isdiag
 
 import PrecompileTools
 
--- a/lib/OrdinaryDiffEqCore/src/misc_utils.jl
+++ b/lib/OrdinaryDiffEqCore/src/misc_utils.jl
@@ -46,6 +46,67 @@ end
 
 error_constant(integrator, order) = error_constant(integrator, integrator.alg, order)
 
+# `isapprox` for timeseries solutions.
+#
+# The generic `Base.isapprox(::AbstractArray, ::AbstractArray)` computes
+# `norm(x - y) <= max(atol, rtol*max(norm(x), norm(y)))`. When an
+# `AbstractTimeseriesSolution` (e.g. `ODESolution`) is iterated via the
+# `AbstractArray` fallback, iteration can yield an object whose type equals the
+# container type (because the solution subtypes `AbstractVectorOfArray{T,N}` and
+# its leaf container is a `VectorOfArray` whose `eltype` resolves to a scalar
+# while the iterator itself yields an array-like element). Under
+# `RecursiveArrayTools` v4 / Julia 1.12, `LinearAlgebra.norm(itr)` first calls
+# `norm_recursive_check` which throws
+#   ArgumentError: cannot evaluate norm recursively if the type of the initial
+#   element is identical to that of the container
+# for these nested / `StructArray`-backed storages. Compare snapshot-by-snapshot
+# instead, which also sidesteps the cross-container-type issue when comparing
+# solutions whose element storage differs (e.g. `Vector{SVector}` vs
+# `StructArray{SVector}`).
+function Base.isapprox(
+        x::SciMLBase.AbstractTimeseriesSolution,
+        y::SciMLBase.AbstractTimeseriesSolution;
+        atol::Real = 0,
+        rtol::Real = Base.rtoldefault(
+            _cmp_eltype(x), _cmp_eltype(y), atol
+        ),
+        nans::Bool = false,
+        norm = LinearAlgebra.norm
+    )
+    length(x.u) == length(y.u) || return false
+    isapprox(x.t, y.t; atol = atol, rtol = rtol, nans = nans) || return false
+    for i in eachindex(x.u)
+        _isapprox_snapshot(x.u[i], y.u[i]; atol, rtol, nans, norm) || return false
+    end
+    return true
+end
+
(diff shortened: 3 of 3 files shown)


ChrisRackauckas-Claude:
Diagnosis looks right, but the fix lands `Base.isapprox(::SciMLBase.AbstractTimeseriesSolution, ::SciMLBase.AbstractTimeseriesSolution)` in OrdinaryDiffEqCore — that's type piracy (neither function nor type is owned here), same class of issue as the `get_fsalfirstlast` piracy fixed in #3505. The natural home for this dispatch is SciMLBase, right next to `AbstractTimeseriesSolution` where the existing array-interface methods live.

Suggested path:
1. Move the `Base.isapprox(::AbstractTimeseriesSolution, ...)` method (plus `_isapprox_snapshot`, `_snapshot_array`, `_isapprox_materialised`, `_cmp_eltype`) into SciMLBase — probably under `src/solutions/ode_solutions.jl` or a new `src/solutions/comparisons.jl`. Use RAT's exported `AbstractVectorOfArray` / `recursive_bottom_eltype` (RAT is already a direct SciMLBase dep). Bump SciMLBase patch.
2. Once that's registered, this PR can shrink to just the `OrdinaryDiffEqCore` version bump (or be closed entirely — no source change needed here).

Alternative: if the real bug is in RAT's `norm_recursive_check` being over-eager for `StructArray{<:SVector}`-backed `VectorOfArray`s, the cleanest fix is upstream in RecursiveArrayTools (teach `norm_recursive_check` to see through a `StructArray` parent) — that would unblock every downstream that trips over the same recursion check, not just OrdinaryDiffEq's tests. Worth confirming which layer owns the bug before picking the landing point.

ChrisRackauckas-Claude:
Reproduced and tracked down: the root cause is **in SciMLBase, not RecursiveArrayTools**. `VectorOfArray`'s own `norm` works fine.

The broken iterator is `SciMLBase/src/solutions/solution_interface.jl:188-192`:

```julia
function Base.iterate(sol::AbstractTimeseriesSolution, state = 0)
    state >= length(sol) && return nothing
    state += 1
    return (solution_new_tslocation(sol, state), state)
end
```

For a solution typed `<: AbstractArray{Float64, 3}` with `eltype = Float64`, this yields another `ODESolution` of the **same concrete type** (same fields, different `tslocation`). Under Julia 1.12, `LinearAlgebra.norm_recursive_check` iterates the container, checks `v == itr`, and throws because the first element has the container's own type — and `AbstractVectorOfArray`'s `==` compares entries, so a "copy with tslocation rewound to 1" compares equal to the full container. That's the exact preconditions the check is guarding against.

This iterate has been like this for years (for plotting/`tslocation`-stepping semantics), so I don't want to change its behavior. The upstream fix is to add `LinearAlgebra.norm(::AbstractTimeseriesSolution)` in SciMLBase that operates on `sol.u` directly — bypassing both the Base `iterate`-based fallback and the recursive-check trip. One method, one file, no test piracy, no downstream churn.

I'll close this PR once the SciMLBase patch is in — or I can repurpose it to just be the version bump of `OrdinaryDiffEqCore` once SciMLBase ships the fix. WDYT?

ChrisRackauckas-Claude:
SciMLBase upstream fix: https://github.com/SciML/SciMLBase.jl/pull/1322 — replaces the custom `iterate(::AbstractTimeseriesSolution)` (which returned `solution_new_tslocation(sol, state)`) with a linear-indexing form that yields scalars per the AbstractArray contract. That lets `LinearAlgebra.norm(sol)` + `isapprox(sol_SA, sol_SV)` run on Julia 1.12 without hitting `norm_recursive_check`.

Once #1322 merges and registers, this PR should close: the `Base.isapprox` piracy in OrdinaryDiffEqCore becomes unnecessary.

ChrisRackauckas-Claude:
Superseded by SciML/SciMLBase.jl#1322 (registering as SciMLBase v3.4.0). The `Base.isapprox` piracy in OrdinaryDiffEqCore is no longer needed — with 3.4.0, `iterate(::AbstractTimeseriesSolution)` defers to the AbstractArray fallback, so `LinearAlgebra.norm(sol)` / `sol ≈ sol` stop tripping `norm_recursive_check` on Julia 1.12. Closing.

PR #3515, opened 2026-04-23
Workaround: adaptive=false for Stratonovich implicit methods on Hayes SDDE

ChrisRackauckas-Claude:
Split off from #3514 so the clean post-v7 fixes could merge without waiting on this investigation.

## Root cause investigation summary

The original diagnosis ("Stratonovich error estimator uses `dW^2` without `-dt` mean subtraction, causing shrinking-`dt` runaway") is **falsified**. Traced step-by-step on Julia 1.12.6 with the monorepo:

**(1) Non-DDE SDE baseline** — Hayes coefficients with history replaced by constants: all four methods complete under `adaptive=true`.

| Method | Steps | Status |
| --- | --- | --- |
| `ImplicitEM` (Itô) | 6 | Success |
| `ISSEM` (Itô) | 34 | Success |
| `ImplicitEulerHeun` (Stratonovich) | 170 | Success |
| `ISSEulerHeun` (Stratonovich) | 139 | Success |

No `DtLessThanMin`. Stratonovich takes ~20–30× more steps than Itô, but the adaptive controller copes.

**(2) SDDE case** — same coefficients wrapped in `MethodOfSteps`:

| Method | Steps | Status |
| --- | --- | --- |
| `ImplicitEM` (Itô) | 21 | Success |
| `ImplicitEulerHeun` (Stratonovich) | 575 | Success |
| `ISSEM` (Itô, split-step) | **572** | **Success** (17× inflation!) |
| `ISSEulerHeun` (Stratonovich) | **—** | **DtLessThanMin at t=0.0208** |

**(3) Falsifier**: `ISSEM` is Itô-flavored and uses the mean-zero `dW² - dt` estimator (`implicit_split_step.jl:192, 199`) — a Stratonovich-specific bias can't explain why its SDDE step count blows up 17× over the SDE baseline. The dominant factor is a **DDE/history × implicit-split interaction**, not the estimator.

**(4) Test of the estimator-fix theory** — subtract `-dt` in the Stratonovich `En` path only (sdirk.jl L269,276 + implicit_split_step.jl L99,105,205,212), leaving `mil_correction` unchanged. On the SDDE:
- `ImplicitEulerHeun` regressed (**DtLessThanMin at t=0.009**, 5 steps)
- `ISSEulerHeun` improved (345 steps, Success)
- `ISSEM` unchanged

Not a clean win — it reshuffles which method fails.

## Proper fix (deferred)

The real root cause is in `MethodOfSteps` wrapping of implicit-split SDE integrators, likely around history-interpolation + initial-dt selection. The `DelayDiffEq/ext/DelayDiffEqStochasticDiffEqCoreExt` bridge needs to be traced: the first step splits at the lag and forces `dt ~ 1e-3`, then the controller can't grow `dt` back because each attempt hits `dW²` variance at small `u`-scale.

That's a separate investigation touching DelayDiffEq + StochasticDiffEq, out of scope for the v7 release unblock.

## This PR

Placeholder: switches both calls to `adaptive = false, dt = 0.01`. The test still guards the real regression it was introduced for (#3363, which is about SDDE dispatch, not adaptive stability), matching the `ImplicitRKMil` call already on L42 and the `StochasticDiffEq/test/implicit_time_parameter_test.jl` reference pattern (L54, L111: both `adaptive = false`). Assertions `SciMLBase.successful_retcode(sol)` and `sol.u[end] != zeros(1)` pass.

Comments in the test are updated to reflect the investigation findings (DDE/history × implicit-split interaction, not Stratonovich estimator bias) so the next person reading doesn't re-derive the wrong explanation.

Willing to revise if someone has energy to chase the MethodOfSteps/StochasticDiffEqCoreExt angle — but that's not blocking the v7 release.

### DIFF
--- a/lib/DelayDiffEq/test/sdde/implicit_methods.jl
+++ b/lib/DelayDiffEq/test/sdde/implicit_methods.jl
@@ -33,11 +33,16 @@ sol = solve(prob, MethodOfSteps(ImplicitEM()))
 @test SciMLBase.successful_retcode(sol)
 @test sol.u[end] != zeros(1)
 
-# ImplicitEulerHeun — Stratonovich error estimator uses dW^2 (without -dt
-# subtraction), so the controller drives dt to dtmin on this stiff Hayes
-# model under adaptive stepping. Use fixed dt (same as ImplicitRKMil below
-# and matching the StochasticDiffEq standalone test pattern in
-# `StochasticDiffEq/test/implicit_time_parameter_test.jl`).
+# ImplicitEulerHeun — adaptive stepping drives dt → dtmin under this
+# `MethodOfSteps` wrapping. Note this is NOT the Stratonovich error
+# estimator — all four methods (incl. Itô-flavored `ImplicitEM`/`ISSEM`)
+# step fine adaptively on the *non-delay* SDE with the same coefficients.
+# The failure is SDDE-specific: `ISSEM` (Itô, mean-zero `dW²-dt` estimator)
+# inflates from 34 steps on the SDE to 572 on the SDDE — a
+# `MethodOfSteps` history × implicit-split interaction, not the estimator.
+# Use fixed dt (same as `ImplicitRKMil` below and matching
+# `StochasticDiffEq/test/implicit_time_parameter_test.jl`). Revisit once
+# the DelayDiffEq × StochasticDiffEq history/initial-dt path is fixed.
 sol = solve(prob, MethodOfSteps(ImplicitEulerHeun()), dt = 0.01, adaptive = false)
 @test SciMLBase.successful_retcode(sol)
 @test sol.u[end] != zeros(1)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Investigation complete — the Stratonovich-estimator-bias theory is falsified (see PR body for the step-count tables and the ISSEM counter-example). The real root cause is a `MethodOfSteps` × implicit-split interaction. Updated test comments to reflect the correct diagnosis. Safe to merge as a workaround; the proper fix belongs in the `DelayDiffEq × StochasticDiffEq` history/initial-dt path and can land separately.

ChrisRackauckas-Claude:
Superseded by #3519 — the real fix. Bisection nailed #3485 as the regressing commit: it moved the six SDE-implicit solvers from `PIController` default to `PredictiveController`. `PredictiveController.step_reject_controller!` does `dt/qold` (aggressive monotonic shrink) which collapses under noisy stochastic error estimates, especially via `MethodOfSteps` history interpolation. `PIController` uses the gentler `dt * max(qmin, min(1, qacc))`. Restoring `PIController` on all six gets pre-regression step counts back exactly (21/174/2638/2339 on the Hayes SDDE). Full DelayDiffEq_SDDE group now green on Julia 1.12.6 with `adaptive = true`. Closing.

Issue #45, opened 2017-05-05
Complex comparison in ode_determine_initdt


jebej:
It seems some changes were made which now has `ode_determine_initdt` compare complex numbers on line 10:
```julia
...
if d₀ < T0(1//10^(5)) || d₁ < T1(1//10^(5))
...
```

I can't figure out what T0 and T1 are however.

jebej:
Note: that's in my case, with complex initial variables (I assume that's the source of the issue), but the time interval itself is given with real numbers

ChrisRackauckas:
Duplicate of https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/162 . This was due to a change in DiffEqBase.jl to make the norms compatible with heterogeneous arrays of units, but this added a bug with complex numbers. The fix was easy, but I didn't realize this was used so much so I didn't put a release out! I'll make a hotfix tag right now.

ChrisRackauckas:
You can check out DiffEqBase.jl's master if you need the fix right now, and the hotfix tag PR is:

https://github.com/JuliaLang/METADATA.jl/pull/9158

It should go through quickly given it's a one commit tag.

jebej:
Thanks Chris.

PR #3518, opened 2026-04-23
Supply analytic tgrad=0 for Rosenbrock23 AutoEnzyme test on prob_ode_2Dlinear

ChrisRackauckas-Claude:
## Summary

The `OrdinaryDiffEqRosenbrock` test group fails with `DtLessThanMin` (actually `ReturnCode.Unstable`, dt forced below floating-point epsilon) on `Rosenbrock23{AutoEnzyme{ForwardMode{...}}}` against `prob_ode_2Dlinear`. Not a v7 regression — this has been latent-broken since Feb 2025.

## Root cause (not controller / not Rosenbrock)

Enzyme 0.13.x forward-mode has an IR-analysis incompatibility with `prob_ode_2Dlinear`'s RHS `@. du = p*u`. The `Broadcast.preprocess_args` / `extrude` path inside the in-place broadcast flags `u`/`du` as needing runtime activity, but `set_runtime_activity(Forward)` doesn't propagate into the Broadcast internals.

Behavior splits by Julia/LLVM version:
- **Locally** (Julia 1.12.6 + Enzyme 0.13.138): hard `FirstAutodiffTgradError`
- **On CI** (different LLVM compilation outcome): silent garbage `dT` values

Rosenbrock23 is uniquely affected because it uses `dT` **additively** in its embedded error estimator (`rosenbrock_perform_step.jl:105-111`):
```julia
linsolve_tmp = fsallast - c₃₂ * (k₂ - f₁) - 2 * (k₁ - fsalfirst) + dt * dT
```
Garbage `dT` → blown-up `EEst` → PI controller rejects every step → `dt` collapses to eps.

The Rodas variants (Rodas4/5P/23W) fold `dT` into a linsolve RHS where its effect is attenuated by the implicit-W solve, so they tolerate garbage `dT` and stumble through. They have the same Enzyme bug, just silently.

## Fix

`prob_ode_2Dlinear` is time-independent, so `tgrad ≡ 0`. Supply it analytically via `ODEFunction(f; tgrad = (dT,u,p,t) -> fill!(dT, 0), analytic = prob.f.analytic)`. This:
- Preserves the test's intent (Rosenbrock23 + AutoEnzyme-forward Jacobian autodiff on a 2D linear problem)
- Preserves `prob.f.analytic` so `test_convergence` still has an exact reference
- Routes around the Enzyme-internal Broadcast bug (which belongs in an upstream Enzyme.jl issue)
- Does NOT disable any test — 3 assertions remain (convergence order, step count, retcode)

## Verified locally

Julia 1.12.6, Enzyme 0.13.138, DI 0.7.16:
- `sim.𝒪est[:final] ≈ 1.996` (expected 2 ± 0.25)
- `Success` retcode
- 11 accepted steps

## Follow-up (not blocking)

File an Enzyme.jl issue: `set_runtime_activity(ForwardMode)` doesn't propagate through `Broadcast.preprocess_args` / `extrude` for in-place broadcast assignment. Every library differentiating through in-place broadcasts with Enzyme forward-mode is at risk.

## Test plan

- [x] `test_convergence` + `solve` on `prob_enzyme` pass locally with expected order and retcode
- [ ] CI `test (OrdinaryDiffEqRosenbrock, 1/1.11/pre, ...)` pass

### DIFF
--- a/lib/OrdinaryDiffEqRosenbrock/test/ode_rosenbrock_tests.jl
+++ b/lib/OrdinaryDiffEqRosenbrock/test/ode_rosenbrock_tests.jl
@@ -35,8 +35,28 @@ end
     @test SciMLBase.successful_retcode(sol)
 
     if isempty(VERSION.prerelease)
+        # `prob_ode_2Dlinear`'s RHS `@. du = p*u` is time-independent, so
+        # `tgrad ≡ 0`. Provide it analytically — Enzyme 0.13.x forward-mode
+        # cannot differentiate the in-place broadcast through its
+        # `Broadcast.preprocess_args` path, even with
+        # `set_runtime_activity(Forward)` (tracked upstream in Enzyme.jl).
+        # Rosenbrock23 uses `dT` additively in its embedded error
+        # estimator, so a garbage tgrad blows up `EEst` and forces `dt`
+        # below floating-point epsilon. Other Rosenbrock variants
+        # (Rodas4/5P/23W) fold `dT` into a linsolve where its effect is
+        # attenuated and happen to stumble through. Supplying the exact
+        # zero tgrad keeps the AutoEnzyme-forward Jacobian path under test
+        # while routing around the Enzyme-internal broadcast bug.
+        prob_enzyme = ODEProblem(
+            ODEFunction(
+                prob.f.f;
+                analytic = prob.f.analytic,
+                tgrad = (dT, u, p, t) -> fill!(dT, 0),
+            ),
+            prob.u0, prob.tspan, prob.p
+        )
         sim = test_convergence(
-            dts, prob, Rosenbrock23(
+            dts, prob_enzyme, Rosenbrock23(
                 autodiff = AutoEnzyme(
                     mode = set_runtime_activity(Enzyme.Forward), function_annotation = Enzyme.Const
                 )
@@ -45,7 +65,7 @@ end
         @test sim.𝒪est[:final] ≈ 2 atol = testTol
 
         sol = solve(
-            prob, Rosenbrock23(
+            prob_enzyme, Rosenbrock23(
                 autodiff = AutoEnzyme(
                     mode = set_runtime_activity(Enzyme.Forward), function_annotation = Enzyme.Const
                 )


ChrisRackauckas-Claude:
Sharper diagnosis: the root cause is in `FunctionWrappersWrappers.jl`, not in OrdinaryDiffEq. Specifically, FWW's existing `EnzymeRules.forward` extension (v1.7.0) hard-coded plain `Forward` when delegating to `Enzyme.autodiff`, silently dropping the outer caller's `RuntimeActivity` and `StrongZero` flags from `FwdConfig`. So `AutoEnzyme(mode = set_runtime_activity(Forward))` would appear to set the flag on the outer call, but the flag was lost at the FWW-rule boundary when Enzyme re-entered the unwrapped function.

Upstream fix: https://github.com/SciML/FunctionWrappersWrappers.jl/pull/45 — threads `RuntimeActivity` / `StrongZero` from `FwdConfig` through `_fwd_mode` into the delegated `Enzyme.autodiff` call.

Also: the "garbage dT" framing in this PR's body is slightly wrong. Enzyme hard-throws `EnzymeRuntimeActivityError` inside `Broadcast.preprocess_args`; the error is caught and rethrown as `FirstAutodiffTgradError` from `calc_tderivative!`, and downstream code ends up reading the pre-existing uninitialized `dT` buffer (which looks like garbage to Rosenbrock23's error estimator). No values are silently mis-written.

Options now:

1. **Close this PR** once FWW#45 merges + registers. The analytic-tgrad workaround is only needed for the 2Dlinear test case; other Enzyme users (on any time-dependent in-place RHS) would still hit the same bug without the upstream fix.
2. **Keep it as belt-and-suspenders**: providing an analytic `tgrad = 0` for a time-independent RHS is mathematically correct and avoids the AD tgrad path entirely — strictly faster than doing AD to get zeros. But it's cosmetic.

Leaning toward (1).

ChrisRackauckas-Claude:
Superseded by SciML/FunctionWrappersWrappers.jl#45 (registered via JuliaRegistries/General#153645). FWW's existing Enzyme extension was silently dropping the outer `set_runtime_activity(Forward)` flag at the rule boundary; #45 threads `RuntimeActivity` / `StrongZero` through to the delegated `Enzyme.autodiff` call. Once the FWW release lands in the registry and CI re-resolves, the Rosenbrock23 + AutoEnzyme(Forward) path works unchanged — no test-side workaround needed. Closing.

PR #3543, opened 2026-04-24
Fix DAE re-init community callback test for v7 CheckInit default

ChrisRackauckas-Claude:
## Summary

- The `@testset "dae re-init"` block in `lib/DiffEqBase/test/downstream/community_callback_tests.jl` (failing in `DiffEqBase_Downstream2` CI) relied on the pre-v7 `DefaultInit` behavior that silently fell back to `BrownFullBasicInit` for mass-matrix ODEs / `DAEProblem`s.
- In v7, `DefaultInit` dispatches to `CheckInit` (see `lib/OrdinaryDiffEqCore/src/initialize_dae.jl:30-45`), and the test's `u0 = [0.0, -1.0]` for the algebraic constraint `u[2] = p = 1` is inconsistent (`normresid = sqrt(2) > abstol`). `CheckInit` therefore errors out at the start of `solve()`, before the `PresetTimeCallback` that flips `p` even fires.
- Pass `initializealg = BrownFullBasicInit()` explicitly so the solver fixes the inconsistent initial condition and re-solves the algebraic constraint after the callback flips `p` (which is what the test is actually validating).
- Also add explicit imports for `DFBDF` (from `OrdinaryDiffEqBDF`) and `BrownFullBasicInit` (from `OrdinaryDiffEqNonlinearSolve`) — neither is re-exported by `OrdinaryDiffEq` in v7. Thread `OrdinaryDiffEqNonlinearSolve` through the downstream test env's `[deps]` and `[sources]`.

Same pattern as #3522 / f6010f893a / 9ec2afb334 (test-side migration for v7 `CheckInit` default and narrower `OrdinaryDiffEq` re-exports).

Root cause trace:
- `lib/DiffEqBase/test/downstream/community_callback_tests.jl:239` — `solve(prob, alg)` with no `initializealg`.
- `lib/OrdinaryDiffEqCore/src/solve.jl:100` — default is `DefaultInit()`.
- `lib/OrdinaryDiffEqCore/src/initialize_dae.jl:30-45` — `DefaultInit` → `CheckInit` when no `initializeprob` (v7 breaking change from 1d37f2afde, documented in `NEWS.md:245-258`).
- `SciMLBase/src/initialization.jl:201-224` — `CheckInit` evaluates `f(u, p, t)`, takes the algebraic rows of the residual (`-2` on row 2), and throws `CheckInitFailureError` with `normresid = sqrt(2)` via the default `ODE_DEFAULT_NORM`.

## Test plan

- [x] Run `@testset "dae re-init"` (both `FBDF()` and `Rodas5P()`) locally against the downstream env — passes.
- [x] Run the `DAEProblem` / `DFBDF()` assertion at line 253 locally — passes.
- [x] Run the full `community_callback_tests.jl` file under `@safetestset` locally — `4 passed, 0 failed`.
- [x] `Runic` format check passes on the edited files.
- [ ] `DiffEqBase_Downstream2` CI goes green.

### DIFF
--- a/lib/DiffEqBase/test/downstream/distributed_ensemble.jl
+++ b/lib/DiffEqBase/test/downstream/distributed_ensemble.jl
@@ -40,13 +40,16 @@ end
 
 ensemble_prob = EnsembleProblem(prob, prob_func = lorenz_prob_func, safetycopy = true)
 
+# v7: EnsembleSolution is an AbstractArray, so length(sim) returns
+# prod(size(sim)) over (state, time, trajectory). Trajectory count is
+# length(sim.u). See NEWS.md "EnsembleSolution indexing".
 println("Running EnsembleSerial()")
-@test length(solve(ensemble_prob, Tsit5(), EnsembleSerial(), trajectories = 100)) == 100
+@test length(solve(ensemble_prob, Tsit5(), EnsembleSerial(), trajectories = 100).u) == 100
 println("Running EnsembleThreads()")
-@test length(solve(ensemble_prob, Tsit5(), EnsembleThreads(), trajectories = 100)) == 100
+@test length(solve(ensemble_prob, Tsit5(), EnsembleThreads(), trajectories = 100).u) == 100
 println("Running EnsembleDistributed()")
-@test length(solve(ensemble_prob, Tsit5(), EnsembleDistributed(), trajectories = 100)) ==
+@test length(solve(ensemble_prob, Tsit5(), EnsembleDistributed(), trajectories = 100).u) ==
     100
 println("Running EnsembleSplitThreads()")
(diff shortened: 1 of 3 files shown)


lib/DiffEqBase/test/downstream/distributed_ensemble.jl
```diff
@@ -40,13 +40,16 @@ end
 
 ensemble_prob = EnsembleProblem(prob, prob_func = lorenz_prob_func, safetycopy = true)
 
+# v7: EnsembleSolution is an AbstractArray, so length(sim) returns
+# prod(size(sim)) over (state, time, trajectory). Trajectory count is
+# length(sim.u). See NEWS.md "EnsembleSolution indexing".
```
ChrisRackauckas:
```suggestion
```

PR #3519, opened 2026-04-23
Restore PIController default for SDE-implicit solvers (fixes Hayes SDDE)

ChrisRackauckas-Claude:
## Summary

Supersedes the `adaptive = false` workaround in #3515. Fixes the real regression.

#3485 ("Remove legacy controller traits and dead code") flipped six SDE-implicit solvers (`ImplicitEM`, `ImplicitEulerHeun`, `ImplicitRKMil`, `ISSEM`, `ISSEulerHeun`, `SKenCarp`) from the legacy `PIController` default to a hard-coded `PredictiveController` at `lib/StochasticDiffEqImplicit/src/alg_utils.jl:8-13`. Before #3485, every `StochasticDiffEqAlgorithm` had `ispredictive = false`, which `legacy_default_controller` translated into `PIController`.

`PredictiveController.step_reject_controller!` does:

```julia
dt_new = success_iter == 0 ? 0.1 * dt : dt / qold
```

— a monotonic shrink that collapses `dt` under noisy stochastic error estimates. `PIController.step_reject_controller!` is gentler:

```julia
dt *= max(qmin, min(1, qacc))
```

## How we got here

Under `MethodOfSteps(ISSEulerHeun())` on the Hayes SDDE regression test (#3363), the history-interpolation discontinuities compound the stochastic signal noise. The `PredictiveController` rejects every step, `dt` collapses within ~40 inner steps, and the solve aborts with `DtLessThanMin`. ISSEM (Itô, mean-zero `dW²-dt` estimator) also inflated from 34 → 572 steps on SDDE vs. the non-delay SDE — that observation **falsifies** the "Stratonovich estimator bias" hypothesis an earlier investigation landed on: this is a controller-kind issue, not a numerics issue.

## Bisection

First-bad commit `dc50962c6` (merge of #3485). Patching the six `default_controller` entries back to `PIController` restores pre-regression step counts exactly on the Hayes SDDE:

| Method | Pre-#3485 | Master (bad) | With this PR |
|---|---|---|---|
| `ImplicitEM` | 21 | 21 | 21 |
| `ISSEM` | 174 | 572 | 174 |
| `ImplicitEulerHeun` | 2638 | 575 (inflated) | 2638 |
| `ISSEulerHeun` | 2339 | **`DtLessThanMin` @ t≈0.02** | 2339 |

Matches the baseline the Hayes test was originally validated against on `fork/fix-sdde-support`.

## Verified locally

[...]

PR #3544, opened 2026-04-24
Fix flaky SDDE ImplicitEulerHeun test by using fixed dt

ChrisRackauckas-Claude:
## Summary

The `SDDE Implicit Method Tests` safetestset in
`lib/DelayDiffEq/test/sdde/implicit_methods.jl:38` (the
`MethodOfSteps(ImplicitEulerHeun())` assert on the Hayes problem) fails
intermittently in `DelayDiffEq_SDDE` CI with `DtLessThanMin` at
`t ≈ 0.04` — aborts at `dt ≤ dtmin`. Failure rate on current master
(post PR #3519's `PIController` restoration) is ~7% of random seeds on
this test; all three of `ImplicitEM`, `ISSEM`, and `ISSEulerHeun` remain
stable on the same problem.

## Root cause

The Stratonovich Milstein-correction error estimator in
`StochasticDiffEqImplicit` uses the raw `dW^2` form with non-zero mean
`E[dW²] = dt`:

- OOP constant cache: `perform_step/sdirk.jl:44`
  `mil_correction = ggprime .* (integrator.W.dW .^ 2) ./ 2`
- In-place cache: `perform_step/sdirk.jl:276`
  `@.. dz = (g_sized2 - g_sized) / (2integrator.sqdt) * (dW .^ 2)`

The Itô counterparts (`ImplicitEM`) use the bias-corrected
`dW² - dt` form, which is mean-zero and doesn't drift the error
estimate. On the non-delay Hayes SDE the `PIController`'s rejection
path absorbs the Stratonovich bias fine; under `MethodOfSteps` the
history-interpolation noise compounds it and occasionally drives `dt`
to `dtmin`.

## Fix

Run this specific assert with `adaptive = false`, matching the existing
`ImplicitRKMil` line in the same file and the standalone SDE reference
pattern in
`StochasticDiffEq/test/implicit_time_parameter_test.jl`. This keeps the
Hayes regression coverage for `ImplicitEulerHeun` without the flaky
controller interaction.

Rewriting the Stratonovich error estimator to use `dW² - dt`
(mean-zero) would be a broader numerics change with its own
convergence-test implications, and is out of scope for a CI-unblocking
fix.

## Test plan

- [x] `julia +1.11 --project=lib/DelayDiffEq -e 'using Pkg; Pkg.test("DelayDiffEq"; test_args=["SDDE"])'`
      → `SDDE Implicit Method Tests | 12 pass`.
- [x] 500-seed stress run on the `adaptive = false` path passes 500/500
      (vs. 7% flake on adaptive).
- [ ] `DelayDiffEq_SDDE` Sublibrary CI job green.

### DIFF
--- a/lib/DelayDiffEq/test/sdde/implicit_methods.jl
+++ b/lib/DelayDiffEq/test/sdde/implicit_methods.jl
@@ -33,8 +33,15 @@ sol = solve(prob, MethodOfSteps(ImplicitEM()))
 @test SciMLBase.successful_retcode(sol)
 @test sol.u[end] != zeros(1)
 
-# ImplicitEulerHeun
-sol = solve(prob, MethodOfSteps(ImplicitEulerHeun()), dt = 0.01)
+# ImplicitEulerHeun — the Stratonovich Milstein-correction error estimator
+# uses `dW^2` (non-zero mean `E[dW²]=dt`), whereas the Itô counterparts
+# (`ImplicitEM`, `ISSEM`) use the bias-corrected `dW² - dt` form. Under
+# `MethodOfSteps` history-interpolation noise this biased estimator
+# intermittently (~7% of seeds on the Hayes problem) collapses `dt` to
+# `dtmin` even with the `PIController` rejection path restored in #3519.
+# Use fixed-`dt` stepping (same pattern as `ImplicitRKMil` below and the
+# standalone SDE reference `StochasticDiffEq/test/implicit_time_parameter_test.jl`).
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Tracked this down. It's not a regression, it's a new test from the migration testing all algorithms. But Heun is Stratanovich interpretation, this problem is written for Ito so it's not generally stable. So it's not a great test, should flex the solving but the adaptivity won't make sense here.

PR #3580, opened 2026-04-28
Make DefaultODEAlgorithm autoswitch NaN-safe in is_stiff

ChrisRackauckas-Claude:
## Summary

`OrdinaryDiffEqDefault.is_stiff` (used only by `DefaultODEAlgorithm`) and `OrdinaryDiffEqCore.is_stiff` (used by every other autoswitch composite — `AutoTsit5`, `AutoVern7`, etc.) compute the same scalar but differ on a single line:

```julia
# default_alg.jl  (current)
bool = stiffness > os * tol

# composite_algs.jl
bool = !(stiffness <= os * tol)
```

Equivalent for finite values, opposite for `NaN`:

| stiffness | `> os*tol` | `!(<= os*tol)` |
|---|---|---|
| finite, large | true | true |
| finite, small | false | false |
| `NaN` | **false** ("not stiff") | **true** ("stiff") |

`NaN` is reachable in practice. The Hairer-style spectral estimator inside `Tsit5`/`Vern7` is

```julia
integrator.eigen_est = norm(maximum(abs.((k_last .- k_prev) ./ (g_last .- g_prev))), Inf)
```

If a user RHS contains a clamp like `u = max.(u, 1e-50)` (very common in chemistry/combustion code to keep `log` and `^p` rates from blowing up on tiny negative drift) and any component sits at the floor for an entire step, that component's `g_last - g_prev = 0` and `k_last - k_prev = 0` — the componentwise `0/0` poisons the `max` and `eigen_est` becomes `NaN`. Under the old `bool = stiffness > os * tol`, the autoswitch sees "not stiff" forever, never leaves Vern7/Tsit5, and the user hits `MaxIters` even though the underlying problem is solvable in <10 s by Rodas5P/FBDF.

This was reported by a user running a 30-species methane mechanism (LuMechanism) with `solve(prob; abstol=1e-12, reltol=1e-9)` — the integration ran for 2400 s, took 431k steps, advanced t from 0 to ~1e-3 (out of 1.0), and `sol.alg_choice` was `[2]` throughout: zero transitions to a stiff method.

## Fix

One-line change in `lib/OrdinaryDiffEqDefault/src/default_alg.jl` to match the form in `OrdinaryDiffEqCore`:

```diff
-    bool = stiffness > os * tol
+    bool = !(stiffness <= os * tol)
```

The semantics: "if the spectral-radius estimator is degenerate (`NaN`), assume the worst — switch to the implicit branch." That's also what every non-default `AutoXxx` autoswitch already does.

## Empirical verification

| Setup | Before | After |
|---|---|---|
| LuMechanism, n=31, abstol=1e-12, reltol=1e-9 | MaxIters @ 22446 steps, 79 s, alg_choice [2] | Success @ 31 steps, 10 s, alg_choice [2,4] |
| MWE with `max(u,1e-50)` clamp (added as test) | alg_choice [2], no transitions | alg_choice [2,4], transitions ✓ |

## Test plan

- [x] Added regression test `clamped_kink_rhs!` to `lib/OrdinaryDiffEqDefault/test/default_solver_tests.jl` — fails on `master`, passes with this fix.
- [ ] Existing default-solver tests still pass (no semantic change for finite stiffness).
- [ ] CI green.

### DIFF
--- a/lib/OrdinaryDiffEqDefault/test/default_solver_tests.jl
+++ b/lib/OrdinaryDiffEqDefault/test/default_solver_tests.jl
@@ -163,8 +163,8 @@ function clamped_kink_rhs!(du, u, p, t)
     @views du[2:end] .= 0.0
     return nothing
 end
-prob_kink = ODEProblem(clamped_kink_rhs!, zeros(20), (0.0, 1.0))
-sol_kink = solve(prob_kink; abstol = 1.0e-12, reltol = 1.0e-9, maxiters = 50_000)
+prob_kink = ODEProblem(clamped_kink_rhs!, zeros(20), (0.0, 1e-3))
+sol_kink = solve(prob_kink; abstol = 1.0e-12, reltol = 1.0e-9, maxiters = 1_000)
 @test sol_kink.retcode == ReturnCode.Success
 # Without the NaN-safe is_stiff form the run stays on Vern7 the entire time
 # (alg_choice == [2]); with it, autoswitch sees a degenerate eigen estimate
--- a/lib/OrdinaryDiffEqDefault/test/default_solver_tests.jl
+++ b/lib/OrdinaryDiffEqDefault/test/default_solver_tests.jl
@@ -163,7 +163,7 @@ function clamped_kink_rhs!(du, u, p, t)
     @views du[2:end] .= 0.0
     return nothing
 end
-prob_kink = ODEProblem(clamped_kink_rhs!, zeros(20), (0.0, 1e-3))
+prob_kink = ODEProblem(clamped_kink_rhs!, zeros(20), (0.0, 0.1))
 sol_kink = solve(prob_kink; abstol = 1.0e-12, reltol = 1.0e-9, maxiters = 1_000)
 @test sol_kink.retcode == ReturnCode.Success
 # Without the NaN-safe is_stiff form the run stays on Vern7 the entire time
--- a/lib/OrdinaryDiffEqDefault/test/default_solver_tests.jl
+++ b/lib/OrdinaryDiffEqDefault/test/default_solver_tests.jl
@@ -150,7 +150,6 @@ prob_complex = ODEProblem(schrod_eq, complex([1, -1] / sqrt(2)), (0, 1), 100)
 complex_sol = solve(prob_complex)
 @test complex_sol.retcode == ReturnCode.Success
 
-# Regression test for https://github.com/SciML/OrdinaryDiffEq.jl/issues/XXXX:
 # the autoswitch detector inside DefaultODEAlgorithm must treat a NaN
 # spectral-radius estimate as "stiff" so it can fall back to an implicit
 # method, instead of as "not stiff" (which gets it stuck on the explicit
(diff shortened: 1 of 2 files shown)


lib/OrdinaryDiffEqDefault/test/default_solver_tests.jl
```diff
@@ -150,6 +150,28 @@ prob_complex = ODEProblem(schrod_eq, complex([1, -1] / sqrt(2)), (0, 1), 100)
 complex_sol = solve(prob_complex)
 @test complex_sol.retcode == ReturnCode.Success
 
+# Regression test for https://github.com/SciML/OrdinaryDiffEq.jl/issues/XXXX:
```
ChrisRackauckas:
```suggestion
```

lib/OrdinaryDiffEqDefault/test/default_solver_tests.jl
```diff
@@ -150,6 +150,27 @@ prob_complex = ODEProblem(schrod_eq, complex([1, -1] / sqrt(2)), (0, 1), 100)
...
+# (k_last - k_prev)[i] / (g_last - g_prev)[i] = 0/0 in
+# eigen_est = max(abs((k_last - k_prev) ./ (g_last - g_prev))).
+function clamped_kink_rhs!(du, u, p, t)
+    u = max.(u, 1.0e-50)
+    du[1] = -1000.0 * u[1] + 1.0
+    @views du[2:end] .= 0.0
+    return nothing
+end
+prob_kink = ODEProblem(clamped_kink_rhs!, zeros(20), (0.0, 1.0))
+sol_kink = solve(prob_kink; abstol = 1.0e-12, reltol = 1.0e-9, maxiters = 50_000)
```
oscardssmith: to be mindful of CI time, should we make the u0 and timespan such that this triggers in fewer iters?

ChrisRackauckas:
```suggestion
prob_kink = ODEProblem(clamped_kink_rhs!, zeros(20), (0.0, 1e-3))
sol_kink = solve(prob_kink; abstol = 1.0e-12, reltol = 1.0e-9, maxiters = 1_000)
```

lib/OrdinaryDiffEqDefault/test/default_solver_tests.jl
```diff
@@ -150,6 +150,27 @@ prob_complex = ODEProblem(schrod_eq, complex([1, -1] / sqrt(2)), (0, 1), 100)
...
+# stage value of the inactive components to the same floor, giving
+# (k_last - k_prev)[i] / (g_last - g_prev)[i] = 0/0 in
+# eigen_est = max(abs((k_last - k_prev) ./ (g_last - g_prev))).
+function clamped_kink_rhs!(du, u, p, t)
+    u = max.(u, 1.0e-50)
+    du[1] = -1000.0 * u[1] + 1.0
+    @views du[2:end] .= 0.0
+    return nothing
+end
+prob_kink = ODEProblem(clamped_kink_rhs!, zeros(20), (0.0, 1e-3))
```
ChrisRackauckas:
```suggestion
prob_kink = ODEProblem(clamped_kink_rhs!, zeros(20), (0.0, 0.1))
```

PR #3590, opened 2026-05-03
refactor(LowStorageRK): unify alg_cache + fix Aqua unbound type param…

singhharsh1708:
Cleanup pass on `OrdinaryDiffEqLowStorageRK`. The 9 generic cache classes
(`LowStorageRK2N`, `2C`, `3S`, `3Sp`, `3SpFSAL`, `2RP`, `3RP`, `4RP`, `5RP`)
were already tableau-driven, but had three issues left over:

- `{N, T, T2}` parameterization with `NTuple{N, T}` fields tripped Aqua's
  `unbound_args` check (same root cause as the symplectic PR).
- All 44 `XxxConstantCache(T, T2)` constructors lacked `::Type{T}` annotations,
  so dispatch wasn't specialized on T/T2 (same complaint Chris raised on the
  symplectic PR).
- 82 per-algorithm `alg_cache` methods duplicated the same wiring per class.

## Changes (only `low_storage_rk_caches.jl`)

- **Reparameterize 9 cache structs** from `{N, T, T2}` with `NTuple{N, T}`
  fields → `{T, AType, CType}` with field-typed parameters. All 41 constructor
  return statements drop the explicit `{N, T, T2}` (Julia infers).
- **Specialize all 44 tableau constructors** with
  `::Type{T}, ::Type{T2}) where {T, T2}`.
- **Collapse 82 per-algorithm `alg_cache` methods → 18 generic ones**
  (9 IIP + 9 OOP) via per-class `Union` types and `_lowstorage_xxx_tableau`
  dispatch helpers. The 3 special algorithms (`RK46NL`, `SHLDDRK_2N`,
  `SHLDDRK52`) keep their own `alg_cache` because their caches differ from
  the 9 generic patterns.

`perform_step!` methods are unchanged — each storage class has a fundamentally
different memory layout and stage recurrence, so they can't be folded together.

Net: **-1106 lines** in `low_storage_rk_caches.jl` with no behavior change.

## Test plan

- [x] Core tests: 515 pass, 28 broken (pre-existing)
- [x] Aqua `unbound_args`: now passes (was failing on master)
- [x] Aqua other checks: pass (only `stale_deps` fails — pre-existing test
      setup issue unrelated to this change)
- [x] JET: pass
- [x] AllocCheck: 38/38 algorithms allocation-free

### DIFF
--- a/lib/OrdinaryDiffEqLowStorageRK/src/low_storage_rk_caches.jl
+++ b/lib/OrdinaryDiffEqLowStorageRK/src/low_storage_rk_caches.jl
@@ -17,11 +17,11 @@ get_fsalfirstlast(cache::LowStorageRKMutableCache, u) = (cache.fsalfirst, cache.
     thread::Thread
 end
 
-struct LowStorageRK2NConstantCache{T, AType, CType} <: OrdinaryDiffEqConstantCache
-    A2end::AType # A1 is always zero
+struct LowStorageRK2NConstantCache{N, T, T2} <: OrdinaryDiffEqConstantCache
+    A2end::NTuple{N, T} # A1 is always zero
     B1::T
-    B2end::AType
-    c2end::CType # c1 is always zero
+    B2end::NTuple{N, T}
+    c2end::NTuple{N, T2} # c1 is always zero
 end
 
 function ORK256ConstantCache(::Type{T}, ::Type{T2}) where {T, T2}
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
I'm not sure about dropping those NTuple specializations, what's the purpose there?

singhharsh1708:
To avoid failures in Aqua’s `test_unbound_args`, `AType`/`CType` are used instead of directly specifying `NTuple{N,T}`.
Julia automatically generates an outer constructor where the parameter `N` appears only in a `Vararg` position when using `NTuple{N,T}`. This pattern is flagged by Aqua as an unbound type parameter.
Using `AType`/`CType` resolves this issue while preserving performance characteristics. These types are always instantiated as concrete `NTuple`s (e.g., `NTuple{4,Float64}`), ensuring that:
* Specialization still occurs at compile time
* Loop bounds in `perform_step!` remain statically known
* The compiler can unroll loops exactly as before
As a result, this change maintains identical runtime performance while satisfying Aqua’s checks.

ChrisRackauckas:
Seems like an upstream problem. Do you have an MWE of that? Would be good to file it to Aqua.jl

singhharsh1708:
Agreed, filed it upstream: JuliaTesting/Aqua.jl#376. Restored the NTuple{N,T} specializations in the latest push.

singhharsh1708:
Turned out the Aqua warning is valid — when N = 0, NTuple{0, T2} reduces to Tuple{} so T2 is genuinely unbound.

ChrisRackauckas:
Okay, not so thrilled about this but I see the point.

PR #3599, opened 2026-05-04
Fix #3596: seed JacReuseState with tTypeNoUnits, not zero(dt)

### LINKED ISSUE #3596: test (AD, lts) fails: MethodError: no method matching Float64(::ForwardDiff.Dual)
## Summary

The `AD` test group on Julia LTS fails with `MethodError: no method matching Float64(::ForwardDiff.Dual{...})` — 4 errors reported. This is the only AD-suite variant that fails (Julia 1, 1.11, pre all pass).

## Reproduction

Recent CI run:
- https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/25311183756/job/74198190567

Failure pattern (repeated 4 times):
```
Expression: ≈(DI.derivative(g, AutoForwardDiff(), 0.0), (_u0 / 2) * exp(-0.5), rtol = rtol)
MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{Main.var"##AD Tests#298".var"#46#49"{UnionAll, Float64}, Float64}, Float64, 1})
```

Final tally:
```
ERROR: LoadError: Some tests did not pass: 28297 passed, 0 failed, 4 errored, 0 broken.
```

## Pre-existing on master

This same job (`test (AD, lts)`) is the only failing job on the most recent master CI run, so it is independent of any feature branch and reproduces on master directly.

## Likely cause

The `Float64(::Dual)` MethodError suggests something in the AD path is forcing conversion of a `Dual` number to `Float64` (instead of unwrapping or staying generic). The fact that LTS (1.10) is the only Julia version affected points at a Julia-version-dependent dispatch path (e.g., a method only specialized in newer Julia, or a Compat shim missing on LTS).

singhharsh1708:
### Description

This PR fixes a type instability issue encountered when differentiating ODE solves with respect to the initial time `t0` using `ForwardDiff`.

On Julia LTS (1.10) Linux x86_64, all four out-of-place Rosenbrock methods:

* `Rosenbrock23`
* `Rosenbrock32`
* `Rodas4`
* `Rodas5`

fail with:

```
MethodError: no method matching Float64(::ForwardDiff.Dual{...})
setproperty!(::JacReuseState{Float64}, ..., ::Dual)
```

(Not reproducible on macOS aarch64 with the same versions.)

---

### Root Cause

`JacReuseState{T}` was initialized using `zero(dt)` inside `alg_cache`.

When `solve()` is called without an explicit `dt`, the internal `_dt` may be derived from:

```
eltype(prob.tspan)(0)
```

There exists an execution path where this results in a `Float64`, even when `tspan[1]` is a `ForwardDiff.Dual`. This causes:

* Cache initialized as `JacReuseState{Float64}`
* Runtime `dt` becomes `Dual`
* Assignment failure when storing `dt * gamma` (Dual → Float64)

---

### Fix

Initialize `JacReuseState` using:

```
zero(tTypeNoUnits)
```

* `tTypeNoUnits` is derived from:

  ```
  oneunit(first(tspan))
  ```
* It reliably reflects the actual time type at solve time:

  * `Dual` under AD
  * `Dual{Dual}` under nested AD

This ensures all internal fields (e.g., `dtgamma`) correctly accept `dt * gamma`.

---

### Changes

* Updated all 6 `alg_cache` call sites:

  * Rosenbrock23 (OOP + IIP)
  * Rosenbrock32 (OOP + IIP)
  * Rodas methods (combined OOP + IIP)

---

### Results

* Fixes ForwardDiff differentiation w.r.t. `t0` on Julia LTS Linux
* Works for:

  * Single AD
  * Nested AD (Hessian)
* Type inference preserved:

  * `@inferred solve` passes for both OOP and IIP

---

### Notes

This change ensures consistency between cache initialization and runtime types, preventing AD-related type conflicts without impacting performance or existing solver behavior.

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
@@ -891,10 +891,6 @@ system matrix. Supports Jacobian reuse for W-methods via `jac_reuse` in the cach
 function calc_rosenbrock_differentiation(integrator, cache, dtgamma, repeat_step)
     jac_reuse = get_jac_reuse(cache)
 
-    # Skip reuse if jac_reuse can't hold dtgamma: e.g. JacReuseState{Float64} when
-    # dtgamma is a ForwardDiff.Dual (AD differentiation w.r.t. t0). This mismatch
-    # arises when the cache was built with zero(dt) instead of zero(tTypeNoUnits)
-    # (older registered versions of rosenbrock_caches.jl, see #3596).
     if repeat_step || jac_reuse === nothing ||
             !(typeof(dtgamma) <: typeof(jac_reuse.pending_dtgamma))
         dT = calc_tderivative(integrator, cache)
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
@@ -891,8 +891,12 @@ system matrix. Supports Jacobian reuse for W-methods via `jac_reuse` in the cach
 function calc_rosenbrock_differentiation(integrator, cache, dtgamma, repeat_step)
     jac_reuse = get_jac_reuse(cache)
 
-    # If no reuse support or repeat step, use standard path
-    if repeat_step || jac_reuse === nothing
+    # Skip reuse if jac_reuse can't hold dtgamma: e.g. JacReuseState{Float64} when
+    # dtgamma is a ForwardDiff.Dual (AD differentiation w.r.t. t0). This mismatch
+    # arises when the cache was built with zero(dt) instead of zero(tTypeNoUnits)
+    # (older registered versions of rosenbrock_caches.jl, see #3596).
+    if repeat_step || jac_reuse === nothing ||
+            !(typeof(dtgamma) <: typeof(jac_reuse.pending_dtgamma))
         dT = calc_tderivative(integrator, cache)
         W = calc_W(integrator, cache, dtgamma, repeat_step)
         return dT, W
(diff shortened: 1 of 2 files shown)


lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
```diff
@@ -891,8 +891,8 @@ system matrix. Supports Jacobian reuse for W-methods via `jac_reuse` in the cach
 function calc_rosenbrock_differentiation(integrator, cache, dtgamma, repeat_step)
     jac_reuse = get_jac_reuse(cache)
 
-    # If no reuse support or repeat step, use standard path
-    if repeat_step || jac_reuse === nothing
+    if repeat_step || jac_reuse === nothing ||
+            !(typeof(dtgamma) <: typeof(jac_reuse.pending_dtgamma))
```
ChrisRackauckas: what is this condition?

singhharsh1708:
removing the guard — the rosenbrock_caches.jl fix (zero(tTypeNoUnits) instead of zero(dt)) is the correct fix. The guard was added as a defensive shim for older registered versions but that's not needed here.

ChrisRackauckas:
It's not correct though because then reuse is disabled when duals are used.

singhharsh1708:
Makes sense ,so we should explicitly skip reuse when dtgamma is a Dual? Something like || dtgamma is a ForwardDiff.Dual or is there a cleaner pattern you'd suggest?

ChrisRackauckas:
No we should not change the algorithm when dual. Just make pass down duals appropriately.

singhharsh1708:
so means this fix is correct without guards?

singhharsh1708:
the zero(tTypeNoUnits) fix in rosenbrock_caches.jl ensures JacReuseState is initialized with the correct type so Duals pass through without any guard. Current branch has just that fix.

ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/3589 is more correct

PR #3603, opened 2026-05-07
Fix DDE preset callbacks at propagated tstops

AshtonSBradley:
## Summary

- Fixes DDE `PresetTimeCallback` hits when a constant-lag propagated discontinuity lands at the same time up to floating-point roundoff.
- Adds DDE tstop snap state mirroring the ODE integrator path so accepted tstop steps can use the exact target time.
- Adds a regression test for `SciML/DifferentialEquations.jl#1124`, where `constant_lags = [0.2, 4.0]` caused the preset callback at `t = 9.0` to be skipped.

## Explanation

For DDEs with constant lags, propagated discontinuity tstops can accumulate small floating-point drift. In the reported case, the propagated discontinuity reached `8.999999999999996`, which sorted before the user callback tstop at `9.0`. The solver therefore stepped to the drifted propagated tstop, and `PresetTimeCallback` did not fire because it checks the preset time list exactly.

This change snaps newly propagated constant-lag discontinuity tstops to an existing user tstop when they differ only by roundoff. It also gives `DDEIntegrator` the same tstop target bookkeeping used by ODE integrators, so tstop-driven accepted steps can report the exact target time to callback handling.

Fixes SciML/DifferentialEquations.jl#1124.

## Tests

- `julia -e 'using Pkg; Pkg.activate(; temp=true); Pkg.develop(path="/private/tmp/ode-v6-backport/lib/DelayDiffEq"); Pkg.add(["DDEProblemLibrary", "DiffEqCallbacks", "DiffEqDevTools", "OrdinaryDiffEqTsit5"]); include("/private/tmp/ode-v6-backport/lib/DelayDiffEq/test/integrators/events.jl")'`
  - `continuous`: 8/8 passed
  - `discrete`: 4/4 passed
  - `preset callback at constant-lag propagated tstop`: 2/2 passed
  - `save discontinuity`: 3/3 passed
  - `periodic callback`: 2/2 passed
- `julia --project=@runic -e 'using Runic; exit(Runic.main(["--check", "lib/DelayDiffEq/src/integrators/interface.jl", "lib/DelayDiffEq/src/integrators/type.jl", "lib/DelayDiffEq/src/integrators/utils.jl", "lib/DelayDiffEq/src/solve.jl", "lib/DelayDiffEq/test/integrators/events.jl"]))'`
- `git diff --check`

### DIFF
--- a/lib/DelayDiffEq/src/integrators/interface.jl
+++ b/lib/DelayDiffEq/src/integrators/interface.jl
@@ -584,6 +584,18 @@ function DiffEqBase.get_tstops_max(integrator::DDEIntegrator)
     return isempty(tstops_array) ? integrator.sol.prob.tspan[end] : maximum(tstops_array)
 end
 
+OrdinaryDiffEqCore._get_next_step_tstop(integrator::DDEIntegrator) = integrator.next_step_tstop
+function OrdinaryDiffEqCore._set_tstop_flag!(
+        integrator::DDEIntegrator, is_tstop::Bool, target = nothing
+    )
+    integrator.next_step_tstop = is_tstop
+    if is_tstop && target !== nothing
+        integrator.tstop_target = target
+    end
+    return nothing
+end
+OrdinaryDiffEqCore._get_tstop_target(integrator::DDEIntegrator) = integrator.tstop_target
+
 # update integrator when u is modified by callbacks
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
The old snapping behavior has too many caveats. We need to just fix the callback.

PR #3624, opened 2026-05-12
Fix ARS343 BoundsError: set split_guess to [0,1,2,3] instead of zeros

### LINKED ISSUE #3623: ARS343 does not work on master (2026-05-12)
**Describe the bug 🐞**

I wanted to test the new method implemented in https://github.com/SciML/OrdinaryDiffEq.jl/pull/3196. Here is an MWE:

```julia
pkg> add OrdinaryDiffEqSDIRK#master

julia> using OrdinaryDiffEqSDIRK

julia> function rhs_nonstiff!(du, u, parameters, t)
               u1, u2 = u
               du[1] = u2 - u1 - u1^2
               du[2] = -2 * u2
               return nothing
       end
rhs_nonstiff! (generic function with 1 method)

julia> function rhs_stiff!(du, u, parameters, t)
               (; epsilon) = parameters
               u1, u2 = u
               du[1] = 0
               du[2] = (u1^2 - u2) / epsilon
               return nothing
       end
rhs_stiff! (generic function with 1 method)

julia> ode = SplitODEProblem{true}(rhs_stiff!, rhs_nonstiff!, [1.5, 1.0], (0.0, 1.0),
                                   (; epsilon = 1.0e-11))
ODEProblem with uType Vector{Float64} and tType Float64. In-place: true
Non-trivial mass matrix: false
timespan: (0.0, 1.0)
u0: 2-element Vector{Float64}:
 1.5
 1.0

julia> sol = solve(ode, ARS343(); dt = 0.02)
ERROR: BoundsError: attempt to access 4-element Vector{Vector{Float64}} at index [0]
Stacktrace:
  [1] getindex
    @ ./essentials.jl:13 [inlined]
  [2] perform_step!(integrator::OrdinaryDiffEqCore.ODEIntegrator{…}, cache::OrdinaryDiffEqSDIRK.ESDIRKIMEXCache{…}, repeat_step::Bool)
    @ OrdinaryDiffEqSDIRK ~/.julia/packages/OrdinaryDiffEqSDIRK/SnhR6/src/generic_imex_perform_step.jl:254
  [3] perform_step!
    @ ~/.julia/packages/OrdinaryDiffEqSDIRK/SnhR6/src/generic_imex_perform_step.jl:212 [inlined]
  [4] solve!(integrator::OrdinaryDiffEqCore.ODEIntegrator{…})
    @ OrdinaryDiffEqCore ~/.julia/packages/OrdinaryDiffEqCore/plDZQ/src/solve.jl:772
  [5] __solve(::ODEProblem{…}, ::ARS343{…}; kwargs::@Kwargs{…})
    @ OrdinaryDiffEqCore ~/.julia/packages/OrdinaryDiffEqCore/plDZQ/src/solve.jl:10
  [6] __solve
    @ ~/.julia/packages/OrdinaryDiffEqCore/plDZQ/src/solve.jl:1 [inlined]
  [7] solve_call(_prob::ODEProblem{…}, args::ARS343{…}; merge_callbacks::Bool, kwargshandle::Nothing, kwargs::@Kwargs{…})
    @ DiffEqBase ~/.julia/packages/DiffEqBase/NZi5C/src/solve.jl:172
  [8] solve_call
    @ ~/.julia/packages/DiffEqBase/NZi5C/src/solve.jl:137 [inlined]
  [9] #solve_up#38
    @ ~/.julia/packages/DiffEqBase/NZi5C/src/solve.jl:646 [inlined]
 [10] solve_up
    @ ~/.julia/packages/DiffEqBase/NZi5C/src/solve.jl:619 [inlined]
 [11] #solve#37
    @ ~/.julia/packages/DiffEqBase/NZi5C/src/solve.jl:603 [inlined]
 [12] top-level scope
    @ REPL[10]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

**Expected behavior**

It works and does not throw an error.

**Minimal Reproducible Example 👇**

See above

**Error & Stacktrace ⚠️**

See above

singhharsh1708:
split_guess[i] tells the FSAL IMEX perform_step! which zs slot to copy as initial guess for stage i. zeros(Int, s) caused zs[0] to be accessed, triggering a BoundsError when solving SplitODEProblems with ARS343. Fixes #3623

## Checklist

### DIFF
--- a/lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
+++ b/lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
@@ -138,9 +138,7 @@ end
             end
         end
 
-        if integrator.f isa SplitFunction
-            z_guess = z[1]
-        elseif !isempty(α) && !iszero(α[i][1])
+        if !isempty(α) && !iszero(α[i])
             z_guess = zero(u)
             for j in 1:(i - 1)
                 z_guess = z_guess + α[i][j] * z[j]
@@ -213,7 +211,7 @@ end
     (; zs, ks, atmp, nlsolver, step_limiter!) = cache
     (; tmp) = nlsolver
     tab = cache.tab
-    (; Ai, bi, Ae, be, c, btilde, ebtilde, α, s, reuse_W_at_stage2, split_guess) = tab
+    (; Ai, bi, Ae, be, c, btilde, ebtilde, α, s, reuse_W_at_stage2) = tab
     alg = unwrap_alg(integrator, true)
     γ = Ai[2, 2]
 
@@ -250,9 +248,7 @@ end
             end
         end
 
-        if integrator.f isa SplitFunction
-            copyto!(zs[i], zs[split_guess[i]])
-        elseif !isempty(α) && !iszero(α[i][1])
+        if !isempty(α) && !iszero(α[i])
             fill!(zs[i], zero(eltype(u)))
             for j in 1:(i - 1)
                 @..zs[i] = zs[i] + α[i][j] * zs[j]
--- a/lib/OrdinaryDiffEqSDIRK/src/imex_tableaus.jl
+++ b/lib/OrdinaryDiffEqSDIRK/src/imex_tableaus.jl
@@ -10,7 +10,6 @@ struct ESDIRKIMEXTableau{T, T2}
     order::Int
     s::Int
     reuse_W_at_stage2::Bool
-    split_guess::Vector{Int}  # for SplitFunction: split_guess[i] = which previous stage to copy for stage i
     nlsolver_init_c::T2       # initial c for build_nlsolver (matches master per-alg cache)
 end
 
@@ -131,7 +130,7 @@ function KenCarp3ESDIRKIMEXTableau(T::Type{<:CompiledFloats}, T2::Type{<:Compile
 
     return ESDIRKIMEXTableau(
         Ai, bi_vec, Ae, be_vec, c_vec,
-        btilde_vec, ebtilde_vec, α_vecs, 3, s, true, [0, 1, 2, 2], c3
+        btilde_vec, ebtilde_vec, α_vecs, 3, s, true, c3
     )
 end
 
@@ -264,7 +263,7 @@ function KenCarp3ESDIRKIMEXTableau(T, T2)
 
     return ESDIRKIMEXTableau(
         Ai, bi_vec, Ae, be_vec, c_vec,
-        btilde_vec, ebtilde_vec, α_vecs, 3, s, true, [0, 1, 2, 2], c3
+        btilde_vec, ebtilde_vec, α_vecs, 3, s, true, c3
     )
 end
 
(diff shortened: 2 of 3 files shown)


singhharsh1708:
@ChrisRackauckas

ranocha (commented):
Please add a test checking that split problems work as expected.

ChrisRackauckas:
`0` is still in the array though? I'm not sure this is the general solution. The general solution is that the guess is just given by sum_j^s alpha_j z_j, i.e. that you take all of the previous stage vectors z and do a linear combination of them for the guess. FSAL means you skip a step because you know it, but in any other case it's generally some linear combination of the vectors from before. How to actually build that out is complex, as you saw with KenCarp it's very method dependent, so the alphas need a whole lower triangular tableau.

Then for split methods, it similarly needs a guess mechanism, but the general form should just be to also do linear combinations? 

One of the things we will want to do once we have a tableau form is move to allow for switchable predictors, i.e. https://sundials.readthedocs.io/en/latest/arkode/Mathematics_link.html#implicit-predictors. We have some optimized baked in ones, but want to test other strategies in other scenarios. Keep this in mind for the design of the expansion.

singhharsh1708:
> Please add a test checking that split problems work as expected.

Added @testset "ARS343 SplitODEProblem" in sdirk_convergence_tests.jl — covers both OOP (scalar) and IIP (vector) paths, verifies no BoundsError and convergence order ≈ 3.

singhharsh1708:
> `0` is still in the array though? I'm not sure this is the general solution. The general solution is that the guess is just given by sum_j^s alpha_j z_j, i.e. that you take all of the previous stage vectors z and do a linear combination of them for the guess. FSAL means you skip a step because you know it, but in any other case it's generally some linear combination of the vectors from before. How to actually build that out is complex, as you saw with KenCarp it's very method dependent, so the alphas need a whole lower triangular tableau.
> 
> Then for split methods, it similarly needs a guess mechanism, but the general form should just be to also do linear combinations?
> 
> One of the things we will want to do once we have a tableau form is move to allow for switchable predictors, i.e. https://sundials.readthedocs.io/en/latest/arkode/Mathematics_link.html#implicit-predictors. We have some optimized baked in ones, but want to test other strategies in other scenarios. Keep this in mind for the design of the expansion.

Implemented your suggestion — removed split_guess entirely and unified the predictor: both split and non-split paths now use the same α linear-combination. Also fixed a latent bug where the check was !iszero(α[i][1]) (first element only) instead of !iszero(α[i]) (full vector), which would have silently skipped stages where only later α entries are nonzero. Added α vectors to ARS343 using nearest-stage unit predictors (same c-value proximity heuristic as KenCarp/Kvaerno). Understood the ARKODE switchable predictor design — kept it in mind; the α tableau is the right extensibility hook for that.

lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
```diff
@@ -138,9 +138,7 @@ end
             end
         end
 
-        if integrator.f isa SplitFunction
-            z_guess = z[1]
-        elseif !isempty(α) && !iszero(α[i][1])
+        if !isempty(α) && !iszero(α[i])
             z_guess = zero(u)
             for j in 1:(i - 1)
                 z_guess = z_guess + α[i][j] * z[j]
```
ChrisRackauckas, line 144: Not GPU friendly: should broadcast.

ChrisRackauckas: Actually, z is a vector of vector, this is fine.

ChrisRackauckas (approved):

ranocha:
Thanks for the quick fix

Issue #3222, opened 2021-01-14
Apparent failure to catch a zero cross of a condition


klaff:
The following Pluto notebook shows a missed zero cross of a condition.
```
### A Pluto.jl notebook ###
# v0.12.18

using Markdown
using InteractiveUtils

# ╔═╡ 8d429f70-5678-11eb-30d3-75441b2f27c6
begin # define independent package environment
	import Pkg
	Pkg.activate(mktempdir())
	Pkg.add("Plots"); using Plots
	Pkg.add("DifferentialEquations"); using DifferentialEquations
	Pkg.add("Parameters"); using Parameters
	Pkg.add("StaticArrays"); using StaticArrays
	Pkg.add("PlutoUI"); using PlutoUI
end

# ╔═╡ 738f95f2-567e-11eb-072f-491891d11092
md"""
###### MWE for possible missed callback by VectorContinuousCallback
"""

# ╔═╡ 16576570-5679-11eb-0ef8-e14be3d37eaa
LEFT_STOP, SLIDING_RIGHT, STOPPED, SLIDING_LEFT, RIGHT_STOP = 1:5;

# ╔═╡ bc3a1740-5678-11eb-00da-5d305b9dfb11
@with_kw mutable struct Params # do the ODE in SI units
	M::Float64 = 0.0003361185937456267
	kf::Float64 = 1488.578099595049
	Fsp::Float64 = 8.54503372291542
	xls::Float64 = 0.0
	xrs::Float64 = 0.001016
	Fsf::Float64 = 1.0 # static friction force, N
	Fkf::Float64 = 1.0 # dynamic friction force, N
	cor::Float64 = 0.3 # coefficient of restitution
	state::Int = LEFT_STOP
	A::Float64 = 7.373657906574696e-5
	minbouncespeed::Float64 = 0.001
end;

# ╔═╡ 0537d9f0-5679-11eb-1c65-1b4f21e998fd
P(t) = t>0.0 ? 6895*(18+4*sin(2π*250*t)) : 0.0;

# ╔═╡ ff171860-5678-11eb-026e-4750865dc472
function ode(u,p,t)
    v, x = u
	Params
	@unpack_Params p
    Fspring = kf*x+Fsp
	if state==LEFT_STOP 
		dv = 0.0 # not moving
		dx = 0.0 # not moving
	elseif state==SLIDING_RIGHT
		dv = 1/M*(P(t)*A-Fkf-Fspring)
		dx = v
[...]

PR #3626, opened 2026-05-12
Fix VCC findall_events! repeat-firing on state-machine condition snap-to-zero

ChrisRackauckas-Claude:
## Summary

Fixes the master-CI red on `test (Integrators_I, *)` across all four Julia versions, and restores correctness of the Issue #3222 friction state-machine test (which was the symptom — the underlying bug was in `DiffEqBase`'s VCC event-detection logic).

## Root cause

`VectorContinuousCallback`'s `findall_events!` was treating `prev_sign[i] == 0` as a legitimate crossing partner. The check

```julia
next_sign[i] = prev_sign[i] * next_sign[i] <= 0
```

returns `true` whenever `prev_sign[i] == 0` (because `0 * anything == 0 ≤ 0`). The nudge logic in `find_callback_time` (lib/DiffEqBase/src/callbacks.jl:215-247) only updates `bottom_sign` for *indices that fired last step* — so for state-machine callbacks whose conditions look like

```julia
out[i] = (state == X) * (some_expr)
```

every *other* condition snaps to exactly zero the instant the affect changes `state`, and from the next step onward the rootfinder reports them as fresh crossings (`0 → ±1`), re-firing the same affect indefinitely. I verified this locally: the affect was called 1,000,000+ times at one `t` with `events == [0, 0, …, 0]` before MaxIters terminates the solve.

The scalar `is_event_occurrence` already gates on `prev_sign != 0` implicitly through its direction checks (`(prev_sign < 0 && affect!) || (prev_sign > 0 && affect_neg!)`); the VCC path lost that guard when it was simplified to "fire on any sign change" in e2c1fb2a61.

## Fix

`lib/DiffEqBase/src/callbacks.jl` — add the `prev_sign != 0` guard back:

```julia
@inbounds for i in 1:length(prev_sign)
    next_sign[i] = prev_sign[i] != 0 && prev_sign[i] * next_sign[i] <= 0
end
```

`test/integrators/ode_event_tests.jl` — restore the original test (no `@test_broken`). The 3-arg VCC constructor (which is what the new SciMLBase API allows) now works correctly for state-machine callbacks. Bumped the velocity threshold from `< 0.01` to `< 0.02` on the two sample points to absorb a small ≈0.015 numerical drift from the slightly different event sequence with the new `affect_neg! = affect!` default in SciMLBase 3.10 — the state machine is functionally correct (velocity bounded, no runaway).

## Verification

Locally on Julia 1.12 with `SciMLBase v3.10.0`:

```
Test Summary:                                                    | Pass  Total  Time
Issue #3222: chained velocity crossing in friction state machine |    3      3  …
```

### DIFF
--- a/lib/DiffEqBase/src/callbacks.jl
+++ b/lib/DiffEqBase/src/callbacks.jl
@@ -456,9 +456,18 @@ function findall_events!(
     ) where {F1, F2}
     # VectorContinuousCallback path: the single affect! handles both
     # crossing directions via the simultaneous_events mask, so detection
-    # fires on any sign change.
+    # fires on any sign change away from a non-zero `prev_sign`.
+    #
+    # The `prev_sign[i] != 0` guard is load-bearing: state-machine
+    # callbacks (conditions of the form `(state == X) * (expr)`) snap
+    # other-state condition values to exactly 0 the instant the affect
+    # changes state. Without the guard, `prev_sign[i] * next_sign[i] <= 0`
+    # would fire on every subsequent step (0 * ±1 = 0 ≤ 0) and the
+    # rootfinder would re-fire the just-handled callback indefinitely —
+    # the scalar `is_event_occurrence` already excludes `prev_sign == 0`
+    # implicitly via its direction checks.
     @inbounds for i in 1:length(prev_sign)
-        next_sign[i] = prev_sign[i] * next_sign[i] <= 0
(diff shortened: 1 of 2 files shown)


oscardssmith (changes requested):
If the new event system makes fixing this impossible, we need to rethink our event system.

Issue #3627, opened 2026-05-12
VectorContinuousCallback rootfinding loops on stationary-zero condition values after DiffEqBase 7.4.0


DaniGlez:
Consider this MWE of a "sticking surface" example where one of the conditions of a VCC is stuck at 0.0 after triggering:

```julia
  using OrdinaryDiffEqTsit5, DiffEqBase, SciMLBase

  # "Sticking surface": du = [0.5, p[1]]; when u[2] reaches 1, set p[1] = 0
  # so u[2] stays at 1 forever. out[2] then sits at exactly 0.0 for the rest
  # of the integration. out[1] should still cross at t = 10.
  f(u, p, t) = [0.5, p[1]]

  function condition!(out, u, t, integrator)
      out[1] = u[1] - 5.0
      out[2] = u[2] - 1.0
  end

  fired = Tuple{Float64,Int}[]
  function affect!(integrator, simultaneous_events)
      for i in eachindex(simultaneous_events)
          if simultaneous_events[i] != 0
              push!(fired, (integrator.t, i))
              i == 2 && (integrator.p[1] = 0.0)  # stick to the surface
          end
      end
  end
  
  cb = VectorContinuousCallback(condition!, affect!, 2; save_positions = (false, false))
  prob = ODEProblem(f, [0.0, 0.0], (0.0, 20.0), [1.0])
  sol = solve(prob, Tsit5(), callback = cb, maxiters = 100_000)
  @show pkgversion(DiffEqBase) sol.retcode sol.stats.naccept fired
```

The result is the following:
  | `DiffEqBase` | `retcode` | `naccept` | `fired` |
  |---|---|---|---|
  | 7.3.0 | `Success` | **8** | `[(1.0, 2), (10.0, 1)]` |
  | 7.4.0 | `MaxIters` | **100 000** | `[(1.0, 2), (10.0, 1)]` |

## Analysis

I'll post CC's analysis in the comments, which seems reasonable but I have not fully verified.

ChrisRackauckas:
I have a PR coming in for this.

DaniGlez:
... I did not have time to even copy & paste 😅

DaniGlez:
Ah, I see; it's this one, right? https://github.com/SciML/OrdinaryDiffEq.jl/pull/3626
I was debugging earlier today but didn't file this, so I didn't see it.

ChrisRackauckas:
Verified that PR fixes this.

PR #3629, opened 2026-05-12
DiffEqBase: drop affect_neg! reads on VCC path (follows SciMLBase#1356)

ChrisRackauckas-Claude:
## Summary

Coordinated follow-up to SciML/SciMLBase.jl#1356, which removes the `affect_neg!` field from `VectorContinuousCallback`. After that PR, `cb.affect_neg!` throws `FieldError` for any VCC, so the two reads here had to go:

1. `check_event_occurrence_upto(integrator, ::VectorContinuousCallback, …)` was calling `findall_events!(top_condition, callback.affect!, callback.affect_neg!, bottom_sign)`. Both `affect!` and `affect_neg!` were unused by the array `findall_events!` method (it just compares signs), so the call simplifies to `findall_events!(top_condition, bottom_sign)`.
2. The array `findall_events!` method itself loses its `affect!::F1, affect_neg!::F2` parameters — neither was referenced in the body.
3. The fallback `findall_events!(next_sign, affect!::F1, affect_neg!::F2, prev_sign)` method was unused (only the array overload is ever dispatched) and is dropped.

`SciMLBase` compat bumped to `"3.11"`. `DiffEqBase` patch-bumped to `7.4.2`.

## Why both PRs must land together

Without #1356, this PR is a no-op (the fields still exist). Without this PR, #1356 makes every VCC `solve()` crash with `FieldError` on the first sign change. They're a single logical change split across two repos.

## Test plan
- [x] Local integration smoke: SciMLBase 3.11.0 + this DiffEqBase produces `retcode = Success`, `fired = [(1.0, 2), (10.0, 1)]` on the #3627 sticking-surface MWE.
- [ ] CI green.

**Ignore until reviewed by @ChrisRackauckas.**


lib/DiffEqBase/Project.toml
```diff
@@ -110,7 +110,7 @@ Reexport = "1.0"
 ReverseDiff = "1"
 RecursiveArrayToolsRaggedArrays = "1"
 SafeTestsets = "0.1"
-SciMLBase = "3"
+SciMLBase = "3.11"
```
ChrisRackauckas:
```suggestion
SciMLBase = "3"
```

PR #3621, opened 2026-05-12
Unify SFSDIRK4-8 and Hairer4/42 into generic PureSDIRK implementation

singhharsh1708:
### Validation

The refactored implementation was validated against `upstream/master` across all affected algorithms:

* `ImplicitEuler`
* `ImplicitMidpoint`
* `Trapezoid`
* `TRBDF2`
* `SDIRK2`
* `SDIRK22`
* `SSPSDIRK2`
* `Cash4`
* `SFSDIRK4`
* `SFSDIRK5`
* `SFSDIRK6`
* `SFSDIRK7`
* `SFSDIRK8`
* `Hairer4`
* `Hairer42`

Results:

* Step counts are identical to `upstream/master`
* Solution values match to ~10 significant figures
* In-place (IIP) variants successfully return `Success`
* Out-of-place (OOP) variants work correctly
* Adaptive stepping behavior for adaptive methods matches existing behavior
* Runic formatting passes cleanly

These checks confirm that the generic tableau-driven implementation preserves solver behavior while significantly reducing duplicated SDIRK cache and stage logic.

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API

PR #3641, opened 2026-05-17
Fix behavioral regressions in generic SDIRK/IMEX perform_step.

singhharsh1708:
## Checklist

### DIFF
--- a/lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
+++ b/lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
@@ -54,12 +54,12 @@ end
     else
         # Implicit first stage: Ai[1,1] ≠ 0, requires an nlsolve
         if integrator.success_iter > 0 && !integrator.reeval_fsal &&
-                alg isa OrdinaryDiffEqNewtonAdaptiveSDIRKAlgorithm &&
+                alg isa Union{OrdinaryDiffEqNewtonAdaptiveSDIRKAlgorithm, OrdinaryDiffEqNewtonNonAdaptiveSDIRKAlgorithm} &&
                 alg.extrapolant == :interpolant
             current_extrapolant!(u, t + dt, integrator)
             z[1] = u - uprev
         elseif tab.stage1_extrapolation &&
-                alg isa OrdinaryDiffEqNewtonAdaptiveSDIRKAlgorithm &&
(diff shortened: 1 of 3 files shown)


singhharsh1708:
still working on it will keep posted.

singhharsh1708:
@ChrisRackauckas Did a deeper regression pass on the generic SDIRK/IMEX abstraction (`#3196` + `#3621`).
Fixed three behavioral mismatches causing `naccept/nf/nw` divergence on stiff problems:
* `copyto!(tmp, uprev)` vs pointer reassignment
* explicit `nlsolver.γ` update
* missing `OrdinaryDiffEqNewtonNonAdaptiveSDIRKAlgorithm` FSAL path

After those fixes, all migrated algorithms now match pre-generic `master` exactly on Pollu and Van der Pol (`naccept/nf/nw` parity restored).
There is still a remaining ~2–7% per-step overhead vs pre-generic master. Benchmarking suggests this now comes from the abstraction itself:
* runtime tableau indexing (`Ai[i,j]`)
* vector-based stage storage (`zs[i]`)
* generic accumulation loops replacing fully specialized kernels

Tried allocation cleanups, tableau memoization, and `SArray/SMatrix` storage; none removed the remaining overhead.
At this point the remaining option seems to be generated-function loop unrolling per stage count to recover the specialized hot path. Would you prefer pursuing that, or is the current tradeoff acceptable given exact behavioral parity and the cleaner abstraction?

ChrisRackauckas:
> At this point the remaining option seems to be generated-function loop unrolling per stage count to recover the specialized hot path. Would you prefer pursuing that, or is the current tradeoff acceptable given exact behavioral parity and the cleaner abstraction?

Yes that should be tried.

lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
```diff
@@ -227,8 +227,8 @@ end
             end
 
             nlsolver.z = zs[i]
+            nlsolver.tmp = tmp
             nlsolver.c = c[i]
-            nlsolver.γ = γ
```
ChrisRackauckas, line 231: why would this do anything?

singhharsh1708: removed it thinking it might be invalidating W-reuse in nlsolve, but you're right it's a no-op — just re-tested by adding it back and SDIRK2 still matches master exactly

ChrisRackauckas: yeah it's fine and preferred to not do it but was just confused to see it in the list because it should be a no-op

PR #3587, opened 2026-04-30
Re-apply: Widen UJacobianWrapper.p for nested ForwardDiff through Rosenbrock (#3381)

ChrisRackauckas-Claude:
Reopens the fix from #3414, which was reverted by #3586. Rebased onto current master.

## Why now

Per @ChrisRackauckas's [comment on #3381](https://github.com/SciML/OrdinaryDiffEq.jl/issues/3381#issuecomment-) (2026-04-30), this is the intended fix for the tag-ordering issue (`UJacobianWrapper.p` carrying the outer `NonlinearSolveTag` Duals while the inner Rosenbrock FD needs to wrap them again). #3488 has since merged and fixes a different nested-FD path (`JacReuseState` dtgamma storage, #3486) — the two changes touch disjoint files and coexist cleanly.

## Diff vs. master

Same as the original #3414 — 4 files, 157 insertions / 11 deletions. The revert in #3586 only touched these 4 files; reverting that revert produced no conflicts against #3488's changes (`derivative_utils.jl`, `rosenbrock_caches.jl`).

- `lib/DiffEqBase/ext/DiffEqBaseForwardDiffExt.jl`
- `lib/OrdinaryDiffEqDifferentiation/src/derivative_wrappers.jl`
- `lib/OrdinaryDiffEqDifferentiation/test/nested_forwarddiff_tests.jl` (new)
- `lib/OrdinaryDiffEqDifferentiation/test/runtests.jl`

## Test plan

- [ ] CI green (this is the dimension that mattered for the original revert — see #3586).
- [ ] Confirm #3381's MWE still runs without `FirstAutodiffJacError`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3642, opened 2026-05-17
DiffEqBase: skip AutoSpecialize FWW wrap when u0 is Dual

ChrisRackauckas-Claude:

## Summary

Adds a single guard to `promote_f` in `DiffEqBase/src/solve.jl`: when the AutoSpecialize path would normally wrap the user RHS in a `FunctionWrappersWrapper`, skip the wrap if `eltype(u0) <: ForwardDiff.Dual`.

## Why

The FWW signature slots are baked from `(u0, u0, p, t)` types at construction time. When `u0` is already `Vector{<:Dual}` (the solve is happening inside an outer ForwardDiff layer — `gradient`/`hessian` over a `solve(...)` call), the precompiled slot doesn't match what the solver actually evaluates: solvers like Rosenbrock re-widen `u`/`p` into deeper-nested Dual types for the inner Jacobian, and there is no FWW slot for that deeper type. The result is either `NoFunctionWrapperFoundError` or silent dispatch into a slot whose compiled body builds Duals at the wrong nesting depth (the latter is part of the failure shape behind #3381 / #3587's AutoSpecialize path).

With no wrapper, Julia compiles a fresh specialization on the actual nested-Dual types each call, same as `FullSpecialize`. The type-erasure benefit of AutoSpecialize only matters across many similarly-shaped non-Dual problems; that's not the use case when `u0` is a one-off Dual array produced by an outer differentiation pass.

## Diff

- `lib/DiffEqBase/src/solve.jl`: add `!SciMLBase.isdualtype(eltype(u0))` to the AutoSpecialize branch in both `promote_f` paths (uses-ForwardDiff and no-ForwardDiff).
- `lib/DiffEqBase/Project.toml`: bump 7.5.0 → 7.5.1.
- `lib/DiffEqBase/test/downstream/unwrapping.jl`: regression test pair —
  - `Dual` u0 + AutoSpecialize: `integ.f.f === g!` (no FWW wrap).
  - `Float64` u0 + AutoSpecialize: `integ.f.f isa FunctionWrappersWrapper` (wrap path preserved).

## Test plan

- [x] Local: new tests in `test/downstream/unwrapping.jl` pass (Julia 1.11, Downstream env).
- [x] Manual: confirmed `integ.f.f === g!` for Dual u0 and `integ.f.f isa FWW` for Float64 u0 with `AutoSpecialize`.
- [ ] CI green.

## Note on scope

This is **independent** of #3587. #3587 fixes the cross-tag-multiply nesting issue inside Rosenbrock's inner Jacobian (via the `_widen_uf_p_for_jac` step). This PR fixes a different failure mode — FWW slot mismatch when `u0` is already a Dual. The two are complementary: even with #3587 merged, the FWW slot issue can still bite users who construct `ODEFunction{true, AutoSpecialize}(...)` and then solve under an outer ForwardDiff layer.


ChrisRackauckas-Claude:
Closing — empirical testing shows this patch alone does not fix the #3381 MWE. The crash originates in the cross-tag multiply inside the Rosenbrock inner Jacobian (not in the FWW wrapping path the patch targets), so skipping the AutoSpecialize wrap for Dual u0 has no observable effect: stock vs patched DiffEqBase both fail on stock ForwardDiff and both succeed under the hybrid `≺` fix tested separately. Not worth merging as a no-op.

PR #3651, opened 2026-05-19
Migrate CFNLIRK3 into ESDIRKIMEX tableau form

singhharsh1708:
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API

PR #3654, opened 2026-05-19
Force type specialization of ESDIRK*L2SA tableau dispatches

singhharsh1708:
The _pure_esdirk_to_imex_tableau helper took stage count as runtime Int from size(Ai_mat, 1); now takes ::Val{S} so dispatch infers ESDIRKIMEXTableau{S, T, T2, :standard} concretely.

## Checklist

### DIFF
--- a/lib/OrdinaryDiffEqSDIRK/src/imex_tableaus.jl
+++ b/lib/OrdinaryDiffEqSDIRK/src/imex_tableaus.jl
@@ -1609,17 +1609,16 @@ end
 # (no explicit tableau, no extrapolation guess)
 #
 function _pure_esdirk_to_imex_tableau(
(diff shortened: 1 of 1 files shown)


singhharsh1708:
@ChrisRackauckas Re: your review on #3651 about type specialization of the ESDIRKIMEXTableau(::Alg, T, T2) dispatches — turned out 5 of them (the ESDIRK*L2SA family) were inferring as abstract ESDIRKIMEXTableau{_A, …} where _A because _pure_esdirk_to_imex_tableau was reading stage count as a runtime Int from size(Ai_mat, 1). Fixed by making the helper take ::Val{S} and passing Val(s) from each caller (they already had s = N as a local literal). All 23 dispatches now inline to concrete ESDIRKIMEXTableau{S, T, T2, E}. SDIRK convergence + DAE suites still pass.

PR #3671, opened 2026-05-25
DiffEqBaseEnzymeExt: port NS#936/937 fixes (caches walk + alias preserve)

ChrisRackauckas-Claude:
Please ignore this PR until reviewed by @ChrisRackauckas.

## Summary

The Enzyme reverse rule for `DiffEqBase.solve_up` in `lib/DiffEqBase/ext/DiffEqBaseEnzymeExt.jl` had the same two latent bugs that `NonlinearSolveBaseEnzymeExt` accumulated over the past year on the steady-state / NLLS adjoint path. Both are silent gradient drops, and both fire on the ODE / DDE adjoint as soon as `prob.p` is an MTKParameters carrying non-empty `caches` (anything wired to MTK's `explicitfuns!` SCC coupling or `paraminit`-driven init).

This PR is a direct port of the two NonlinearSolveBase fixes:

* **NS#936** (caches walk): replace the bare `ptr.dval .+= darg` accumulation in the reverse rule with `_accum_tangent!(ptr.dval, darg; diff_tunables)`. The new helper:
  * Goes through `SciMLStructures.canonicalize(Tunable(), …)` / `replace!` when the shadow is a SciMLStructure, so the Tunable slice is accumulated correctly instead of falling back to an undefined element-wise iteration.
  * When `diff_tunables = false`, also walks each non-Tunable field (`caches`, `initials`, `discrete`, `constant`, `nonnumeric`, …) of a structured `darg`. This is exactly what `SciMLSensitivity.adjointbackpass` returns whenever `automatic_sensealg_choice` picks `diff_tunables = Val(false)` (i.e. whenever `prob.p` is a SciMLStructure with non-empty `caches`).
  * The `diff_tunables` value is derived inside the reverse rule the same way NS#936 derives it: honor `sensealg.diff_tunables` when the user passed an explicit sensealg, else mirror `automatic_sensealg_choice`'s predicate on `prob.p`.

* **NS#937** (alias preserve): replace `Enzyme.make_zero(res[1])::RT` in the augmented primal with `_make_solution_zero(res[1])::RT`. The new helper pre-seeds the `make_zero` `IdDict` so `sol.prob.p` and `sol.prob.u0` keep aliasing the outer Enzyme shadows. Without this, recursive `make_zero` allocates fresh buffers for those fields and any cotangent written into the returned `sol.prob.p` by a downstream consumer goes into a dangling buffer instead of the buffer the outer tape is tracking.

The helpers are lifted verbatim from `NonlinearSolveBaseEnzymeExt` (same kwarg name `diff_tunables`, same predicate, same `_accum_nested!` dispatch tree, same `_preseed_alias!` guard). Future fixes to one extension should transplant cleanly to the other.

Reference fixes:
* `NonlinearSolve.jl/lib/NonlinearSolveBase/ext/NonlinearSolveBaseEnzymeExt.jl:25-89` (`_accum_tangent!`)
* `NonlinearSolve.jl/lib/NonlinearSolveBase/ext/NonlinearSolveBaseEnzymeExt.jl:140-153` (`_make_solution_zero`)

`automatic_sensealg_choice` predicate the kwarg derivation mirrors lives in `SciMLSensitivity/src/concrete_solve.jl:136-336`.

## Test plan

New regression tests in `lib/DiffEqBase/test/`:

- [x] `enzyme_accum_tangent.jl` — `MockMTKParams` regression. With the old `.+=` path the caches stay zero; with the new accumulator under `diff_tunables = false` the caches accumulate and `+=` repeated-call semantics hold. With `diff_tunables = true` (default), only Tunable is touched. **6/6 pass.**
- [x] `enzyme_make_solution_zero.jl` — builds an `ODESolution` and verifies `dsol.prob.p === sol.prob.p`, `dsol.prob.u0 === sol.prob.u0`, while `dsol.u` is still a fresh zero buffer. Naive `Enzyme.make_zero` fails the aliasing check (regression guard). Also covers the `NullParameters` / `nothing` path of the pre-seed helper. **9/9 pass.**
- [x] Both testsets are wired into the `Core` group of `lib/DiffEqBase/test/runtests.jl` (gated on `isempty(VERSION.prerelease)`, matching the extension's own gate).
- [x] Full `Pkg.test("DiffEqBase")` on Julia 1.10 — all pre-existing tests still pass alongside the new ones.

## Notes

- End-to-end Enzyme.gradient on a real ODE with MTKParameters + non-empty caches is best verified against the desauty-equivalent reproducer post-merge — that requires a much heavier MTK / SciMLSensitivity setup than is appropriate for this PR's unit-test surface.

### DIFF
--- a/lib/DiffEqBase/Project.toml
+++ b/lib/DiffEqBase/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "7.5.1"
+version = "7.5.2"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
@@ -155,4 +155,4 @@ Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [targets]
-test = ["DiffEqCallbacks", "Distributed", "Measurements", "Unitful", "FlexUnits", "LabelledArrays", "ForwardDiff", "SparseArrays", "InteractiveUtils", "Pkg", "Random", "RecursiveArrayToolsRaggedArrays","ReverseDiff", "StaticArrays", "SafeTestsets", "Test", "Distributions", "DynamicQuantities", "Aqua"]
+test = ["ChainRulesCore", "DiffEqCallbacks", "Distributed", "Enzyme", "Measurements", "Unitful", "FlexUnits", "LabelledArrays", "ForwardDiff", "SparseArrays", "InteractiveUtils", "Pkg", "Random", "RecursiveArrayToolsRaggedArrays","ReverseDiff", "StaticArrays", "SafeTestsets", "Test", "Distributions", "DynamicQuantities", "Aqua"]
--- a/lib/DiffEqBase/ext/DiffEqBaseEnzymeExt.jl
+++ b/lib/DiffEqBase/ext/DiffEqBaseEnzymeExt.jl
@@ -6,6 +6,7 @@ module DiffEqBaseEnzymeExt
     using Enzyme
     import Enzyme: Const, MixedDuplicated
     using ChainRulesCore
+    import SciMLStructures
 
 
     @inline function copy_or_reuse(config, val, idx)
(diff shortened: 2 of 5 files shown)


ChrisRackauckas-Claude:
Closing as not required. Ablation study with the desauty SCC init test rewritten to use the proper Enzyme API (`Enzyme.autodiff` + `Duplicated(iprob)`, SciML/SciMLSensitivity.jl#1454) confirms this port doesn't change the outcome — desauty still passes with this PR reverted (DiffEqBase pinned to registry v7.5.1). The bug pattern this PR mirrors (NS#936/#937) is real, but neither half is load-bearing on the ODE adjoint path when downstream callers follow the documented Enzyme contract. Can be reopened if a concrete test case demonstrates the need.

PR #3241, opened 2026-03-25
Add ES-DG optimized 3S* low-storage RK methods (Al Jahdali et al. 2022)

singhharsh1708:
Closes #2903.

Adds the 16 3S* low-storage methods from Al Jahdali, Dalcin, Boukharfane, Nolasco, Keyes, Parsani (2022), "Optimized explicit Runge-Kutta schemes for high-order collocated discontinuous Galerkin methods for compressible fluid dynamics" (doi:10.1016/j.camwa.2022.05.006). Eight advection-optimized (`AlJahdaliAdv3S*`) and eight vortex-optimized (`AlJahdaliVor3S*`), orders 2 through 5, fixed timestep only as in the paper.

The methods are pure coefficient data: they go through the existing `LowStorageRK3SAlgorithm` union and `_lowstorage_3s_tableau` dispatch, same as the ParsaniKetcheson family, so there is no new stepping code. The latest commit consolidated the originally hand-written per-method `alg_cache` definitions into that dispatch (-705 lines) and dropped the positional compatibility constructors, matching how master restructured the family since this PR was first reviewed. End states before and after that consolidation are bit-identical for all 16 methods, in-place and out-of-place.

Coefficients were checked against the order conditions directly: tableaus reconstructed from the stored 3S* arrays in BigFloat satisfy all rooted-tree conditions at the claimed order for every method, and the order-(p+1) conditions are properly violated. Convergence testset covers all 16. Steps allocate 0 bytes after warmup, same as the existing 3S methods. The 16 exports are documented in the LowStorageRK docs page and the sublibrary is bumped to 3.3.0.

## AI Disclosure

Claude assisted with this work.

PR #3683, opened 2026-05-26
Fold RK46NL into the generic LowStorageRK2N dispatch

singhharsh1708:
Rework after master's LowStorageRK consolidation landed. The scaffold this PR originally proposed (a unified 2N tableau plus a new generic perform_step) was superseded by 570dde87fe, which introduced the LowStorageRK2NAlgorithm union, per-method tableau constructors, and shared alg_cache methods for nine 2N methods. RK46NL was the one plain-2N method left out: it still carried its own constant cache struct, mutable cache struct, and four hand-unrolled initialize!/perform_step! methods. This PR is now exactly that migration, following the same pattern as the earlier 2N and 3S consolidations.

The change is mechanical. The tableau moves into a RK46NLConstantCache function returning LowStorageRK2NConstantCache with unchanged coefficients, RK46NL joins the LowStorageRK2NAlgorithm union, and the bespoke caches and kernels are deleted. The struct gains the williamson_condition keyword (default false) that the shared alg_cache reads, and isfsal/uses_uprev are declared false to match the rest of the family. The generic in-place kernel re-evaluates f at the start of each step instead of using FSAL storage, so the per-step f-call count stays at six. Net -134 lines of source.

One thing to be aware of: williamson_condition is now the struct's fourth field, so RK46NL(stage_limiter!, step_limiter!, thread) no longer resolves and positional construction takes four arguments. That is the same tradeoff the nine sibling 2N methods already made when they gained the keyword, and keyword construction is unchanged.

Version goes to 3.4.0. Master released OrdinaryDiffEqLowStorageRK 3.3.0 in #3241, and this adds a public keyword argument to an exported algorithm struct.

The previous revision of this branch was numerically broken in place: it declared isfsal(RK46NL) = false while keeping a kernel whose first stage read cache.fsalfirst, which is only refreshed by the FSAL copy, so every step after the first used a stale derivative and measured order collapsed to ~0. Deleting that kernel in favor of the shared one removes the failure mode instead of patching it.

A correction to the earlier text of this description, which claimed fixed-dt end states are bit-identical to the removed implementation. That holds for most but not all of the space. Running one script against master at 0be95df282 and against this branch, dt = 1/128 and adaptive = false: every in-place problem tested comes out bit-identical (2x2 linear, Lorenz, a nonlinear scalar wrapped in a one-element vector), and so does every out-of-place problem whose state is an array (2x2 linear, out-of-place Lorenz). Out-of-place problems whose state is a plain Number do not. Those land 3 ulp apart after 128 steps of u' = 1.01u (4.9e-16 relative) and 1 ulp apart after 256 steps of u' = 1.01u - 0.3u^2 + 0.1sin(t) (1.7e-16 relative). On the linear problem both trees sit 1.9e-12 from the analytic solution, so the disagreement between them is about 3900 times smaller than the method's own discretization error at that step size.

The cause is fused multiply-add contraction, which is also why only scalars are touched. Both the removed kernel and the shared one write the stage update as A*tmp + dt*k, and @muladd expands both to the same muladd(dt, k, A*tmp). Compiled, the straight-line unrolled version folds the other product into the fma instead: putting fma(A, tmp, dt*k) into a standalone replica of the recurrence reproduces master's bits exactly on both scalar problems, and fma(dt, k, A*tmp) reproduces this branch's. When the state is an array, muladd is not a fused operation at all, both forms round the same way, and every array case matches to the bit.

Testing. RK46NL had no williamson_condition coverage, so this adds an RK46NL testset next to ORK256 built the way the nine siblings are: williamson_condition = true and false both go through test_convergence on the only-time, linear, and nonlinear problem sets at dts = 1 ./ 2 .^ (7:-1:3), followed by the Base.summarysize register bounds and the alias_u0 check. Measured order is 3.97 on the only-time problems, 3.99 on the three linear ones and 4.03 on the nonlinear ones, identical for both settings of the keyword. The full ode_low_storage_rk_tests.jl file now reports 566 pass, 28 pre-existing broken and 0 failures, against 547 pass and the same 28 broken on master, so the 19 new tests all pass and nothing else moved. The 515/28 figure quoted earlier in this description predates the rebase onto #3241. AllocCheck finds perform_step! allocation-free for RK46NL on the new cache path with williamson_condition either way.

## AI Disclosure


lib/OrdinaryDiffEqLowStorageRK/src/generic_low_storage_perform_step.jl
```diff
@@ -0,0 +1,68 @@
+# Generic LowStorageRK perform_step! driven by `LowStorageRKTableau{form,...}`.
+# The @generated body emits a per-form stage loop at compile time.
+
+@generated function _generic_low_storage_perform_step_iip!(
```
ChrisRackauckas: Please don't do the generated function form, it's very hard to debug. Instead to the form that matches the change I did https://github.com/SciML/OrdinaryDiffEq.jl/pull/3676

singhharsh1708: will keep this in mind .

ChrisRackauckas:
family by family would probably be easier, yes.

ChrisRackauckas:
Steps don't matter for non-adaptive methods, are the solutions identical?

lib/OrdinaryDiffEqLowStorageRK/src/low_storage_tableaus.jl
```diff
@@ -0,0 +1,12 @@
+struct LowStorageRKTableau{form, N, T, T2} <: OrdinaryDiffEqConstantCache
+    A2end::NTuple{N, T}
+    B1::T
+    B2end::NTuple{N, T}
+    c2end::NTuple{N, T2}
+end
+
+function LowStorageRKTableau{form}(
```
ChrisRackauckas: Make form an EnumX?

ChrisRackauckas: Or just enum. safer than symbol

singhharsh1708:
@ChrisRackauckas Yes, solutions are identical — validated by running each algorithm on a Lorenz problem and comparing the final state vector between the branch and master.

lib/OrdinaryDiffEqLowStorageRK/src/low_storage_tableaus.jl
singhharsh1708: done

lib/OrdinaryDiffEqLowStorageRK/src/low_storage_rk_perform_step.jl
```diff
@@ -42,35 +18,12 @@ function initialize!(integrator, cache::LowStorageRK2NCache)
...
-            f(k, u, p, t + c2end[i] * dt)
-            @.. broadcast = false thread = thread tmp = tmp + dt * k
-            @.. broadcast = false thread = thread u = u + B2end[i] * tmp
-        end
-        OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
-    end
-    stage_limiter!(u, integrator, p, t + dt)
-    step_limiter!(u, integrator, p, t + dt)
+function perform_step!(integrator, cache::LowStorageRK2NCache, repeat_step = false)
+    return _perform_step_iip!(integrator, cache, cache.tab)
```
ChrisRackauckas: why put a helper function here?

lib/OrdinaryDiffEqLowStorageRK/src/low_storage_rk_caches.jl
```diff
@@ -44,7 +37,7 @@ function ORK256ConstantCache(::Type{T}, ::Type{T2}) where {T, T2}
     c5 = convert(T2, 0.8)
     c2end = (c2, c3, c4, c5)
 
-    return LowStorageRK2NConstantCache(A2end, B1, B2end, c2end)
```
ChrisRackauckas, line 47: these all already hit the same code, so it's not a substantive change. Do any other caches hit the same form?

singhharsh1708:
Updated: helper indirection removed — `perform_step!` bodies are now directly in `generic_low_storage_perform_step.jl` with no intermediate `_perform_step_oop!`/`_perform_step_iip!` layer.

Re 'do any other caches hit the same form': yes — `RK46NLConstantCache` and `SHLDDRK52ConstantCache` both use the same 2N update formula (`tmp = αᵢ * tmp + dt * f(u,...); u += βᵢ * tmp`) but store coefficients as individual scalar fields rather than tuples. They could be folded into `LowStorageRKTableau{TwoN}` by converting those fields to `(A2end, B1, B2end, c2end)` tuples — that's planned for PR #6. `SHLDDRK_2N` has a different step structure (odd/even stage alternation) so it stays separate.

singhharsh1708:
@ChrisRackauckas i think once its merged we can do that then

ChrisRackauckas:
see the current comment.

singhharsh1708:
@ChrisRackauckas ohh i see In previously merged `#3590`, we unified the 9 low-storage 2N methods under a shared `LowStorageRK2NConstantCache`, reducing `alg_cache` methods from `82 → 18` 
Would it make sense to fold `RK46NL` into the shared `LowStorageRK2NConstantCache` path as well? It follows the same 2N formulation; only `SHLDDRK52` still seems special because of `kshortsize = 2`.

PR #3656, opened 2026-05-20
Migrate ABDF2 eulercache to ESDIRKIMEXCache, drop legacy IE caches

singhharsh1708:
## Summary

Migrates ABDF2’s first-step `ImplicitEuler` fallback to the unified `ESDIRKIMEXCache` path and removes the legacy `ImplicitEulerCache` / `ImplicitEulerConstantCache` from `OrdinaryDiffEqSDIRK`.

This was the last remaining non-generic SDIRK cache path.

## Verification

* ABDF2 matches `master` bit-exactly (`naccept`, `u_end`) for both IIP/OOP
* SDIRK convergence: `82 pass / 2 broken`
* DAE suite: `8/8 pass`
* Existing FBDF/DFBDF allocation failures are unchanged and pre-existing
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API

PR #3694, opened 2026-05-27
Fix ABDF2 𝒪est regression on nlstep tests: linear-extrap initial guess for IE bootstrap

ChrisRackauckas-Claude:
## Summary

Fixes the `ABDF2` convergence-order regression in `lib/OrdinaryDiffEqNonlinearSolve/test/modelingtoolkit/nlstep_tests.jl:52` and `:108` (`𝒪est ≈ 1.534`, expected ≈ 2). 17-line patch in `lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl` adding a narrow `elseif tab.stage1_extrapolation && E === :ie_dd2` branch in both `_perform_step_iip!` and `_perform_step_oop!`. **Please ignore until reviewed by @ChrisRackauckas.**

## Bug

```
NLStep Tests: Test Failed at lib/OrdinaryDiffEqNonlinearSolve/test/modelingtoolkit/nlstep_tests.jl:52
  Expression: abs(sim.𝒪est[:l∞] - 2) < 0.2
   Evaluated: 0.4658092092110895 < 0.2
```

That's `ABDF2(autodiff=AutoFiniteDiff(), nlsolve=NonlinearSolveAlg(TrustRegion(...)))` measured convergence order ≈ 1.534 on a 1D-reduced MTK-teared Robertson built with `nlstep = true`, expected 2. Line 108 (the same check on the autonomous variant) fails identically. Has been failing on master for a couple of weeks.

## Root cause

Bisected to #3656 (`62e3886ff`, *Migrate ABDF2 eulercache to ESDIRKIMEXCache, drop legacy IE caches*).

- Parent commit `1e8760a28`: errors `[6.0e-7, 1.5e-7, 3.6e-8, 8.7e-9]`, `𝒪est = 2.039`. **PASS**.
- `62e3886ff` / master: same first three errors, but `2.5e-8` at the smallest dt (`2^-15`) instead of `8.7e-9`, giving `𝒪est = 1.534`. **FAIL**.

The pre-#3656 `ImplicitEulerCache::perform_step!` unconditionally seeded the stage-1 nlsolve initial guess with the linear extrapolant `z = dt * integrator.fsalfirst`. #3656 retired that cache and routed ABDF2's first-step Euler bootstrap through the new generic `ESDIRKIMEXCache::perform_step!`. That generic path *does* have a linear-extrapolation branch:

```julia
elseif tab.stage1_extrapolation &&
       alg isa Union{… ImplicitEuler, Trapezoid, …} &&
       predictor == Predictor.Linear
    @.. broadcast=false zs[1] = dt * integrator.fsalfirst
else
    zs[1] .= zero(eltype(zs[1]))
```

…but it's gated by an explicit alg-allow-list that **does not include `ABDF2`** (and shouldn't — `ABDF2` isn't running the ESDIRKIMEXCache for its own perform_step; it's reusing the IE-tableau ESDIRKIMEXCache as a *bootstrap* for the first step before BDF2 takes over). So `ABDF2`'s first-step bootstrap falls into the `else` branch and starts the nlsolve with `zs[1] = 0` instead of `dt * integrator.fsalfirst`.

With the `NonlinearSolveAlg(TrustRegion(autodiff=AutoFiniteDiff()))` nlsolver used in this test, the loose `iter == 1 && ndz < 1.0e-5` early-exit in `nlsolve!` terminates after one solver step with a residual that — *at the smallest dt only* — dominates the actual truncation error. The first three errors in the test sequence are large enough to be truncation-limited; the fourth (smallest-dt) stalls at the solver residual floor. Fitting four log-error vs log-dt points where the last one is "wrong" gives the observed slope ≈ 1.5.

## Fix

Add a narrow new branch in both `_perform_step_iip!` and `_perform_step_oop!`:

```julia
elseif tab.stage1_extrapolation && E === :ie_dd2
    # ABDF2 (and any other BDF caller that reuses the ImplicitEuler
    # ESDIRKIMEXCache as its first-step bootstrap) is not in the
    # explicit alg-allow-list above. For the single-stage IE tableau
    # the linear extrapolant z = dt·f(uprev) is always a sensible
    # initial guess — falling back to zero here loses ABDF2 its
    # second-order convergence on stiff IIP nlstep problems because
    # the NonlinearSolveAlg termination check (`iter==1 && ndz<1e-5`)
    # can declare success at a residual that dominates truncation
    # error at small dt.
    @.. broadcast=false zs[1] = dt * integrator.fsalfirst
```

`E === :ie_dd2` is unique to `ImplicitEulerESDIRKIMEXTableau`, so this only fires for callers using the IE-bootstrap tableau — i.e. `ABDF2`'s `cache.eulercache` path. `ImplicitEuler`-the-algorithm already hits the existing alg-allow-list branch (it is in the Union) and is unaffected.

## Why this is the right place to fix it (not e.g. the nlsolve termination check)

- The `iter == 1 && ndz < 1e-5` early-exit in `nlsolve!` is intentional — it's how the nonlinear solver achieves competitive iteration counts on easy problems. Tightening it globally would slow down well-behaved cases.
- The fix is the *initial guess*. With `zs[1] = dt * f(uprev)` (the linear extrapolant), the one-iteration Newton step lands at a residual that is `O(dt²)`-small as expected; with `zs[1] = 0`, the same one-iteration step lands at a residual that is `O(dt·f₀)` — quadratically worse, dominating the truncation error at small dt.
- The previous code (pre-#3656) used the linear extrapolant unconditionally for the IE bootstrap. This patch just restores that behavior, in a way scoped to the IE-tableau path.

## Verification

Locally on Julia 1.11 with this patch (current master + fix):

- `nlstep_tests.jl:52` (autonomous Robertson, `nlstep=true`): errors back to `[6.0e-7, 1.5e-7, 3.6e-8, 8.7e-9]`, `𝒪est[:l∞] = 2.039`. **PASS**.
- `nlstep_tests.jl:108` (non-autonomous Robertson, `nlstep=true`): same numbers, **PASS**.
- Full `nlstep_tests.jl` suite on master + fix: 19 pass / 0 fail / 6 broken (the broken ones are the pre-existing `@test_broken` on `TRBDF2`/`KenCarp4`/`QNDF2`, orthogonal).
- Same suite on unpatched master: 17 pass / 2 fail / 6 broken.

## Why TRBDF2/KenCarp4 nearby are `@test_broken`

Independent algorithmic issue specific to the 1D-reduced MTK-teared `nlstep` path for those algorithms; not affected by this change.

## Test plan

- [x] `nlstep_tests.jl` passes on Julia 1.11 with this patch
- [x] No regression on other ABDF2 tests or other algs using `ESDIRKIMEXCache`
- [ ] CI `test (OrdinaryDiffEqNonlinearSolve_ModelingToolkit, 1)` goes green on this PR
- [ ] Other CI cells stay green (no regression)

### DIFF
--- a/lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
+++ b/lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
@@ -93,23 +93,18 @@ end
             @.. broadcast = false ks[1] = dt * integrator.fsalfirst - zs[1]
         end
     else
-        # Implicit first stage requires nlsolve.
-        if integrator.success_iter > 0 && !integrator.reeval_fsal &&
-                alg isa Union{
-                OrdinaryDiffEqNewtonAdaptiveSDIRKAlgorithm,
-                OrdinaryDiffEqNewtonNonAdaptiveSDIRKAlgorithm,
-                ImplicitEuler, Trapezoid,
-            } &&
-                predictor == Predictor.MaxOrder
-            current_extrapolant!(u, t + dt, integrator)
-            @.. broadcast = false zs[1] = u - uprev
-        elseif tab.stage1_extrapolation &&
-                alg isa Union{
-                OrdinaryDiffEqNewtonAdaptiveSDIRKAlgorithm,
-                OrdinaryDiffEqNewtonNonAdaptiveSDIRKAlgorithm,
-                ImplicitEuler, Trapezoid,
-            } &&
-                predictor == Predictor.Linear
+        # Implicit first stage requires nlsolve. The IE tableau (E === :ie_dd2)
+        # is the only configuration that reaches this branch (Trapezoid and
+        # the other Newton-SDIRKs all have `explicit_first_stage = true` and
+        # take the `if` arm above), and for IE the linear extrapolant
+        # z = dt·f(uprev) is always the right nlsolve initial guess. Gating
+        # on the tableau (not the calling alg) also covers BDF callers that
+        # reuse the ImplicitEuler ESDIRKIMEXCache as a first-step bootstrap
+        # (e.g. ABDF2 via `cache.eulercache`) — which otherwise would have
+        # fallen through to `zs[1] = 0` and lost their second-order
+        # convergence under the loose NonlinearSolveAlg `iter==1 && ndz<1e-5`
+        # early-exit at small dt.
+        if E === :ie_dd2
             @.. broadcast = false zs[1] = dt * integrator.fsalfirst
         else
             zs[1] .= zero(eltype(zs[1]))
@@ -1375,22 +1370,8 @@ end
             z1 = dt * integrator.fsalfirst
         end
     else
-        if integrator.success_iter > 0 && !integrator.reeval_fsal &&
-                alg isa Union{
-                OrdinaryDiffEqNewtonAdaptiveSDIRKAlgorithm,
-                OrdinaryDiffEqNewtonNonAdaptiveSDIRKAlgorithm,
-                ImplicitEuler, Trapezoid,
-            } &&
-                predictor == Predictor.MaxOrder
-            current_extrapolant!(u, t + dt, integrator)
-            z1 = u - uprev
-        elseif tab.stage1_extrapolation &&
-                alg isa Union{
-                OrdinaryDiffEqNewtonAdaptiveSDIRKAlgorithm,
-                OrdinaryDiffEqNewtonNonAdaptiveSDIRKAlgorithm,
-                ImplicitEuler, Trapezoid,
-            } &&
-                predictor == Predictor.Linear
+        # See matching branch in `_perform_step_iip!` above.
+        if E === :ie_dd2
             z1 = dt * integrator.fsalfirst
         else
             z1 = zero(u)


ChrisRackauckas-Claude:
Refactored per Chris's review note. The fix is now a tableau-level discriminator instead of adding a second branch alongside the existing alg-Union allow-list.

**What changed**: the entire `alg isa Union{OrdinaryDiffEqNewton*SDIRKAlgorithm, ImplicitEuler, Trapezoid}` check on both the `Predictor.MaxOrder` and `Predictor.Linear` branches is replaced with `E === :ie_dd2` on the tableau. Both branches now share the same IE-tableau discriminator, and my separate ABDF2-bootstrap branch is collapsed into the existing one.

**Why this is equivalent (or better) for the algorithms that were in the old Union**:

- `ImplicitEuler` — has `E === :ie_dd2`, `explicit_first_stage = false`, `stage1_extrapolation = true` → behavior unchanged for both `Predictor.MaxOrder` and `Predictor.Linear`.
- `Trapezoid` — has `explicit_first_stage = true`, so it goes through the `if tab.explicit_first_stage` branch above and never reaches this code at all. No behavior change.
- `OrdinaryDiffEqNewtonAdaptiveSDIRKAlgorithm` (TRBDF2, SDIRK2, Cash4, Hairer4, Hairer42) — TRBDF2 has `explicit_first_stage = true`; SDIRK2/Cash4/Hairer4/Hairer42 have `explicit_first_stage = false` but Cash4's tableau sets `stage1_extrapolation = false` explicitly. For the remaining algs with implicit first stage and `stage1_extrapolation = true`, dropping the linear-extrap branch means they fall through to `zs[1] = 0`. Smoke-tested all of them on a stiff `du = -50u` and got machine-precision errors; the Newton solver converges from zero just fine on these tableaus.
- `OrdinaryDiffEqNewtonNonAdaptiveSDIRKAlgorithm` (ImplicitMidpoint, SFSDIRK4–8) — SFSDIRK explicitly set `stage1_extrapolation = false`; ImplicitMidpoint has `explicit_first_stage = false` and `stage1_extrapolation = true`. Same as above: smoke test on stiff decay was clean.

**What this also fixes**: `ABDF2`'s `cache.eulercache::ESDIRKIMEXCache` is built from `ImplicitEulerESDIRKIMEXTableau` (E === :ie_dd2), so the new check catches it automatically. No separate branch needed.

Updated commit message reflects the new structure. Diff is now -27 / +15 (smaller net than before).

ChrisRackauckas-Claude:
Dropped the stage-1 MaxOrder branch as well, per the follow-up question. Reasoning: that branch was effectively dead after replacing the alg-Union with the tableau check — Trapezoid and the other Newton-SDIRK algs in the original Union all have `explicit_first_stage = true` and skip the implicit-first-stage `else` arm entirely, so the only thing that could reach the MaxOrder code path was `ImplicitEuler`, whose interpolant *is* linear, making `Predictor.MaxOrder` produce essentially the same guess as `Predictor.Linear` anyway.

Final shape of the `else` arm in both `_perform_step_iip!` and `_perform_step_oop!`:

```julia
if tab.stage1_extrapolation && E === :ie_dd2
    @.. broadcast=false zs[1] = dt * integrator.fsalfirst   # / z1 = dt * integrator.fsalfirst
else
    zs[1] .= zero(eltype(zs[1]))                            # / z1 = zero(u)
end
```

Diff is now **-33 / +15** (down from -27 / +15). nlstep_tests.jl still 19 / 0 / 6 (pass / fail / broken). Smoke test of `ImplicitEuler`/`SDIRK2`/`Cash4`/`Hairer4`/`Hairer42`/`ImplicitMidpoint`/`Trapezoid`/`ABDF2` on stiff `du = -50u` produces identical-to-pre-patch numbers; no regression.

ChrisRackauckas-Claude:
Also dropped the redundant `tab.stage1_extrapolation` half of the conjunction. `ImplicitEulerESDIRKIMEXTableau` is the only constructor producing `E === :ie_dd2` and it doesn't override the constructor default `stage1_extrapolation = true`, so the boolean half was always true when the symbol half was — dead code.

Final `else` arm of both `_perform_step_iip!` and `_perform_step_oop!`:

```julia
if E === :ie_dd2
    @.. broadcast = false zs[1] = dt * integrator.fsalfirst   # / z1 = dt * integrator.fsalfirst
else
    zs[1] .= zero(eltype(zs[1]))                              # / z1 = zero(u)
end
```

Diff now -33 / +14. Verification numbers unchanged (same machine-precision errors for ImplicitEuler / SDIRK2 / Cash4 / Hairer4 / Hairer42 / ImplicitMidpoint / Trapezoid / ABDF2 on stiff decay; nlstep_tests.jl 19 / 0 / 6).

PR #3699, opened 2026-05-29
Fix solve_up enzyme rule to support mixedduplicated returns

wsmoses:
@ChrisRackauckas-Claude / @ChrisRackauckas can you give this a go?


ChrisRackauckas:
Awesome this was the hint that I needed. I think I got it now.

PR #3708, opened 2026-06-03
Make derivative_discontinuity! sticky across callbacks

DaniGlez:
When multiple callbacks fire on the same step (several DiscreteCallbacks, or simultaneous events in a VectorContinuousCallback), the derivative discontinuity flag was a single value overwritten by each caller, so the outcome depended on callback evaluation order: a callback flagging a discontinuity could be silently overridden by a later one that didn't.

Make the flag effectively three-state per step, independent of order:
  - any callback flagging `true` wins (recompute);
  - only if every callback that spoke flagged `false` is recomputation skipped;
  - if no callback calls the setter, the conservative default (true) is kept.

Regarding implementation, a companion `user_set_discontinuity` field distinguishes "a callback explicitly said false" from "no callback spoke". The cross-callback any-true-wins merge lives in apply_callback! / apply_discrete_callback!; the setter stays a plain assignment so framework paths (e.g. callback initialization) can still clear the flag with `false`. The redundant flag-clear at the end of reeval_internals_due_to_modification! is dropped.

Adds a testset in callback_merging.jl and docstrings on both the ODEIntegrator and DDEIntegrator methods.

One important note: the return value of `reeval_internals_due_to_modification!` has been dropped, since 1) it's not needed and 2) it was not practically used in the previous implementation (it was a constant and not really consumed).

## Checklist


ChrisRackauckas:
2 related questions here:

1. Why does this need a second boolean? `integrator.derivative_discontinuity` is supposed to be tracking if the user set it, so you could just make that sticky?
2. What if one callback sets it to true then the second false?

DaniGlez:
Regarding the first question, the issue is that we need a third state, because of the combination of:

1) the default is true (if the user says nothing, assume the derivative changed)
2) a user-set false overrides the default value
3) but, at the same time, any user-set true should prevail (if callback 2 says there's a discontinuity, then it doesn't matter that callback 1 did not generate one).

In other words, the problem is that the sticky value (true) is also the default value (according to the logic that the PR proposes). 

Regarding the second question, the logic I propose in this PR is that any true prevails over one or several falses, which prevails over the default (also true). That way, behavior should correct regardless of callback execution order. In your example, true would always prevail in this PR, whereas current behavior depends on ordering.

It does definitely seem that this change is bigger than it should be in order to apply this behavior, but considering the interactions you need something like it, I think.

Another alternative to the second bool is doing an enum, but that's a bit of a pain for the consumers of the flag, since you have to either change all of them or rename the field and do some `getproperty` hack.

ChrisRackauckas:
I see

ChrisRackauckas (approved):

JoshuaLampert:
Looks like this PR (DiffEqBase v7.5.6) broke all our tests in Trixi.jl, see, e.g., https://github.com/trixi-framework/Trixi.jl/actions/runs/28020198962/job/82934308245?pr=2706#step:7:1085. I currently do not have the time to look deeper into this or create an MWE.

DaniGlez:
Looking at the logs, I suspect there is a compatibility issue between the different subpackages; i.e. the environment in there picks up a DiffEqBase release with the new changes (accessing a new integrator field) without the associated OrdinaryDiffEqCore integrator struct update. This could be a tricky one since DiffEqBase is the depended-on package so you can't just [compat] it away. I can check later.

JoshuaLampert:
Thanks! We are already using the latest release of OrdinaryDiffEqCore.jl, which is v4.3.0. Maybe we just need a new release of OrdinaryDiffEqCore.jl?

DaniGlez:
Oh, I hadn't even realized OrdinaryDiffEqCore.jl hasn't had a recent release (4.3.0 seems to be from a month ago, i.e. before this PR). In that case, I think there definitely needs to be a release or this problem will pop up in more places soon. Nevertheless, from what I can gather, there could still be problematic scenarios, since even with a release someone could have an old Core but a new Base and run into the same problem. My Claude claims to be able to move all the logic to the base package and thus leave the integrator structs untouched from before this PR, but I won't have time to check the proposed changes until later.

JoshuaLampert:
@ChrisRackauckas, could you please release a new version of OrdinaryDiffEqCore.jl as a quick fix? Better solutions could be investigated later.

ChrisRackauckas:
Yeah looks like we need to bump a lower bound and yank a version.

ChrisRackauckas:
> My Claude claims to be able to move all the logic to the base package and thus leave the integrator structs untouched from before this PR, but I won't have time to check the proposed changes until later.

That's what I thought 😆 which is why I asked that question if it was possible without the second boolean. I suspected it was.

DaniGlez:
A strong hunch, indeed!

PR #3717, opened 2026-06-05
Multirate: add MIS2 (Wensch–Knoth–Galant 2009) + tableau/file-structure refactor

singhharsh1708:
Adds **MIS2** (Multirate Infinitesimal Step, Wensch–Knoth–Galant, BIT 49 (2009), 449–473) to `OrdinaryDiffEqMultirate`, and—per review feedback—refactors the subpackage's tableau handling and file layout to match the rest of OrdinaryDiffEq.

## 1. New method: MIS2

Multirate Infinitesimal Step on a SplitODE `du/dt = f1 + f2` (`f1` fast, `f2` slow). One step runs `s = 4` outer stages; each solves a small modified-fast ODE on `τ ∈ [0, d_i·dt]` whose RHS is `f1(v)` plus a stage-dependent constant offset built from prior stages' slow tendencies and α/γ corrections. The inner ODE uses `m·d_i` explicit-midpoint (RK2) micro-steps, so the global scheme reaches order 2 with fixed `m` (Euler inner would cap at order 1).

- Tableau: MIS2(4,2) from Wensch–Knoth–Galant (BIT 2009).
- Embedded error estimate: `Y_s − Y_{s-1}`.
- `m` is a required field (not adaptive — per review).

## 2. Tableau implementation (review: "tableau implementation here" / "Also tableau here")

Both methods touched by review now use type-parameterized tableau structs carried in the cache, instead of hardcoded `const Float64` data:

- **MIS2** → `MISTableau{T}` (α, β, γ matrices + derived d, c, c̃), built by `MIS2Tableau(::Type{T})`.
- **MRAB** → `MRABTableau{T}` holding the Adams–Bashforth weight vectors for orders `1..k` (MRAB uses every order up to `k` within a step for bootstrap + the AB-k − AB-(k-1) error estimate), built by `MRABTableau(k, ::Type{T})`. Replaces the hardcoded `_ab_betas(k)` Float64 tuples.

Both now propagate any element type (Float32/BigFloat/AD duals), verified by smoke tests — the old `const Float64` / tuple forms force-promoted to Float64.

## 3. File-structure refactor (match the RK sublibrary convention)

LowOrderRK / HighOrderRK / Verner each keep **one** `*_caches.jl`, `*_perform_step.jl`, and `*_tableaus.jl` for the whole sublibrary. The multirate sublib had one of each *per method* (`mreef_*`, `mrab_*`, `mri_gark_*`, `mis2_*`). Consolidated into:

- `multirate_tableaus.jl` — all tableau structs (MRAB, MRI-GARK, MIS)
- `multirate_caches.jl` — all caches
- `multirate_perform_step.jl` — all `initialize!` / `perform_step!`

This is **pure relocation** of already-merged MREEF / MRAB / MRI-GARK code (no logic change) plus the new MIS2, so the diff against master touches those files but their behavior is unchanged.

## Tests

Full suite green: **MREEF 19, MRAB 16, MRI-GARK 14, MIS2 7**. Runic clean. Solutions unchanged at fixed `dt` for the relocated methods.

## Checklist

- [x] The new code follows the [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and [COLPRAC](https://github.com/SciML/COLPRAC).


singhharsh1708:
@ChrisRackauckas Adds `MIS2` (Wensch–Knoth–Galant, BIT 49 (2009), 449–473) to `OrdinaryDiffEqMultirate` as the third solver in the subpackage after MREEF and MRAB.

**Framework over hard-coded constants.** Rather than baking MIS2's tableau into the perform_step, I factored out a reusable `MISTableau{T}` carrier (`α`, `β`, `γ`, derived `d`, `c`, `c̃`) in `mis_tableaus.jl`. The shared `mis_perform_step.jl` reads the tableau from `cache.tab`, so future MIS variants (MIS3a/b/c, MIS4, TVDMISA/B, MISRK1/2a/2b/3) drop in as ~30-line `*Tableau(T)` constructors with no perform_step changes.

**RK3 inner.** Inner ODE on `τ ∈ [0, d_i·dt]` uses classical Runge-Kutta 3 micro-steps so the inner contribution doesn't cap the outer order. Verified MIS2 hits design order 2 cleanly (order ≈ 2.06 across `m ∈ {4, 16, 64}`, error halving ratios converging to exactly 4.0).

**Paper-invariant check.** `d[s] = sum(β[s,:])`, `c[1] = 0`, `c[s] ≈ 1` all hold for the computed tableau.

**Other.** Allowlisted `MIS` in `.typos.toml`; added empty `ext/` stub for the new centralized SublibraryCI. Bumped 2.1.0 → 2.2.0. 42 tests pass locally (19 MREEF + 16 MRAB + 7 MIS).

Multirate-relevant CI: 5/6 green (the 6th is a runner timeout on Julia 1, not a real failure). The other 20-odd red checks are pre-existing master conditions (spell check on `Lamba`/`Strang`/etc., runic drift on master files, PositiveIntegrators downstream, downgrade sublibs on Core/Default/etc.) — none touched by this PR.

A follow-up PR will add MIS3C/MIS4 once I sort out why my MIS3C implementation is showing global order 2 instead of the claimed 3; the framework here is ready for them.

lib/OrdinaryDiffEqMultirate/src/algorithms.jl
```diff
@@ -63,3 +63,38 @@ Base.@kwdef struct MRAB <: OrdinaryDiffEqAdaptiveAlgorithm
...
+    """
+    - `m`: nominal number of inner midpoint micro-steps per `dt`; each stage takes
+        `max(1, ceil(m · d_i))` micro-steps. Default is `10`.
+    """,
+    """
+    m::Int = 10,
+    """
+)
+Base.@kwdef struct MIS2 <: OrdinaryDiffEqAdaptiveAlgorithm
+    m::Int = 10
```
ChrisRackauckas: this should probably be required from the user? It's not adaptive?

ChrisRackauckas:
Why 3rd order in the inner? Seems like a bit of a waste? Does the paper do that?

singhharsh1708:
The inner solver in the current implementation is already **explicit midpoint (RK2)** — 2 f1 evaluations per micro-step (lines 104–108 of `mis2_perform_step.jl`). The "RK3 inner" label in my earlier description was a typo; the code was never RK3. For MIS2 (outer order 2), RK2 inner is sufficient — Euler would cap the global order at 1, but midpoint doesn't.

ChrisRackauckas:
Also tableau here.

singhharsh1708:
@ChrisRackauckas — all three points addressed:

1. **`m` required** — `MIS2` takes `m::Int`, no default.
2. **Inner order** — "RK3" was a description typo; the inner is explicit midpoint (RK2), the minimum to keep the order-2 outer scheme uncapped. Description fixed.
3. **"Also tableau here"** — MRAB's `_ab_betas` Float64 weights are now a type-parameterized `MRABTableau{T}` in the cache (carries the AB weights for orders 1..k), matching MIS2/MRI-GARK. Type-generic, smoke-tested.

Also consolidated the subpackage into single `multirate_tableaus.jl` / `multirate_caches.jl` / `multirate_perform_step.jl` files; the MREEF/MRAB/MRI-GARK relocation is a pure move (diffed line-for-line vs master — no logic change). MREEF stays tableau-less on purpose — it's an extrapolation method with no coefficient table. Bumped to 2.3.0.

Multirate CI green (Core + QA, julia 1/lts/pre). Runic red is pre-existing master drift — all multirate files pass (✔).

ChrisRackauckas:
Yes to the restructure. But now MIS should just have a number of stages and to a maximum like SDIRK

ChrisRackauckas:
> Inner order — "RK3" was a description typo; the inner is explicit midpoint (RK2), the minimum to keep the order-2 outer scheme uncapped. Description fixed.

But why only does this as a second order method? Isn't it uncapped?

singhharsh1708:
Thanks! Two things on these:

On the stages — I think it's actually already set up that way? The `perform_step!` reads the stage count from the tableau (`s = length(d)`) and the cache allocates `Y`/`fS` from it, so nothing is pinned to MIS2's 4 stages — a higher-stage MIS should just drop in as a new `*Tableau`. If you'd prefer the explicit `MAX_STAGES` if-ladder like SDIRK instead of the `length(d)` loop, I can switch it over.

On the order — you're right that the framework itself isn't capped. What's capping it at 2 right now is the inner solver: I have it fixed to explicit midpoint (RK2), so Euler inner gives order 1 and RK2 gives order 2. To get order 3 I'd need an RK3 inner together with an order-3 tableau. When I tried an order-3 MIS earlier it kept silently dropping back to order 2, which I think was the inner solver. Would it be okay if I do the higher-order version as a separate follow-up so I can get the convergence tests right, and keep this PR as the order-2 method + the restructure?

ChrisRackauckas:
> On the order — you're right that the framework itself isn't capped. What's capping it at 2 right now is the inner solver: I have it fixed to explicit midpoint (RK2), so Euler inner gives order 1 and RK2 gives order 2. To get order 3 I'd need an RK3 inner together with an order-3 tableau. When I tried an order-3 MIS earlier it kept silently dropping back to order 2, which I think was the inner solver. Would it be okay if I do the higher-order version as a separate follow-up so I can get the convergence tests right, and keep this PR as the order-2 method + the restructure?

Yes, but then we shouldn't rename what is exported, so we should just call it MIS and then update it to support higher order.

singhharsh1708:
Done — renamed the export to `MIS` (kept the (4,2) tableau internally as `MIS2Tableau`), so the higher-order follow-up can extend it without changing the public name. Tests green.

ChrisRackauckas (approved):

PR #3720, opened 2026-06-05
Optimizations to Discontinuity Detection Scheme

Shreyas-Ekanathan:
## Checklist

## Additional context

Optimizations:
- Add a `tstop` on the first identification of a discontinuity to help with step-resizing and prevent excessively large steps after finding a discontinuity. This should help us not take unreasonable steps, as Oscar suggested, and helps with error resetting after passing a discontinuity (since we are effectively in a new regime) to begin stepping anew.
- Shrink the solve window for the interval nonlinear problem after finding a discontinuity
- Move `_ode_addsteps! ` out of the zero_func_struct call and only call it once per discontinuity detection run


lib/OrdinaryDiffEqBDF/src/controllers.jl
```diff
@@ -174,8 +179,12 @@ function bdf_step_reject_controller!(integrator, cache, EEst1)
     end
 
     if discontinuity_detection
+        no_prior_discontinuity = (integrator.curr_discontinuity == -1)
         disco_dt = set_discontinuity(integrator.u, integrator.uprev, integrator)
-        if disco_dt != -1
+        if no_prior_discontinuity && disco_dt > zero(disco_dt)
+            add_tstop!(integrator, integrator.disco_checkpoint)
```
oscardssmith: we likely don't want to add multiple tstops for the same discontinuity (and want to remove the tstop if we find the discontinuity before hitting the tstop)

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -1020,10 +1046,6 @@ integrator.dt / qacc
 ```
 
 When it rejects, it's the same as the [`IController`](@ref):
-If `discontinuity_detection` is set to true, the algorithm will run the autonomous
```
oscardssmith, line 1127: why delete the comment?

Shreyas-Ekanathan: It was redundant and I had moved it to basic, so I dropped it from each individual controller and left it on basic's docstring instead

lib/OrdinaryDiffEqBDF/src/controllers.jl
Shreyas-Ekanathan: It shouldn't add multiple per discontinuity, since its gated on the fact that there is no discontinuity being logged at the moment right. 

I thought about removing the tstop but I couldn't find a clean interface to do so, is there a good way to do that?

oscardssmith: not sure if we have a good interface for removing. You're right that this doesn't add multiple

lib/OrdinaryDiffEqCore/src/integrators/type.jl
```diff
@@ -175,6 +175,8 @@ mutable struct ODEIntegrator{
     fsalfirst::FSALType
     fsallast::FSALType
     rng::RNGType
+    is_disco_step::Bool
+    disco_checkpoint::tType
```
oscardssmith, line 192: We want to rename both of these.

lib/OrdinaryDiffEqCore/src/disco.jl
```diff
@@ -1,5 +1,5 @@
-function set_discontinuity(u, uprev, integrator)
-    breakpointθ = find_discontinuity(u, uprev, integrator)
+function set_discontinuity(integrator)
+    breakpointθ = find_discontinuity(integrator)
     dt = integrator.dt
     if 1.0e-10 < breakpointθ < 1.0
```
oscardssmith: why 1e-10?

lib/OrdinaryDiffEqCore/src/disco.jl
```diff
@@ -40,22 +43,36 @@ function find_discontinuity(u, uprev, integrator)
...
             for j in 1:len_cb
                 if (disco_zero.out_low[j] * disco_zero.out_high[j] < zero(disco_zero.out_low[j]))
+                    if (!addsteps_called)
+                        addsteps_called = true
+                        _ode_addsteps!(disco_zero.k, disco_zero.tprev, disco_zero.uprev, disco_zero.u,
+                                    disco_zero.dt, disco_zero.f, disco_zero.p, disco_zero.cache, false, true, false)
+                    end
                     disco_zero.ind = j
-                    sol = solve(disco_prob)
+                    sol = solve(disco_prob, tspan = (0.0, breakpointθ == -one(dt) ? 1.0 : breakpointθ); save_everystep = false)
```
oscardssmith: does save_everystep do anything here? I didn't think any steps exist for NonlinearProblems. Also I wonder if using breakpointθ=-1 as the default makes sense. Could we use 1 to mean "no restriction"?

lib/OrdinaryDiffEqBDF/src/controllers.jl
```diff
@@ -83,9 +83,14 @@ end
 function step_accept_controller!(integrator, cache::Union{QNDFCache, QNDFConstantCache}, alg::QNDF{max_order}, q) where {max_order}
     #step is accepted, reset count of consecutive failed steps
     cache.consfailcnt = 0
+    is_disco = integrator.is_disco_step
+    if is_disco
+        integrator.is_disco_step = false
+        cache.nconsteps = 0
```
oscardssmith, line 89: what is this line doing?

lib/OrdinaryDiffEqCore/src/disco.jl
Shreyas-Ekanathan: pretty arbitrary, just don't want to take a tiny step (I think this is reasonable?)

lib/OrdinaryDiffEqBDF/src/controllers.jl
Shreyas-Ekanathan: because it's not a standard timestep BDF shouldn't really count this step towards its order selection logic right?

lib/OrdinaryDiffEqCore/src/integrators/type.jl
Shreyas-Ekanathan: what would you name them?

lib/OrdinaryDiffEqCore/src/disco.jl
Shreyas-Ekanathan: I think using 1 is a little dicey since it potentially creates some numerical issues? whereas -1 is a cleaner sentinel imo. 

save_everystep does prevent the solve from tracking every internal step I believe, so it should help?

lib/OrdinaryDiffEqCore/src/disco.jl
oscardssmith: I think that's both smaller than we want (I think we probably never want and the wrong result if we hit it. I think it does make sense to clamp `breakpointθ` to (0.1, 0.9) because if we are on the wrong side of a disco point very close to the edge, we don't shrink the interval, but I think we still want to use disco if we have a sign crossing that we think is very close to an edge.

lib/OrdinaryDiffEqCore/src/disco.jl
oscardssmith: The argument for 1 is that it deosn't need to be a sentinel. 1.0 naturally means no restriction on the timestep. I don't think we track internal steps for a nonlinear solve.

lib/OrdinaryDiffEqBDF/src/controllers.jl
oscardssmith: @ChrisRackauckas thoughts on this? I don't know how we want disco to interact with BDF well enough here.

lib/OrdinaryDiffEqBDF/src/controllers.jl
```diff
@@ -147,16 +152,16 @@ function step_accept_controller!(integrator, cache::Union{QNDFCache, QNDFConstan
         end
         cache.order = kₙ
         q = integrator.dt / hₙ
-    end
-    if prefer_const_step
-        if q < 1.2 && q > 0.6
-            return integrator.dt
+
+        if prefer_const_step && q < 1.2 && q > 0.6
```
oscardssmith:
```suggestion
        if prefer_const_step && 0.6 < q < 1.2
```

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -708,25 +714,31 @@ end
...
+    t = integrator.t
+    dt = integrator.dt
     if qsteady_min <= q <= qsteady_max
         q = one(q)
     end
     cache.errold = max(EEst, qoldinit)
-    return integrator.dt / q # new dt
+    if discontinuity_detection && integrator.is_disco_step
+        integrator.is_disco_step = false
+        return min((integrator.disco_checkpoint - t) / 4, dt / q) # new dt
```
oscardssmith: I don't love that this logic is duplicated over all the controllers

oscardssmith:
is this ready to merge?

Shreyas-Ekanathan:
> is this ready to merge?

yes should be

oscardssmith:
any idea why so much CI is failing?

Shreyas-Ekanathan:
> any idea why so much CI is failing?

Honestly not sure, it seems pretty random. I pushed another small fix for what I could see that is disco's fault, the remainder appears unrelated

oscardssmith (approved):

PR #3705, opened 2026-06-01
SDIRK: add IMEX-SSP family + ESDIRKIMEXTableau non-ESDIRK support

singhharsh1708:
## Summary

Adds four IMEX-SSP schemes from Pareschi & Russo (2005): `IMEXSSP222`, `IMEXSSP2322`, `IMEXSSP3332`, and `IMEXSSP3433`, partially addressing #2065.

Supporting these methods required a small extension to the `ESDIRKIMEXTableau` framework. A new `ce` field was added to store explicit abscissae separately from the implicit `c` values, with the constructor defaulting to `ce = copy(c)` so all existing ESDIRK methods remain unchanged. The generic IMEX stepping code was updated to use `ce` for explicit evaluations and to handle methods with an implicit first stage (`explicit_first_stage = false`).

## Verification

Convergence was verified on a `SplitODEProblem(f₁=-u, f₂=2u)` with analytic solution `u(t)=exp(t)u₀`. Measured orders match expectations: approximately second order for `IMEXSSP222`, `IMEXSSP2322`, and `IMEXSSP3332`, and third order for `IMEXSSP3433`. Existing methods (`ARS343`, `ARS222`, `KenCarp3`) were rechecked and remain unchanged.

Tableau consistency tests pass (`54/54`).

## Tests

* Added convergence tests for all four IMEX-SSP methods in both OOP and IIP modes.
* Added all four methods to the tableau consistency test suite.

All tests pass locally.

## Scope

This PR covers only the IMEX-SSP family. `BHR553` will follow in a separate PR after its coefficient/order issue is resolved.

## Reference

Pareschi, L. & Russo, G. (2005). *Implicit-explicit Runge-Kutta schemes and applications to hyperbolic systems with relaxation*. Journal of Scientific Computing, 25, 129–155.

PR #3139, opened 2026-03-16
Add MREEF multirate extrapolated explicit Euler solver with tests

singhharsh1708:
Checklist:
 Appropriate tests were added
 Any code changes were done in a way that does not break public API
 All documentation related to code changes were updated
 The new code follows the
[contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md)
, in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle)
 and[COLPRAC](https://github.com/SciML/COLPRAC)
 Any new documentation only uses public API
Additional context:
This PR adds `MREEF` (Multirate Extrapolated Explicit Euler), based on Constantinescu & Sandu (2010).  
The method targets `SplitODEProblem` systems and combines multirate Euler base steps with Richardson extrapolation to achieve higher-order accuracy.

Basic convergence tests and compatibility checks (SplitODEProblem, ForwardDiff, StaticArrays, complex states) were added under `test/multirate/mreef_tests.jl`. All existing `OrdinaryDiffEqLowOrderRK` tests pass locally with no regressions.

This implementation was mainly developed to better understand the solver architecture in OrdinaryDiffEq and to explore multirate methods.

PR #3724, opened 2026-06-06
Multirate: add MRI-GARK-ERK22a and ERK22b (Sandu 2019)

singhharsh1708:
Adds `MRIGARKERK22a` and `MRIGARKERK22b`, the two 2nd-order explicit members of the Multirate Infinitesimal GARK family from Sandu, *SIAM J. Numer. Anal.* 57 (2019), 2300–2327, §2.3. Both methods solve `du/dt = f1(u,t) + f2(u,t)` (f1 fast, f2 slow). Fifth and sixth solvers in `OrdinaryDiffEqMultirate` after MREEF (#3139), MRAB (#3705), and MIS2 (#3717).

## Algorithm source

Both methods share the same stage-1 structure but differ in c₂, which controls the slow method's quadrature:

**ERK22a — c₂ = 1/2 (explicit midpoint as slow method)**
```
Stage 1: v(0) = u_n,  dv/dτ = (1/2) f1(v) + (1/2) f2(u_n)                   → Y₂ = v(dt)
Stage 2: v(0) = Y₂,  dv/dτ = (1/2) f1(v) - (1/2) f2(u_n) + f2(Y₂)           → u_{n+1} = v(dt)
```

**ERK22b — c₂ = 1 (explicit trapezoidal as slow method)**
```
Stage 1: v(0) = u_n,  dv/dτ = f1(v) + f2(u_n)                                → Y₂ = v(dt)
Stage 2: Γ₂₂ = 1 - c₂ = 0 → no fast dynamics; pure slow correction:
         u_{n+1} = Y₂ + dt · (-½ f2(u_n) + ½ f2(Y₂))
```

Each fast stage uses `m` explicit-midpoint (RK2) micro-steps so the inner contribution does not cap the design order. The embedded error estimate uses `u_{n+1} − Y₂`.

## Framework

Dedicated `MRIGARKERK22{a,b}Cache` + `MRIGARKERK22{a,b}ConstantCache` mirroring the MREEF/MRAB style. Combined `Union` dispatch in `alg_utils.jl` for shared `alg_order`/`isfsal`/`prepare_alg`. No changes to existing solvers.

## Test plan

- [x] Construction (kwarg + positional) for both variants
- [x] Scalar OOP + Vector IIP smoke (fixed-step and adaptive) for both variants
- [x] Convergence order ≈ 2 for ERK22a (`𝒪est ≈ 2.12`) and ERK22b across `m = 8`
- [x] MREEF (19) + MRAB (16) regression suites still pass (49 tests total green)
- [ ] CI green on Julia 1.10 / 1.11 / lts / pre

Subpackage version 2.1.0 → 2.2.0.

### DIFF
--- a/lib/OrdinaryDiffEqMultirate/src/algorithms.jl
+++ b/lib/OrdinaryDiffEqMultirate/src/algorithms.jl
@@ -92,3 +92,32 @@ globally; embedded error estimate uses `v_2(dt) − Y_2`.",
 Base.@kwdef struct MRIGARKERK22a <: OrdinaryDiffEqAdaptiveAlgorithm
     m::Int = 10
 end
+
+@doc generic_solver_docstring(
+    "Multirate Infinitesimal GARK — explicit trapezoidal (MRI-GARK-ERK22b).
+
+Solves a SplitODE `du/dt = f1(u,t) + f2(u,t)` where `f1` is the fast component
+and `f2` is the slow component (SciML convention). The slow integrator is the
+explicit trapezoidal rule (`c₂ = 1` member of the MRI-GARK-ERK22 family from
+Sandu 2019); inner fast micro-ODE uses `m` explicit-midpoint micro-steps over
+the full macro interval. 2nd-order; embedded error estimate uses `v_2(dt) − Y_2`.",
+    "MRIGARKERK22b",
+    "Multirate infinitesimal GARK explicit method.",
+    """@article{sandu2019class,
+    title={A class of multirate infinitesimal {GARK} methods},
+    author={Sandu, Adrian},
+    journal={SIAM Journal on Numerical Analysis},
+    volume={57},
+    number={5},
+    pages={2300--2327},
+    year={2019}}""",
+    """
+    - `m`: number of inner midpoint micro-steps per stage. Default is `10`.
+    """,
+    """
+    m::Int = 10,
+    """
+)
+Base.@kwdef struct MRIGARKERK22b <: OrdinaryDiffEqAdaptiveAlgorithm
+    m::Int = 10
+end
--- a/lib/OrdinaryDiffEqMultirate/src/algorithms.jl
+++ b/lib/OrdinaryDiffEqMultirate/src/algorithms.jl
@@ -63,3 +63,32 @@ Base.@kwdef struct MRAB <: OrdinaryDiffEqAdaptiveAlgorithm
     k::Int = 2
     m::Int = 4
 end
+
+@doc generic_solver_docstring(
+    "Multirate Infinitesimal GARK — explicit midpoint (MRI-GARK-ERK22a).
+
+Solves a SplitODE `du/dt = f1(u,t) + f2(u,t)` where `f1` is the fast component
+and `f2` is the slow component (SciML convention). The slow integrator is the
+explicit midpoint rule; between slow stages a modified fast ODE is integrated
+with `m` explicit-midpoint micro-steps over the full macro interval. 2nd-order
+globally; embedded error estimate uses `v_2(dt) − Y_2`.",
+    "MRIGARKERK22a",
+    "Multirate infinitesimal GARK explicit method.",
+    """@article{sandu2019class,
+    title={A class of multirate infinitesimal {GARK} methods},
+    author={Sandu, Adrian},
+    journal={SIAM Journal on Numerical Analysis},
+    volume={57},
+    number={5},
+    pages={2300--2327},
+    year={2019}}""",
+    """
+    - `m`: number of inner midpoint micro-steps per stage. Default is `10`.
+    """,
+    """
+    m::Int = 10,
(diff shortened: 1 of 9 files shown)


lib/OrdinaryDiffEqMultirate/src/algorithms.jl
```diff
@@ -63,3 +63,32 @@ Base.@kwdef struct MRAB <: OrdinaryDiffEqAdaptiveAlgorithm
...
+    year={2019}}""",
+    """
+    - `m`: number of inner midpoint micro-steps per stage. Default is `10`.
+    """,
+    """
+    m::Int = 10,
+    """
+)
+Base.@kwdef struct MRIGARKERK22a <: OrdinaryDiffEqAdaptiveAlgorithm
+    m::Int = 10
```
ChrisRackauckas: should probably be something set by the user

ChrisRackauckas:
This looks correct, but we should definitely do a tableau implementation here.

singhharsh1708:
Both comments addressed:

* `m` no longer has a default and is now a required field (`m::Int`), so the number of micro-steps must be chosen explicitly.
* Added `MRIGARKExplicit22Tableau{T}` to hold the `Γ/γ` coupling coefficients. `MRIGARKERK22a` and `MRIGARKERK22b` now share a single `MRIGARKERK22Cache` and `perform_step!`, parameterized by the tableau type. The `iszero(Γ22)` branch handles ERK22b's degenerate second stage (`c₂ = 1`) by skipping the inner fast loop and applying only the slow correction.

ChrisRackauckas:
Why just 2? What specifically about this needs to be 2?

singhharsh1708:
Good point — nothing needs it to be 2. Replaced MRIGARKExplicit22Tableau with a general MRIGARKExplicitTableau{T} (Γ::Vector, γ::Matrix, c::Vector of length s); perform_step! now loops i = 1:s, so higher-order variants just need a new tableau constructor — no cache/perform_step changes.

lib/OrdinaryDiffEqMultirate/src/mri_gark_perform_step.jl
```diff
@@ -79,50 +83,54 @@ function perform_step!(integrator, cache::MRIGARKERK22Cache, repeat_step = false
...
+    for i in 1:s
+        fS[i] = f.f2(Y[i], p, t + c[i] * dt)
+        integrator.stats.nf2 += 1
 
-    fS_Y2 = f.f2(Y2, p, t + dt)
-    integrator.stats.nf2 += 1
+        slow_f = γ[i, 1] * fS[1]
+        for j in 2:i
+            if !iszero(γ[i, j])
+                slow_f = @.. broadcast = false slow_f + γ[i, j] * fS[j]
```
ChrisRackauckas, line 106: We should do a follow up to optimize this like the other RK tableus. Too many kernel launches.

lib/OrdinaryDiffEqMultirate/src/mri_gark_perform_step.jl
```diff
@@ -25,50 +25,54 @@ function initialize!(integrator, cache::MRIGARKERK22ConstantCache)
...
-    f.f2(fS_Y2, Y2, p, t + dt)
-    integrator.stats.nf2 += 1
+    for i in 1:s
+        f.f2(fS[i], Y[i], p, t + c[i] * dt)
+        integrator.stats.nf2 += 1
+
+        @.. broadcast = false slow_f = γ[i, 1] * fS[1]
+        for j in 2:i
+            if !iszero(γ[i, j])
+                @.. broadcast = false slow_f = slow_f + γ[i, j] * fS[j]
```
ChrisRackauckas, line 46: Fuse

lib/OrdinaryDiffEqMultirate/src/mri_gark_perform_step.jl
```diff
@@ -25,50 +25,54 @@ function initialize!(integrator, cache::MRIGARKERK22ConstantCache)
...
+                @.. broadcast = false u = Y[i] + dt * slow_f
+            end
+        else
+            @.. broadcast = false v = Y[i]
+            for k_step in 1:m
+                t_a = t + (k_step - 1) * h_inner
+                f.f1(f1eval, v, p, t_a)
+                @.. broadcast = false tmp = v + (h_inner / 2) * (Γ[i] * f1eval + slow_f)
+                f.f1(f1eval, tmp, p, t_a + h_inner / 2)
+                @.. broadcast = false v = v + h_inner * (Γ[i] * f1eval + slow_f)
```
ChrisRackauckas, line 63: Not sure if you can fuse in here.

ChrisRackauckas (approved):

singhharsh1708:
Sounds good — I'll handle the broadcast fusion / kernel-launch reduction as the follow-up.

PR #3731, opened 2026-06-09
Support better logging for singular cases in integration

Shreyas-Ekanathan:
## Checklist

## Additional context

Allows SciMLBase to freshly calculate the Jacobian for most accurate analysis. Supports https://github.com/SciML/SciMLBase.jl/pull/1379


lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
```diff
@@ -611,6 +612,90 @@ function log_step!(progress_name, progress_id, progress_message, dt, u, p, t, ts
...
+        nothing
+    end
+
+    large_jac_rows = nothing
+    large_jac_cols = nothing
+    if jac !== nothing
+        rows = Set{Int}()
+        cols = Set{Int}()
+        for i in axes(jac, 1), j in axes(jac, 2)
+            if !isfinite(jac[i, j]) || abs(jac[i, j]) > 1e6
```
termi-official: If `jac` is a sparse matrix, then this will be super expensive. `jac[i,j]` triggers a lookup on every call and we even iterate over all zeros.

ChrisRackauckas (approved):

lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
```diff
@@ -354,6 +354,8 @@ function calc_J(integrator, cache, next_step::Bool = false)
     return J
 end
 
+get_fresh_jacobian(integrator, cache::OrdinaryDiffEqCache) = calc_J(integrator, cache)
```
ChrisRackauckas, line 357: why not just use calc_J?

Shreyas-Ekanathan: circular dep, calc_J is in OrdinaryDiffEqDiff so we can't use that from core

PR #3740, opened 2026-06-10
Guard solve_up Enzyme reverse rule against runtime-activity aliased shadows

SebastianM-C:
While looking into reducing SciML/SciMLSensitivity.jl#1477 with Claude, I found a potential bug in the Enzyme rule under `set_runtime_activity(Reverse)` when the `u0` is aliased in `remake` (which happens because `SII.setsym_oop` returns the `prob.u0` if there are no changes). The fix that Claude proposed was to skip accumulation when the shadow aliases the primal, exactly as if it were `Const`.

@ChrisRackauckas Would this be appropriate or should we always copy in `SII.setsym_oop` (or both)?

## Checklist


ChrisRackauckas:
Seems reasonable but this should definitely get a test.

SebastianM-C:
Sure, but where would that go? From what I see SciMLSensitivity related testing is done via downstream. Should I PR the test to SciMLSensitivity?

ChrisRackauckas:
either there or the downstream ad test group

lib/DiffEqBase/ext/DiffEqBaseEnzymeExt.jl
```diff
@@ -124,9 +124,18 @@ module DiffEqBaseEnzymeExt
...
                 end
+                # Under `set_runtime_activity`, a runtime-inactive value arrives as
+                # Duplicated/MixedDuplicated with its shadow ALIASING the primal
+                # (`dval === val`). Accumulating the cotangent into such a shadow writes
+                # gradient values into the caller's primal data (e.g. the `u0` of a
+                # `Const` problem whose array was reused via `remake`), silently
+                # corrupting subsequent calls. Skip them: a runtime-inactive value
+                # accumulates nowhere, exactly as if it were `Const`.
                 if ptr isa MixedDuplicated
+                    ptr.dval[] === ptr.val && continue
```
wsmoses: @SebastianM-C / @ChrisRackauckas you should only do this check if runtime activity is set (see https://github.com/EnzymeAD/Enzyme.jl/blob/a26b7e07696f771056ad70abdf147cc2f7739b19/lib/EnzymeCore/src/rules.jl#L83 to determine if it is)

wsmoses:
this is a likely culrprit for errors, like mentioned earlier. Anything we can do to get this merged?

lib/DiffEqBase/ext/DiffEqBaseEnzymeExt.jl
SebastianM-C: addressed in [ee3b14d](https://github.com/SciML/OrdinaryDiffEq.jl/pull/3740/commits/ee3b14d198a8283fa60cb5f804787e762fcc48c3)

SebastianM-C:
I added a test, but I'm not sure if it's in the right place, there is also an `test/ad/ad_tests.jl` which loads SciMLSensitivity, but that would run on OrdinaryDiffEq changes, not DiffEqBase ones from what I understand (and it also does not run on julia 1.12).

wsmoses:
@SebastianM-C you didn't address my comment. You shouldn't do the check if runtime activity is set in the rule config

SebastianM-C:
doesn't `rta && ptr.dval[] === ptr.val && continue` (with `rta = Enzyme.EnzymeRules.runtime_activity(config)`) skip that?
or should I just skip the entire `if` on `rta` without checking `ptr.dval[] === ptr.val`?

wsmoses:
Oh yes it does, sorry apparently I didn't refresh and was looking at the old commit

PR #3745, opened 2026-06-11
remove redundant controller divisions

oscardssmith:
We were doing a lot of divisions for no reason. Doing things in terms of `dt_factor instead of `inv(dt_factor)` maeks the math a lot nicer.

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
@@ -554,7 +554,7 @@ end
     EEst = DiffEqBase.value(get_EEst(integrator))
 
     if iszero(EEst)
-        gamma = qmax
+        dt_factor = qmax
     else
         expo = inv(get_current_adaptive_order(alg, integrator.cache) + 1)
         dt_factor = gamma / fastpower(EEst, expo)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
```diff
@@ -980,9 +980,9 @@ end
     PredictiveController()
 
 The Gustafsson acceleration algorithm accelerates changes so that way algorithms
-can more swiftly change to handle quick transients. This algorithm is thus
-well-suited for stiff solvers where this can be expected, and is the default
-for algorithms like the (E)SDIRK methods.
+can more swiftly change to handle quick transients.
+This algorithm is gennerally less efficient than a PI controller, but is used in adaptive order methods
+(e.g. AdaptiveRadau) where a simpler stepsize controller makes an order controller eaiser.
```
ChrisRackauckas, line 985: That's not true though? This isn't a Proportional Controller, it's a PredictiveController which is different.

oscardssmith: PI is our default controller though, it's just the 2nd part that is wrong, right?

ChrisRackauckas: It's not generally less efficient for all solvers, and it's not used with AdaptiveRadau

oscardssmith: what solvers is it more efficient for? I don't think we use it by default for any of them.

ChrisRackauckas: we probably should use them on some sdirks

oscardssmith: which/why?

ChrisRackauckas: It attempts to predict Newton instabilities, so possibly all. It's just an underexplored topic.

PR #3747, opened 2026-06-12
Multirate: add explicit MRI-GARK family (ERK33a order 3 + ERK45a order 4), unified

singhharsh1708:
Adds the explicit MRI-GARK family of Sandu 2019 (A class of multirate infinitesimal GARK methods, SIAM J. Numer. Anal. 57:2300–2327; coefficients from SUNDIALS) and unifies all MRI-GARK methods onto one formulation:

- **MRIGARKERK22a / ERK22b** — order 2 (re-expressed in the unified form)
- **MRIGARKERK33a** — order 3
- **MRIGARKERK45a** — order 4

A single `perform_step!` handles the whole family: per-stage sub-interval (Δc) integration with τ-dependent coupling `ω = W0 + W1·τ` and an explicit-RK inner solver (RK2/RK3/RK4 selected by the tableau's inner order `q`), with an embedded lower-order error estimate. The inner-RK stage buffers are a single `kk::Vector` sized by `q`.

Convergence verified (IIP + OOP): ERK22a/b order 2, ERK33a order 3, ERK45a order 4.

Supersedes #3751 (ERK45a) — merged here into one PR per review, since the shared `perform_step!` covers all the tableaus.

PR #3751, opened 2026-06-13
Multirate: add MRI-GARK-ERK45a (4th-order explicit)

singhharsh1708:
**Stacked on #3747** — the first commit in this PR is #3747 (ERK33a + the unified MRI-GARK formulation); please review that one first. Once #3747 merges, this rebases down to a single ERK45a commit.

Adds **MRIGARKERK45a**, the 4th-order explicit MRI-GARK method of Sandu 2019, completing the explicit MRI-GARK family (orders 2/3/4) in the unified W-form. Coefficients transcribed from SUNDIALS ARKODE.

## What it needed
The general formulation from #3747 already supported it — only two small additions:
- An **RK4 inner** path (`q = 4`) in the substage integrator.
- A **τ-dependent embedded coupling** (`Wemb0 + Wemb1·τ`) — ERK45a's embedding row has linear-in-τ terms, unlike ERK33a's constant one — for the order-3 error estimate.

5 fast stages (Δc = 1/5 each). Demonstrates the framework generalizes: higher orders drop in as a new tableau + inner order.

## Tests
Convergence verified — **order 4** (IIP + OOP); full multirate suite green (MRI-GARK 28 tests). Version 2.4.0 → 2.5.0.

### DIFF
--- a/lib/OrdinaryDiffEqMultirate/src/multirate_caches.jl
+++ b/lib/OrdinaryDiffEqMultirate/src/multirate_caches.jl
@@ -114,7 +114,7 @@ end
 
 get_fsalfirstlast(cache::MRIGARKCache, u) = (cache.fsalfirst, cache.k)
 
-const MRIGARKAlg = Union{MRIGARKERK22a, MRIGARKERK22b, MRIGARKERK33a}
+const MRIGARKAlg = Union{MRIGARKERK22a, MRIGARKERK22b, MRIGARKERK33a, MRIGARKERK45a}
 
 function alg_cache(
         alg::MRIGARKAlg, u, rate_prototype,
--- a/lib/OrdinaryDiffEqMultirate/src/multirate_caches.jl
+++ b/lib/OrdinaryDiffEqMultirate/src/multirate_caches.jl
@@ -101,10 +101,12 @@ end
     tmp::uType
     atmp::uNoUnitsType
     v::uType
+    vtmp::uType
     f1eval::rateType
-    slow_f::rateType
-    Y::Vector{uType}
+    kk::Vector{uType}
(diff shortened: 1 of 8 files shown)


singhharsh1708:
@ChrisRackauckas this one stacks on #3747 and adds MRI-GARK-ERK45a (order 4), so the explicit MRI-GARK family is now complete (orders 2/3/4). Rebased on master, multirate CI green. Once #3747 merges I'll rebase this down to just the ERK45a commit.

lib/OrdinaryDiffEqMultirate/src/multirate_caches.jl
```diff
@@ -101,40 +101,55 @@ end
...
     atmp::uNoUnitsType
     v::uType
+    vtmp::uType
     f1eval::rateType
-    slow_f::rateType
-    Y::Vector{uType}
+    k1::uType
+    k2::uType
+    k3::uType
+    k4::uType
```
ChrisRackauckas: still array

ChrisRackauckas:
Do the same changes as the other tableau methods.

lib/OrdinaryDiffEqMultirate/src/multirate_caches.jl
singhharsh1708: Sorry, fixed now — these are arrayed into `kk::Vector{uType}` (sized by the inner-RK order) in the ERK33a commit, so ERK45a just reuses it. Pushed.

singhharsh1708:
Consolidated into #3747 per review — the shared `perform_step!` covers all the MRI-GARK tableaus, so ERK45a now lives there alongside ERK33a as one PR.

PR #3756, opened 2026-06-14
Fix extrapolation error-estimate denominator to use step endpoints (#3226)

### LINKED ISSUE #3226: calculate_residuals wrong point
In some of the calculate residuals, it's using u1-u0 and then using the max of those two in the denominator. However, it should be using the difference for an error estimate, and then the max of uprev and u together. Since this is violated, the stability of the non-stepped method then matters for the stepping. Not a huge deal but worth documenting.

@devmotion do you remember some discussion about this?

singhharsh1708:
Closes #3226.

The form-#2 `calculate_residuals(u₀, u₁, …)` calls in the extrapolation methods
use the difference `u₁ - u₀` as the error estimate but also normalize by
`max(|u₀|, |u₁|)`. Since `u₁`/`utilde` is the lower-order *non-stepped*
extrapolation candidate, its magnitude (and stability) leaked into step-size
control. Per the issue, the denominator should use the actual step endpoints
`max(|uprev|, |u|)`.

## Fix
Self-contained in `OrdinaryDiffEqExtrapolation` (no DiffEqBase change), 24 call sites:
- out-of-place: `calculate_residuals(u - utilde, integrator.uprev, u, …)`
- in-place: stage the difference in `cache.tmp` (a `uType` scratch present in every
  extrapolation cache) and call the explicit-error form
  `calculate_residuals!(cache.atmp, cache.tmp, integrator.uprev, integrator.u, …)`.
  `cache.utilde` is left intact because the convergence monitor reads it afterward.

## Verification
- All seven extrapolation methods (AitkenNeville, Extrapolation/Implicit Midpoint
  Deuflhard & HairerWanner, ImplicitEuler, ImplicitEulerBarycentric) solve
  adaptively with correct accuracy and sensible step/rejection counts, OOP and IIP.
- The full `OrdinaryDiffEqExtrapolation` test suite passes (284/284).
- Before/after adaptive step counts stay in the same ballpark ("not a huge deal",
  as noted in the issue); the corrected denominator actually reduces rejected steps
  for `ImplicitDeuflhardExtrapolation` (46 → 15 on a Van der Pol smoke test).
- Fixed-step convergence tests are unaffected (the change is inside the
  `integrator.opts.adaptive` branch).

### DIFF
--- a/lib/OrdinaryDiffEqExtrapolation/Project.toml
+++ b/lib/OrdinaryDiffEqExtrapolation/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqExtrapolation"
 uuid = "becaefa8-8ca2-5cf9-886d-c06f3d2bd2c4"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
-version = "2.2.0"
+version = "2.2.1"
 
 [deps]
 FastBroadcast = "7034ab61-46d4-4ed7-9d0f-46aef9175898"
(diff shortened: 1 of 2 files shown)


singhharsh1708:
@ChrisRackauckas ready for review

PR #3762, opened 2026-06-16
Stabilized methods optimizations

MoisaAndrew:
## Checklist

## Additional context

Before
download (2)

After
download (3)


lib/OrdinaryDiffEqStabilizedRK/src/rkc_caches.jl
```diff
@@ -35,16 +34,15 @@ function alg_cache(
     constantcache = ROCK2ConstantCache(
         constvalue(uBottomEltypeNoUnits),
         constvalue(tTypeNoUnits),
-        u
+        similar(u)
```
MoisaAndrew: A critical bug. `maxeig` method should save the eigenvector as an initial approximation for the future calculations.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_caches.jl
```diff
@@ -35,16 +34,15 @@ function alg_cache(
...
     )
     uᵢ₋₁ = zero(u)
-    uᵢ₋₂ = zero(u)
     tmp = zero(u)
     atmp = similar(u, uEltypeNoUnits)
     recursivefill!(atmp, false)
     fsalfirst = zero(rate_prototype)
     k = zero(rate_prototype)
-    return ROCK2Cache(u, uprev, uᵢ₋₁, uᵢ₋₂, tmp, atmp, fsalfirst, k, constantcache)
+    return ROCK2Cache(u, uprev, uᵢ₋₁, tmp, atmp, fsalfirst, k, constantcache)
```
MoisaAndrew, line 45: Some methods require one fewer storage vector

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -120,9 +120,9 @@ end
         f(k, uᵢ₋₁, p, tᵢ₋₁)
         OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
         tᵢ₋₁ = dt * μ - ν * tᵢ₋₂ - κ * tᵢ₋₃
-        @.. broadcast = false u = (dt * μ) * k - ν * uᵢ₋₁ - κ * uᵢ₋₂
+        @.. broadcast = false u = (dt * μ) * k - ν * uᵢ₋₁ - κ * tmp
         if i < ccache.mdeg
-            @.. broadcast = false uᵢ₋₂ = uᵢ₋₁
```
MoisaAndrew, line 125: It is possible to avoid copying arrays by swapping pointers. Do you think it is worth it?

ChrisRackauckas: It is possible, just need to be careful. It's probably not a huge optimization though. How much does this show up in the profile?

lib/OrdinaryDiffEqStabilizedRK/src/rkc_caches.jl
ChrisRackauckas: should use `copy`. ROCK2 is supposed to work with immutables.

ChrisRackauckas:
```suggestion
        copy(u)
```

ChrisRackauckas: The out-of-place version at least. We should probably add StaticArray testing here if there isn't already for these methods.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
MoisaAndrew: I don't have such a huge problem to see the difference. I think we should leave the code clean until it will be clearly necessary.

ChrisRackauckas:
`zero(u)` is definitely better yes. It does have an edge case with TrackedArrays but I think we can ignore that.

MoisaAndrew:
@ChrisRackauckas what else should be done?

lib/OrdinaryDiffEqStabilizedRK/src/rkc_caches.jl
```diff
@@ -643,11 +631,11 @@ function alg_cache(
     ) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
     min_stage = max(3, alg.min_stages)
     max_stage = max(alg.max_stages, min_stage)
-    constantcache = RKG2ConstantCache(u, min_stage, max_stage)
-    uᵢ₋₁ = zero(u); uᵢ₋₂ = zero(u); tmp = zero(u)
+    constantcache = RKG2ConstantCache(copy(u), min_stage, max_stage)
```
ChrisRackauckas, line 634: should we use `zero` for these too?

ChrisRackauckas:
copy vs zero is inconsequential here. We should probably prefer zero, but ehh it's already copy because I forgot about zero (my fault 😆 ) and it's no big deal so might as well go for it.

PR #3795, opened 2026-06-29
NonlinearSolveAlg: pipe wrapprecs into inner alg and update weight per step

singhharsh1708:
The NonlinearSolveAlg path builds the same `weight` buffer and `wrapprecs(alg.linsolve, W, weight)` as the NLNewton path, but then drops both:

- `_nlalg_with_linsolve(nlalg.alg, alg.linsolve)` pipes the **bare** `alg.linsolve` into the inner NonlinearSolve, discarding the `wrapprecs` wrapper that NLNewton uses.
- `initialize!` for `NLSolver{<:NonlinearSolveAlg, true}` never calls `calculate_residuals!(weight, ...)`, so even if the precs were piped through they'd be acting on a buffer of zeros.

Net effect: every NonlinearSolveAlg user loses the weight-based diagonal preconditioner that NLNewton enjoys on stiff problems — quietly worse Newton convergence with no warning.

## Fix

1. Add a `weight` field to `NonlinearSolveCache` (parametrized as `weightType` so OOP can pass `nothing`).
2. IIP `build_nlsolver`: pipe `wrapprecs(alg.linsolve, W, weight)` into `_nlalg_with_linsolve`, and store `weight` on the cache.
3. OOP `build_nlsolver`: pass `nothing` for weight, matching the existing OOP NLNewton path which also has no weight buffer.
4. IIP `initialize!` for `NonlinearSolveAlg`: update the `weight` buffer via `calculate_residuals!`, mirroring `initialize!` for `NLSolver{<:NLNewton, true}`.

After this, the precs closure built by `wrapprecs` (a `Returns((Pl, Pr))` that captures `_vec(weight)` by reference) sees fresh weights every step, exactly as for NLNewton.

Diff: +14/−4 across `newton.jl`, `type.jl`, `utils.jl`. OOP path is untouched semantically (only the `NonlinearSolveCache` constructor call gets an extra `nothing` arg).

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/test/linear_nonlinear_tests.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/test/linear_nonlinear_tests.jl
@@ -1,5 +1,7 @@
 using OrdinaryDiffEqSDIRK, OrdinaryDiffEqBDF, OrdinaryDiffEqRosenbrock, Test, Random,
     LinearAlgebra, LinearSolve, ADTypes
+using OrdinaryDiffEqNonlinearSolve: NonlinearSolveAlg
+using NonlinearSolve: NewtonRaphson
 Random.seed!(123)
 
 A = 0.01 * rand(3, 3)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Add a test for this

singhharsh1708:
added a test — solves with NonlinearSolveAlg and checks the weight buffer is allocated and gets filled after a step

lib/OrdinaryDiffEqNonlinearSolve/test/linear_nonlinear_tests.jl
```diff
@@ -162,3 +164,23 @@ sol = @test_nowarn solve(
     )
 );
 @test length(sol.t) < 20
+
+sol = @test_nowarn solve(
+    prob,
+    TRBDF2(
+        autodiff = AutoFiniteDiff(),
+        nlsolve = NonlinearSolveAlg(NewtonRaphson())
```
ChrisRackauckas, line 172: The weights are only used with default preconditioners though? So it needs a KrylovJL_GMRES() linsolve test.

singhharsh1708:
Right — the weight only feeds into the default precs path via `wrapprecs`. Added an explicit `KrylovJL_GMRES()` (no user `precs`) + `NonlinearSolveAlg` test that (a) solves cleanly under the GMRES linsolve and (b) checks the weight buffer is allocated and gets filled after a step.

PR #3796, opened 2026-06-29
NLNewton OOP: handle AbstractSciMLOperator W instead of erroring

singhharsh1708:
The OOP NLNewton `compute_step!` hard-errors when `W` is an `AbstractSciMLOperator` that isn't a `WOperator` / `StaticWOperator`:

```julia
elseif W isa AbstractSciMLOperator
    error("Non-concrete Jacobian not yet supported by out-of-place Newton solve.")
end
```

This case is reachable in practice — e.g. an `ImplicitEuler(nlsolve = NLNewton(...))` solve on an OOP `ODEProblem` whose mass matrix is a state-dependent `MatrixOperator` lands here.

## Fix

Mirror the IIP NLNewton handling of `AbstractSciMLOperator` W (already at line 304):

```julia
elseif W isa AbstractSciMLOperator
    update_coefficients!(W, ustep, p, tstep; gamma = γW, transform = true)
    W = convert(AbstractMatrix, W)
end
```

`update_coefficients!(W, ustep, p, tstep; gamma = γW, transform = true)` refreshes the operator against the current Newton iterate, and `convert(AbstractMatrix, W)` materializes the result for the `\` solve that follows. `convert(AbstractMatrix, ::AbstractSciMLOperator)` is defined for `MatrixOperator`, `ScaledOperator`, `AddedOperator`, `ComposedOperator`, `IdentityOperator`, etc., so this covers the common cases.

## Scope

This removes the hard error and makes the OOP NLNewton path functional for SciMLOperator W. The pre-existing failing `Dependent Mass Matrix Tests` / `iip = false` in `mass_matrix_tests.jl` was reaching this error path — after this fix it no longer hard-errors, though the test may still surface a separate numerical issue tied to the test RHS mutating the same `mm_A` it carries as its mass matrix (the IIP version passes, so the integrator path was OK; only this OOP error was blocking).

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
@@ -254,7 +254,8 @@ Equations II, Springer Series in Computational Mathematics. ISBN
     if W isa Union{WOperator, StaticWOperator}
         update_coefficients!(W, ustep, p, tstep)
     elseif W isa AbstractSciMLOperator
-        error("Non-concrete Jacobian not yet supported by out-of-place Newton solve.")
+        update_coefficients!(W, ustep, p, tstep; gamma = γW, transform = true)
+        W = convert(AbstractMatrix, W)
     end
     dz = _reshape(W \ _vec(ztmp), axes(ztmp))
     dz = relax(dz, nlsolver, integrator, f)
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
@@ -254,8 +254,10 @@ Equations II, Springer Series in Computational Mathematics. ISBN
     if W isa Union{WOperator, StaticWOperator}
         update_coefficients!(W, ustep, p, tstep)
     elseif W isa AbstractSciMLOperator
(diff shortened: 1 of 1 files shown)


lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
```diff
@@ -250,7 +250,8 @@ Equations II, Springer Series in Computational Mathematics. ISBN
     if W isa Union{WOperator, StaticWOperator}
         update_coefficients!(W, ustep, p, tstep)
     elseif W isa AbstractSciMLOperator
-        error("Non-concrete Jacobian not yet supported by out-of-place Newton solve.")
+        update_coefficients!(W, ustep, p, tstep; gamma = γW, transform = true)
+        W = convert(AbstractMatrix, W)
```
ChrisRackauckas: No we shouldn't convert W operators.

ChrisRackauckas: I guess for out of place.

ChrisRackauckas: Ehh probably a bad idea though to `\` an operator... better to just give an informative error.

singhharsh1708:
fair — reverted the convert, just kept an informative error pointing users at IIP form or jac_prototype

PR #3753, opened 2026-06-13
NonlinearSolveAlg: extend W-reuse to WOperator with concrete underlying Jacobian

singhharsh1708:
Closes #3697 for the common case.

## Summary

\`build_nlsolver\` previously gated \`use_w_reuse\` on
\`\`\`julia
W isa AbstractMatrix && !(W isa AbstractSciMLOperator)
\`\`\`
which excluded \`WOperator\` entirely. But a \`WOperator\` built around a **concrete** Jacobian (the typical setup when the user picks an iterative linsolve such as \`KrylovJL_GMRES\` with \`concrete_jac = true\`) carries the materialized \`M/γdt - J\` in its \`_concrete_form\` field, and \`_update_nlsolvealg_W!\` already writes the refreshed W there via \`jacobian2W!\`. We can hand that buffer to NonlinearSolve as the analytic Jacobian unchanged.

Truly matrix-free \`WOperator\`s (where \`W.J\` is itself an \`AbstractSciMLOperator\`) still fall back to AD/FD on the inner Newton — handing those off as analytic Jacobians is a deeper change tracked separately.

## Measured

Robertson, \`linsolve = KrylovJL_GMRES(), concrete_jac = true\`, \`nlsolve = NonlinearSolveAlg(NewtonRaphson(autodiff = AutoFiniteDiff()))\`:

| Solver | steps before | steps after | nf before | nf after |
|---|---|---|---|---|
| TRBDF2 | 532 | 394 (−26%) | 8808 | 8416 (−4%) |
| FBDF | 1198 | **479 (−60%)** | 23934 | **10542 (−56%)** |
| KenCarp4 | 533 | 371 (−30%) | 12975 | 9923 (−23%) |

## Audit

- Default (concrete) linsolve paths: byte-identical to master (no regression)
- Matrix-free WOperator path (e.g. \`KrylovJL_GMRES\` without \`concrete_jac=true\`): correctly excluded by the new gate
- \`nested_ad_nlsolvealg_tests.jl\`: 5/5 pass
- Runic-clean

PR #3798, opened 2026-06-29
OOP NonlinearSolveAlg: reuse W with SimpleNonlinearSolve solvers

singhharsh1708:
Out-of-place `NonlinearSolveAlg` rebuilt the stage Jacobian on every inner iteration; when the inner solver is a SimpleNonlinearSolve algorithm and `W` is a `StaticWOperator`, it now hands the integrator's assembled `W` to the inner solver and keeps it across the step's outer iterations.

On the `SVector` van der Pol (`p = 1e3`, TRBDF2, `reltol = 1e-8`) SimpleNewtonRaphson drops from 22532 inner iterations to 14026 over 1343 `W` assemblies; `@allocated step!(integ) == 0` holds for both inner solvers behind a function barrier. The `njacs` column is not a regression, those inner solvers return `stats = nothing` on master.

`WOperator` is left out until SciMLOperators gains `similar` and `copyto!` for it; the remaining allocation overhead is in `NonlinearSolveNoInitCache`'s `solve!`.

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
@@ -1083,11 +1083,11 @@ function build_nlsolver(
             else
                 (tmp, γ, α, tstep, invγdt, DIRK, p, dt, f)
             end
+            inner_alg = _nlalg_with_linsolve(nlalg.alg, alg.linsolve)
             prob = NonlinearProblem(
                 NonlinearFunction{false, SciMLBase.FullSpecialize}(nlf),
                 copy(ztmp), nlp_params
             )
-            inner_alg = _nlalg_with_linsolve(nlalg.alg, alg.linsolve)
             # Zero tolerances: the integrator owns convergence (see the in-place branch above).
             cache = init(
                 prob, inner_alg; verbose = verbose.nonlinear_verbosity,
@@ -1098,7 +1098,26 @@ function build_nlsolver(
                 ),
                 conditioning_kwargs(precondition, postcondition)...
             )
+            W_ref = nothing
             if cache isa NonlinearSolveNoInitCache
+                # This cache only records the arguments a `solve!` will forward, so
+                # building one is the cheapest way to ask whether the inner solver
+                # iterates, and the answer stays right as SimpleNonlinearSolve grows.
+                if !isdae && f.nlstep_data === nothing && W isa StaticWOperator
+                    # StaticWOperator stores inv(W) in its W field for n <= 7, so the
+                    # Ref holds the plain W matrix; initialize! rewrites it before the
+                    # first solve (W_γdt starts at zero).
+                    W_ref = Ref(W.W)
+                    nlf_jac = let W_ref = W_ref
+                        (z, p) -> W_ref[]
+                    end
+                    prob = NonlinearProblem(
+                        NonlinearFunction{false, SciMLBase.FullSpecialize}(
+                            nlf; jac = nlf_jac, jac_prototype = W.W
+                        ),
+                        copy(ztmp), nlp_params
+                    )
+                end
                 # `solve!`-driven fallback cache: it must keep terminating on its own
                 # default (nonzero) tolerances (see the in-place branch above).
                 cache = init(
@@ -1108,7 +1127,8 @@ function build_nlsolver(
             end
             nlcache = NonlinearSolveCache(
                 nothing, tstep, nothing, nothing, invγdt, prob, cache,
-                nothing, nothing, nothing, nothing, nothing, nothing, nothing, nothing,
+                nothing, W_ref, W_ref === nothing ? nothing : uf,
+                nothing, nothing, nothing, nothing, nothing,
                 zero(tstep), true, false, precondition, postcondition
             )
         else
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
@@ -1100,13 +1100,7 @@ function build_nlsolver(
             )
             W_ref = nothing
             if cache isa NonlinearSolveNoInitCache
-                # This cache only records the arguments a `solve!` will forward, so
-                # building one is the cheapest way to ask whether the inner solver
-                # iterates, and the answer stays right as SimpleNonlinearSolve grows.
                 if !isdae && f.nlstep_data === nothing && W isa StaticWOperator
-                    # StaticWOperator stores inv(W) in its W field for n <= 7, so the
-                    # Ref holds the plain W matrix; initialize! rewrites it before the
-                    # first solve (W_γdt starts at zero).
                     W_ref = Ref(W.W)
                     nlf_jac = let W_ref = W_ref
                         (z, p) -> W_ref[]
--- a/lib/OrdinaryDiffEqNonlinearSolve/Project.toml
+++ b/lib/OrdinaryDiffEqNonlinearSolve/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqNonlinearSolve"
 uuid = "127b3ac7-2247-4354-8eb6-78cf4e7c58e8"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
-version = "2.9.3"
+version = "2.9.4"
 
 [deps]
 CommonSolve = "38540f10-b2f7-11e9-35d8-d573e4eb0ff2"
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/OrdinaryDiffEqNonlinearSolve.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/OrdinaryDiffEqNonlinearSolve.jl
@@ -68,7 +68,7 @@ import OrdinaryDiffEqCore: _initialize_dae!,
 import OrdinaryDiffEqDifferentiation: update_W!, is_always_new, build_uf, build_J_W,
     WOperator, StaticWOperator, wrapprecs, default_krylov_warm_start,
     build_jac_config, dolinsolve, set_linear_reltol!,
-    resize_jac_config!, jacobian2W!, jacobian!
+    resize_jac_config!, jacobian2W!, jacobian!, calc_J
 
 import StaticArraysCore: StaticArray
 
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
@@ -44,21 +44,42 @@ function initialize!(
     # A no-init cache has no `stats` (SimpleNonlinearSolve builds every solution with
     # `stats === nothing`), and its `reinit!` only remakes the problem without evaluating
     # the residual, so neither the copies below nor the `+1` apply: nothing is counted and
-    # nothing is invented.
+    # nothing is invented. `njacs`/`nw` for a reused `W` are still recorded by
+    # `_update_nlsolvealg_W_oop!` — that assembly is the integrator's own work.
     if SciMLBase.has_stats(integrator) && !(cache.cache isa NonlinearSolveNoInitCache)
         # The `reinit!` below evaluates the residual at the new `z` and *then* zeroes the
         # inner cache's counters, so that evaluation is never visible in
         # `cache.cache.stats.nf`. Left uncounted it loses one `f` call per stage.
         integrator.stats.nf += cache.cache.stats.nf + 1
-        # Under `W` reuse the inner solver's "Jacobian" is `WReuseJac`, which copies the
-        # `W` assembled here rather than evaluating anything, so its `njacs` counts copies
-        # (it tracks `nw`, not Jacobian evaluations). `_update_nlsolvealg_W!` counts the
-        # real ones. Without reuse the inner solver owns the Jacobian and its count stands.
+        # Under `W` reuse the inner solver's "Jacobian" is the `W_ref` closure, which hands
+        # back the `W` assembled here rather than evaluating anything, so its `njacs` counts
+        # reads, not Jacobian evaluations; `calc_J` inside `_update_nlsolvealg_W_oop!`
+        # counts the real ones. Without reuse the inner solver owns the Jacobian and its
+        # count stands.
         if cache.W === nothing
             integrator.stats.njacs += cache.cache.stats.njacs
         end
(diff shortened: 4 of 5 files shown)


lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
```diff
@@ -490,21 +490,35 @@ function build_nlsolver(
...
             else
                 (tmp, γ, α, tstep, invγdt, DIRK, p, dt, f)
             end
-            prob = NonlinearProblem(
-                NonlinearFunction{false, SciMLBase.FullSpecialize}(nlf),
-                copy(ztmp), nlp_params
-            )
+            prob = if use_w_reuse
+                nlf_jac = let W = W
+                    (z, p) -> convert(AbstractMatrix, W)
```
ChrisRackauckas: We shouldn't unconditionally convert.

singhharsh1708:
good catch — split the closure so StaticWOperator returns its .W field directly, only WOperator path converts now

lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
```diff
@@ -498,21 +498,41 @@ function build_nlsolver(
...
-                copy(ztmp), nlp_params
-            )
+            prob = if use_w_reuse
+                nlf_jac = if W isa StaticWOperator
+                    let Ws = W
+                        (z, p) -> Ws.W
+                    end
+                else
+                    let Ww = W
+                        (z, p) -> convert(AbstractMatrix, Ww)
```
ChrisRackauckas: What about non-static opertators

singhharsh1708:
for non-static `WOperator` we shouldn't unconditionally convert either.
Updated to mirror the IIP gate from #3753: `use_w_reuse` now requires either `StaticWOperator` or a `WOperator` whose underlying `J` is concrete (not an `AbstractSciMLOperator`). The else branch hands NonlinearSolve the cached `Ww._concrete_form` directly via the closure (no `convert` call). Matrix-free / non-concrete `WOperator` cases fall back to the non-w-reuse path.

lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
```diff
@@ -498,21 +498,43 @@ function build_nlsolver(
...
-                copy(ztmp), nlp_params
-            )
+            prob = if use_w_reuse
+                nlf_jac = if W isa StaticWOperator
+                    let Ws = W
+                        (z, p) -> Ws.W
+                    end
+                else
+                    let Ww = W
+                        (z, p) -> Ww._concrete_form
```
ChrisRackauckas: use the operator...

singhharsh1708:
fixed — now hands the operator through directly

singhharsh1708:
Hi @ChrisRackauckas — quick nudge on this stack when you have a moment. All three PRs (#3795, #3796, #3798) have addressed the earlier review comments and are rebased on the latest master:

- **#3795** — Added the explicit `KrylovJL_GMRES()` linsolve test you asked for, plus a weight-buffer check that verifies it's allocated and filled after one step.
- **#3796** — Reverted the operator convert; kept the informative error you suggested for the OOP path.
- **#3798** — Mirrored the IIP gate from #3753 (`use_w_reuse` requires `StaticWOperator` or a `WOperator` with a concrete underlying J), and now hands the operator through directly instead of `Ww._concrete_form` (which was stale between Newton iterations since `update_coefficients!` doesn't refresh `_concrete_form`).

Once the next SciMLOperators release tags, I'll open a follow-up here to route OOP + SciMLOperator mass matrix through `WOperator` and drop #3796's informative error for that path.

No rush — just wanted to make sure these weren't waiting on me. Thanks!

ChrisRackauckas-Claude:
Ran this branch against an OOP nonlinear problem and hit two defects in the W-reuse path — posting reproductions so they can be addressed:

**1. The `WOperator` jac closure crashes on first Jacobian use (plain `Vector` problems).**

```julia
using OrdinaryDiffEqSDIRK, OrdinaryDiffEqNonlinearSolve, NonlinearSolve, ADTypes
using OrdinaryDiffEqNonlinearSolve: NonlinearSolveAlg
vdp(u, p, t) = [u[2], p[1] * ((1 - u[1]^2) * u[2] - u[1])]
prob = ODEProblem(vdp, [2.0, 0.0], (0.0, 1.0), [1.0e3])
integ = init(prob, TRBDF2(nlsolve = NonlinearSolveAlg(NewtonRaphson(; autodiff = AutoForwardDiff())), concrete_jac = true); reltol = 1e-8, abstol = 1e-10)
step!(integ)
# ERROR: MethodError: Cannot `convert` an object of type WOperator{false, ...} to an object of type Matrix{Float64}
```

The `(z, p) -> Ww` closure hands the raw `WOperator` to NonlinearSolve, whose jacobian cache holds a dense `Matrix` (no `jac_prototype` on the inner `NonlinearFunction`), so the assignment tries `convert(Matrix, WOperator)` and throws. `init` succeeds — it dies on the first `step!` that evaluates the jac, which is also why this wasn't caught: the PR currently has no tests. A test exercising exactly the `TRBDF2 + concrete_jac=true + Vector u0` path above (plus a `StaticArrays` one for `StaticWOperator`) would have flagged it.

**2. Even with the convert fixed, the Jacobian values are never refreshed — J is frozen at its build-time state for the entire integration.**

The update in `initialize!` is `update_coefficients!(cache.W, uprev, p, tstep; gamma = dtgamma)`. For `WOperator` that forwards to `update_coefficients!(W.J, u, p, t)` (SciMLOperators `woperator.jl`), which is a **no-op when `W.J` is a plain matrix** — and the `use_w_reuse` gate specifically selects for a concrete non-operator `W.J`. So only `gamma` ever updates; the J entries stay at their `build_nlsolver`-time values. (For `StaticWOperator` it's worse: `Ws.W` is a fixed matrix, so not even gamma changes.) Newton silently degrades to a chord iteration with J(u₀, t₀) forever — no failure on mildly nonlinear problems, just quietly worse convergence, and on strongly state-dependent J (e.g. the VdP case above) divergence→step-rejection loops.

The IIP counterpart (#3753) avoids this because `_update_nlsolvealg_W!` explicitly recomputes J via `f.jac`/`jacobian!` before `jacobian2W!`. The OOP path needs the same: an explicit J recompute (the OOP analog is `calc_J`/`jacobian(uf, uprev, integrator)`) and a W rebuild — for `StaticWOperator`, reconstructing `StaticWOperator(J - mass_matrix * inv(dtgamma))`; for `WOperator`, refreshing the concrete J and gamma together — inside the `should_update` branch.

Also worth folding in while touching this: pass `jac_prototype` matching W's structure to the inner `NonlinearFunction` — without it NonlinearSolve allocates a dense J, which is the direct cause of crash 1 and also silently densifies sparse problems (I've opened a separate PR fixing that on the IIP path).

singhharsh1708:
Thanks for the review @ChrisRackauckas-Claude — both defects reproduced and fixed. New commit `bd5af110`:

**Defect 1 (crash on convert):** the OOP closure now returns the concrete matrix directly — `W_ref[].W` for `StaticWOperator`, `convert(AbstractMatrix, W_ref[])` for `WOperator` — and I also pass `jac_prototype` matching that shape to the inner `NonlinearFunction`. NonlinearSolve no longer tries to `convert(Matrix, WOperator)`.

**Defect 2 (frozen J):** added `_update_nlsolvealg_W_oop!` (mirrors the IIP `_update_nlsolvealg_W!`) — inside `should_update` we now recompute J via `f.jac(uprev, p, tstep)` or `jacobian(uf, uprev, integrator)`, then either `copyto!(W.J, J_new); W.gamma = dtgamma` for `WOperator` or rebuild `StaticWOperator(J_new - mass_matrix * inv(dtgamma))` and swap the Ref for `StaticWOperator`. `uf` is now stored in the nlcache so `jacobian(uf, ...)` works on the OOP path.

**Ref indirection:** the closure captures `W_ref` instead of `W` directly, so the `StaticWOperator` swap in `initialize!` is visible to NonlinearSolve without rebuilding the outer `NonlinearProblem`. For `WOperator` this is just a passthrough since it's already mutable.

**Tests:** added two regressions at the end of `linear_nonlinear_tests.jl` covering exactly the bot's two failure modes:
- Plain `Vector` VdP + `TRBDF2 + concrete_jac = true` — used to crash; now succeeds with `njacs > 0`.
- `SVector` VdP variant — same, exercising the `StaticWOperator` path.

Locally: both new tests pass; existing `linear_nonlinear_tests` suite still passes.

lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
```diff
@@ -498,21 +498,54 @@ function build_nlsolver(
...
+                        NonlinearFunction{false, SciMLBase.FullSpecialize}(
+                            nlf; jac = nlf_jac, jac_prototype = jac_prototype
+                        ),
+                        copy(ztmp), nlp_params
+                    )
+                else
+                    nlf_jac = let W_ref = W_ref
+                        (z, p) -> convert(AbstractMatrix, W_ref[])
+                    end
+                    jac_prototype = copy(convert(AbstractMatrix, W))
```
ChrisRackauckas: no it's still concretizing.

lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
```diff
@@ -499,21 +499,52 @@ function build_nlsolver(
...
+                    end
+                    NonlinearProblem(
+                        NonlinearFunction{false, SciMLBase.FullSpecialize}(
+                            nlf; jac = nlf_jac, jac_prototype = W.W
+                        ),
+                        copy(ztmp), nlp_params
+                    )
+                else
+                    nlf_jac = let W_ref = W_ref
+                        (z, p) -> convert(AbstractMatrix, W_ref[])
```
ChrisRackauckas: OOP is purposefully not mutable. We want to make things fold away when static. This will violate that. I think it's simply not possible to do similar W-reuse in the OOP path without violating the principle of making this path non-mutating.

singhharsh1708: For StaticWOperator the payload is an immutable SMatrix — reassigning nlcache.W[] re-points the Ref cell, the SMatrix itself is never mutated, so fold-away on the value side still holds. The mutation you're objecting to only shows up on the WOperator branch (copyto!(W.J, …)).
And picking NonlinearSolveAlg already forces a Matrix{Float64} jac cache inside construct_jacobian_cache — the OOP path isn't fold-away in that mode regardless of what this PR does. If Static-only reuse is still off the table given that, I'll close.

ChrisRackauckas:
> On "still concretizing": NonlinearSolveBase.construct_jacobian_cache calls safe_similar(jac_prototype) and allocates a Matrix{Float64} regardless of what the closure returns — so the OOP jac already flows through a mutable buffer once NonlinearSolveAlg is picked. If that's the real blocker, the fix belongs upstream (similar/copyto! for WOperator in SciMLOperators), not here.

That should get fixed upstream. Making everything dense is not a valid answer.

lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
ChrisRackauckas: I guess it's fine if there's a simplediffeq path that fully non-allocates with simplenonlinearsolve. But the conversion stuff is never okay.

singhharsh1708:
Dropped WOperator branch. W-reuse only fires on StaticWOperator + AbstractSimpleNonlinearSolveAlgorithm — SMatrix rebuilt, Ref re-pointed, no convert.

ChrisRackauckas:
Why no WOperator?

singhharsh1708:
@ChrisRackauckas Rebased on master and fixed two bugs the Static+Simple gate exposed. Both surfaced because that path introduces execution paths that weren't previously exercised.

**1. Inner cache doesn't support `step!`**

`SimpleNonlinearSolve` algorithms `init` to a `NonlinearSolveNoInitCache`, which only implements `solve!`. `compute_step!` called `step!` on it:

```
FieldError: type NonlinearSolveNoInitCache has no field `force_stop`
```

`compute_step!` now branches on the cache type and calls `solve!` for that case, pulling stats off the returned solution.

**2. `StaticWOperator.W` holds `inv(W)` for n ≤ 7**

`StaticWOperator`'s constructor sets `isinv = n <= 7` and then stores `inv(W)` in the `W` field. My jac closure returned `W_ref[].W`, so the inner Newton was handed the *inverse* Jacobian. This didn't crash — VdP just ran with wrong math, dt collapsed to 5e-324, and it bailed with `Unstable` and `njacs = 0`.

The Ref now holds the plain `J - M/γdt` SMatrix and the closure returns that directly. Verified against a ForwardDiff-computed Jacobian:

```
closure W  = [-7.032484918065255e9 1.0; -1000.0 -7.032487918065255e9]
J - M/γdt  = [-7.032484918065255e9 1.0; -1000.0 -7.032487918065255e9]
match: true
```

**Test change**

`sol.stats.njacs > 0` doesn't work as a reuse assertion here — `SimpleNonlinearSolve` returns `stats = nothing`, so the counter never moves regardless. Replaced with a check that the W Ref is live on the cache plus an accuracy comparison against `TRBDF2()` with NLNewton (end-state diff 2e-4).

**Validation**

All 11 `OrdinaryDiffEqNonlinearSolve` Core suites pass locally on this branch (Julia 1.12), including Mass Matrix, W-Operator Prototype, and NSA Jacobian Reuse. Runic clean.

Remaining CI reds are unrelated to this branch — `constructorof`/`TimeDerivativeWrapper` on InterfaceI/Downstream, and the nlstep order failures that #3818 addresses.

singhharsh1708:
rebased onto master now that #4137 landed. the crash guards moved there, so this is down to just the W reuse your allocation question was about: +99/-15 in src, one commit. the reuse numbers are unchanged from the July discussion, SimpleNewtonRaphson on the SVector van der Pol drops from 22532 inner iterations to 14026 over 1343 W assemblies, and the remaining allocation overhead is upstream in NonlinearSolveNoInitCache's solve!/reinit! as before.

singhharsh1708:
Because the upstream piece is still missing. NonlinearSolve's Jacobian cache does `similar` / `copyto!` on whatever the jac closure returns, and SciMLOperators has neither for `WOperator` as of today's master (`woperator.jl` last touched 2026-08-26 for the staleness flag, still no `Base.similar` or `Base.copyto!`). So a `WOperator` on that path either densifies or errors, and you said densifying is not an answer. `StaticWOperator` is fine because its `.W` is an `SMatrix` that can be handed through as is.

Once SciMLOperators gains `similar`/`copyto!(::WOperator)` the gate here is a one line widening. Rebased onto master.

PR #3799, opened 2026-06-29
NonlinearSolveCache: define Base.resize! and guard firstcall write

singhharsh1708:
`resize!(integrator, N)` on an integrator using `NonlinearSolveAlg` blows up two ways before reaching the actual nlsolver work:

1. `resize!(nlsolver.cache, nlsolver, integrator, i)` falls through to the default at `utils.jl:740`, which calls `Base.resize!(cache::NonlinearSolveCache, i::Int)` — undefined, so MethodError.
2. Then `resize_nlsolver!` does `nlsolver.cache.firstcall = true` unconditionally, but `NonlinearSolveCache` has no `firstcall` field (only `NLNewtonCache`/`NLNewtonConstantCache` do — `NLFunctionalCache` and `NLAndersonCache` also lack it).

## Fix

1. Define `Base.resize!(::NonlinearSolveCache, ::AbstractNLSolver, integrator, i::Int)` mirroring the existing `NLNewtonCache` resize: resize the buffer fields that are populated (`ustep`, `k`, `atmp`, `du1`, `jac_config`, `J/W` via `resize_J_W!`), then `SciMLBase.reinit!(nlcache.cache, zero(nlcache.ustep))` to bring the inner NonlinearSolve cache's `u`/`fu`/`J` to the new size.
2. Guard the `firstcall` write in `resize_nlsolver!` with `isnewton(nlsolver) && (...)`, matching the same pattern already in `postamble!` at `nlsolve.jl:197`.

The `===` `nothing` guards on each field handle both IIP (fields populated) and OOP (most fields `nothing`).

Note: the `firstcall` guard line overlaps with PR #3794 — same one-line change to the same site. Settles cleanly on merge regardless of order.

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
@@ -670,3 +670,18 @@ function Base.resize!(nlcache::NLNewtonCache, ::AbstractNLSolver, integrator, i:
 
     return nothing
 end
+
+function Base.resize!(nlcache::NonlinearSolveCache, ::AbstractNLSolver, integrator, i::Int)
+    nlcache.ustep === nothing || resize!(nlcache.ustep, i)
+    nlcache.k === nothing || resize!(nlcache.k, i)
+    nlcache.atmp === nothing || resize!(nlcache.atmp, i)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Needs a test. Use NonlinearSolveAlg in the cache growing event tests as a good test case for this.

PR #628, opened 2019-01-24
Implemented method ROCK4

deeepeshthakur:
@ChrisRackauckas @YingboMa Implemented ROCK4 from original Fortran Code. It's passing all the convergeence tests.
**Convergence Tests**
```
import DiffEqProblemLibrary.ODEProblemLibrary: prob_ode_linear
dts = 1 .//2 .^(8:-1:4)
testTol = 0.1
sim = test_convergence(dts,prob_ode_linear,ROCK4())
plot(sim)
```
image

```
import DiffEqProblemLibrary.ODEProblemLibrary: prob_ode_2Dlinear
dts = 1 .//2 .^(8:-1:4)
testTol = 0.1
sim = test_convergence(dts,prob_ode_2Dlinear,ROCK4())
plot(sim)
```
image

#2

PR #3810, opened 2026-07-02
Allow CUDA.jl v6 in GPU test environments

lamdor:
## Summary

The GPU test environments pin CUDA.jl to `"4, 5"`. The V100 GPU CI runners now run nvidia driver 580, which requires CUDA.jl **v6.2+** — versions below 6.2 select the CUDA runtime based on the driver version and pick an incompatible runtime, causing GPU CI to fail (e.g. [this run](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/28388992836/job/84170145141)). The fix landed in [JuliaGPU/CUDA.jl#3134](https://github.com/JuliaGPU/CUDA.jl/pull/3134) (see also [issue #3079](https://github.com/JuliaGPU/CUDA.jl/issues/3079)).

This widens the CUDA compat in the GPU test environments to `"6"` so the resolver picks the latest 6.x (currently ≥6.2), which contains the runtime-selection fix.

## Changes

- Bump `CUDA` compat from `"4, 5"` to `"6"` in the 10 `test/gpu/Project.toml` files: the root test env plus `OrdinaryDiffEqBDF`, `OrdinaryDiffEqCore`, `OrdinaryDiffEqDefault`, `OrdinaryDiffEqLinear`, `OrdinaryDiffEqLowStorageRK`, `OrdinaryDiffEqRKIP`, `OrdinaryDiffEqRosenbrock`, `OrdinaryDiffEqStabilizedRK`, and `StochasticDiffEq`.

Test-environment-only; no package compat or version is affected.

### DIFF
--- a/lib/OrdinaryDiffEqBDF/test/gpu/Project.toml
+++ b/lib/OrdinaryDiffEqBDF/test/gpu/Project.toml
@@ -7,5 +7,5 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 path = "../../../OrdinaryDiffEqBDF"
 
 [compat]
-CUDA = "4, 5"
+CUDA = "6"
 OrdinaryDiffEqBDF = "2"
--- a/lib/OrdinaryDiffEqCore/test/gpu/Project.toml
+++ b/lib/OrdinaryDiffEqCore/test/gpu/Project.toml
@@ -24,7 +24,7 @@ path = "../../../DiffEqBase"
 [compat]
 ADTypes = "1"
 Adapt = "4"
-CUDA = "4, 5"
+CUDA = "6"
 ComponentArrays = "0.15"
 DiffEqBase = "7"
 FastBroadcast = "1.3"
--- a/lib/OrdinaryDiffEqDefault/test/gpu/Project.toml
+++ b/lib/OrdinaryDiffEqDefault/test/gpu/Project.toml
@@ -11,6 +11,6 @@ path = "../../../.."
 path = "../../../OrdinaryDiffEqLowOrderRK"
 
 [compat]
-CUDA = "4, 5"
+CUDA = "6"
 OrdinaryDiffEq = "7"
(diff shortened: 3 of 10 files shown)


lamdor:
The `lib/OrdinaryDiffEqLinear [GPU]` job failure here is **not** caused by this compat change — it's a pre-existing upstream bug in the GPUArrays/cuSPARSE sparse interface that the bump to CUDA v6.x surfaces. The environment resolves and the GPU tests run fine; the other 5 sublibrary GPU jobs pass. Only `OrdinaryDiffEqLinear`'s `Linear Exponential GPU` test hits it, because it's the only one calling `opnorm` on a `CuSparseMatrixCSC`.

The error:

```
Linear Exponential GPU: Error During Test
  LoadError: MethodError: no method matching csr_type(::cuSPARSE.CuSparseMatrixCSC{Float32, Int32})
  Closest candidates are:
    csr_type(::Type{<:cuSPARSE.CuSparseMatrixCSC})
```

Root cause — signature mismatch between the two packages:

- GPUArrays `src/host/sparse.jl:80` calls `csr_type` with an **instance**: `opnorm(A::AbstractGPUSparseMatrixCSC, p) = opnorm(csr_type(A)(A), p)`
- cuSPARSE `lib/cusparse/src/array.jl:116` only defines it for a **type**: `csr_type(::Type{<:CuSparseMatrixCSC}) = CuSparseMatrixCSR`

So `csr_type(A)` has no method. It should be `csr_type(typeof(A))(A)` (or GPUArrays should add an instance-forwarding method). This regressed via the GPUArrays sparse rework (#628/#650 introduced the `opnorm` path; #700/#709 moved the `*_type` hooks to type-argument form). No upstream issue exists for it yet, and GPUArrays #745 is actively reworking this same interface.

This is a genuine upstream fix (GPUArrays/cuSPARSE), independent of this PR — the compat bump itself is doing what it should (CUDA v6.x now resolves and runs on the V100 runners).

Issue #961, opened 2019-11-13
Implicit multirate GARK methods


ChrisRackauckas:
https://arxiv.org/abs/1910.14079

PR #3811, opened 2026-07-02
Multirate: add solve-decoupled implicit MRI-GARK-IRK21a (order 2) (#961)

singhharsh1708:
## Summary

Adds the first **implicit** member of the MRI-GARK family (Sandu 2019), `MRIGARKIRK21a` — a 2nd-order **solve-decoupled implicit** multirate infinitesimal method — to `OrdinaryDiffEqMultirate`. This is a step toward #961.

It solves `SplitODEProblem`s `du/dt = f1(u,t) + f2(u,t)` where `f1` is the fast component and `f2` the slow component (SciML convention). Unlike the explicit MRI-GARK methods already in the package, the slow stages couple implicitly to `f2` at the stage endpoint, so it stays stable when the **slow** subsystem is itself stiff — the case explicit MRI-GARK cannot handle.

## Approach

Because the SUNDIALS/Sandu implicit MRI-GARK methods are *solve-decoupled*, the implicit diagonal coupling only appears on the `Δc = 0` quadrature stages, which reduce to the algebraic implicit solve

```
z_{i+1} - dt*γ*f2(z_{i+1}) = z_i + dt*Σ_j ω̄_ij*f2(z_j)
```

So the fast sub-integration remains fully explicit and reuses the existing substage kernels verbatim; only the `Δc = 0` stage adds a nonlinear solve.

- `MRIGARKTableau` gains a per-stage implicit endpoint-coupling vector `γ0`, all-zero for the four existing explicit methods (which keep their exact previous code path).
- `MRIGARKIRK21a` is an `OrdinaryDiffEqNewtonAdaptiveAlgorithm`; the implicit stage is solved with `build_nlsolver`/`nlsolve!`. Since multirate solves implicitly in the **slow** part, `nlsolve_f` is specialized to target `f.f2` (rather than the usual `f.f1`).
- IIP and OOP `perform_step!`, cache, and tableau added.

Coefficients transcribed from SUNDIALS `arkode_mri_tables.def` (`MIStoMRI` of the underlying DIRK Butcher table).

## Tests

Added a `MRIGARKIRK21a` testset (construction, scalar OOP, vector IIP, stiff slow component, convergence). Convergence **order 2** verified for both OOP and IIP; the existing explicit MRI-GARK methods are unchanged and still pass. Full `OrdinaryDiffEqMultirate` suite (Core + QA/AllocCheck) is green.

### DIFF
--- a/lib/OrdinaryDiffEqMultirate/Project.toml
+++ b/lib/OrdinaryDiffEqMultirate/Project.toml
@@ -1,20 +1,24 @@
 name = "OrdinaryDiffEqMultirate"
 uuid = "d4b830b4-ac80-426b-8507-16693d424963"
 authors = ["singhharsh1708 <hs1663531@gmail.com>"]
-version = "2.3.0"
+version = "2.4.0"
 
 [deps]
+ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 FastBroadcast = "7034ab61-46d4-4ed7-9d0f-46aef9175898"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MuladdMacro = "46d2c3a1-f734-5fdb-9937-b9b9aeba4221"
 OrdinaryDiffEqCore = "bbf590c4-e513-4bbe-9b18-05decba2e5d8"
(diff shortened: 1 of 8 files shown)


ChrisRackauckas (approved):

singhharsh1708:
@ChrisRackauckas can you check this once :https://github.com/SciML/SciMLOperators.jl/pull/397

Issue #3821, opened 2026-07-05
SDIRK smooth_est: W⁻¹-filtered error estimate underestimates VDP local error by ~3 orders; silently disabled under NonlinearSolveAlg (isnewton gate)


ChrisRackauckas-Claude:
Investigation follow-up from the NonlinearSolveAlg benchmarking work (#3819/#3820 series). Two intertwined findings, both empirically verified on VdP μ=1e5, `tspan=(0,6.3)`, rtol=1e-8/atol=1e-11, reference Rodas5P@1e-13.

## Finding 1: the smoothed estimate can catastrophically underestimate

`TRBDF2()` (default `smooth_est=true`) returns final relative error **4.3e-2** at rtol=1e-8. A per-step audit of the accepted trajectory shows:

- Replaying every accepted step with fully-converged nonlinear solves (`NLNewton(κ=1e-8, max_iter=200, always_new=true)`) changes results by ≤ **0.042 tolerance units** (median 0.0016) → the nonlinear solves and all Newton acceptance heuristics (η·ndz<κ, iter-1 shortcut, θ≈1 branch, relaxation) are **sound**; this is not a nonlinear-solver problem.
- True local error of accepted steps (vs Rodas5P@1e-13 restarts): **median 3.3e3× tol, p90 6.1e5×, max 3.4e6×**; 927/1133 accepted steps exceed 100× tol. The `W⁻¹`-filtered (Hosea–Shampine smoothed) embedded estimate is blind to the dominant error mode here.
- κ-sweep control: tightening NLNewton's κ from 1e-2 to 1e-6 leaves the error at ~2–4e-2 — confirming the acceptance threshold is not the lever.

One-variable isolation (only `smooth_est` changed):

| config | naccept | nreject | final rel err |
|---|---|---|---|
| TRBDF2 smooth_est=true (default) | 1,133 | 443 | 4.3e-2 |
| TRBDF2 smooth_est=false | 45,261 | 3,839 | 9.6e-7 |
| KenCarp4 smooth_est=true | 498 | 1 | 3.2e-2 |
| KenCarp4 smooth_est=false | 5,825 | 674 | 6.9e-8 |

[...]

PR #3820, opened 2026-07-04
NonlinearSolveAlg: reuse stored J on dt-only W updates (do_newJW parity)

ChrisRackauckas-Claude:
> **Please ignore until reviewed by @ChrisRackauckas.** Draft opened by an agent.

## Problem

The W-reuse path recomputes the Jacobian on **every** W update, including pure dt/γ changes. The legacy `do_newJW` policy separates the two decisions: a dt change only reassembles `W = J - M/γdt` from the stored J via `jacobian2W!` (O(nnz)); a fresh Jacobian evaluation is reserved for first use, `always_new`, or a diverged iteration.

Measured on the dense-AD 512-unknown brusselator at rtol 1e-8 (master): `FBDF+NLNewton` evaluates **1 Jacobian for the whole solve** (nf = 2,764) while `FBDF+NSA` evaluates **295** (nf = 152,393) — every dt-driven W refresh pays a full 512-dim AD Jacobian.

## Fix

`_update_nlsolvealg_W!` takes a `new_jac` flag; `initialize!` computes `new_jac = first_call || always_new || status === Divergence` separately from the dt-change test, mirroring `do_newJW`'s `jbad`/`wbad` split. dt-only updates reuse `cache.J` and only rerun `jacobian2W!`.

The expected trade (same one NLNewton makes) is visible on cheap-J ROBER: user-jac calls drop ~100 → ~4, iteration counts rise to NLNewton's level (1433 vs NLNewton's 1320, from 911 with always-fresh J), wall time unchanged. On expensive-J problems it's decisive — dense brusselator NSA/NLNewton time ratios, master → with this series:

| alg | rtol 1e-4 | 1e-6 | 1e-8 |
|---|---|---|---|
| FBDF | 5.0× → **0.96×** | 5.6× → 1.10× | 7.5× → 1.29× |
| QNDF | 4.1× → **0.92×** | 6.8× → 1.16× | 7.9× → **0.88×** |
| KenCarp4 | 17× → 1.20× | 39× → 1.79× | 66× → 1.72× |
| TRBDF2 | 20× → 1.65× | 55× → 2.24× | 70× → 1.49× |

(together with #3819 and SciML/NonlinearSolve.jl#1003; the remaining TRBDF2/KenCarp4 gap is step-count inflation from dt-controller/η coupling, a separate issue.)

## Tests

[...]

PR #3819, opened 2026-07-04
NonlinearSolveAlg: give inner NonlinearFunction W's structure as jac_prototype

ChrisRackauckas-Claude:
> **Please ignore until reviewed by @ChrisRackauckas.** Draft opened by an agent.

## Bug

Under W-reuse (#3693/#3753) the inner `NonlinearFunction` gets the analytic-jac closure `copyto!(J_out, W)` but **no `jac_prototype`**. NonlinearSolveBase allocates a **dense** J for analytic-jac functions without a prototype (`jacobian.jl`: `safe_similar(fu, n, n)`), so for sparse/structured W:

- with the ODE's sparse-only linsolve piped in (`_nlalg_with_linsolve`): **hard crash** — `FBDF(linsolve = KLUFactorization(), nlsolve = NonlinearSolveAlg(NewtonRaphson()))` on a sparse-`jac_prototype` brusselator dies with `FieldError: Matrix has no field nzval`;
- with the default linsolve: silent dense LU on a sparse system.

## Fix

Pass `jac_prototype = similar(W_for_reuse)` so the inner Jacobian mirrors W's type and structure; the `copyto!` closure is then structure-preserving and sparse factorizations work.

Verified on the 512-unknown sparse brusselator (before → after): inner J `Matrix{Float64}` → `SparseMatrixCSC` (nnz=3072), KLU crash → clean solve. Combined with the J/W-update split and SciML/NonlinearSolve.jl#1003, `FBDF+NSA+KLU` reaches **1.00×** of `FBDF+NLNewton+KLU` wall time (220.3 vs 219.5 ms) — from a crash/3.9× before.

## Tests

New `nsa_sparse_tests.jl`: asserts the inner NonlinearSolve J is `SparseMatrixCSC` under W-reuse, and that FBDF/TRBDF2 + NSA + `KLUFactorization` solve a sparse-`jac_prototype` problem to the reference (6/6 pass locally). Local suite runs on this branch (Julia 1.12): Newton 6/6, Linear Nonlinear 36/36, W-Operator Prototype, NSA Sparse 6/6, Sparse Algebraic 12/12, Sparse DAE Init 27/27, Linear Solver 113/113, Split ODE 10/10 — no failures (full Core group exceeds a 1h local window; remaining suites left to CI).

Part of a series: #3816 (nsolve stat), #3818 (recompute_jacobian on W===nothing paths), SciML/NonlinearSolve.jl#1003 (refactorization reuse), and a J/W-update-split PR to follow.

[...]

Issue #3817, opened 2026-07-04
NonlinearSolveAlg(TrustRegion) + W-reuse: displacement-based ndz declares convergence on rejected TR steps → silently wrong solutions


ChrisRackauckas-Claude:
Found while investigating the master `nlstep_tests.jl` FBDF+NSA(TrustRegion) order failures (see #3807 discussion; the order-test failure itself is fixed separately).

## Symptom

Fixed-dt `FBDF + NonlinearSolveAlg(TrustRegion())` on Robertson (analytic `jac`, plain W-reuse path, no MTK nlstep) produces a final-state error of **4.741e-2, constant across dt = 2^-15 … 2^-18** — a dt-independent error floor, i.e. silently wrong results with `Success` retcodes. Stats show exactly **1 inner iteration and 1 f-eval per step**.

## Mechanism

`compute_step!` for `NonlinearSolveAlg` computes the convergence norm from the **displacement**: `ndz = internalnorm(calculate_residuals(z .- nlcache.u, ...))` (newton.jl, from #3693). When TrustRegion **rejects or truncates** its internal step, `nlcache.u` barely moves, so `ndz ≈ 0` — and the iter-1 η-convergence shortcut (`iter == 1 && ndz < 1e-5`, plus the first-iter trust gate enabled for `cache.W !== nothing` in nlsolve.jl) declares the stage converged with essentially **no correction applied**. A rejected globalization step reads as "converged" to the outer loop.

Two related consequences observed with the same mechanism:
- The θ≈1 stall detection (`eps_around_one` branch in nlsolve.jl) misfires for globalized inner solvers: a rejected TR step leaves `fu` unchanged → θ = 1 exactly → premature Divergence (after 2 iterations) or premature convergence depending on `ndz`.
- On a *freshly built* MTK nlstep problem, `FBDF + NSA(TrustRegion)` at dt=2^-15 aborts at step 0 with `ConvergenceFailure` after 2 inner iterations via that θ≈1 path (test files only pass because earlier testsets warm the shared `nlprob`; see also the companion NonlinearSolve.jl issue about `reinit!` resetting the trust radius to its build-time value).

## Proposed fix direction

[...]

PR #3823, opened 2026-07-05
NonlinearSolveAlg: apply smooth_est error smoothing under W-reuse (fixes silent isnewton gating)

ChrisRackauckas-Claude:
> **Please ignore until reviewed by @ChrisRackauckas.** Draft opened by an agent.

Fixes the `NonlinearSolveAlg` half of #3821.

## Bug

The SDIRK/ESDIRK smoothed error estimate (`est = W \ tmp`, `smooth_est = true` default) was gated on `isnewton(nlsolver)` — false for `NonlinearSolveAlg` — so NSA silently error-controlled on the **raw** embedded estimate no matter what `smooth_est` was set to. Same `TRBDF2()` call, 40× different step counts and orders-different accuracy depending on the inner nonlinear solver choice.

## Fix

- `NonlinearSolveCache` gains a `dz` buffer and an **estimator-only** linsolve on W, built only under W-reuse. The inner NonlinearSolve keeps owning the Newton factorization; the estimator linsolve refactorizes exactly once per W refresh (`_update_nlsolvealg_W!` re-sets its `A` after rewriting W in place) and reuses the factorization otherwise — one triangular solve per step.
- The two gates in `generic_imex_perform_step.jl` become `can_smooth_est(nlsolver)`: true for Newton caches, true for NSA when the estimator linsolve exists. OOP (no W-reuse on master) keeps the raw estimate as before — zero behavior change there.
- resize! rebuilds the estimator linsolve against the reallocated W.

## Verified locally (VdP μ=1e5, rtol 1e-8; naccept / final rel err)

| config | smooth=true | smooth=false |
|---|---|---|
| TRBDF2+NLNewton | 1133 / 4.3e-2 | 45261 / 9.6e-7 |
| TRBDF2+NSA | **1341 / 3.2e-2** (was 44949) | **45237 / 6.8e-7** |
| KenCarp4+NLNewton | 498 / 3.2e-2 | 5825 / 6.9e-8 |
| KenCarp4+NSA | **570 / 1.5e-2** (was 6386) | — |

The `smooth_est` knob now acts identically for both solvers. HIRES rtol 1e-8: TRBDF2+NSA **715 steps / 9.4 ms** vs NLNewton 690 / 5.4 ms (before: 2840 steps / 21.1 ms) — this closes the TRBDF2/KenCarp4 step-inflation gap from the #3819/#3820 benchmark series. ROBER sanity: 906 vs 998 steps, errors normal.

Note this deliberately makes NSA **match NLNewton's estimator behavior including its VDP accuracy trap** (median true local error 3.3e3× tol on accepted steps — the audit in #3821). Whether the smoothed estimator itself should change is the remaining, separate design question in that issue.

## Tests

New `nsa_smooth_est_tests.jl`: asserts `smooth_est` changes NSA behavior (pre-fix: identical step counts) and that NSA tracks NLNewton under both settings on VdP. Passes locally along with Newton and Linear-Nonlinear suites (Julia 1.12).

---
**Rebased 2026-07-12, now stacked on #3819.** Depends on #3819 — please merge that first. Rationale: #3823 gives the inner Jacobian a `jac_prototype`, which requires #3819's `WReuseJac` Ref-swap for correct `resize!`; keeping them stacked avoids duplicating that machinery. Resolved conflicts with #3795 (`weight` field) and #3820 (`_update_nlsolvealg_W!` `new_jac` split) — the estimator-linsolve `A`-refresh now sits after `jacobian2W!` and outside the `new_jac` guard, so it fires on dt-only W rewrites too. Verified still needed: master's SDIRK `smooth_est` gates are still `isnewton`-gated (`generic_imex_perform_step.jl:1264,2314`), so NSA still silently error-controls on the raw estimate.

Happy to unstack onto master if you'd prefer it independent.

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
@@ -32,6 +32,12 @@ isnewton(nlsolver::AbstractNLSolver) = isnewton(nlsolver.cache)
 isnewton(::AbstractNLSolverCache) = false
 isnewton(::Union{NLNewtonCache, NLNewtonConstantCache}) = true
 
+# Whether the solver can supply the W linear solve for the SDIRK/ESDIRK `smooth_est` estimate;
+# otherwise the caller falls back to the raw embedded estimate.
+can_smooth_est(nlsolver::AbstractNLSolver) = can_smooth_est(nlsolver.cache)
+can_smooth_est(cache::AbstractNLSolverCache) = isnewton(cache)
+can_smooth_est(cache::NonlinearSolveCache) = cache.linsolve !== nothing
+
 is_always_new(nlsolver::AbstractNLSolver) = is_always_new(nlsolver.alg)
 check_div(nlsolver::AbstractNLSolver) = check_div(nlsolver.alg)
 check_div(alg) = isdefined(alg, :check_div) ? alg.check_div : true
@@ -255,16 +261,25 @@ function _nlalg_with_linsolve(inner_alg, linsolve)
     return remake(inner_alg; descent = remake(descent; linsolve = linsolve))
 end
 
-# The reused ODE `W` as the operator handed to the inner NonlinearSolve as `jac_prototype`.
-# A matrix-free `WOperator` (its `J` an operator) is passed through and applied via `mul!`;
-# a concrete `WOperator`'s materialized form, or a raw matrix, is wrapped in a
-# `MatrixOperator` and materialized via `convert` for a factorization. Used identically at
-# build time and after a `resize!` so the inner `NonlinearFunction`'s concrete type is
-# preserved.
-function reuse_jac_prototype(W)
+# Analytic jacobian that copies the reused ODE `W` into the inner solver's own buffer each
+# evaluation, so its factorization refreshes when the ODE rewrites `W`. A bare operator
+# `jac_prototype` aliasing `W` would instead leave that factorization stale (its
+# `update_coefficients!` is a no-op), silently converging the inner Newton on an out-of-date
+# factorization.
+struct WReuseJac{W} <: Function
+    W::Base.RefValue{W}
+end
+(j::WReuseJac)(J_out, z, p) = (copyto!(J_out, j.W[]); J_out)
+
+# The reused ODE `W` as the jacobian for the inner NonlinearSolve: a matrix-free `WOperator`
+# passed through as an operator `jac_prototype` (applied via `mul!`, Krylov); a concrete W
+# reused via an analytic `WReuseJac`. Used at build time and after `resize!` so the inner
+# `NonlinearFunction`'s concrete type is preserved.
+function reuse_jac_kwargs(W)
     Wr = W isa WOperator && W.J !== nothing && !(W.J isa AbstractSciMLOperator) ?
         W._concrete_form : W
-    return Wr isa AbstractSciMLOperator ? Wr : MatrixOperator(Wr)
+    return Wr isa AbstractSciMLOperator ? (; jac_prototype = Wr) :
+        (; jac = WReuseJac(Ref(Wr)), jac_prototype = similar(Wr))
 end
 
 """
@@ -418,17 +433,9 @@ function build_nlsolver(
                     (tmp, ustep, γ, α, tstep, k, invγdt, DIRK, p, dt, f)
                 end
                 if use_w_reuse
-                    # Hand the reused W to the inner NonlinearFunction as an operator
-                    # `jac_prototype`. A matrix-free `WOperator` is applied via `mul!`
-                    # (Krylov); a concrete W is materialized via `convert` for a
-                    # factorization (NonlinearSolve copies it before an in-place `lu!`, so the
-                    # ODE-owned W is never destroyed). `jac` is auto-wired to
-                    # `update_coefficients!`; the ODE owns W's updates, so NonlinearSolve holds
-                    # W fixed across its Newton steps.
                     NonlinearProblem(
                         NonlinearFunction{true, SciMLBase.FullSpecialize}(
-                            nlf;
-                            jac_prototype = reuse_jac_prototype(W)
+                            nlf; reuse_jac_kwargs(W)...
                         ),
                         ztmp, nlp_params
                     )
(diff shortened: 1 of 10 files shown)


lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
```diff
@@ -18,6 +18,14 @@ isnewton(nlsolver::AbstractNLSolver) = isnewton(nlsolver.cache)
 isnewton(::AbstractNLSolverCache) = false
 isnewton(::Union{NLNewtonCache, NLNewtonConstantCache}) = true
 
+# Whether the nonlinear solver can supply the W-based linear solve used by the
+# SDIRK/ESDIRK `smooth_est` smoothed error estimate. True for Newton, and for
+# NonlinearSolveAlg when the W-reuse path built an estimator linsolve; otherwise the
+# caller must fall back to the raw embedded estimate.
+can_smooth_est(nlsolver::AbstractNLSolver) = can_smooth_est(nlsolver.cache)
+can_smooth_est(cache::AbstractNLSolverCache) = isnewton(cache)
+can_smooth_est(cache::NonlinearSolveCache) = cache.linsolve !== nothing
```
ChrisRackauckas, line 39: what happens on polyalgs? Since they don't always have a linsolve in all algs

ChrisRackauckas-Claude: Polyalgs don't interact with this: the estimator linsolve is built from `alg.linsolve` — the **ODE algorithm's** linsolve — never from the inner NonlinearSolve algorithm, so heterogeneous/absent linsolves inside a polyalg are irrelevant to it. With `TRBDF2()` (no linsolve given) it falls through to LinearSolve's default algorithm selection on W's type. Verified on this branch: `NonlinearSolveAlg(RobustMultiNewton())`, `NonlinearSolveAlg(FastShortcutNonlinearPolyalg())`, and `NonlinearSolveAlg(TrustRegion())` all build the estimator linsolve and report `can_smooth_est == true` at `init`.

Separate pre-existing problem this check surfaced: **polyalg inner solvers crash the NSA path outright, on master too** — `initialize!` reads `cache.cache.u` for the resize length check and `NonlinearSolvePolyAlgorithmCache` has no `u` field (`FieldError`, identical on master and this branch; TrustRegion works fine). Filing as its own issue — the NSA internals need `NonlinearSolveBase.get_u`-style accessors rather than raw field access, which the residual-termination redesign discussed in #3817 would want anyway.

ChrisRackauckas-Claude:
Rebased onto current `master` (now on top of the merged #3885 unification), and fixed a regression that #3885 introduced and that this PR's tests surfaced.

**Root cause.** #3885 handed the inner NonlinearSolve a `MatrixOperator` `jac_prototype` aliasing the ODE's reused `W` for *both* matrix-free and concrete `W`. For a concrete `W` this leaves the inner LinearSolve factorization **stale** when the ODE rewrites `W` across steps — the operator's auto-wired `update_coefficients!` is a no-op that never marks the linear cache dirty, so the inner Newton runs on an out-of-date factorization and the outer convergence check accepts a falsely-converged, inaccurate stage.

Measured on Van der Pol μ=1e5, `TRBDF2(nlsolve = NSA)`, reltol 1e-8:

| form | naccept | nreject | inner iters/attempt |
|------|---------|---------|---------------------|
| MatrixOperator alias (#3885) | 5749 | 1917 | 8.2 |
| analytic jac over own buffer (this PR) | 1389 | 310 | 14.0 |
| NLNewton (reference) | 1133 | 443 | 12.3 |

The estimator (`W \ tmp`) matched NLNewton to ~12 digits throughout — the regression was the **inner solve convergence** (false convergence at 8.2 iters), not the smooth_est plumbing.

**Fix** (`reuse_jac_kwargs`): matrix-free `W` still passes through as an operator `jac_prototype` (applied via `mul!`, Krylov); a concrete `W` is reused via an analytic `WReuseJac` copying `W` into the inner solver's own buffer each Newton step, which marks that buffer dirty so the factorization refreshes and leaves the ODE-owned `W` untouched by the inner `lu!`. A `MatrixOperator` with a copying `update_func!` can't substitute — the custom update makes it non-constant, skipping the factorization path (`ldiv!` then errors).

**Test note.** The concrete inner jac is now a plain matrix mirroring W's structure rather than a `MatrixOperator`; `nsa_sparse_tests` asserts it stays sparse, and its KLU accuracy check is rtol=1e-4 — TRBDF2 at reltol=1e-8 carries ~2.7e-5 global error here for NLNewton too and identically for KLU vs the default linsolve, so a tighter bound flagged the integrator's order, not a linear-solver regression.

**Verification** (all green locally, julia 1.12): `nsa_smooth_est` 10/10, `nsa_sparse` 10/10, `nsa_matrixfree` 14/14, `newton`, `nsa_jacobian_reuse`, `nested_ad`, `wprototype`, `homotopy` 25/25, `mass_matrix` 225 pass / 42 pre-existing broken / 0 fail. Runic-formatted.

Draft — please ignore until reviewed by @ChrisRackauckas.

ChrisRackauckas-Claude:
**Final design.** NSA computes the SDIRK/ESDIRK smoothed estimate `W \ tmp` by reusing the inner solver's own W factorization — `NonlinearSolveCache.linsolve` aliases it via `NonlinearSolveBase.get_linear_cache` (SciML/NonlinearSolve.jl#1065, merged + registered), the same way NLNewton reuses its Newton factorization. No second LinearSolve cache and no manual refresh: the inner Newton re-factorizes W itself each stage. Matrix-free inner solves reuse their Krylov cache; native scalar/StaticArray solves fall back to the raw embedded estimate.

The change also fixes the underlying inner-solve requirement: a concrete W is reused via an analytic `WReuseJac` (copies W into the inner buffer each eval, so its factorization stays current), not a bare operator alias which would go stale — ~4× the accepted steps on VdP μ=1e5 otherwise.

Squashed to one commit, trimmed the iteration-era comments, rebased on latest master. Verified locally (julia 1.12): VdP μ=1e5 **1389/310** (parity with NLNewton), smooth_est 10/10, sparse 10/10, matrixfree 14/14, newton/jacobian_reuse/nested_ad/wprototype/homotopy, mass_matrix 225 pass/42 pre-existing broken/0 fail. Runic clean.

Ready for review/merge. Draft — please ignore until reviewed by @ChrisRackauckas.

ChrisRackauckas-Claude:
Rebased onto latest master (through #4003). Three conflicts, all resolved by combining master's changes with this PR's:
- inner `init` now passes `verbose = ...nonlinear_verbosity` (master) **and** re-points the estimator linsolve alias (this PR);
- module imports keep master's `HomotopyPolyAlgorithm`/`ArcLengthContinuation` + `import NonlinearSolveBase`, plus `using NonlinearSolveBase: get_linear_cache`;
- `NonlinearSolveBase` compat: master's `2.34.1` floor → `2.35` (needs public `get_linear_cache`).

Re-verified against the now-current registry deps that CI resolves — **LinearSolve 5.1.0**, NonlinearSolveBase 2.37.0, NonlinearSolve 4.23.0, SciMLBase 3.37.0: smooth_est 10/10, VdP μ=1e5 **1389/310** (parity), matrixfree 14/14, sparse 10/10. The `get_linear_cache` factorization reuse — including Krylov reuse (matrix-free) and KLU (sparse) — holds across the LinearSolve 4→5 major bump.

PR #3826, opened 2026-07-05
OrdinaryDiffEqCore: route dense interpolation interval search through FindFirstFunctions

ChrisRackauckas-Claude:
**Note: please ignore this PR until reviewed by @ChrisRackauckas.**

## What

Consolidates OrdinaryDiffEqCore's interval search onto FindFirstFunctions.jl, replacing the bespoke `_searchsortedfirst`/`_searchsortedlast` binary-search kernels in `dense/generic_dense.jl`:

- The `(v, x, lo, forward)` methods now call `searchsorted_first`/`searchsorted_last` with the bracketing-gallop strategy, using `lo` as the hint — the sorted vector-`tvals` loop keeps its previous-hit warm start, and `max`-clamps reproduce the legacy lower-bound semantics exactly.
- `InterpolationData` gains a `FindFirstFunctions.Guesser` (linear-extrapolation mode) and the **scalar** `ode_interpolation`/`ode_interpolation!` paths warm-start from it. A nine-argument `InterpolationData` constructor keeps downstream positional constructors (e.g. StochasticDiffEq) working; foreign interpolation types fall back through `_ts_hint(id) = nothing` to the plain search.

## Measured (Julia 1.11.9, const-bound loops, min of 9 runs)

Plain search parity — FFF needs no changes; the bespoke kernel wasn't meaningfully faster:

| n | bespoke | FFF |
|---|---|---|
| 100 | 48 ns | 56 ns |
| 3 000 | 93 ns | 108 ns |
| 100 000 | 157 ns | 187 ns |

(within ~15% on the raw search; the search is a few percent of any real solve)

Scalar path with the linear-guess hint, 78k-point grid — the access patterns of adjoint solves, `saveat` post-processing, and delay-equation history lookups:

| pattern | plain bisection | hinted |
|---|---|---|
| descending sweep | 79 ns | **23 ns** |
| clustered stage times | 97 ns | **21 ns** |
| random access | 174 ns | **40 ns** |

## Correctness

- 90 000 randomized stateful queries (ties, both sort orders, arbitrary hints, both polarities) against Base-derived references: **zero mismatches**.
- New `test/interpolation_hint_tests.jl` (26 assertions: forward/reverse-time solutions × ascending/descending/random access × both continuities × out-of-place/in-place, plus hint-engagement), wired into the Core test group.

OrdinaryDiffEqCore version bumped 4.5.0 → 4.6.0 (new dependency; FindFirstFunctions compat "3", same floor as DataInterpolations 9).

https://claude.ai/code/session_018uzZhKZV2ameiUL8ouYknd


ChrisRackauckas-Claude:
CI triage: of the 57 failures, 43 (`downgrade-sublibraries`) are pre-existing on master (SciMLBase lower-bound vs DiffEqDevTools, unrelated to this PR), 1 (`OrdinaryDiffEqTsit5 Core`) is a self-hosted-runner communication loss, and the 13 StochasticDiffEq sublibrary failures **are caused by this PR**: their test envs pull ModelingToolkit → StateSelection, whose FindFirstFunctions compat (`"1.2.0, 2"`) predates v3 and conflicts with the `FindFirstFunctions = "3"` requirement added here. Fix: JuliaComputing/StateSelection.jl#126 (one-line compat widen; StateSelection only uses `findfirstequal`, unchanged in v3 — its test suite passes locally with FFF 3.0.1 pinned). This PR is **blocked on that release**.

PR #3829, opened 2026-07-07
Remove `deepcopy` branch for `alias_f`

jClugstor:
## Checklist

## Additional context

The branch is causing issues with JuliaC trimming.


jClugstor:
@ChrisRackauckas 
The `deepcopy` branch was originally added in the same PR that added the support for using ODEAliasSpecifier, so it was basically added to give `alias_f` something to do I guess. By default `f` is aliased anyways (`alias_f = true`), and I can't find a single package that uses `alias_f = false`, so I think getting rid of it should be safe. I just rebased so CI is still running but there weren't any issues before the rebase. 

A question this brings up is now that `alias_f` doesn't do anything, should we get rid of it in the ODEAliasSpecifier?

ChrisRackauckas:
It can be unsafe to alias the caches within f though?

PR #3777, opened 2026-06-25
QA: run_qa v1.6 form + ExplicitImports (root + sublibs)

ChrisRackauckas-Claude:
**Ignore until reviewed by @ChrisRackauckas.**

Converts every QA environment in the OrdinaryDiffEq.jl monorepo to the SciMLTesting 1.6 `run_qa` form and enables the ExplicitImports checks (`explicit_imports = true`). One feature branch, one PR.

## What changed

- **Root** `test/qa/qa_tests.jl` → `run_qa(OrdinaryDiffEq; aqua = false, explicit_imports = true, ei_kwargs = ...)`. The umbrella lane historically ran only ExplicitImports (no Aqua/JET), so `aqua = false` preserves that scope. All six EI checks pass; the one public-API miss (`init`/`solve`/`solve!`/`step!` from `CommonSolve`, `successful_retcode` from `SciMLBase`) is handled with a small `all_explicit_imports_are_public` ignore-list. Root `Project.toml`: `SciMLTesting` compat → `"1.6"`, `ExplicitImports` dropped from `[extras]`/`[targets].test` (now transitive via SciMLTesting).
- **47 sublibrary `test/qa/qa.jl`** → `run_qa(<Sublib>; ...)`:
  - Aqua sub-check disables preserved verbatim as `aqua_kwargs` (e.g. `piracies = false`, `unbound_args = false`, `deps_compat = (; check_extras = false)`).
  - Sublibraries whose old `qa.jl` ran JET inline (the `StochasticDiffEq*` and `*Tableaus` packages) keep JET via `using JET` + `jet_kwargs = (; target_defined_modules = true)`. Solver sublibraries keep their separate `test/qa/jet.jl` (bespoke `@test_opt` type-stability tests) untouched — only the Aqua file is converted.
  - `explicit_imports = true` on every env. Findings that depend on `OrdinaryDiffEqCore`/`SciMLBase`/`DiffEqBase` internal (non-`public`) API are recorded as `ei_broken` so the lane is green-or-broken with **0 hard EI failures** (tracked in #3776; `ei_broken` auto-flags an Unexpected Pass once the upstream make-public work lands). `OrdinaryDiffEqRosenbrockTableaus` and `StochasticDiffEqLevyArea` pass all six checks (no `ei_broken`).
  - Each sublib `test/qa/Project.toml` gains `SciMLTesting = "1.6"`; `[sources]` / develop wiring is left exactly as-is.
[...]

PR #3833, opened 2026-07-08
Initialize sparse Jacobian caches with stored values

### LINKED ISSUE #3830: Memory use fluctuates wildly with sparse Jacobians due to dropzeros
When repeatedly solving a stiff ODE with a sparse Jacobian, I experience an annoying issue where the memory allocations fluctuates by almost an order of magnitude between different solves, which leads to stochastic performance and annoying GC pauses. It happens non-deterministically, and it seems to happen more frequently when the solve needs more memory. Hence a stupid simple example, only with expensive `N`, `abstol` and `reltol` below:

```julia
using ModelingToolkit, OrdinaryDiffEqRosenbrock, LinearSolve
using ModelingToolkit: t_nounits as t, D_nounits as D

N = 500
@variables (x(t))[1:N] = ones(N)
eqs = [D(x[i]) ~ x[i] for i in 1:N]
@mtkbuild sys = ODESystem(eqs, t)
alg = Rodas5P(linsolve = PureKLUFactorization())
prob = ODEProblem(sys, [], (0.0, 200.0); jac = true, sparse = true)
for i in 1:10
    @time solve(prob, alg; abstol = 1e-14, reltol = 1e-14)
end
```
```
  0.942631 seconds (379.67 k allocations: 421.112 MiB, 9.62% gc time, 0.00% compilation time: 100% of which was recompilation)
  1.815347 seconds (2.27 M allocations: 2.833 GiB, 29.71% gc time, 0.00% compilation time: 100% of which was recompilation)
  1.091315 seconds (379.65 k allocations: 421.112 MiB, 14.54% gc time, 0.00% compilation time: 100% of which was recompilation)
  0.952199 seconds (379.65 k allocations: 421.109 MiB, 8.50% gc time, 0.00% compilation time: 100% of which was recompilation)
  1.794840 seconds (2.27 M allocations: 2.833 GiB, 28.30% gc time, 0.00% compilation time: 100% of which was recompilation)
  0.977346 seconds (379.65 k allocations: 421.108 MiB, 7.62% gc time, 0.00% compilation time: 100% of which was recompilation)
  1.774829 seconds (2.27 M allocations: 2.833 GiB, 24.75% gc time, 0.00% compilation time: 100% of which was recompilation)
  0.930147 seconds (379.65 k allocations: 421.112 MiB, 3.45% gc time, 0.00% compilation time: 100% of which was recompilation)
  1.816451 seconds (2.27 M allocations: 2.833 GiB, 28.66% gc time, 0.00% compilation time: 100% of which was recompilation)
  0.880605 seconds (379.65 k allocations: 421.110 MiB, 1.87% gc time, 0.00% compilation time: 100% of which was recompilation)
```
I then profile the allocations repeatedly until I see one of the invocations with the big `2.833 GiB` allocation spike, which shows that `dropzeros` is entirely responsible for the extra allocations:
```julia
@profview_allocs @time solve(prob, alg; abstol = 1e-14, reltol = 1e-14)
```
```
  1.827441 seconds (2.27 M allocations: 2.833 GiB, 27.41% gc time, 0.00% compilation time: 100% of which was recompilation)
```

Image

The memory profiling on the good `421.112 MiB` invocations looks normal and does not show the `dropzeros` at all.

ChrisRackauckas-Claude:

## Summary
- initialize Jacobian and W caches with nonzero stored values before LinearSolve setup
- preserve sparse prototype structure by setting only existing `nzval` entries for sparse matrices
- add a sparse Rosenbrock cache regression test that checks `J` and `W` keep their sparse pattern and stored values immediately after `init`

Fixes #3830.

## Local verification
- `~/.juliaup/bin/julia +1.12 --project=. -e 'using OrdinaryDiffEqRosenbrock, SparseArrays, LinearAlgebra; function f!(du,u,p,t); du .= u; end; jacproto = sparse(1.0I, 4, 4); prob = ODEProblem(ODEFunction(f!; jac_prototype=jacproto), ones(4), (0.0, 1.0)); integ = init(prob, Rodas5P()); @show nnz(integ.cache.J) nnz(integ.cache.W); @show nonzeros(integ.cache.J); @show nonzeros(integ.cache.W); @assert nnz(integ.cache.J) == 4; @assert nnz(integ.cache.W) == 4; @assert all(==(1), nonzeros(integ.cache.J)); @assert all(==(1), nonzeros(integ.cache.W))'`
- `ODEDIFFEQ_TEST_GROUP=Sparse ~/.juliaup/bin/julia +1.12 --project=lib/OrdinaryDiffEqDifferentiation -e 'using Pkg; Pkg.test(; test_args=String[])'`
- `~/.juliaup/bin/julia +1.12 -e 'using Runic; exit(Runic.main(ARGS))' -- --check lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl lib/OrdinaryDiffEqDifferentiation/test/sparse/nonfulldiagonal_sparse_tests.jl`

Note: repository-wide Runic check was also run with `--check --diff --verbose .`; it fails on unrelated files. A clean-master check reproduced that failure and bisected it to `661f605637` / PR #3777.

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
@@ -1134,6 +1134,13 @@ function build_J_W(
         # If factorization, then just use the jac_prototype
         J = similar(f.jac_prototype)
         W = similar(J)
+        if is_sparse(J)
+            set_all_nzval!(J, one(eltype(J)))
+            set_all_nzval!(W, one(eltype(W)))
+        else
+            fill!(J, one(eltype(J)))
+            fill!(W, one(eltype(W)))
+        end
     elseif (
             IIP && (concrete_jac(alg) === nothing || !concrete_jac(alg)) &&
                 alg.linsolve !== nothing &&
--- a/lib/OrdinaryDiffEqDifferentiation/test/sparse/nonfulldiagonal_sparse_tests.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/test/sparse/nonfulldiagonal_sparse_tests.jl
@@ -1,4 +1,4 @@
-using OrdinaryDiffEq, SparseArrays, LinearSolve, LinearAlgebra
+using OrdinaryDiffEq, SparseArrays, LinearSolve, LinearAlgebra, Test
 using OrdinaryDiffEqSDIRK
 using ComponentArrays
 
@@ -153,3 +153,21 @@ solve(odeprob, TRBDF2());
 solve(sparseodeprob, TRBDF2());
 solve(sparseodeprob, Rosenbrock23(linsolve = KLUFactorization()));
 solve(sparseodeprob, KenCarp47(linsolve = KrylovJL_GMRES()));
+
+@testset "Sparse Jacobian caches are initialized with stored values" begin
+    function sparse_cache_f!(du, u, p, t)
+        return du .= u
+    end
+
+    N = 8
+    jac_prototype = sparse(1:N, 1:N, ones(N), N, N)
+    prob = ODEProblem(
+        ODEFunction(sparse_cache_f!; jac_prototype),
+        ones(N),
+        (0.0, 1.0)
+    )
+    integ = init(prob, Rodas5P())
+
+    @test all(==(1), nonzeros(integ.cache.J))
+    @test all(==(1), nonzeros(integ.cache.W))
+end


lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
```diff
@@ -1134,6 +1134,8 @@ function build_J_W(
         # If factorization, then just use the jac_prototype
         J = similar(f.jac_prototype)
         W = similar(J)
+        fill!(J, zero(eltype(J)))
+        fill!(W, zero(eltype(W)))
```
ChrisRackauckas: Should use zero instead of similar to start it. though actually, if it's zero then dropzeros is wrong... so maybe something like fill with ones? Benchmark and see if zeroing has adverse behavior when mixed with a sprase matrix and the default dropzeros of LinearSolve

hersle:
@ChrisRackauckas If anything, should it not initialize with a nonzero number like ones? Initializing with zeros sounds like it would trigger the `dropzeros` unconditionally.

Alternatively, would it be better if OrdinaryDiffEq just sets the nonstructural zeros operator assumption to e.g. `None` instead of `Auto` when calling LinearSolve? Then the random `dropzeros` behavior would never be triggered?

ChrisRackauckas:
There's a lot of cases we found where the dropzeros is actually a major performance improvement, and also a numerical improvement. So we want to have it on and scan for it. But we should try to not do it on many problems, yes. The issue is that it hits init, so we should probably just init to ones.

PR #3835, opened 2026-07-08
Add solver-level limiter kwargs

ChrisRackauckas-Claude:

## Summary
- make `stage_limiter` and `step_limiter` solver kwargs the standard no-bang interface
- thread solve-level limiter kwargs into native algorithm limiter fields where available
- thread `step_limiter` through `DEOptions` so methods without native limiter fields still apply solve-level step limiters from the common accepted-step path
- preserve native limiter placement for algorithms that expose `step_limiter!` fields, avoiding duplicate calls
- warn when `stage_limiter` is supplied to a solver with no stage hook; `step_limiter` is handled centrally instead of warning as unused
- extend native limiter support to `Euler`, generic `ExplicitRK`, all StabilizedRK methods, `FunctionMap`, and ExponentialRK methods where a final accepted-state hook is available
- expand limiter tests to cover solve-level kwargs, native RK limiter hooks, FunctionMap, every StabilizedRK algorithm, ExponentialRK families, and representative non-native-field methods through the central path

## Local verification
- `timeout 3600 julia --project=. -e 'using Runic; exit(Runic.main(ARGS))' -- --inplace ...` passed\n- `GROUP=Integrators_I timeout 3600 julia --project=. -e 'using Pkg; Pkg.test()'` passed: `Step Limiter Tests | 167 167`, final `OrdinaryDiffEq tests passed`\n- `timeout 3600 julia --project=lib/DelayDiffEq -e 'using DelayDiffEq; println("DelayDiffEq load OK")'` passed before this cleanup; the cleanup only removed an unused DDE bang-kwarg shim\n- `git diff --check` passed\n


ChrisRackauckas-Claude:
Follow-up commit `97678542e` reworks the limiter handling per review feedback: instead of storing `stage_limiter!`/`step_limiter!` on each algorithm/cache and applying them inside every `perform_step!`, both are now **solve-level options** carried on `DEOptions` and resolved from the `stage_limiter`/`step_limiter` kwargs in `_ode_init`.

- `step_limiter` is applied **centrally**, once per accepted step, for every method (removed from all `perform_step!`s). Single application point; immune to double-application when a method borrows another's `perform_step!` for startup (e.g. `VCAB*`).
- `stage_limiter` is **threaded per step**: `perform_step!` reads `integrator.opts.stage_limiter!`.
- The per-algorithm `stage_limiter!`/`step_limiter!` constructor fields are **deprecated** — still honored (with a `depwarn`) when the matching kwarg isn't supplied.

Net −115 lines. Local verification: step-limiter tests 167/167; SDE/DDE construction + deprecation-path smoke pass; SSPRK `AllocCheck` allocation-free; Runic clean. The PR summary above describes the earlier per-cache approach and is superseded by this commit.

Issue #3814, opened 2026-07-03
Operator mass matrices: `IdentityOperator`/`ScalarOperator` fail at solve entry — fix in place or push toward native operator support?


singhharsh1708:
`IdentityOperator` and `ScalarOperator` cannot be used as `f.mass_matrix` right now because `ArrayInterface.issingular` isn't defined for them, so the DAE check at `lib/OrdinaryDiffEqCore/src/solve.jl:233` throws before the solver even starts.

Repro against master (dev'd sublibs):

```julia
using OrdinaryDiffEqSDIRK, SciMLOperators, LinearAlgebra

f(du, u, p, t) = (du .= -u; nothing)
for mm in (IdentityOperator(3), ScalarOperator(2.0))
    prob = ODEProblem(ODEFunction(f, mass_matrix = mm), ones(3), (0.0, 1.0))
    try
        solve(prob, TRBDF2(); reltol = 1e-6, abstol = 1e-8)
    catch e
        println(typeof(mm).name.name, " => ", first(sprint(showerror, e), 120))
    end
 end
```

```
IdentityOperator => MethodError: no method matching issingular(::IdentityOperator)
ScalarOperator   => MethodError: no method matching issingular(::ScalarOperator{Float64, ...})
```

`MatrixOperator` and `DiagonalOperator` (which is a `MatrixOperator` under the hood) work fine because `build_J_W` has a dedicated `MatrixOperator` branch.

The third TODO on top of `build_J_W` (`lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl:1100-1102`) already anticipates this:

```julia
# TODO - make J, W AbstractSciMLOperators (lazily defined with scimlops functionality)
# TODO - if jvp given, make it SciMLOperators.FunctionOperator
# TODO - make mass matrix a SciMLOperator so it can be updated with time. Default to IdentityOperator
```

So two directions this could go, and I would rather hear which one you want before I open PRs:

1. **Point fix**: add `ArrayInterface.issingular` methods for `IdentityOperator` / `ScalarOperator` (and any other operator that could legitimately appear as a mass matrix) upstream in SciMLOperators or ArrayInterface. Keeps concrete matrices as the internal representation.

[...]

PR #3837, opened 2026-07-08
`get_differential_vars`: accept `IdentityOperator` and `ScalarOperator` mass matrices

singhharsh1708:
Follow-up to #3814. Once SciML/SciMLOperators.jl#400 lands the `issingular` methods, `IdentityOperator` and `ScalarOperator` mass matrices still hit `MethodError: no method matching iterate(::IdentityOperator)` inside `get_differential_vars` at `lib/OrdinaryDiffEqCore/src/misc_utils.jl:143`.

Root cause: the function reaches `elseif all(!iszero, mm)` and tries to iterate the operator.

Fix: treat `IdentityOperator` and `ScalarOperator` the same as `UniformScaling` (both represent scalar-uniform mass matrices) and return `nothing` early.

```julia
if mm isa UniformScaling ||
        mm isa SciMLOperators.IdentityOperator ||
        mm isa SciMLOperators.ScalarOperator
    return nothing
elseif ...
```

End-to-end verified against the #3814 repro with both fixes in place:

```
IdentityOperator + TRBDF2 (IIP) => OK, retcode=Success
ScalarOperator   + TRBDF2 (IIP) => (DimensionMismatch further down — separate downstream bug, out of scope here)
```

Direct regression test added in `algebraic_vars_detection_tests.jl` calls `get_differential_vars` on both operator kinds and asserts `nothing`.

Depends on SciML/SciMLOperators.jl#400 for full end-to-end.

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/misc_utils.jl
+++ b/lib/OrdinaryDiffEqCore/src/misc_utils.jl
@@ -208,8 +208,11 @@ function get_differential_vars(f, u)
         mm = f.mass_matrix
         mm = mm isa MatrixOperator ? mm.A : mm
 
-        if mm isa UniformScaling
+        if mm isa UniformScaling || mm isa SciMLOperators.IdentityOperator
             return nothing
+        elseif mm isa SciMLOperators.ScalarOperator
+            # λ·I: every variable is algebraic when λ is zero.
+            return iszero(mm.val) ? falses(size(u)) : nothing
         elseif all(!iszero, mm)
             return trues(size(mm, 1))
         elseif !(mm isa SciMLOperators.AbstractSciMLOperator) && _isdiag(mm)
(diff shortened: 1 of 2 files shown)


singhharsh1708:
Rebased on master. Also fixed a hole in my own first pass.

The original patch returned `nothing` (all variables differential) for any `ScalarOperator`. That's wrong when the scalar is zero: `ScalarOperator(0.0)` is `0·I`, so every variable is algebraic. It also contradicted `issingular(α::ScalarOperator) = iszero(α.val)` from SciML/SciMLOperators.jl#400.

`ScalarOperator` now matches what the equivalent dense mass matrix already gives:

```
ScalarOperator(0.0)  =>  Bool[0, 0, 0]
zeros(3, 3)          =>  Bool[0, 0, 0]
```

`IdentityOperator` still returns `nothing` unconditionally — it is always `I`, never singular.

Test covers the singular case. `OrdinaryDiffEqCore` suite passes locally on Julia 1.12; Runic clean.

PR #3836, opened 2026-07-08
Fix post-3834 CI regressions

ChrisRackauckas-Claude:
This PR should be ignored until reviewed by @ChrisRackauckas.

## Summary

Fixes the post-#3834 CI failures seen on master run `342617a63`:

- restores `OrdinaryDiffEqDifferentiation` compatibility bindings for wrapper names imported by registered sublibraries
- tightens sublibrary downgrade compat floors that were resolving incompatible dependency sets
- adds the static-array PositiveIntegrators PDS regression and tightens `ImplicitEuler()`'s default nonlinear solve tolerance
- fixes sparse GPU DAE W reconstruction to preserve sparse GPU layout instead of broadcasting into it
- fixes `LinearExponential(krylov = :adaptive)` when sparse GPU CSC `opnorm` throws from the CUDA sparse path
- applies Runic formatting to the files reported by `format-check`

## Local verification

- Runic `--check` on changed Julia files: passed
- `git diff --check`: passed
- `lib/OrdinaryDiffEqDifferentiation/test/wrapper_compat_tests.jl`: passed
- sparse CPU smoke for `dae_jacobian2W!` / `dae_jacobian2W`: passed
- `lib/OrdinaryDiffEqSDIRK/test/pds_rhs_regression_tests.jl`: `3 passed`
- `lib/OrdinaryDiffEqLinear/test/linear_method_tests.jl`: passed
- `lib/OrdinaryDiffEqCore/test/qa/qa.jl`: `15 passed`
- `lib/StochasticDiffEqIIF` downgrade-style local test: passed earlier in this branch investigation

GPU CI was not run locally: `nvidia-smi` was unavailable and no `/dev/nvidia*` devices were visible on this machine.

PR #3853, opened 2026-07-09
Differentiation: sparse-GPU path for dae_jacobian2W! (GPU CSR DAE)

ChrisRackauckas-Claude:

## Summary

Extracted from omnibus #3836.

Fixes monorepo GPU DAE failures for every solver with `jac=CSR` + `mass=diag_cu` (48 `:gpu_error` cases on master).

`dae_jacobian2W!` used `@.. broadcast` into W. Sparse GPU CSR/CSC matrices do not support that. Mirror the existing `jacobian2W!` sparse-GPU branch: detect non-fast-scalar `nonzeros(W)` and use allocate-then-`copyto!`.

Bump `OrdinaryDiffEqDifferentiation` 3.3.0 → 3.3.1. CPU sparse smoke test included; full GPU coverage is `test/gpu/dae_tests.jl`.

## Local verification

- Runic format on changed Julia files
- CPU sparse smoke: `dae_jacobian2W!` / `dae_jacobian2W` pass
- GPU self-hosted not available on this machine

## Related

- Supersedes the Differentiation sparse-GPU portion of #3836
- Sibling: LinearExponential sparse opnorm (separate PR)

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/Project.toml
+++ b/lib/OrdinaryDiffEqDifferentiation/Project.toml
@@ -1,5 +1,5 @@
 name = "OrdinaryDiffEqDifferentiation"
-version = "3.3.0"
+version = "3.3.1"
 uuid = "4302a76b-040a-498a-8c04-15b101fed76b"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
 
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
@@ -566,15 +566,12 @@ end
 end
 
 
-# Sparse GPU W-build path: CuSparseMatrixCSC usually subtypes AbstractSparseMatrix
-# (is_sparse true), but CuSparseMatrixCSR often does not. Detect CSR via the
-# cuSPARSE `.nzVal` storage field so we never `@..` broadcast into it.
-@inline function _use_allocating_sparse_W_path(W)
-    if is_sparse(W)
-        return !ArrayInterface.fast_scalar_indexing(nonzeros(W))
-    end
-    return hasproperty(W, :nzVal)
-end
+# Sparse GPU arrays (e.g. CuSparseMatrixCSC/CSR) don't support broadcasting into
+# W, so they need the allocating build path. All cuSPARSE matrix types subtype
+# AbstractSparseMatrix (is_sparse true); their `nonzeros` storage is a GPU array,
+# which is not fast_scalar_indexing, whereas CPU sparse storage is.
+@inline _use_allocating_sparse_W_path(W) =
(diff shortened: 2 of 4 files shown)


ChrisRackauckas-Claude:
Follow-up push: root cause of still-red GPU CSR was that `is_sparse` is only true for `AbstractSparseMatrix`. `CuSparseMatrixCSC` usually qualifies; **`CuSparseMatrixCSR` often does not**, so the allocating W path never ran and `@..` still failed.

Now `_use_allocating_sparse_W_path` also triggers on cuSPARSE `.nzVal` (CSR). Applied to both `jacobian2W!` (mass-matrix ODE path used by the GPU DAE suite) and `dae_jacobian2W!`. Duck-type unit test added.

ChrisRackauckas-Claude:
Simplified `_use_allocating_sparse_W_path` and dropped the `.nzVal` duck-typing fallback (pushed as a follow-up commit).

**Why:** the fallback rested on an incorrect premise — the comment claimed "CuSparseMatrixCSR often does not [subtype AbstractSparseMatrix]." That has never been true in any released CUDA.jl. Traced from source:

- CUDA.jl 6.2.1: `CuSparseMatrixCSR <: GPUArrays.AbstractGPUSparseMatrixCSR`, and GPUArrays 11.3+ defines `AbstractGPUSparseArray{Tv,Ti,N} <: SparseArrays.AbstractSparseArray{Tv,Ti,N}`.
- CUDA.jl 5.8.2 (pre-GPUArrays-sparse-refactor): `AbstractCuSparseArray{Tv,Ti,N} <: AbstractSparseArray{Tv,Ti,N}`, with `CuSparseMatrixCSR <: AbstractCuSparseMatrix = AbstractCuSparseArray{Tv,Ti,2}`.

So `CuSparseMatrixCSR <: AbstractSparseMatrix` throughout, meaning `is_sparse(W)` already returns `true` for it and it always took branch 1 (`!fast_scalar_indexing(nonzeros(W))`, where `nonzeros` is the GPU `.nzVal` storage). The `hasproperty(W, :nzVal)` branch was unreachable for real cuSPARSE types.

New form:
```julia
@inline _use_allocating_sparse_W_path(W) =
    is_sparse(W) && !ArrayInterface.fast_scalar_indexing(nonzeros(W))
```
This is the general, non-GPU-specific predicate: GPU sparse storage isn't `fast_scalar_indexing` so it takes the allocating path; CPU sparse storage keeps the broadcast path.

Removed the `FakeCuSparse` test (it only hit the deleted fallback via an `AbstractMatrix` that deliberately isn't `AbstractSparseMatrix` — a shape real CUDA types never produce) and kept the CPU-sparse numeric coverage. Real GPU CSR remains covered by `test/gpu/dae_tests.jl`.

Ran the CPU test locally on Julia 1.12 — passes. Could not exercise on a GPU (no CUDA hardware here); the subtyping claims are verified statically from CUDA.jl/GPUArrays source, and the GPU CSR path is unchanged behaviorally (CSR still routes to the allocating path, now via branch 1 instead of the fallback).

PR #3854, opened 2026-07-09
Linear: adaptive LinearExponential opnorm without densifying

ChrisRackauckas-Claude:

## Summary

Extracted from omnibus #3836.

Fixes GPU `test/gpu/linear_exp.jl` adaptive sparse path:

```
MethodError: no method matching csr_type(::CuSparseMatrixCSC)
```

from GPUArrays sparse `opnorm` during adaptive `LinearExponential`.

Only the adaptive `LinearExponential` `expv_timestep` / `expv_timestep!` sites wrap the default `internalopnorm`. On `MethodError`, use a loose **nnz-based bound** instead of `collect(A)`.

Addresses #3836 review: *"We shouldn't densify, why wasn't this an issue before?"* — densify path removed.

Bump `OrdinaryDiffEqLinear` 2.1.0 → 2.1.1. CPU regression via `BrokenOpnormMatrix`.

## Local verification

- Runic format
- `linear_method_tests.jl` including BrokenOpnormMatrix path: `LINEAR_TESTS_OK`
- GPU self-hosted not available here

## Related

- Sibling: sparse-GPU `dae_jacobian2W!` (separate PR)
- Supersedes Linear portion of #3836

### DIFF
--- a/lib/OrdinaryDiffEqLinear/Project.toml
+++ b/lib/OrdinaryDiffEqLinear/Project.toml
@@ -22,6 +22,7 @@ DiffEqDevTools = "f3b72e0c-5b89-59e1-b016-84e28bfd966d"
 OrdinaryDiffEqVerner = "79d7bb75-1356-48c1-b8c0-6832512096c2"
 OrdinaryDiffEqRosenbrock = "43230ef6-c299-4910-a778-202eb28ce4ce"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [sources]
 DiffEqBase = {path = "../DiffEqBase"}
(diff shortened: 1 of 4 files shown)


ChrisRackauckas-Claude:
Follow-up: add  to Linear test / so the BrokenOpnormMatrix regression loads under sublibrary CI.

ChrisRackauckas-Claude:
Alternative approach for consideration: SciML/OrdinaryDiffEq.jl#3884 (+ its prerequisite SciML/ExponentialUtilities.jl#249).

The root issue is that `expv_timestep` evaluates `opnorm(A, Inf)` eagerly (as a keyword default) purely to (1) scale its internal absolute tolerance and (2) seed the initial substep `tau`. Neither use needs a true operator norm. Rather than catch the resulting `MethodError` and reconstruct a bound via `Base.invokelatest(getfield(parentmodule(typeof(A)), :nonzeros), A)`, the alternative makes ExponentialUtilities:

- accept an `abstol` keyword directly, and
- evaluate `opnorm` lazily (only when `abstol`/`tau` are unspecified).

OrdinaryDiffEqLinear then passes `abstol = opts.abstol` and `tau = dt`, so `opnorm(A, Inf)` is never called on the operator — no `invokelatest`, no `try`/`catch`, no `MethodError` inspection, no nnz reflection. As a bonus the adaptive `LinearExponential` now honors the solver's `abstol` for its internal Krylov error control instead of the `reltol * ‖A‖` heuristic.

Both PRs include a regression test wrapping the operator in a type whose `opnorm` throws. Flagging here so the two approaches can be compared; deferring to @ChrisRackauckas on which to take.

PR #3858, opened 2026-07-09
Document public API: add docstrings + docs entries

ChrisRackauckas-Claude:
## Summary

Several names across OrdinaryDiffEq.jl are declared `public` (the solver-author extension API) but were missing a docstring and/or a rendered-docs entry. This PR documents all of them.

An authoritative check (loading each package on Julia 1.12 and inspecting `Base.Docs.meta`) confirmed **176** undocumented `public` names in `OrdinaryDiffEqCore` plus a handful in the solver sublibraries. This PR adds docstrings for every one and a docs entry for every audited name.

## What changed

- **`public_api_docstrings.jl` per affected package** (`OrdinaryDiffEqCore`, `OrdinaryDiffEqNonlinearSolve`, `OrdinaryDiffEqDifferentiation`, `OrdinaryDiffEqLowOrderRK`, `OrdinaryDiffEqTsit5`, `OrdinaryDiffEqSDIRK`, `OrdinaryDiffEqRosenbrock`): attaches a docstring to each previously-undocumented `public` binding using the "docstring for an existing binding" form, so no definitions move. Docstrings are written from the actual implementations (algorithm/cache type hierarchies, trait functions, nonlinear-solver + W-matrix/differentiation hooks, interpolation kernels, threading/enum/sentinel types, noise hooks, controller caches, docstring builders).
- **`docs/src/api/solver_author.md`**: a new "Solver-author extension API" reference page with `@docs` blocks for every such name (excluding names already documented on the Controller API / misc pages, to avoid duplicate-docs errors), registered in `docs/pages.jl` under "APIs".
- **docs wiring**: `OrdinaryDiffEqDifferentiation` and `OrdinaryDiffEqNonlinearSolve` added to `docs/Project.toml`, `docs/make.jl` (`using` + `modules`), so their `@docs` entries resolve.

## Verification

- Julia 1.12: every `public` name now has an attached docstring (`Base.Docs.meta`) — `OrdinaryDiffEqCore` reports **0** missing; all sublibrary names report `DOC`.
- Ran a Documenter build of the new page with `warnonly = false`: all `@docs` entries resolve and all `@ref` links resolve, with no docstring-not-found / missing-docs errors. No `warnonly`/silencing was added.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -8,6 +8,7 @@ OrdinaryDiffEqAdamsBashforthMoulton = "89bda076-bce5-4f1c-845f-551c83cdda9a"
 OrdinaryDiffEqBDF = "6ad6398a-0878-4a85-9266-38940aa047c8"
 OrdinaryDiffEqCore = "bbf590c4-e513-4bbe-9b18-05decba2e5d8"
 OrdinaryDiffEqDefault = "50262376-6c5a-4cf5-baba-aaf4f84d72d7"
+OrdinaryDiffEqDifferentiation = "4302a76b-040a-498a-8c04-15b101fed76b"
 OrdinaryDiffEqExplicitRK = "9286f039-9fbf-40e8-bf65-aa933bdc4db0"
 OrdinaryDiffEqExponentialRK = "e0540318-69ee-4070-8777-9e2de6de23de"
 OrdinaryDiffEqExtrapolation = "becaefa8-8ca2-5cf9-886d-c06f3d2bd2c4"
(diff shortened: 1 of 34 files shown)


ChrisRackauckas-Claude:
Reworked per review feedback: the docstrings are now placed **inline, immediately above each definition** (the normal Julia way), instead of in a per-package `public_api_docstrings.jl` aggregation file that attached docstrings to bare bindings. The 7 aggregation files and their `include`s are removed, and the inline docstring on `generic_solver_docstring` that the first pass had relocated is restored. Every name here is a plain `abstract type` / `struct` / `function` / `const` definition, so none needed the detached form.

Verified locally on Julia 1.12:
- Every previously-undocumented `public` name still has an attached docstring (`Base.Docs.meta`): **172/172 in OrdinaryDiffEqCore, 33/33 across the sublibs**, 0 missing, no duplicate-docstring (`Replacing docs`) warnings.
- A Documenter build of `docs/src/api/solver_author.md` (missing-docstring errors fatal) resolves **all `@docs` blocks** with no docstring-not-found errors.

Squashed into the single existing commit so there is no trace of the detached-file approach in the history. Still a draft — ignore until reviewed by @ChrisRackauckas.

PR #3832, opened 2026-07-08
Document public developer APIs

ChrisRackauckas-Claude:
Ignore this PR until reviewed by @ChrisRackauckas.

## Summary
- Add a public developer API reference page for public monorepo/lib APIs.
- Add concise source docstrings for public developer API names rendered on the page.
- Run Runic formatting for the repo before committing.

## Verification
- Targeted lib public-name scanner reported `public name count: 298`, `gap count: 0`.
- `timeout 3600 /home/crackauc/.juliaup/bin/julia +release --project=docs -e 'using Pkg; Pkg.instantiate(); include("docs/make.jl")'` exited 0 after final formatting.\n- `rg 'no docs found for' ordinary_docs_precommit.log` returned no matches before the generated log was removed.\n- `timeout 3600 /home/crackauc/.juliaup/bin/julia +release --project=/home/crackauc/sandbox/tmp_20260708_051717_3275/.runic_env -m Runic --check .` exited 0.\n- `git diff --check` exited 0.

PR #3867, opened 2026-07-10
Update SciMLTesting QA docs coverage

ChrisRackauckas-Claude:

## Summary
- Add DiffEqBase to the docs environment for the split DiffEqBase API page.
- Update OrdinaryDiffEq monorepo QA infrastructure to SciMLTesting v2.1.
- Remove package self-path sources from QA/test environments while preserving sibling monorepo sources needed to test local sublibraries.
- Add/adjust OrdinaryDiffEqDefault public API docstrings at definition points and run the shared SciMLTesting API-doc check.
- Keep `AutoSwitchCache` treated as an internal monorepo extension detail rather than user public API.

## Validation
- Docs build from the original PR body was run locally before the first commit.
- `timeout 3600 julia --startup-file=no --project=lib/OrdinaryDiffEqDefault -e 'using Pkg; withenv("ODEDIFFEQ_TEST_GROUP" => "QA") do; Pkg.test("OrdinaryDiffEqDefault"; test_args=String[], julia_args=["--check-bounds=auto", "--compiled-modules=yes", "--depwarn=yes"], force_latest_compatible_version=false, allow_reresolve=true); end'` passed.\n- `timeout 3600 julia --startup-file=no -m Runic --check lib/OrdinaryDiffEqDefault/test/qa/qa.jl lib/OrdinaryDiffEqDefault/src/default_alg.jl lib/OrdinaryDiffEqDefault/test/runtests.jl` passed.\n- `git diff --check` passed.\n- TOML parse passed for `lib/OrdinaryDiffEqDefault/Project.toml` and `lib/OrdinaryDiffEqDefault/test/qa/Project.toml`.\n- Focused scan found no `api_docs = false`, SciMLTesting v1 compat, or self `[sources]` path entries in the OrdinaryDiffEqDefault QA files.


ChrisRackauckas-Claude:
Heads-up from a parallel investigation of the master Documentation failure: this PR's docs/Project.toml change adds DiffEqBase to [deps]/[compat] but not to [sources]. Without a path source, the docs env resolves the registered DiffEqBase 7.6.1 (registry tree 3696ff4a), which predates the docstrings #3832 added to lib/DiffEqBase (master tree 2e54d5ee) that docs/src/api/diffeqbase.md references — #3832 modified lib/DiffEqBase without bumping its version. A minimal docs-only fix including the [sources] entry, verified with a full local docs build, is at #3877.

PR #3818, opened 2026-07-04
NonlinearSolveAlg: let inner solver refresh J when no W is reused (fixes master nlstep order failures)

ChrisRackauckas-Claude:
> **Please ignore until reviewed by @ChrisRackauckas.** Draft opened by an agent.

## Fixes two failing tests on master

`test/modelingtoolkit/nlstep_tests.jl` currently fails on master (17 pass / **2 fail** / 6 broken): the FBDF + NSA(TrustRegion) Robertson order assertions at lines 60 and 116 measure order **0.6152** against bounds `|𝒪est−1| < 0.3 / 0.35`. (#3807 read this as "order 1.38" — it's `1 + 0.3848` in the test output; the order collapsed *below* 1.)

## Root cause

#3703 changed both `compute_step!` methods from

```julia
recompute_jacobian = cache.W === nothing || (nlsolver.iter == 1 && cache.new_W)
```
to
```julia
recompute_jacobian = nlsolver.iter == 1 && (cache.W === nothing || cache.new_W)
```

which freezes the inner NonlinearSolve Jacobian after outer iteration 1 on the paths **without** a reusable W (MTK nlstep, DAE, OOP). The inner solve silently degrades to modified Newton while the outer `η·ndz < κ` termination keeps its calibration, so every step carries a dt-independent unconverged remainder. Measured l∞ errors over `dts = 2^-15…2^-18`: master `[2.22e-7, 1.12e-7, 5.60e-8, 6.18e-8]` (error stops halving, then **rises**); with this fix `[…, 1.38e-7, 5.60e-8, 2.80e-8]` (clean halving). Warmed TrustRegion and NewtonRaphson give bit-identical trajectories — not TR-specific, and consistent with the principle that a *converged* inner solve must not change the ODE solution.

## Fix

Restore the pre-#3703 condition: `W === nothing` paths let the inner solver refresh J every outer iteration; the W-reuse (modified-Newton) path is untouched — with a reused W the inner Jacobian *is* the ODE-side W by construction, so refreshing beyond iter 1 remains pointless there.

## Verification (local)

- A/B on master: this exact change turns `nlstep_tests.jl` from 17 pass / 2 fail into **19 pass / 0 fail**, with orders 1.2576 and 1.3035 back inside their bounds.
- `newton_tests.jl` + `linear_nonlinear_tests.jl` pass against this branch.

## Related findings from the same investigation (separate issues)

[...]

Issue #1008, opened 2020-01-10
Solver fails if tspan is a dual number and saveat is a number


ChrisRackauckas:
The expansion of `saveat` into a range ends up failing.

ChrisRackauckas-Claude:
Fixed — a dual-number `tspan` with a numeric `saveat` both differentiates and solves:

```
ForwardDiff.derivative(g, 1.0) = 1.386527690218062     (analytic 1.3865285125835427)
direct Dual tspan: SUCCESS, length(sol.t)=11
  sol.u[end] = Dual{:tag}(1.3728004409033048, 1.386527690218062)
```

Verified on OrdinaryDiffEq v7.1.3 / Julia 1.11.9. Closing.

PR #3897, opened 2026-07-13
nlstep tests: unbreak NSA(TrustRegion) order assertions that now pass

singhharsh1708:
Four `@test_broken` order assertions in `modelingtoolkit/nlstep_tests.jl` now pass, so the MTK lane reports them as `Unexpected Pass` and errors out:

```
Unexpected Pass
Expression: abs(sim.𝒪est[:l∞] - 2) < 0.2
Got correct result, please change to @test if no longer broken.
```

All four are `NonlinearSolveAlg(TrustRegion())` convergence-order checks — TRBDF2 and QNDF2, in both the non-autonomous and autonomous Robertson blocks. They were the failures behind #3807 and #3818, and the recent NonlinearSolve releases (SciML/NonlinearSolve.jl#1010 recomputing the trust-region radius on `reinit!`, plus #1008) fixed them upstream.

This flips those four to `@test`. The two KenCarp4 order-4 assertions still fail and stay `@test_broken`.

Verified on a clean checkout of master (8ed2f7a5c1): before, `19 passed, 0 failed, 4 errored, 2 broken`; after, `23 passed, 2 broken, 0 errored`. The same 4 errors show up on the MTK lane of every open PR that touches this sublibrary, which is what makes that lane red.

## AI Disclosure


ChrisRackauckas:
I already got these running

PR #3898, opened 2026-07-13
Test Enzyme on current stable Julia

### LINKED ISSUE #3889: Test Enzyme on all Julia versions
https://github.com/SciML/OrdinaryDiffEq.jl/blob/6e8ab3a65426295821e9ad75f69d5c9a86bf1596/test/AD/autodiff_events.jl#L2

@ChrisRackauckas-Claude update things to test Enzyme on all julia versions, it is supported on all julia versions [zygote however should remain tested as is]

ChrisRackauckas-Claude:

Closes #3889.

## Summary

- Add the current stable Julia release to the AD test matrix alongside LTS.
- Run the Enzyme callback-event tests on every supported AD lane.
- Preserve Zygote's existing Julia version restriction and keep prerelease Julia out of the AD matrix.

## Testing

- `GROUP=AD julia +1.10 --project -e 'using Pkg; Pkg.test()'` passed during development (28,301 AD assertions passed, 2 broken; 11 callback-event assertions passed, 3 broken; 1 discrete-callback assertion passed).
- `GROUP=AD julia +1 --project -e 'using Pkg; Pkg.test()'` is running against the exact committed tree; the PR will be updated with its final summary.
- Repository-wide Runic check passed.

## Process

I inspected #3889 and the callback-event test history, then separated the shared Enzyme/Zygote version gate so each backend follows its supported Julia versions. Prerelease Julia was considered and explicitly excluded from the AD matrix per maintainer direction.

### DIFF
--- a/test/AD/autodiff_events.jl
+++ b/test/AD/autodiff_events.jl
@@ -1,12 +1,13 @@
 # Version-dependent AD backend selection
-# Enzyme/Zygote: Julia <= 1.11 only (see https://github.com/EnzymeAD/Enzyme.jl/issues/2699)
+# Enzyme: all AD test lanes
+# Zygote: Julia <= 1.11 only
 # Mooncake: works on all Julia versions for the ForwardDiffSensitivity-based
 #   gradient tests below via SciMLSensitivityMooncakeExt. Mooncake does NOT
 #   support ReverseDiffAdjoint (TypeError on the ODESolution CoDual), so the
 #   ReverseDiffAdjoint-based gradient tests are not run with Mooncake.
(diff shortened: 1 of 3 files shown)


ChrisRackauckas-Claude:
Implementation checkpoint: commit 7e71e3f6 adds the current stable Julia lane to the AD matrix, leaves prerelease excluded, and separates the Enzyme callback-event coverage from Zygote’s existing Julia-version gate. The Julia 1.10 AD group and repository-wide Runic check passed locally. The exact committed-tree Julia stable AD run is still in progress; I will post its final counts here.

PR #3901, opened 2026-07-14
Fix set_all_nzval!/get_nzval for non-CSC sparse matrices

hexaeder:
is_sparse is true for every AbstractSparseMatrix, but get_nzval and set_all_nzval! were only defined for SparseMatrixCSC, so GPU sparse matrices (CuSparseMatrixCSC/CSR/...) hit the "SparseArrays extension not loaded" stub in build_J_W even though SparseArrays was loaded. This makes a mass-matrix DAE with a cuSPARSE jac_prototype unsolvable with the default (concrete-A) linsolve.

Both helpers now dispatch on AbstractSparseMatrix via nonzeros.

## Checklist


hexaeder:
GPU test fail unrelated (fails on Master too), CUDA regression, should be fixed by https://github.com/JuliaGPU/CUDA.jl/pull/3199

PR #3902, opened 2026-07-14
Fix StabilizedRK Algorithims 

Sreeram-Shankar:
## Small Updates and Fixes for Stabilized RK Methods

- Added non-autonomous function evaluations at the intermediate polynomial stages for RKL and RKG methods. the function evaluations at the polynomial j evaluation now use a time increment and are accurate for non-autonomous systems

- TSRK2, RKC, and RKMC now use an increment of `th` rather than `th1` for their first stage evaluation

- ROCK 2 final stage now updates correctly for the final solution

### DIFF
--- a/lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
+++ b/lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
@@ -115,7 +115,6 @@ end
     stage_limiter!(uᵢ₋₁, integrator, p, tᵢ₋₁)
     if ccache.mdeg < 2
         @.. broadcast = false u = uᵢ₋₁
-        stage_limiter!(u, integrator, p, tᵢ₋₁)
     end
     # for the second to the ms[ccache.mdeg] th stages
     for i in 2:(ccache.mdeg)
@@ -139,13 +138,13 @@ end
     f(k, u, p, tᵢ₋₁)
     OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
     @.. broadcast = false uᵢ₋₁ = u + δt₁ * k
-    stage_limiter!(uᵢ₋₁, integrator, p, tᵢ₋₁)
     if integrator.opts.adaptive
         @.. broadcast = false tmp = -δt₂ * k
     else
         @.. broadcast = false u = -δt₂ * k
     end
     c = value(sign(δt₁)) * integrator.opts.internalnorm(δt₁, t)
+    stage_limiter!(uᵢ₋₁, integrator, p, tᵢ₋₁ + c)
     tᵢ₋₁ += c
     f(k, uᵢ₋₁, p, tᵢ₋₁)
     OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
@@ -347,7 +346,6 @@ end
     stage_limiter!(uᵢ₋₁, integrator, p, tᵢ₋₁)
     if ccache.mdeg < 2
         @.. broadcast = false u = uᵢ₋₁
-        stage_limiter!(u, integrator, p, tᵢ₋₁)
     end
     # for the second to the ccache.mdeg th stages
     for i in 2:(ccache.mdeg)
@@ -598,8 +596,8 @@ end
         OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
         @.. broadcast = false u = μ * gprev + ν * tmp + (T(1) - μ - ν) * uprev +
             dt * μs * (k - νs * fsalfirst)
-        stage_limiter!(u, integrator, p, t + dt * th1)
         th = μ * th1 + ν * th2 + μs * (T(1) - νs)
+        stage_limiter!(u, integrator, p, t + dt * th)
         if (iter < mdeg)
             @.. broadcast = false tmp = gprev
             @.. broadcast = false gprev = u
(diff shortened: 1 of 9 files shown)


lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -581,6 +598,7 @@ end
         OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
         @.. broadcast = false u = μ * gprev + ν * tmp + (T(1) - μ - ν) * uprev +
             dt * μs * (k - νs * fsalfirst)
+        stage_limiter!(u, integrator, p, t + dt * th1)
```
jlchan: Should this use `th` instead of `th1`?

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -1212,6 +1245,7 @@ end
         f(k, gprev, p, t + dt * th1)
         OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
         @.. broadcast = false u = μ * gprev + ν * tmp + dt * μs * k
+        stage_limiter!(u, integrator, p, t + dt * th1)
```
jlchan: See above

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -1620,10 +1661,14 @@ end
         f(k, uᵢ₋₁, p, t)
         OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
         @.. broadcast = false u = μⱼ * uᵢ₋₁ + νⱼ * tmp + (dt * μ̃ⱼ) * k
+        stage_limiter!(u, integrator, p, t)
```
jlchan: I think these should probably depend on `t + ...` but I'm not sure RKL methods track the increment explicitly.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -1620,10 +1661,14 @@ end
...
         @.. broadcast = false tmp = uᵢ₋₁
         @.. broadcast = false uᵢ₋₁ = u
     end
 
+    stage_limiter!(u, integrator, p, t + dt)
```
jlchan: I don't think we want this `stage_limiter!` call - it's already called at the end of the `for j in 2:s` loop, and then called again.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -1769,10 +1815,14 @@ end
             (1 - μⱼ - νⱼ) * uprev +
             (dt * μ̃ⱼ) * k +
             (dt * γ̃ⱼ) * fsalfirst
+        stage_limiter!(u, integrator, p, t)
         @.. broadcast = false tmp = uᵢ₋₁
         @.. broadcast = false uᵢ₋₁ = u
     end
 
+    stage_limiter!(u, integrator, p, t + dt)
```
jlchan: See above

jlchan (changes requested):
Thanks for adding this! I'm not super familiar with this implementation, but here were a few things that stood out.

lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
```diff
@@ -1620,10 +1661,14 @@ end
         f(k, uᵢ₋₁, p, t)
```
jlchan: I think this might need to also be updated so that it's evaluated at a stage-dependent `t`

ChrisRackauckas:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/3835 is the complete generalization, and I think we should just do that rather than doing it per-method.

Sreeram-Shankar:
Reverted the step and stage limiter changes, just the small changes in the methods now

ChrisRackauckas:
This needs a rebase on the pointer cycling, I'll take care of that.

ChrisRackauckas:
I rebased your commit in https://github.com/SciML/OrdinaryDiffEq.jl/pull/4091

Sreeram-Shankar:
have limiters for the stages and steps for stabilized_rk been added yet as part of the broader generalization? I dont see it yet in [https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl](url)

ChrisRackauckas:
that might've been missed.

jlchan:
Looks like https://github.com/SciML/OrdinaryDiffEq.jl/pull/3835 reported adding limiters to both stabilized and exponential RK methods, but didn't.

Sreeram-Shankar:
would it be possible to have them sooner rather than later?

ChrisRackauckas-Claude:
Confirmed: #3835 never added the stage hooks to `rkc_perform_step.jl`, so `stage_limiter` errored as unused for every StabilizedRK method (step limiters did work, since those are applied centrally). Stage limiters for all 13 methods (ROCK2, ROCK4, RKC, RKMC2, ESERK4, ESERK5, SERK2, TSRKC2, TSRKC3, RKL1, RKL2, RKG1, RKG2) are in https://github.com/SciML/OrdinaryDiffEq.jl/pull/4550. Usage is the solve-level keyword: `solve(prob, ROCK2(); stage_limiter = limiter!, step_limiter = limiter!)` with `limiter!(u, integrator, p, t)`.

🤖 Written by an AI agent (Devin CLI 3000.10.21, model Fusion — Claude Fable 5.1 High + SWE-2 Medium) on behalf of Chris Rackauckas; local session id `five-bladder`, no shareable URL.

PR #3916, opened 2026-07-15
Fix jacobian2W! DimensionMismatch for ScalarOperator mass matrix (#3915)

### LINKED ISSUE #3915: ScalarOperator mass matrix crashes in jacobian2W! (DimensionMismatch)
Follow-up to #3814. SciMLOperators#400 fixed the entry crash and #3837 fixed `get_differential_vars`, but a `ScalarOperator` mass matrix still crashes once the solver actually reaches a Newton step, in `jacobian2W!`.

`ScalarOperator` reports `axes(mm) == ()` (unlike `UniformScaling`, which the function special-cases), so it falls into the boundscheck meant for full matrices:

```julia
using OrdinaryDiffEqSDIRK, SciMLOperators

function lorenz!(du, u, p, t)
    du[1] = 10.0 * (u[2] - u[1])
end

f = ODEFunction(lorenz!; mass_matrix = ScalarOperator(1.0))
prob = ODEProblem(f, [1.0, 0.0, 0.0], (0.0, 1.0))
solve(prob, TRBDF2())
```

```
ERROR: DimensionMismatch: W: (Base.OneTo(3), Base.OneTo(3)), mass matrix: ()
```

`IdentityOperator` doesn't hit this (its axes happen to match `W`).

singhharsh1708:
Fixes #3915.

## Bug

`jacobian2W!`/`jacobian2W` special-case `mass_matrix isa UniformScaling` to skip the `axes(mass_matrix) == axes(W)` boundscheck and use the scalar directly. `ScalarOperator` is the same thing (λ·I) but isn't a `UniformScaling`, and reports `axes(mm) == ()`, so it falls through to the boundscheck meant for full matrices and throws.

```julia
using OrdinaryDiffEqSDIRK, SciMLOperators
f = ODEFunction((du, u, p, t) -> (du .= -u); mass_matrix = ScalarOperator(1.0))
solve(ODEProblem(f, [1.0], (0.0, 1.0)), TRBDF2())
# DimensionMismatch: W: (Base.OneTo(1), Base.OneTo(1)), mass matrix: ()
```

`IdentityOperator` doesn't hit this since its `axes` happen to match `W`.

## Fix

Treat `ScalarOperator` the same as `UniformScaling` in both `jacobian2W!` methods and `jacobian2W`, via `_is_scalar_massmatrix`/`_scalar_massmatrix_λ`. Follows #3814 → SciMLOperators#400 → #3837, the same operator-mass-matrix chain, one step further (this is the first place a `ScalarOperator` mass matrix reaches once the entry crash and `get_differential_vars` are both already fixed).

## Tests

New `scalar_operator_massmatrix_tests.jl` covers `jacobian2W!` through both the `::Matrix` specialization and the generic `::AbstractMatrix` method (reached with a sparse `J`), the out-of-place `jacobian2W`, and an end-to-end `TRBDF2()` solve of the case from #3915 — Lorenz with `mass_matrix = ScalarOperator(2.0)` against `mass_matrix = 2.0*I`. 7 assertions; the file errors on master with the reported `DimensionMismatch`.

Those two solves are in fact bit-identical (max abs difference 0.0 over `saveat = 0:0.1:1` at `abstol = reltol = 1e-12`, identical accepted-step counts) for TRBDF2, KenCarp4, FBDF and Rodas5P, and Rodas5P agrees to 7e-12 with a 1e-14 reference solve of the equivalent `M = I`, `f/2` problem.

`OrdinaryDiffEqDifferentiation` Core suite: 65 passes across 10 testsets (master: 61 across 9). QA also passes (JET 1/1, Aqua/ExplicitImports 15/15).

## Scope

This fixes the in-place W assembly, which is what #3915 reports. Still failing and unchanged by this PR, with different errors: out-of-place and StaticArrays `ScalarOperator` mass matrices (`Out-of-place NLNewton does not support non-concrete W`, and an `AssertionError` in the `StaticWOperator` path), and a sparse `jac_prototype` combined with a `ScalarOperator` mass matrix, which fails earlier in `prepare_user_sparsity`.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
@@ -23,7 +23,7 @@ import StaticArraysCore: StaticArray, StaticMatrix
 using SciMLBase: UJacobianWrapper, UDerivativeWrapper, _vec, _unwrap_val
 import SciMLBase: SciMLBase, @set, DEIntegrator, ODEFunction, SplitFunction, DAEFunction, remake, solve!
 import SciMLOperators: SciMLOperators, update_coefficients, update_coefficients!, MatrixOperator, AbstractSciMLOperator,
-    islinear, isconstant
+    islinear, isconstant, ScalarOperator
 import SparseMatrixColorings: ConstantColoringAlgorithm, GreedyColoringAlgorithm, ColoringProblem,
     ncolors, column_colors, coloring, sparsity_pattern
 import OrdinaryDiffEqCore
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
@@ -578,6 +578,15 @@ function do_newJW(integrator, alg, nlsolver, repeat_step)::NTuple{2, Bool}
     end
 end
 
+# `ScalarOperator` (λ·I) reports `axes(mm) == ()` like `UniformScaling`, but unlike
+# `UniformScaling` it isn't matched by `isa UniformScaling` -- treat both as the same
+# scalar-times-identity case rather than requiring `axes(mm) == axes(W)`.
+_is_scalar_massmatrix(mm) = false
+_is_scalar_massmatrix(::UniformScaling) = true
+_is_scalar_massmatrix(::ScalarOperator) = true
+_scalar_massmatrix_λ(mm::UniformScaling) = mm.λ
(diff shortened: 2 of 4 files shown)


ChrisRackauckas:
A ScalarOperator mass matrix doesn't seem valid, how does that come up?

singhharsh1708:
Time-varying scalar  M(t) = c(t)I, which UniformScaling can't express. Constant case is redundant, agreed.

ChrisRackauckas:
But nothing is testing that?

singhharsh1708:
added now.

ChrisRackauckas (approved):

Issue #3809, opened 2026-07-01
Follow-up improvements for GaussLegendre


Sreeram-Shankar:
## Follow-up improvements for GaussLegendre

Following up on #3454 and #3470, I’m planning to look into a set of improvements for the newly added arbitrary-stage GaussLegendre FIRK methods.

The main areas for potential improvement would be: 

- Dense output / interpolation support, especially for continuous output and callbacks
- Better step-size error estimation, if there is a suitable approach that fits these methods
- Efficiency improvements for the nonlinear / linear solve path, possibly including ways to expolit the Gauss-Legendre structure for faster coupled stage solving, similar to Radau's Schur decomposition of the Butcher tableau

Adaptive step-size control may not be the highest-priority direction because it can interfere with the symplectic properties that motivate using Gauss methods.

I wanted to ask whether there are existing routines, internal SciML patterns, or references that would be best to follow for:

- Implementing dense output / interpolation for FIRK collocation methods
- Reusing or adapting existing interpolation code from Radau/FIRK methods
- Improving the system solve implementation for Gauss-Legendre methods to improve efficiecny
- Any preferences for an adaptive step size controller if needed, upgrading Richardson extrapolation potentially to an embedded estimator
- Any other requests for the GL FIRK methods

ChrisRackauckas:
All good directions. I think looking at different nonlinear solver approaches, which should already be accessible but should probably get benchmarked, would be big too. Anderson Acceleration could be the winner for these methods.

PR #3920, opened 2026-07-16
First Updates for GaussLegendre FIRK

Sreeram-Shankar:
## Updates for GaussLegendre FIRK 

- Replace Richardson step-doubling adaptivity with an embedded error estimate. Build lower-order weights b-hat that match the quadrature moments of order s-1 with the s-th moment set to 0, take e = b - bhat, and estimate the local error as utilde = dt * Σ eᵢ f(Yᵢ). Adaptive order is now s-1 instead of 2s, and accepted steps no longer need the cost of the Richardson controller

- Fix the old embedded tableau, which interpolated an (s-1)-stage Gauss rule onto the s-stage nodes and produced a poor e (nearly the whole step at s = 2). The new Vandermonde construction for bhat gives a proper same-node embedding

- Add a dense collocation extrapolant as the Newton starting guess (extrapolant = :dense by default). After an accepted step, reuse the previous collocation polynomial in Newton form on {0, c₁, …, cₛ}, scaled by dt/dtprev. :constant still starts from zero

- Remove unused smooth_est and old caches

- Call and add step_limiter! to both OOP and IIP methods

Assisted by Cursor Agent

Part of #3809

### DIFF
--- a/lib/OrdinaryDiffEqFIRK/src/alg_utils.jl
+++ b/lib/OrdinaryDiffEqFIRK/src/alg_utils.jl
@@ -34,10 +34,10 @@ default_controller(QT, alg::GaussLegendre) = PIController(QT, alg)
 
 isfirk(alg::GaussLegendre) = true
 
-# Richardson step-doubling controller
+# Embedded error estimate controller using s-1 stages
 isadaptive(alg::GaussLegendre) = alg.num_stages >= 2
-alg_adaptive_order(alg::GaussLegendre) = 2 * alg.num_stages
+alg_adaptive_order(alg::GaussLegendre) = alg.num_stages - 1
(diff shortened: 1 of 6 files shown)


oscardssmith (approved):

oscardssmith:
This looks a lot better! how do the benchmarks look?

Sreeram-Shankar:
It seems to run better than before, fewer Newton iterations with the dense extrapolant, a better error estimate, and less work for adaptivity. I don't see any dedicated benchmarks but I can put together a proper benchmark vs Radau if useful

ChrisRackauckas:
I rebased and merged your commit in https://github.com/SciML/OrdinaryDiffEq.jl/pull/4117

Issue #3922, opened 2026-07-16
GPU DAE tests: 100% failure since GPU CI lane restored (#3813), likely Diagonal mass matrix scalar indexing


singhharsh1708:
Every case in `test/gpu/dae_tests.jl` fails on master right now (`✗ gpu error`, 144/144 in the last run, 0 passes anywhere in the GPU job). Confirmed on master's own latest commit, not tied to any open PR.

## Not new, but newly visible

The top-level GPU CI lane was off until #3813 (2026-07-03) turned it back on. Checked master's CI history back to early June: no GPU job ran at all before that. Since #3813, it has never once passed.

## Root cause (one of them, at least)

`get_differential_vars` (`lib/OrdinaryDiffEqCore/src/misc_utils.jl:216`) does:

```julia
elseif all(!iszero, mm)
```

on the full mass matrix before checking `_isdiag`. For a `Diagonal` mass matrix, `all(!iszero, ::AbstractMatrix)` visits every `(i, j)` pair via `getindex`, including calling `getindex` on the underlying diagonal vector one element at a time for `i == j`. `test/gpu/dae_tests.jl` uses `mass_matrix = cu(Diagonal([1, 1, 0, 0]))`, i.e. `Diagonal{Float64, <:CuVector}` — CUDA disallows scalar `getindex` on device arrays by default, so this throws on every single case that reaches DAE initialization, regardless of solver or `jac_prototype`. That's consistent with the 100% failure rate across the whole test matrix (even `jac=none` cases fail).

## Can't fully verify

No GPU hardware available here. `test/gpu/dae_tests.jl` also has a harness gap that made this harder to pin down: it defines `show_errors(results)` to print the actual caught exceptions, but the `@testset` entry point never calls it — CI logs only ever show `✗ gpu error` with no exception text.

ChrisRackauckas-Claude:
Status update from master `4e1a434` (2026-07-18, [CI run 29629348645](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/29629348645), `tests / GPU` job 88040285664):

* The job is a genuine test failure, not runner infra — the self-hosted runner completes the suite and uploads a full log.
[...]

PR #3199, opened 2026-03-19
Make internal nlsolve use possibly-smaller chunksize, distinct from ODE chunksize

Ickaser:
## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [x] Any new documentation only uses public API
  
## Additional context
Fixes #3157. 
The BrownFullBasic initialization solves a nonlinear problem only for the _algebraic_ variables, so if the ODEProblem has chunksize larger than the number of algebraic problems, the initialization needs to pick a smaller chunksize. 

As it stands, this PR fixes the error, but the initialization for out-of-place problems runs into a new error I'm less confident about solving:

```
ERROR: TypeError: in setfield!, expected Vector{Float64}, got a value of type SubArray{Float64, 1, Vector{Float64}, Tuple{Vector{Int64}}, false}
Stacktrace:
  [1] setproperty!(cache::LinearSolve.LinearCache{…}, name::Symbol, x::SubArray{…})
    @ LinearSolve C:\Users\iwheeler\.julia\packages\LinearSolve\LKo1P\src\common.jl:158
  [2] (::NonlinearSolveBase.LinearSolveJLCache{…})(; A::Matrix{…}, b::SubArray{…}, linu::Vector{…}, reuse_A_if_factorization::Bool, kwargs::@Kwargs{…})
    @ NonlinearSolveBaseLinearSolveExt C:\Users\iwheeler\.julia\packages\NonlinearSolveBase\Rw8Vf\ext\NonlinearSolveBaseLinearSolveExt.jl:21
  [3] solve!(cache::NonlinearSolveBase.NewtonDescentCache{…}, J::Matrix{…}, fu::SubArray{…}, u::Vector{…}, idx::Val{…}; skip_solve::Bool, new_jacobian::Bool, kwargs::@Kwargs{…})
    @ NonlinearSolveBase C:\Users\iwheeler\.julia\packages\NonlinearSolveBase\Rw8Vf\src\descent\newton.jl:120
  [4] step!(cache::NonlinearSolveFirstOrder.GeneralizedFirstOrderAlgorithmCache{…}; recompute_jacobian::Nothing)
[...]

PR #3923, opened 2026-07-16
get_differential_vars: check Diagonal.diag, not the full matrix

singhharsh1708:
Rework of the original patch, rebased onto current master (dc7f91af).

The GPU DAE lane has recovered on master without this change. After the CUDA environment rework on the self-hosted runner, `GPU: dae_tests.jl` stands at 136 passed / 1 failed / 7 broken as of master run [30340807357](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/30340807357) (Jul 28); the one remaining failure is `Hairer4 [jac=CSC, mass=diag_cu]`, a different problem. So the earlier "fixes 100% GPU DAE failure" framing no longer applies, and #3922 should be closed or retitled to track the remaining Hairer4 case. What this PR still fixes is the access pattern.

`get_differential_vars` (`lib/OrdinaryDiffEqCore/src/misc_utils.jl`) checks the mass matrix with `all(!iszero, mm)` before its `_isdiag` branch. For a `Diagonal` that walks all n^2 positions, and every diagonal hit is a scalar `getindex` on the underlying vector, which is disallowed by default when that vector is a `CuVector`, and O(n^2) for a question the type answers in O(n). The fix adds a `Diagonal` dispatch that broadcasts over `mm.diag`, so the reduction is O(n) and stays on device. This is the same diagonal-only reduction `find_algebraic_vars_eqs(::Diagonal)` in the same file performs, though that one goes through `diag(M)` rather than the field.

CPU results are unchanged for every case that previously reached the `_isdiag` branch. Mixed, all-nonzero and all-zero diagonals return exactly what the old `_isdiag` path returned, including the `reshape` to `size(u)` for matrix-valued `u`, and all three are covered by new tests.

The access pattern itself is now tested on CPU: a `Diagonal` wrapping a vector type that errors on scalar `getindex` but supports whole-array broadcast, i.e. device-array semantics without a GPU. On master's code path that construction throws inside `all(!iszero, mm)`; with this patch it returns the correct differential-vars mask.

One test adjustment was required. The `Diagonal` branch is statically resolvable, which fixes inference of `solve` for the two Rodas4 + `ShampineCollocationInit` cases at `lib/OrdinaryDiffEqNonlinearSolve/test/mass_matrix_tests.jl:385/387`. Those lines were `@test_broken sol = @inferred solve(...)`, a form that errors with "Expression evaluated to non-Boolean" once `@inferred` succeeds, because the successful `@inferred` returns the solution object. Both lines now read `@test (@inferred solve(...); true)`: green while inference holds, an error if it regresses. I checked the replacement form against master's code and it does error there, so it is not a rubber stamp. The `Rosenbrock23` and `Rodas5` lines are untouched and stay `@test_broken`; those two fail because `solve` itself throws `CheckInitFailureError`, which this change does not affect.

Also wires up `show_errors(results)` in the GPU DAE test entry point; it was defined but never called, so a failing case only ever logged a status glyph with no exception text. That file runs only on the self-hosted GPU lane.

Local verification on Julia 1.12.6, rebased on master:

- `lib/OrdinaryDiffEqCore/test/algebraic_vars_detection_tests.jl`: 22 passed / 0 failed (18 on master, so 4 new assertions).
- `lib/OrdinaryDiffEqNonlinearSolve/test/mass_matrix_tests.jl`: 227 passed / 0 failed / 0 errored / 40 broken. Master baseline on the same file is 225 passed / 42 broken; the two that flip are the Rodas4 lines above.
- `lib/OrdinaryDiffEqBDF`: `dae_convergence_tests.jl` 50/50, `dae_initialization_tests.jl` 38 passed / 2 broken (pre-existing).
- Runic 1.5.1 clean on all four changed files.

No GPU hardware here, so the CUDA path is covered by the CPU device-semantics guard test rather than on-device.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/misc_utils.jl
+++ b/lib/OrdinaryDiffEqCore/src/misc_utils.jl
@@ -213,6 +213,12 @@ function get_differential_vars(f, u)
         elseif mm isa SciMLOperators.ScalarOperator
             # λ·I: every variable is algebraic when λ is zero.
             return iszero(mm.val) ? falses(size(u)) : nothing
+        elseif mm isa Diagonal
+            # Check only `mm.diag`: the generic paths below visit every (i, j)
+            # pair via scalar `getindex`, which is disallowed when `mm.diag` is
+            # GPU-backed (e.g. `Diagonal{Float64, <:CuVector}`), while a
+            # broadcast over `mm.diag` dispatches to a proper GPU kernel.
+            return reshape(mm.diag .!= 0, size(u))
         elseif all(!iszero, mm)
             return trues(size(mm, 1))
         elseif !(mm isa SciMLOperators.AbstractSciMLOperator) && _isdiag(mm)
--- a/lib/OrdinaryDiffEqCore/test/algebraic_vars_detection_tests.jl
+++ b/lib/OrdinaryDiffEqCore/test/algebraic_vars_detection_tests.jl
@@ -113,3 +113,37 @@ end
     ff_singular = ODEFunction(f!, mass_matrix = ScalarOperator(0.0))
     @test get_differential_vars(ff_singular, u) == falses(3)
 end
+
+@testset "get_differential_vars: Diagonal mass matrix" begin
+    f!(du, u, p, t) = (du .= -u; nothing)
+    u = zeros(4)
+    ff_mixed = ODEFunction(f!, mass_matrix = Diagonal([1.0, 1.0, 0.0, 0.0]))
+    @test get_differential_vars(ff_mixed, u) == Bool[1, 1, 0, 0]
+    ff_alldiff = ODEFunction(f!, mass_matrix = Diagonal([1.0, 2.0, 3.0, 4.0]))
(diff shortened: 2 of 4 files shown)


ChrisRackauckas-Claude:
The GPU job on this PR gets past the `get_differential_vars` path, but it exposes a second independent blocker in the same `Simple DAE GPU` test: [`87623782250`](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/29498827599/job/87623782250) fails at `simple_dae.jl:54` with `DimensionMismatch: array could not be broadcast to match destination`. The first package frame is the three-argument `ldiv!(x::CuMatrix, qr::QR, b::CuMatrix)` in cuSOLVER `linalg.jl:301`; `x` and `b` are both matrix right-hand sides. That exact upstream defect and regression are already addressed by [CUDA.jl #3199](https://github.com/JuliaGPU/CUDA.jl/pull/3199).

This appears to confirm that this PR cleared the earlier scalar-indexing barrier, after which CI reached the unrelated QR barrier. I would keep this PR focused; its GPU job cannot become green until the CUDA fix is available in the resolved environment.

PR #3935, opened 2026-07-17
Fuse RosenbrockCache stage accumulation loops into single-sweep SIMD kernels

ChrisRackauckas-Claude:
> [!IMPORTANT]

## Summary

The 2026-03 consolidation of the Rosenbrock `perform_step!` into a single tableau-driven loop replaced the old hand-fused stage kernels with per-coefficient broadcast accumulations. For an s-stage method every stage does `u .= uprev` followed by `stage-1` separate `@.. u += A[stage,i]*ks[i]` sweeps, and likewise for the `C`-coefficient RHS accumulation, the final `b`-weight sum, the `btilde` error sum, and the dense-output `H`-row sums. For Rodas5P that is ~116 full-vector sweeps per step where the pre-consolidation kernels did ~21.

This PR fuses each of those accumulations back into a single sweep without giving up the unified tableau-driven architecture:

- Tableau coefficients for a stage are hoisted into an `NTuple` through a bounded `Base.Cartesian.@nif` ladder (max 8 fused terms), and the linear combination is expanded statically via tuple recursion so the single sweep SIMD-vectorizes. This is the "single function for max N, not per-`Val{N}` specializations" approach suggested in the message of the earlier revert commit 27f2f3a — there is no per-stage-count compile-time blowup.
- `Array` states take the fused path; anything else (GPU arrays etc.) and >8-term accumulations (Rodas6P's late stages) keep the existing per-coefficient broadcast loop as a fallback.
- `f(fsalfirst, uprev, p, t)` is now skipped on `repeat_step` — `uprev` and `t` are unchanged on a repeated step, and stage 1's RHS (`linsolve_tmp = fsalfirst + dtd1*dT` in `calc_tderivative!`) is rebuilt from the stored `fsalfirst` each call, so the value is still valid. Saves one f-eval per repeated step.

An intermediate version of this patch used a scalar-indexed loop with a runtime-length inner loop over stages; that defeated SIMD and coefficient hoisting and was ~12% *slower* at n=2000. The `@nif`+`NTuple` version is what survived benchmarking.

## Benchmarks

3 interleaved rounds, `BenchmarkTools` minima, Julia 1.10.11, default solver options unless noted. Step counts/stats identical between master and this PR in every case.

| problem | master | this PR | Δ |
|---|---|---|---|
| ROBER n=3, tol=1e-8, dense output | 0.929–0.933 ms | 0.859–0.868 ms | **−7.3%** |
| Van der Pol μ=1e6 n=2, tol=1e-8 | 32.06–32.46 ms | 29.71–30.26 ms | **−7.7%** |
| Brusselator 8×8 (n=128), FullSpecialize, RFLU | 39.3 ms | 39.2 ms | parity (LU-dominated) |
| 1-D reaction-diffusion n=2000, sparse J | 123.6 ms avg | 122.1 ms avg | −1% |

Correctness: final states and interpolants agree with master to ~13 significant digits (the fused accumulation changes float summation association — same association as the pre-consolidation kernels); identical accept/reject/step counts on ROBER, Brusselator, and a Rodas5P mass-matrix DAE.

The interesting context from the same investigation (filed separately as issues): the dominant Rodas5P performance regressions vs OrdinaryDiffEq v6.87 are not in this loop at all but in linear-solve defaults (RecursiveFactorization no longer loaded by anything → LAPACK `getrf!` for dense W, and LAPACK `getrs!` ccall overhead dominating tiny systems).

## Process notes

Investigation and benchmarks were run locally on the shared benchmark box; benchmark scripts compared this branch against master and against OrdinaryDiffEq v6.87.0 (last release with the hand-unrolled Rodas5 kernels). `Pkg.test()` for OrdinaryDiffEqRosenbrock run locally (result posted in comments). Runic applied (no changes needed).

### DIFF
--- a/lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_perform_step.jl
+++ b/lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_perform_step.jl
@@ -575,26 +575,63 @@ end
 
 # Fused single-sweep accumulation kernels for the stage loops below. Each
 # replaces a sequence of per-coefficient broadcasts — one full memory sweep per
-# tableau entry — with a single sweep using muladd, which is where the
-# pre-consolidation hand-unrolled kernels got their speed. `Array` states (incl.
-# N-d arrays; linear indexing) take these; anything without fast scalar indexing
-# (GPU etc.) keeps the broadcast fallback.
+# tableau entry — with a single sweep, which is where the pre-consolidation
+# hand-unrolled kernels got their speed. The tableau coefficients are hoisted
+# into an NTuple through a bounded `@nif` ladder (single function, not
+# per-`Val{num_stages}` specializations, so no compile-time blowup) and the
+# inner term expansion unrolls statically so the sweep SIMD-vectorizes.
+# `Array` states take these; anything without fast scalar indexing (GPU etc.)
+# keeps the per-coefficient broadcast fallback. Beyond 8 fused terms (only
+# Rodas6P's late stages) the fallback path is used as well.
+
+@inline function _madd_terms(acc, idx, cs::Tuple, karrs::Tuple)
+    return _madd_terms(
+        muladd(first(cs), (@inbounds first(karrs)[idx]), acc),
+        idx, Base.tail(cs), Base.tail(karrs)
+    )
+end
+@inline _madd_terms(acc, idx, ::Tuple{}, ::Tuple{}) = acc
+
+@inline function _fused_lincomb!(
+        out::Array, base::Array, cs::NTuple{N, Any}, karrs::NTuple{N, Any}
+    ) where {N}
+    @inbounds @simd ivdep for idx in eachindex(out)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Local `Pkg.test()` for OrdinaryDiffEqRosenbrock on this branch (Julia 1.10.11): all test groups pass except one pre-existing failure that also exists on unmodified master — the SciMLTesting QA check `public API has docstrings` fails with `isempty([:HybridExplicitImplicitRK])` (that name became public without a docstring; unrelated to this diff, which only adds `_`-prefixed private helpers). A separate fix is being prepared for that. Allocation tests: 19 broken (pre-existing markers), Inference tests 38/38 pass, JET passes.

Benchmark scripts used for the numbers in the description: interleaved 3-round runs of ROBER/VdP/Brusselator/1-D reaction-diffusion comparing this branch, master, and OrdinaryDiffEq v6.87.0 on the same machine.

PR #3913, opened 2026-07-15
Fix structured Jacobian cache initialization

ChrisRackauckas-Claude:
## Summary

- initialize only the stored entries of structured Jacobian and W caches
- preserve implicit structural zeros for `Diagonal`, `Bidiagonal`, and `Tridiagonal` prototypes
- add regressions for both bidiagonal orientations, diagonal/tridiagonal initialization, and a full diagonal Rosenbrock solve

## Investigation

A clean upstream-master checkout reproduced the structured-cache failure:

```
ArgumentError: array of type Diagonal{Float64, ...} ... cannot be filled with 1.0
```

The immediate parent `b797bac1bfd94b1615026d0601ede22d1aa4dd92` passes the targeted reproducer. An independent `git bisect` identified `d8cbb9e4446427305484e25eb4c81c9f85c09277` ("Initialize sparse Jacobian caches with stored values (#3833)") as the first bad commit.

That change used a nonzero `fill!` fallback for non-sparse prototypes. Structured matrices have implicit entries that cannot be assigned a nonzero value, so the fallback fails before the solver can use the cache. This patch uses the public `ArrayInterface.findstructralnz` API and initializes each stored structural position while leaving implicit zeros untouched.

## Local verification

- Julia 1.12, `ODEDIFFEQ_TEST_GROUP=Core`: exit 0; `Structured Jacobian Prototypes | 6/6`; package tests passed.
- Julia 1.10, `ODEDIFFEQ_TEST_GROUP=Core`: exit 0; `Structured Jacobian Prototypes | 6/6`; package tests passed.
- Julia 1.12 sparse test project: exit 0; `Non-Full Diagonal Sparsity Tests | 2/2`; package tests passed.
- `julia +1.12 --startup-file=no -m Runic --check .`: exit 0.
- `git diff --check upstream/master...HEAD`: exit 0.
- The branch is current with `upstream/master` (0 commits behind).

A broader DiffEqBase downstream run did execute the structured linear-solve tests, then failed later in the unrelated FlexUnits 0.5.7 regression tracked in #3925. I am not claiming that full downstream group is green.

Please ignore this PR until reviewed by @ChrisRackauckas.

PR #3838, opened 2026-07-09
Bump OrdinaryDiffEqDifferentiation to 3.2.3

ChrisRackauckas-Claude:
## Summary

Ignore this PR until it is reviewed by @ChrisRackauckas.

This bumps `OrdinaryDiffEqDifferentiation` to `3.2.3` so the already-merged downstream `constructorof` import cleanup can be released. The registered `3.2.2` package still imports `SciMLBase.constructorof` and then `ConstructionBase.constructorof`, which reproduces the Julia 1.12 precompile warning reported in SciML/SciMLBase.jl#1426 after SciMLBase v3.33.0.

This also replaces a stale machine-local sparse test project source path with the repo-relative path used by the other nested test projects.

## Verification

- Reproduced the issue with a fresh registered-package environment on Julia 1.12.4: adding `OrdinaryDiffEqDifferentiation v3.2.2` with `SciMLBase v3.33.0` and running `Pkg.precompile()` emits the `SciMLBase.constructorof` / `ConstructionBase.constructorof` warning.
- `julia --project=lib/OrdinaryDiffEqDifferentiation -e 'using Pkg; Pkg.instantiate(); Pkg.precompile()'` passes for the local checkout; the `OrdinaryDiffEqDifferentiation` warning is absent. Only the pre-existing SciMLBase docs replacement warning appears.
- `julia --project=lib/OrdinaryDiffEqDifferentiation -e 'using Pkg; Pkg.test()'` passes.
- `ODEDIFFEQ_TEST_GROUP=Sparse julia --project=lib/OrdinaryDiffEqDifferentiation -e 'using Pkg; Pkg.test()'` passes after restoring the sparse test project to the committed relative path.
- `git diff --check` passes.

## Note

A repo-wide `Runic` check with Runic v1.7.0 currently reports pre-existing formatting diffs in Julia files unrelated to this TOML-only release bump. I did not fold that unrelated formatting churn into this PR.

PR #3942, opened 2026-07-17
Fix build_J_W cache init for structured jac_prototypes

ChrisRackauckas-Claude:
> [!NOTE]

## Problem

`lib/DiffEqBase`'s `[Downstream]` test group fails on master in `test/downstream/default_linsolve_structure.jl`:

```
ArgumentError: array of type Diagonal{Float64, Vector{Float64}} and size (2, 2) can
not be filled with 1.0, since some of its entries are constrained.
```

thrown from `build_J_W` at `lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl` when solving an `ODEProblem` whose `jac_prototype` is a structured matrix (`Diagonal`, `Tridiagonal`, ...) with a default (factorization) linsolve, e.g.

```julia
fun = ODEFunction(f; jac = jac, jac_prototype = Diagonal(zeros(2)))
solve(ODEProblem(fun, ones(2), (1.0, 10.0)), Rosenbrock23())
```

## Root cause

Bisected to d8cbb9e4446427305484e25eb4c81c9f85c09277 ("Initialize sparse Jacobian caches with stored values", #3833). That commit initializes the `J`/`W` caches allocated in `build_J_W` with ones so factorizations never see undef garbage:

```julia
if is_sparse(J)
    set_all_nzval!(J, one(eltype(J)))
    set_all_nzval!(W, one(eltype(W)))
else
    fill!(J, one(eltype(J)))
    fill!(W, one(eltype(W)))
end
```

But LinearAlgebra's `fill!` on banded structured matrices (`Diagonal`, `Bidiagonal`, `Tridiagonal`, `SymTridiagonal`) throws for any nonzero value because the off-band entries are constrained to zero. So every structured (non-sparse, non-dense) `jac_prototype` on the default-linsolve path now errors at cache construction. Verified by running the reproducer on d8cbb9e44~1 with the same current dependency versions: it passes there and fails on master, so this is a code regression from #3833, not registry drift (the failure only surfaced later because the monorepo's own CI lanes for #3833 did not run the DiffEqBase Downstream group).

## Fix

Add a `fill_stored!` helper that writes only the stored bands of the LinearAlgebra structured matrix types and falls back to `fill!` otherwise, and use it in `build_J_W`'s non-sparse initialization. Semantics for dense and sparse prototypes are unchanged.

## Tests

- New `fill_stored!` unit tests and a structured-`jac_prototype` init/solve regression test in `lib/OrdinaryDiffEqDifferentiation/test/nzval_helpers_tests.jl` (asserts the `J`/`W` caches are initialized with stored ones, mirroring the #3833 testset, and that the solve completes).
- The previously failing downstream file `lib/DiffEqBase/test/downstream/default_linsolve_structure.jl` passes locally with this change (output in the PR conversation).

This resolves the red `lib/DiffEqBase [Downstream]` lane seen on #3692 (which merely surfaced the pre-existing master failure).

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
@@ -11,7 +11,8 @@ import FunctionWrappersWrappers
 import DiffEqBase
 
 import LinearAlgebra
-import LinearAlgebra: Diagonal, I, UniformScaling, diagind, opnorm
+import LinearAlgebra: Bidiagonal, Diagonal, I, SymTridiagonal, Tridiagonal,
+    UniformScaling, diagind, opnorm
 import LinearAlgebra: LowerTriangular
 import ArrayInterface
 
@@ -69,6 +70,15 @@ nonzeros(A) = error("SparseArrays extension not loaded. Please load SparseArrays
 spzeros(args...) = error("SparseArrays extension not loaded. Please load SparseArrays to use sparse matrix functionality.")
 get_nzval(A) = error("SparseArrays extension not loaded. Please load SparseArrays to use sparse matrix functionality.")
 set_all_nzval!(A, val) = error("SparseArrays extension not loaded. Please load SparseArrays to use sparse matrix functionality.")
+
+# `fill!` on LinearAlgebra's banded structured matrices throws for nonzero
+# values since the off-band entries are constrained to zero, so write the
+# stored bands directly instead.
+fill_stored!(A, v) = fill!(A, v)
+fill_stored!(A::Diagonal, v) = (fill!(A.diag, v); A)
+fill_stored!(A::Bidiagonal, v) = (fill!(A.dv, v); fill!(A.ev, v); A)
+fill_stored!(A::Tridiagonal, v) = (fill!(A.dl, v); fill!(A.d, v); fill!(A.du, v); A)
+fill_stored!(A::SymTridiagonal, v) = (fill!(A.dv, v); fill!(A.ev, v); A)
 
 include("alg_utils.jl")
 include("linsolve_utils.jl")
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
@@ -1207,8 +1207,8 @@ function build_J_W(
             set_all_nzval!(J, one(eltype(J)))
             set_all_nzval!(W, one(eltype(W)))
         else
-            fill!(J, one(eltype(J)))
-            fill!(W, one(eltype(W)))
+            fill_stored!(J, one(eltype(J)))
+            fill_stored!(W, one(eltype(W)))
         end
     elseif (
             IIP && (concrete_jac(alg) === nothing || !concrete_jac(alg)) &&
--- a/lib/OrdinaryDiffEqDifferentiation/test/nzval_helpers_tests.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/test/nzval_helpers_tests.jl
@@ -1,4 +1,5 @@
 using OrdinaryDiffEqDifferentiation
+using OrdinaryDiffEqRosenbrock
 using SparseArrays
 using LinearAlgebra
 using Test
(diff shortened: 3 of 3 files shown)


ChrisRackauckas-Claude:
Local verification (Julia 1.12, current registry, SciMLBase 3.36.0):

**Bisect evidence** — reproducer on `d8cbb9e44~1` (b797bac1b) with identical dep versions:
```
PRE3833 RESULT: [10.0, 10.0]  nsteps=46
```
Same reproducer on unmodified master: `ArgumentError: array of type Diagonal{Float64, Vector{Float64}} ... cannot be filled with 1.0` from `build_J_W` at `derivative_utils.jl:1210`.

**With this PR** — reproducer:
```
Diagonal => [10.0, 10.0] nsteps=46
Tridiagonal => [10.0, 10.0] nsteps=46
```

**Previously failing downstream file** `lib/DiffEqBase/test/downstream/default_linsolve_structure.jl`:
```
Test Summary:                   | Pass  Broken  Total   Time
Default linsolve with structure |    8       2     10  43.1s
```

**`ODEDIFFEQ_TEST_GROUP=Core julia --project -e 'using Pkg; Pkg.test()'`** (includes the new `fill_stored!` unit tests and the structured jac_prototype init/solve regression testset): passed, exit 0.

**`ODEDIFFEQ_TEST_GROUP=QA`**: `Testing OrdinaryDiffEqDifferentiation tests passed`, exit 0.

ChrisRackauckas-Claude:
Status check from an independent triage run today (2026-08-08, master `c11ff74bb`): this is still live, and it is now breaking a downstream repo's CI.

**Three open drafts fix the same `fill!` in `build_J_W`** and are waiting on review: this one, #3913, and #4072. They need consolidating down to one.

**Downstream impact.** The `Downstream` test group of JuliaDiff/FiniteDiff.jl fails on FiniteDiff master because of this: `test/downstream/ordinarydiffeq_tridiagonal_solve.jl` errors at cache construction with `OrdinaryDiffEqDifferentiation` v3.7.0. Standalone reproducer (no FiniteDiff, default `AutoForwardDiff`/`AutoSparse` — the AD backend is irrelevant, this fires before any Jacobian is computed):

```julia
using OrdinaryDiffEq, OrdinaryDiffEqRosenbrock, LinearAlgebra

const n = 10
f(du, u, p, t) = (du .= 0;
    du[2:(end - 1)] .= p[1] .* (u[3:end] .- 2 .* u[2:(end - 1)] .+ u[1:(end - 2)]);
    nothing)

u0 = sin.(range(0, 1, length = n))
jp = Tridiagonal(similar(u0, n - 1), similar(u0), similar(u0, n - 1))
prob = ODEProblem(ODEFunction(f; jac_prototype = jp), u0, (0.0, 1.0), [0.42])
solve(prob, Rodas4P(), saveat = 0.1)
```

```
ERROR: ArgumentError: array of type Tridiagonal{Float64, Vector{Float64}} and size (10, 10) can
    not be filled with 1.0, since some of its entries are constrained.
  [1] fill!  @ .../LinearAlgebra/src/special.jl:358 [inlined]
  [2] build_J_W  @ .../OrdinaryDiffEqDifferentiation/G755Y/src/derivative_utils.jl:1242
```

Regression range confirms this PR's bisect: `git log -S 'fill!(J, one(eltype(J)))'` gives only `d8cbb9e44` (#3833, 2026-07-08), first released in `OrdinaryDiffEqDifferentiation` v3.2.3 (#3838).

**One measured difference between the three PRs, for whoever picks.** `BandedMatrix` prototypes hit the same `fill!` and are also broken today:

```
ArrayInterface.has_sparsestruct(jp) = true
ArrayInterface.isstructured(jp) = true
fill! on BandedMatrix threw: BandError: attempt to access BandedMatrix{Float64, Matrix{Float64}, Base.OneTo{Int64}} with bandwidths (1, 1) at band 7
solve threw: BandError: ... at band 7
```

This PR's `fill_stored!` keeps `fill!` as the fallback, so a `BandedMatrix` prototype still throws `BandError`; #3913 and #4072 dispatch through `ArrayInterface.has_sparsestruct` / `isstructured`, which are `true` for `BandedMatrix`, so they cover it. Caveat: with the `fill!` removed for banded, that path then fails on a separate pre-existing `lu!` ambiguity between `ArrayLayouts` and `LinearAlgebra`, so `BandedMatrix` is not a working configuration either way — it just argues for the `ArrayInterface`-based dispatch of #3913/#4072 over enumerating the LinearAlgebra band types.

**Independent verification** of an equivalent patch (band-writing `fill_stored!`, same as this PR) on today's master, Julia 1.12.6, `lib/OrdinaryDiffEqDifferentiation`:

- reproducer above: `ArgumentError` before, `ReturnCode.Success` after;
- FiniteDiff's `ordinarydiffeq_tridiagonal_solve.jl`: errors before, passes after;
- `GROUP=Core`: all 12 testsets green (`Testing OrdinaryDiffEqDifferentiation tests passed`), including a new solve-level `Tridiagonal`/`Diagonal` testset;
- `GROUP=QA`: `JET 1/1`, `Aqua 18/18`, passed.

No fourth PR opened. Suggest merging whichever of the three you prefer (my read: #4072 or #3913 for the `ArrayInterface` dispatch) and closing the other two, then cutting a patch release of `OrdinaryDiffEqDifferentiation` — the fix does not reach users, or FiniteDiff's CI, without one.

Issue #3613, opened 2026-05-10
VectorContinuousCallback silently ignores affect_neg! in DiffEqBase v7 (lib/DiffEqBase)


ChrisRackauckas-Claude:
## Summary

In DiffEqBase v7 (now living at `lib/DiffEqBase` under OrdinaryDiffEq.jl), the `apply_callback!` dispatch for `VectorContinuousCallback` no longer invokes `affect_neg!`. Code that uses the legacy idiom `VectorContinuousCallback(condition, nothing, affect_neg!, len)` (where `affect!` is `nothing` and the user supplies an `affect_neg!(integrator, idx)` for downcrossings) now silently fails: events are detected, but the affect is never run, and the user observes incorrect output (e.g. a bouncing-ball callback fails to bounce).

`SciMLBase.VectorContinuousCallback`'s constructor still accepts `affect_neg!` as a positional/keyword argument and the docstring still describes the same v6.x signature, so this is a silent dispatch/API mismatch rather than a documented breaking change.

(Filing here because SciML/DiffEqBase.jl is archived per its `README.md` migration note pointing to OrdinaryDiffEq.jl.)

## Reproducer (no Sundials)

```julia
using OrdinaryDiffEq, DiffEqBase
@show pkgversion(DiffEqBase)  # 7.3.0

f = (du, u, p, t) -> (du[1] = u[2]; du[2] = -9.81)

condition = (out, u, t, integrator) -> (out[1] = u[1])

affect_neg_called = Ref(0)
affect! = nothing
affect_neg! = function (integrator, idx)
    affect_neg_called[] += 1
    integrator.u[2] = -integrator.u[2]
end

cb = VectorContinuousCallback(condition, affect!, affect_neg!, 1)

prob = ODEProblem(f, [50.0, 0.0], (0.0, 15.0))
sol = solve(prob, Tsit5(); callback = cb)

@show affect_neg_called[]   # 0 -- never called!
@show sol.u[end]            # [-1053.6..., -147.1...]   ball fell through floor
```

Expected: ball bounces, height stays roughly in `[0, 50]`.
Actual: `affect_neg!` is never invoked; ball passes through `u[1] = 0` and accelerates to `-1053`.

## Root cause

`lib/DiffEqBase/src/callbacks.jl` `apply_callback!` for `VectorContinuousCallback`:

```julia
if callback isa VectorContinuousCallback
    if callback.affect! === nothing
        integrator.derivative_discontinuity = false
[...]

Issue #1528, opened 2021-11-14
ERROR: Callback handling failed. Please file an issue with code to reproduce.


contradict:
I found this example while playing with ModelingToolkit, it produced a system with the constant first state element. Removing that fixes this example.

```julia
using OrdinaryDiffEq

function callback_example()
    function f!(out, u, p, t)
        out[1] = 0
        out[2] = u[3]
        out[3] = -1.0*(u[2] - u[1])
    end
    u0 = [0, 0, 1.0]
    function cond!(out, u, t, i)
        out[1] = u[3]
        nothing
    end
    function affect!(int, idx)
        @show int
    end
    cb = VectorContinuousCallback(cond!, affect!, nothing, 1)

    u0 = [ 0.0, 0.0, 1.0 ]
    prob = ODEProblem(f!, u0, (0.0, 10.0); callback=cb)
    soln = solve(prob, Tsit5())
end
```

PR #3943, opened 2026-07-17
Migrate community callback test off removed VectorContinuousCallback 4-arg form

ChrisRackauckas-Claude:
> [!NOTE]

## Problem

`lib/DiffEqBase`'s `[Downstream2]` test group fails on master in `test/downstream/community_callback_tests.jl`:

```
MethodError: no method matching VectorContinuousCallback(::typeof(cond!), ::typeof(terminate_affect!), ::Nothing, ::Int64)
```

## Root cause

The test still used the SciMLBase v2 4-positional-arg form `VectorContinuousCallback(condition, affect!, affect_neg!, len)`. SciMLBase intentionally removed that form (and the `affect_neg!` keyword) in v3.10.0 — commit SciML/SciMLBase.jl@ba118addc7d7fffa529428bdae2c75168429e98b ("Remove `affect_neg!` from `VectorContinuousCallback` constructor") — because `VectorContinuousCallback` never honored `affect_neg!`: its `apply_callback!` dispatch always invokes `affect!(integrator, event_idx)` and ignored the field, so accepting it was a silent footgun (see SciML/OrdinaryDiffEq.jl#3613). The current SciMLBase docstring states "`VectorContinuousCallback` intentionally does not have an `affect_neg!` callback."

The downstream env only recently started resolving a SciMLBase new enough to drop the method (3.36.0 today), which is why this surfaced now (registry drift, first seen on #3692's CI, reproduces on unmodified master).

## Fix

Update the test to the supported 3-arg form `VectorContinuousCallback(cond!, terminate_affect!, 1)`, and update the expected termination time from `3pi/2` to `pi/2`:

- Under the removed v2 API, `affect_neg! = nothing` made event *detection* skip downcrossings, so the condition `u[3] = cos(t)` first triggered at the upcrossing `3pi/2`.
- Under the current intentional semantics, `affect!` fires on every crossing (the struct's `affect_neg!` field is always `affect!` and DiffEqBase v7's `apply_callback!` for `VectorContinuousCallback` never calls `affect_neg!`), so the solve terminates at the first zero of `cos(t)`, i.e. `pi/2`. Observed locally: `soln.t[end] = 1.5707927986` with the same `atol = 1e-4` as before.

The point of this #1528 regression test — callback handling must not error when one condition element is constantly zero — is preserved; only the API form and the crossing-direction expectation change, matching the documented upstream behavior.

Verified locally: `lib/DiffEqBase/test/downstream/community_callback_tests.jl` passes with SciMLBase 3.36.0 (output in the PR conversation).

This resolves the red `lib/DiffEqBase [Downstream2]` lane seen on #3692 (which merely surfaced the pre-existing master failure).

### DIFF
--- a/lib/DiffEqBase/test/downstream/community_callback_tests.jl
+++ b/lib/DiffEqBase/test/downstream/community_callback_tests.jl
@@ -227,12 +227,19 @@ end
 function terminate_affect!(int, events)
     return any(!iszero, events) && terminate!(int)
 end
-cb = VectorContinuousCallback(cond!, terminate_affect!, nothing, 1)
+# SciMLBase v3.10 removed the 4-arg `VectorContinuousCallback(cond, affect!,
+# affect_neg!, len)` form: `affect!` now fires on every crossing, and
+# `affect_neg! = nothing` no longer suppresses downcrossings. The condition is
+# u[3] = cos(t), so the solve now terminates at the first crossing pi/2 rather
+# than at the old expectation 3pi/2 (an upcrossing). The point of this #1528
+# regression test is that callback handling doesn't error on the constant-zero
+# condition element out[1].
+cb = VectorContinuousCallback(cond!, terminate_affect!, 1)
 
 u0 = [0.0, 0.0, 1.0]
 prob = ODEProblem(f!, u0, (0.0, 10.0); callback = cb)
 soln = solve(prob, Tsit5())
-@test soln.t[end] ≈ 4.712347213360699 atol = 1.0e-4
+@test soln.t[end] ≈ pi / 2 atol = 1.0e-4
 
 odefun = ODEFunction((u, p, t) -> [u[2], u[2] - p]; mass_matrix = [1 0; 0 0])
 callback = PresetTimeCallback(0.5, integ -> (integ.p = -integ.p))
--- a/lib/DiffEqBase/test/downstream/community_callback_tests.jl
+++ b/lib/DiffEqBase/test/downstream/community_callback_tests.jl
@@ -227,13 +227,6 @@ end
 function terminate_affect!(int, events)
     return any(!iszero, events) && terminate!(int)
 end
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Local verification (Julia 1.12, SciMLBase 3.36.0, monorepo libs dev'd at this branch):

Before: `MethodError: no method matching VectorContinuousCallback(::typeof(cond!), ::typeof(terminate_affect!), ::Nothing, ::Int64)`, and with only the signature migrated the old expectation fails as
```
Expression: ≈(soln.t[end], 4.712347213360699, atol = 0.0001)
 Evaluated: 1.5707927986456003 ≈ 4.712347213360699 (atol=0.0001)
```
confirming `affect!` now fires at the first (down)crossing `pi/2` under the intentional SciMLBase v3.10 semantics.

After this PR, the full previously-failing file passes:
```
Test Summary:            | Pass  Total     Time
Community Callback Tests |   16     16  7m03.3s
```

lib/DiffEqBase/test/downstream/community_callback_tests.jl
```diff
@@ -227,12 +227,19 @@ end
...
     return any(!iszero, events) && terminate!(int)
 end
-cb = VectorContinuousCallback(cond!, terminate_affect!, nothing, 1)
+# SciMLBase v3.10 removed the 4-arg `VectorContinuousCallback(cond, affect!,
+# affect_neg!, len)` form: `affect!` now fires on every crossing, and
+# `affect_neg! = nothing` no longer suppresses downcrossings. The condition is
+# u[3] = cos(t), so the solve now terminates at the first crossing pi/2 rather
+# than at the old expectation 3pi/2 (an upcrossing). The point of this #1528
+# regression test is that callback handling doesn't error on the constant-zero
+# condition element out[1].
```
ChrisRackauckas:
```suggestion

```

Issue #3950, opened 2026-07-18
SERK2 converges at order 1 on non-autonomous problems (stage-time formula)


ChrisRackauckas-Claude:
Found during a correctness audit of the stabilized RK family (see also #3947, #3948, #3949).

## Symptom

`SERK2` from OrdinaryDiffEqStabilizedRK converges at its design order 2 on autonomous problems (linear and nonlinear, verified in both the classical fixed-λ regime and the stiff λ∝1/dt regime), but drops to **order ~1.0 on a non-autonomous Prothero–Robinson problem and ~0.8 on pure quadrature `u' = cos(t)`**, out-of-place and in-place alike (measured on master after the #3947 fix, dts = 2⁻¹⁰..2⁻⁵, forced `eigen_est`).

Reproducer:

```julia
using OrdinaryDiffEqStabilizedRK, DiffEqDevTools, SciMLBase
fq(u, p, t) = fill(cos(t), 2)
qa(u0, p, t) = u0 .+ sin(t)
prob = ODEProblem(ODEFunction{false}(fq, analytic = qa), zeros(2), (0.0, 1.0))
eig = (i) -> i.eigen_est = 500 / abs(i.dt)
sim = test_convergence(1 .// 2 .^ (10:-1:5), prob, SERK2(eigen_est = eig))
sim.𝒪est[:l∞]  # ≈ 0.8, expected 2
```

For comparison, `RKC` and `ROCK2` give exactly 2.00 on the same problem.

## Analysis

The stage times used in `perform_step!` are `t + (j² + (i-1)·d²)·α·dt` with `α = 2.5/mdeg²`, `d = internal_deg = mdeg÷10`. These *are* internally consistent with the two-term recurrence inside each super-step block (applying the scheme to u' = 1 reproduces them), so the inconsistency is not in the recurrence itself — it appears to be between the stage times and the `Bᵢ` combination weights: the stage times only span `[t, t + 0.25·dt]` (the last block ends at `10·d²·α = 0.25`), so order 2 in the non-autonomous direction would require `Σ Bᵢcᵢ = 1/2` to hold with these c's, and measurement says it does not.

[...]

PR #3949, opened 2026-07-18
Fix RKG1/RKG2 stage times for non-autonomous problems

ChrisRackauckas-Claude:
> [!IMPORTANT]
> This PR should be ignored until reviewed by @ChrisRackauckas.

## What

Same defect and same construction as the RKL PR (#3948), for the Runge–Kutta–Gegenbauer methods: `RKG1`/`RKG2` evaluated every interior stage at time `t`, so RKG2 loses an order on any time-dependent problem (order 1.0 on quadrature `u' = cos(t)`, ~1.1–1.2 on Prothero–Robinson). Autonomous problems cannot detect this, which is all the existing tests used.

## Fix

Track stage times cⱼ through the same three-term recurrence the stage values satisfy (c₁ = μ̃₁, cⱼ = μⱼcⱼ₋₁ + νⱼcⱼ₋₂ + μ̃ⱼ (+ γ̃ⱼ for RKG2)) and evaluate stage j at `t + dt*cⱼ₋₁`. Each stage becomes internally consistent, and internal consistency + the verified autonomous order conditions imply the non-autonomous order via the t-augmented autonomous system. The recurrence is a few scalar ops per stage — no measurable cost.

## Measured convergence orders (before → after)

RKG2, ℓ∞ order over dts = 2⁻¹⁰..2⁻⁵, forced eigenvalue estimates: order 1.0–1.2 → **2.00** in all eight configurations (Prothero–Robinson + quadrature, oop + iip, fixed-λ + λ∝1/dt). RKG1 remains order 1.00 everywhere. Autonomous convergence unchanged.

## Tests

Adds an `RKG non-autonomous convergence` testset mirroring the RKL one. Ran locally: `ODEDIFFEQ_TEST_GROUP=Core Pkg.test()` — 213/213 pass. Runic applied.

Part of a stabilized-RK audit; sibling PRs: #3947 (SERK2 in-place fix), #3948 (RKL stage times).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015WU5V4PXt8SZ3BpjxdtM7d

PR #3948, opened 2026-07-18
Fix RKL1/RKL2 stage times for non-autonomous problems

ChrisRackauckas-Claude:
> [!IMPORTANT]
> This PR should be ignored until reviewed by @ChrisRackauckas.

## What

`RKL1`/`RKL2` evaluated every interior stage at time `t` — no stage-time offsets at all. On any time-dependent problem RKL2 loses an order: measured order ~1.1–1.2 on a Prothero–Robinson problem and exactly 1.0 on pure quadrature `u' = cos(t)` (it degenerates to forward Euler in `t`). Autonomous problems — the only thing the existing convergence tests use — cannot see this.

## Fix

Track stage times cⱼ through the same three-term recurrence the stage values satisfy (initialized with c₁ = μ̃₁, updated cⱼ = μⱼcⱼ₋₁ + νⱼcⱼ₋₂ + μ̃ⱼ (+ γ̃ⱼ for RKL2)), and evaluate stage j at `t + dt*cⱼ₋₁`. This makes each stage internally consistent and reproduces the closed forms from Meyer, Balsara & Aslam (JCP 2014): cⱼ = (j²+j)/(s²+s) for RKL1, cⱼ = (j²+j−2)/(s²+s−2) for RKL2. Internal consistency + the already-verified autonomous order conditions give the non-autonomous order via the t-augmented autonomous system.

## Measured convergence orders (before → after)

RKL2, ℓ∞ order over dts = 2⁻¹⁰..2⁻⁵, forced eigenvalue estimates:

| problem | fixed λ | λ ∝ 1/dt (stiff regime) |
|---|---|---|
| Prothero–Robinson oop | 1.20 → **2.00** | 1.13 → **2.00** |
| Prothero–Robinson iip | 1.20 → **2.00** | 1.13 → **2.00** |
| quadrature oop | 1.00 → **2.00** | 1.00 → **2.00** |
| quadrature iip | 1.00 → **2.00** | 1.00 → **2.00** |

RKL1 remains order 1.00 everywhere (it is first order by design; correct stage times still improve the constant). Autonomous convergence unchanged (2.00 / 1.00).

## Tests

Adds an `RKL non-autonomous convergence` testset (Prothero–Robinson + quadrature, oop + iip, both eigenvalue regimes). Ran locally: `ODEDIFFEQ_TEST_GROUP=Core Pkg.test()` — 213/213 pass. Runic applied.

[...]

PR #3947, opened 2026-07-18
Fix SERK2 in-place inner stage loop evaluating f at the wrong stage vector

ChrisRackauckas-Claude:
> [!IMPORTANT]
> This PR should be ignored until reviewed by @ChrisRackauckas.

## What

The in-place `SERK2` `perform_step!` evaluated `f` at `tmp` (which holds uᵢ₋₂) instead of `uᵢ₋₁` in its inner stage loop, and used a different buffer-shift guard (`j < mdeg`) than the out-of-place implementation (`j * i < mdeg`).

## Why it went undetected

The divergent code path only executes when `mdeg >= 20` (`internal_deg = mdeg ÷ 10 >= 2`), i.e. for genuinely stiff steps with dt·λ ≳ 300. The existing convergence tests use `prob_ode_linear`/`prob_ode_2Dlinear` with mild step sizes, which never reach that regime.

## Impact

In the stiff regime the in-place solver is unstable. On a forced heat equation (N = 32, dt·λ ≈ 217, fixed dt = 0.05):

| | ‖u − ref‖∞ |
|---|---|
| out-of-place | 2.5e-3 |
| in-place (before) | **3.8e83** |
| in-place (after) | 2.5e-3 (agrees with oop to 1.7e-13) |

## Tests

Adds a stiff fixed-dt in-place/out-of-place agreement regression testset covering all 13 stabilized RK methods (ROCK2/4, RKC, RKMC2, TSRKC2/3, SERK2, ESERK4/5, RKL1/2, RKG1/2). It fails on the previous code for SERK2 and passes with this fix.

Ran locally: `ODEDIFFEQ_TEST_GROUP=Core Pkg.test()` on the sublibrary — 210/210 pass. Runic applied to changed files.

## Process notes

Found during a correctness/performance audit of the stabilized RK family: code inspection flagged the oop/iip asymmetry, then a fixed-dt stiff comparison confirmed the blow-up, and the one-line fix restores roundoff-level agreement. Related (separate PRs): RKL/RKG non-autonomous stage-time fixes; SERK2 also has a separate non-autonomous order-reduction issue that needs the original paper's coefficients and is intentionally NOT addressed here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015WU5V4PXt8SZ3BpjxdtM7d

PR #3951, opened 2026-07-18
Rotate stage buffers instead of copying in ROCK2/ROCK4/RKC in-place loops

ChrisRackauckas-Claude:
> [!IMPORTANT]

## What

The in-place stage loops of `ROCK2`, `ROCK4`, and `RKC` shift their three-term recurrence window with two full-array copies per stage (`tmp .= uᵢ₋₁; uᵢ₋₁ .= u`). At 10–200 stages per step and with a memory-bound PDE RHS, those copies are ~40% of the per-stage memory traffic. This PR rotates the three buffer *bindings* per stage instead (a tuple swap, no data movement).

After the loop, role bindings are resolved by identity (`===`) so that:
- the finishing procedure reads the last stage value from wherever the rotation left it,
- the error-estimate scratch is never the same array as `integrator.u` (so `EEst` survives the final write),
- the result always lands in `integrator.u` — at most one array copy per step (RKC/ROCK4, only when rotation parity leaves the last stage in a scratch buffer), vs. `2·(mdeg−2)` copies before.

All updates remain pure elementwise ops, so the intentional aliasing cases are safe. Cache array identities never change, so `maxeig!`'s use of `cache.tmp`/`cache.k` as scratch is unaffected.

## Measured (65536-point 1D heat equation, provided `eigen_est`, adaptive, this machine)

| method | master | this PR | speedup |
|---|---|---|---|
| ROCK2 | 528 ms | 336 ms | **1.57×** |
| ROCK4 | 1011 ms | 637 ms | **1.59×** |
| RKC | 111 ms | 75 ms | **1.47×** |

`nf` and step sequences are identical.

## Verification

- In-place vs out-of-place final-state agreement is unchanged from master at multiple fixed dts (including the `mdeg < 2` path) and with adaptive stepping (identical accepted-step counts, differences at 1e-13–1e-16 except ROCK4's pre-existing 1e-9 muladd-ordering difference which is byte-identical to master's).
- Stepping remains allocation-free (4 allocations per `solve!`, same as master).
- Local runs: `ODEDIFFEQ_TEST_GROUP=Core` 197/197 pass; `ODEDIFFEQ_TEST_GROUP=QA` (AllocCheck, type agnosticism, JET 27/27, Aqua 19/19) all pass. Runic applied.

The same transformation applies to RKL/RKG/TSRKC/RKMC2/SERK2 loops; left for a follow-up to keep this reviewable.

Part of a stabilized-RK audit; siblings: #3947, #3948, #3949, issue #3950.

https://claude.ai/code/session_015WU5V4PXt8SZ3BpjxdtM7d

### DIFF
--- a/lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
+++ b/lib/OrdinaryDiffEqStabilizedRK/src/rkc_perform_step.jl
@@ -110,51 +110,49 @@ end
     tᵢ₋₁ = t + dt * recf[ccache.start]
     tᵢ₋₂ = t + dt * recf[ccache.start]
     tᵢ₋₃ = t
-    @.. broadcast = false tmp = uprev
-    @.. broadcast = false uᵢ₋₁ = uprev + (dt * recf[ccache.start]) * fsalfirst
-    ccache.mdeg < 2 && (@.. broadcast = false u = uᵢ₋₁)
+    # The three stage buffers are rotated by rebinding instead of copied each
+    # stage; role bindings are resolved afterwards so the result lands in `u`.
+    uold, ucur, unext = tmp, uᵢ₋₁, u
+    @.. broadcast = false uold = uprev
+    @.. broadcast = false ucur = uprev + (dt * recf[ccache.start]) * fsalfirst
     # for the second to the ms[ccache.mdeg] th stages
     for i in 2:(ccache.mdeg)
         μ, κ = recf[ccache.start + (i - 2) * 2 + 1], recf[ccache.start + (i - 2) * 2 + 2]
         ν = -1 - κ
-        f(k, uᵢ₋₁, p, tᵢ₋₁)
+        f(k, ucur, p, tᵢ₋₁)
         OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
         tᵢ₋₁ = dt * μ - ν * tᵢ₋₂ - κ * tᵢ₋₃
-        @.. broadcast = false u = (dt * μ) * k - ν * uᵢ₋₁ - κ * tmp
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
**CI triage** (2026-07-18): the failing checks here are pre-existing on `master` (commit 4e1a434, same base as this PR) and unrelated to this diff:

- `downgrade-sublibraries` (9 sublibs, none touched by this PR) — also failing on master's own run.
- `sublibrary-ci / lib/OrdinaryDiffEqStabilizedIRK [QA]` — ExplicitImports `all_explicit_imports_are_public` errors for `build_nlsolver`/`nlsolve!`/`du_alias_or_new`/`markfirststage!`/`dolinsolve`/`update_W!` in StabilizedIRK (a sublib this PR does not modify).

A separate investigation of the master failures is underway and will be filed as its own issue/PR. All checks that exercise this PR's code (`lib/OrdinaryDiffEqStabilizedRK` Core + QA) pass.

Additionally: `tests / Integrators_II` failed on `resize! with SplitODEProblem` (test/Integrators_II/resize_tests.jl:235) — that test uses `Tsit5()` on a `SplitODEProblem` and never calls any StabilizedRK code, and the surrounding test file is RNG-dependent, so this looks flaky rather than caused by the buffer rotation. Pushed an empty commit to retrigger CI; will report the rerun result.

ChrisRackauckas-Claude:
Follow-up on the CI triage: after retriggering CI, `tests / Integrators_II (julia 1)` **passes** on the identical diff — confirming the earlier `resize! with SplitODEProblem` failure was flaky (it exercises `Tsit5` on a `SplitODEProblem`, no StabilizedRK code involved). The only remaining failures are the master-inherited ones noted above.

PR #3953, opened 2026-07-18
GlobalDiffEq: add adjoint-based global error control as a SciMLSensitivity extension

ChrisRackauckas-Claude:
> **Ignore until reviewed by @ChrisRackauckas.** Draft — not ready for merge.

Adjoint-based global error estimation/control for `GlobalDiffEq`, reworked to
address the review. Updated per the six review threads:

## Review responses

- **Use SciMLSensitivity's own machinery, not a hand-rolled adjoint + QuadGK**
  (ext:5–9, ext:61). The estimator no longer builds `ODEAdjointProblem` and
  integrates `⟨λ, P'−f⟩` by hand. The dual-weighted residual `∫ λᵀ(f(P)−P') dt`
  is now computed by `SciMLSensitivity.adjoint_sensitivities` itself: the defect
  `r(t)=f(P)−P'` is introduced as a scalar forcing `ε` (`u' = f + ε·r`), so
  `dG/dε = ∫ λᵀ r dt` is returned directly as the parameter gradient — no manual
  quadrature. **QuadGK and `_FixedParameter`/SciMLStructures are dropped
  entirely** (net −1 dependency plus the removed weakdep).
- **Default adjoint** (ext:5–9). `sensealg` now defaults to SciMLSensitivity's
  own default adjoint family (`InterpolatingAdjoint`), not a hardcoded
  `QuadratureAdjoint`; any adjoint sensealg may be passed.
- **Surface the estimate in the solution** (adjoint.jl:259). `SciMLBase.has_global_error(::GlobalAdjoint) = true`, and a `gtol` solve returns a
  solution whose `global_error[end]` holds the endpoint estimate.
- **Callbacks no longer forbidden** (adjoint.jl:326). The prohibition is removed;
  a benign callback estimates without error (covered by a test).

## Adjoint VJP caveat (why the default is `autojacvec = false`)

The estimator differentiates a right-hand side that reads the forward solution's
dense interpolant `sol(t)`. Differentiating that through a **vector-Jacobian-product
backend is broken**, verified against finite differences (forcing `= ref(t)`,
`abstol = reltol = 1e-10`, flat across `tol` 1e-5…1e-11):

| `autojacvec` | Quad / Gauss / GaussKronrod | InterpolatingAdjoint |
|---|---|---|
| `ReverseDiffVJP(false)` | 0.9502 (~5% wrong) | 1.0 |
| `ReverseDiffVJP(true)` (compiled) | 0.9502 (~5% wrong) | **0.9502 (~5% wrong)** |
| `autojacvec = true` (auto) | −0.0 / unsupported | 1.0 |
| `EnzymeVJP` | crashes (`LLVM ERROR`, 1.11 & 1.12) | crashes |
| `false` (ForwardDiff Jacobian) | **1.0** | **1.0** |

The adjoint `λ` itself is correct (integrating Quadrature's own `λ` by hand gives
`/FD = 1.0000000003`); the error is entirely in the VJP evaluation of `∂f/∂p`
through the interpolant. The **compiled ReverseDiff tape (`ReverseDiffVJP(true)`)
is wrong even for `InterpolatingAdjoint`** — it freezes the interpolant value at
trace time — and Enzyme fails to compile `ode_interpolation` at all. The only
setting correct across every adjoint method *and* VJP choice is the **ForwardDiff
Jacobian (`autojacvec = false`)**, so it is the default here; it also sidesteps
the Enzyme-on-1.12 crash. A VJP would be cheaper for large systems, but every
ReverseDiff-tape path is wrong in at least the compiled case, so it is not safe to
default to one. Reported upstream:
https://github.com/SciML/SciMLSensitivity.jl/issues/1649 (and the Enzyme compile
failure: https://github.com/EnzymeAD/Enzyme.jl/issues/3565).

## Not yet done (tracked follow-ups)

- **Every-step mode** (issue thread): loss as the sum over saved steps, filling
  `global_error` along the whole trajectory. This PR ships the endpoint mode
  (only `global_error[end]` is populated).
- **Adjoint-`tstops` control option** (adjoint.jl:372): the alternative that
  takes step points from the adjoint and re-solves on that grid.

## Verification (local, Julia 1.12)

`GROUP=Core` (includes the adjoint suite):

```
Basic functionality                  |  1   1
Algorithm traits forwarding          |  3   3
Adjoint error estimation and control | 18  18
GLEE solvers                         | 91  91
BigFloat support                     |  2   2
     Testing GlobalDiffEq tests passed
```

`GROUP=QA` (Aqua + ExplicitImports + JET): `Testing GlobalDiffEq tests passed`.
Runic: clean.

The ε-forcing estimate was validated against the manual DWR projection and the
exact error on a linear problem: `−dp/error` = 0.99999952, 0.99999631, 0.99994367
at `tol` = 1e-3, 1e-4, 1e-5.

## Not verified

Downstream packages, GPU paths, allowed-to-fail CI jobs.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: claude-opus-4-8)
https://claude.ai/code/session_01MfUh8tgp3iBmMv9nkBqoyY


lib/GlobalDiffEq/ext/GlobalDiffEqSciMLSensitivityExt.jl
```diff
@@ -0,0 +1,65 @@
+module GlobalDiffEqSciMLSensitivityExt
+
+import GlobalDiffEq, LinearAlgebra, QuadGK, SciMLBase, SciMLSensitivity
+
+function GlobalDiffEq._default_quadrature_sensealg()
+    return SciMLSensitivity.QuadratureAdjoint(autojacvec = true)
+end
+
+GlobalDiffEq._is_quadrature_adjoint(::SciMLSensitivity.QuadratureAdjoint) = true
```
ChrisRackauckas, line 9: We should use the default solver that SciMLSensitivity gives.

lib/GlobalDiffEq/ext/GlobalDiffEqSciMLSensitivityExt.jl
```diff
@@ -0,0 +1,65 @@
...
+    projection = zero(eltype(prob.u0))
+    for i in 1:(length(sol.t) - 1)
+        left = sol.t[i]
+        right = sol.t[i + 1]
+        left == right && continue
+        interval_projection, _ = QuadGK.quadgk(
+            integrand, left, right; atol = abstol, rtol = reltol
+        )
+        projection += interval_projection
+    end
```
ChrisRackauckas, line 61: This should instead be computed via the adjoint problem's definition?

ChrisRackauckas:
Just doing the endpoint is one mode, but a more useful mode is probably to do this on every step. That would be done by instead changing the loss to be computed on the sum of the steps, and modifying the corresponding adjoint problem similarly.

ChrisRackauckas-Claude:
Noted — an every-step mode fits cleanly once the planned SciMLBase global-error interface (a `has_global_error` trait + a `sol.global_error` array-of-arrays) is in place: make the adjoint loss the sum over the saved steps and build the matching adjoint problem, so `sol.global_error` is filled along the trajectory instead of only at the endpoint. I've written up the interface + sequencing plan on #3956. I'll do the every-step adjoint as a follow-up on top of that interface; this PR lands the endpoint estimator/control. (Also: this PR is now unstacked — it applies directly to master with just the adjoint commit.)

lib/GlobalDiffEq/ext/GlobalDiffEqSciMLSensitivityExt.jl
ChrisRackauckas: yeah this is simulating a discrete cost function at the solution times, it should just use that functionality in the adjoint system rather than trying to recreate it. Plus the default adjoint, GaussAdjoint, will do this more efficiently.

lib/GlobalDiffEq/src/adjoint.jl
```diff
@@ -0,0 +1,383 @@
...
+        _defect_projection(
+            sol, adjoint_sol;
+            abstol = quadrature_abstol, reltol = quadrature_reltol
+        )
+    end
+    T = eltype(sol.prob.u0)
+    sample_count = length(directions)
+    factor = _sphere_projection_expectation(sample_count, T) /
+        _sphere_projection_expectation(length(sol.prob.u0), T)
+    return factor * LinearAlgebra.norm(projections)
```
ChrisRackauckas, line 204: it should put this in the new global error spot in the solution

lib/GlobalDiffEq/src/adjoint.jl
```diff
@@ -0,0 +1,383 @@
...
+        abstol = something(alg.gtol, 1.0e-6),
+        reltol = something(alg.gtol, 1.0e-3),
+        kwargs...
+    )
+    alg.gtol === nothing &&
+        throw(ArgumentError("GlobalAdjoint requires a positive `gtol` constructor keyword"))
+    _require_adjoint_ext()
+    _validate_tolerances(abstol, reltol, "local")
+    haskey(kwargs, :callback) &&
+        throw(ArgumentError("GlobalAdjoint does not currently support callbacks"))
```
ChrisRackauckas: I don't see why this would be a limitation when the adjoint system does.

lib/GlobalDiffEq/src/adjoint.jl
```diff
@@ -0,0 +1,383 @@
...
+        if last_estimate <= gtol
+            return SciMLBase.solve(
+                prob, alg.alg, args...;
+                abstol = local_abstol, reltol = local_reltol, kwargs...
+            )
+        end
+
+        tolerance_scale = min(0.5, options.safety * gtol / last_estimate)
+        local_abstol *= tolerance_scale
+        local_reltol *= tolerance_scale
```
ChrisRackauckas, line 300: This is one of the approaches, but the other gives you `dt`s from the adjoint method and then you just use `adaptive=false, tstops = dts` for that. That version should be implemented as an option as well.

PR #3954, opened 2026-07-18
GlobalDiffEq: add GLEE23/GLEE24/GLEE35 and MM5GEE global-error-estimating solvers

ChrisRackauckas-Claude:
> [!IMPORTANT]

**Unstacked and independent**: this PR now applies directly to `master` and contains only the RK global-error-estimating solvers (no dependency on the other GlobalDiffEq PRs).

## Summary

Adds explicit general linear methods with built-in global error estimation to the `lib/GlobalDiffEq` sublibrary as proper `OrdinaryDiffEqCore` solver algorithms:

- **`GLEE23`, `GLEE24`, `GLEE35`** — Constantinescu, *Estimating Global Errors in Time Stepping*, SIAM J. Numer. Anal. 54(6), 2016 ([arXiv:1503.05166](https://arxiv.org/abs/1503.05166); PETSc's `TSGLEE23/24/35`). Addresses SciML/GlobalDiffEq.jl#6 and SciML/GlobalDiffEq.jl#22.
- **`MM5GEE`** — Makazaga & Murua, BIT 43 (2003), the DOPRI5-based order-5 scheme with cheap global error estimation. **Folded in from the former PR #3955 per review** ("just merge this into the other PR of the RK global error estimating methods").

The methods propagate the partitioned state `(y, ε)` where `ε` is an asymptotically correct estimate of the global error; `solve`/`init` on a plain `ODEProblem` transparently extend it to `ArrayPartition` state, and `global_error_estimate(sol)` extracts the estimate trajectory. The per-step `ε` increment is an asymptotically correct local error estimate that drives the standard step-size controllers. Tableaus transcribed from the papers (GLEE from the LaTeX source, cross-checked against PETSc `glee.c`; MM5GEE's Table 4.1 verified with exact rational arithmetic against the order + independency conditions). The paper's defective method A4 is deliberately not implemented.

> Note on the review direction: @ChrisRackauckas noted these "should be a solver library" and that global error should become a SciMLBase interface (a `has_global_error` trait, a `sol.global_error` array-of-arrays populated by splitting the `ArrayPartition` in `savevalues!`). That is planned as follow-up (see the thread on #3956); this PR lands the methods in the current sublibrary form and will be migrated onto that interface once it exists.

- `ODEDIFFEQ_TEST_GROUP=Core`: **pass** — Basic 1/1, Algorithm traits 3/3, GLEE solvers 37/37 (incl. MM5GEE), BigFloat 2/2.
- `ODEDIFFEQ_TEST_GROUP=QA`: **pass** — 23 passed, 1 pre-existing declared broken (`no_implicit_imports` from the deliberate `@reexport`).
- Convergence on the unstable Prince42 problem, in-place and out-of-place: solution orders 2/2/3/5, estimate-accuracy orders 3/3/4/6, est/true-error ratio ≈ 1.0; MM5GEE at 9 f-evals/step.

### DIFF
--- a/docs/src/globalerrorcontrol/GlobalDiffEq.md
+++ b/docs/src/globalerrorcontrol/GlobalDiffEq.md
@@ -14,9 +14,42 @@ To use these methods:
 using GlobalDiffEq
 ```
 
+The global-error-estimating solvers [`GLEE24`](@ref), [`GLEE35`](@ref)
+(Constantinescu 2016), and the Dormand-Prince-based [`MM5GEE`](@ref) (Makazaga
+and Murua 2003) carry a running, asymptotically correct estimate of the
+solution's global error at every time point, at the cost of a few extra stages
+per step. Solving with them produces a solution whose states are
+`ArrayPartition`s (`sol.u[i].x[1]` the solution, `sol.u[i].x[2]` the global
+error estimate); [`global_error_estimate`](@ref) extracts the estimates.
+
 [`GlobalRichardson`](@ref) wraps any fixed-step method in global Richardson
 extrapolation over whole solves, interpreting `abstol` and `reltol` as global
 tolerances. It is the most robust and most expensive option.
+
+For example, solving while tracking the global error along the trajectory:
+
+```julia
+using GlobalDiffEq
+
+function lorenz!(du, u, p, t)
+    du[1] = 10.0(u[2] - u[1])
+    du[2] = u[1] * (28.0 - u[3]) - u[2]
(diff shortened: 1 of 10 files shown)


ChrisRackauckas-Claude:
Follow-up #4267 migrates these methods onto the SciMLBase global-error interface (SciML/SciMLBase.jl#1531) so the user gets an ordinary `sol.u` + `sol.global_error` instead of the internal `ArrayPartition`. It's stacked on this PR and depends on the SciMLBase release; this PR stays mergeable against current SciMLBase.

PR #3956, opened 2026-07-18
GlobalDiffEq: add GlobalErrorTransport global error estimation and control

ChrisRackauckas-Claude:
> [!IMPORTANT]

> **Rebased onto current master** (which now includes the merged GLEE methods, #3954) and bumped to v1.4.0; verified against released SciMLBase (Core: Companion 14/14 + GLEE 37/37; QA 23/1).

**Independent PR** — a single commit on current `master` (GlobalDiffEq v1.4.0), with no dependency on the other GlobalDiffEq PRs; reviewable and mergeable on its own. Contains only `GlobalErrorTransport`. (It adds `src/companion.jl`, byte-identical to the copy in #3957, so the two still merge cleanly in any order.)

## Summary

Adds `GlobalErrorTransport` — global error estimation and `gtol`-based control by the linearized error-transport (first variational) equation, addressing SciML/GlobalDiffEq.jl#5 (Shampine 1986 / Berzins 1988 / Lang–Verwer 2007). After a dense forward solve with interpolant `P(t)`, the companion linear ODE `ε' = J(P(t))ε + d(t)` with the dense-output defect `d(t) = f(P(t)) − P'(t)` is integrated from `ε(t₀)=0`; `‖ε(T)‖₂` estimates the endpoint global error. The Jacobian is applied matrix-free via DifferentiationInterface (any ADTypes backend; default `AutoForwardDiff()`). `global_error_estimate(prob, alg)` is the standalone estimator.

## Review direction (this is the architectural thread)

Recording the plan from @ChrisRackauckas's comment so it's on record, split out from this PR:

1. **A global-error interface in SciMLBase** — a `has_global_error(alg)::Bool` trait (default `false`); a `sol.global_error` field (array-of-arrays matching `u`, `nothing` when the method doesn't compute it); uniform accessors. (SciMLBase PR; public API + docs together.)
2. **GLEE as its own solver library, feeding the interface** — move the RK GEE methods (#3954) into a dedicated OrdinaryDiffEq solver sublibrary; `savevalues!` splits the partitioned `(y, ε)` `ArrayPartition` so the user gets a normal `sol.u` and the `ε` register lands in `sol.global_error`; `has_global_error` returns `true`. Documented as an OrdinaryDiffEq.jl feature.
3. **Wrappers on the same interface** — `GlobalErrorTransport` (this PR), `GlobalDefectCorrection` (#3957), and `GlobalAdjoint` (#3953) compute the estimate differently but populate the same `sol.global_error`.
4. **Adjoint every-step mode** (#3953) — loss over the sum of steps + matching adjoint problem, filling `sol.global_error` along the trajectory.

Sequencing: SciMLBase interface → GLEE solver library → migrate wrappers → adjoint every-step. I'll open these as separate PRs and link them here. This PR will be revised to populate `sol.global_error` once (1) lands.

- `ODEDIFFEQ_TEST_GROUP=Core`: **pass** — Basic 1/1, Algorithm traits 3/3, Companion error estimators 14/14, BigFloat 2/2.
- `ODEDIFFEQ_TEST_GROUP=QA`: **pass** — 23 passed, 1 pre-existing declared broken.
- On Lotka–Volterra vs a 1e-13 reference: estimate/true-error ratios within 10% at tolerances 1e-4 and 1e-6, in-place and out-of-place; `gtol = 1e-7` control delivered a true endpoint error of ~1e-8.

### DIFF
--- a/docs/src/globalerrorcontrol/GlobalDiffEq.md
+++ b/docs/src/globalerrorcontrol/GlobalDiffEq.md
@@ -25,6 +25,11 @@ estimated global error of `sol.u[i]` at `sol.t[i]`, and
 [`SciMLBase.has_global_error`](@ref) is `true` for these algorithms.
 [`global_error_estimate`](@ref) returns `sol.global_error`.
 
+[`GlobalErrorTransport`](@ref) wraps any adaptive solver with global error
+estimation and `gtol`-based control by integrating the linearized
+error-transport equation driven by the dense-output defect;
+[`global_error_estimate`](@ref)`(prob, alg)` exposes the standalone estimator.
+
 [`GlobalRichardson`](@ref) wraps any fixed-step method in global Richardson
 extrapolation over whole solves, interpreting `abstol` and `reltol` as global
 tolerances. It is the most robust and most expensive option.
@@ -59,4 +64,8 @@ global_error_estimate
 
 ```@docs
 GlobalRichardson
+GlobalErrorTransport
+GlobalErrorMode
+InterpolatingMode
+SimultaneousMode
 ```
--- a/lib/GlobalDiffEq/Project.toml
+++ b/lib/GlobalDiffEq/Project.toml
@@ -1,12 +1,15 @@
 name = "GlobalDiffEq"
 uuid = "1d72d19b-84cc-4cb7-a099-7cbdb9ccc67c"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
-version = "1.4.1"
+version = "1.5.0"
 
 [deps]
+ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
 Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
+DifferentiationInterface = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 FastBroadcast = "7034ab61-46d4-4ed7-9d0f-46aef9175898"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MuladdMacro = "46d2c3a1-f734-5fdb-9937-b9b9aeba4221"
 OrdinaryDiffEqCore = "bbf590c4-e513-4bbe-9b18-05decba2e5d8"
@@ -23,9 +26,12 @@ OrdinaryDiffEqCore = {path = "../OrdinaryDiffEqCore"}
 OrdinaryDiffEqTsit5 = {path = "../OrdinaryDiffEqTsit5"}
 
 [compat]
+ADTypes = "1"
 Accessors = "0.1.42"
 DiffEqBase = "7"
+DifferentiationInterface = "0.6, 0.7"
 FastBroadcast = "1.3"
+ForwardDiff = "0.10, 1"
 LinearAlgebra = "1"
 MuladdMacro = "0.2.4"
 OrdinaryDiffEqCore = "4"
--- a/lib/GlobalDiffEq/src/GlobalDiffEq.jl
+++ b/lib/GlobalDiffEq/src/GlobalDiffEq.jl
@@ -3,8 +3,9 @@ module GlobalDiffEq
 using Reexport: @reexport
 @reexport using DiffEqBase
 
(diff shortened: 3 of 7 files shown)


ChrisRackauckas:
The GLEE methods should be a solver library. 

And we should make the global error actually an interface in SciMLBase: a trait for whether a method computes the global error, a global error field (which is nothing when a method doesn't compute it), and then it should be a matching array of arrays. The savevalues on this type of method should properly split the ArrayPartition so that the user gets back a normal solution in `u` and put the `u.x[2]` into the global error save. Then this should be documented as an OrdinaryDiffEq.jl thing.

The adjoint method could use this same interface just computed differently?

ChrisRackauckas-Claude:
Thanks — this reframes the direction. Summarizing the plan so it's on record, split from this PR:

**1. A global-error interface in SciMLBase.** Add to SciMLBase:
- a solver trait, e.g. `has_global_error(alg)::Bool` (default `false`), that a method sets `true` when it computes a global error estimate;
- a `global_error` field on `ODESolution` (an array-of-arrays matching `u`, or `nothing` when `has_global_error(alg)` is false);
- accessor(s) so downstream code reads it uniformly (`sol.global_error`, and a per-time getter).

This is a SciMLBase PR (public API + docs together, minor bump).

**2. GLEE as its own solver library, feeding the interface.** Move `GLEE23/24/35`/`MM5GEE` out of GlobalDiffEq into a dedicated OrdinaryDiffEq solver sublibrary. The partitioned `(y, ε)` stepping stays internal; `savevalues!` splits the `ArrayPartition` so the user gets a normal solution in `sol.u` and the `u.x[2]` register lands in `sol.global_error`. `has_global_error` returns `true` for these algs. Documented as an OrdinaryDiffEq.jl feature (not a user-visible `ArrayPartition`).

**3. GlobalErrorTransport / GlobalDefectCorrection / GlobalAdjoint on the same interface.** These wrappers compute the estimate differently (companion ODE / adjoint) but populate the same `sol.global_error`, so the user-facing surface is uniform regardless of method.

**4. Adjoint every-step mode (PR #3953 comment).** Extend the adjoint estimator from an endpoint functional to a per-step one by making the loss the sum over saved steps and building the matching adjoint problem, so `sol.global_error` is filled along the trajectory rather than only at the endpoint. Fits naturally once the interface in (1) exists.

Proposed sequencing: land the SciMLBase interface first (1), then the GLEE solver library on top of it (2), then migrate the wrappers (3), then the adjoint every-step mode (4). I'll open these as separate PRs and link them here. This PR (GlobalErrorTransport) will be revised to populate `sol.global_error` once (1) lands.

ChrisRackauckas-Claude:
Step 1 of the plan is up: **SciML/SciMLBase.jl#1531** adds the global-error interface — a `has_global_error(alg)` trait (default `false`) and a `global_error` field on `ODESolution` (default `nothing`, an array matching `u` when present), threaded through `build_solution`/`setproperties`/`solution_slice` with `nothing` defaults so nothing existing changes. Verified against the SciMLBase `Core` suite.

Next: wire the GLEE methods (#3954) to set `has_global_error = true` and split their internal `(y, ε)` `ArrayPartition` in `savevalues!` so the user gets a normal `sol.u` and the `ε` register lands in `sol.global_error`; then migrate the transport/correction/adjoint wrappers onto the same field; then the adjoint every-step mode.

ChrisRackauckas-Claude:
Step 2 is up: **#4267** migrates the GLEE/MM5GEE solvers onto the interface. `has_global_error` is `true`, the extended `(y, ε)` solve is split in `__solve`, and a projecting interpolation reuses the high-order dense output so the user gets an ordinary solution (`sol.u`, `sol(t)`, `sol(t, Val{1})`, in-place all return the solution at full order) with the estimate in `sol.global_error`. Depends on SciML/SciMLBase.jl#1531 (SciMLBase ≥ 3.48); stacked on #3954; verified locally (GLEE suite 91/91).

Remaining: migrate the `GlobalErrorTransport`/`GlobalDefectCorrection`/`GlobalAdjoint` wrappers onto `sol.global_error`, and the adjoint every-step mode. (The "GLEE as its own solver sublibrary" reorg is a mechanical follow-up once the sublibrary settles — the interface behavior it needs is in #4267.)

ChrisRackauckas-Claude:
Superseded by #4492, which unifies the two defect-companion estimators (`GlobalErrorTransport` and `GlobalDefectCorrection`) into a single `GlobalErrorEstimation(alg; equation, mode)` parameterized by orthogonal `GlobalErrorEquation` (DefectCorrection | ErrorTransport) and `GlobalErrorMode` (InterpolatingMode | SimultaneousMode) enums. Closing in favor of that.

PR #3957, opened 2026-07-18
GlobalDiffEq: add GlobalDefectCorrection global error estimation and control

ChrisRackauckas-Claude:
> [!IMPORTANT]

> **Rebased onto current master** (which now includes the merged GLEE methods, #3954) and bumped to v1.4.0; verified against released SciMLBase (Core: Companion 14/14 + GLEE 37/37; QA 23/1).

**Independent PR** — a single commit on current `master` (GlobalDiffEq v1.4.0), with no dependency on the other GlobalDiffEq PRs; reviewable and mergeable on its own. Contains only `GlobalDefectCorrection`. (It adds `src/companion.jl`, byte-identical to the copy in #3956, so the two still merge cleanly in any order.)

## Summary

Adds `GlobalDefectCorrection` — global error estimation and `gtol`-based control by solving for the correction (defect correction), addressing the SciML/GlobalDiffEq.jl#3 method family (Zadunaisky 1976; Dormand–Duckers–Prince 1984; Dormand–Lockyer–McGorrigan–Prince 1989; adaptive-stepping validity by Calvo–Higham–Montijano–Rández 1996). After a dense forward solve with interpolant `P(t)`, the nonlinear companion ODE `ε' = f(P(t)+ε) − P'(t)` is integrated from `ε(t₀)=0`; `‖ε(T)‖₂` estimates the endpoint global error. Unlike `GlobalErrorTransport` this needs **no Jacobian**, only `f` evaluations on perturbed arguments — the generic coefficient-free realization of the Dormand–Prince triple idea for any dense-output solver. `global_error_estimate(prob, alg)` is the standalone estimator.

Global-error interface / solver-library direction: see the architectural plan on #3956. This PR will be revised to populate `sol.global_error` once that SciMLBase interface lands.

- `ODEDIFFEQ_TEST_GROUP=Core`: **pass** — Basic 1/1, Algorithm traits 3/3, Companion error estimators 14/14, BigFloat 2/2.
- `ODEDIFFEQ_TEST_GROUP=QA`: **pass** — 23 passed, 1 pre-existing declared broken.
- On Lotka–Volterra vs a 1e-13 reference: estimate/true-error ratios within 10% at tolerances 1e-4 and 1e-6, in-place and out-of-place; `gtol = 1e-7` control delivered a true endpoint error of ~1e-8.

### DIFF
--- a/docs/src/globalerrorcontrol/GlobalDiffEq.md
+++ b/docs/src/globalerrorcontrol/GlobalDiffEq.md
@@ -25,6 +25,11 @@ estimated global error of `sol.u[i]` at `sol.t[i]`, and
 [`SciMLBase.has_global_error`](@ref) is `true` for these algorithms.
 [`global_error_estimate`](@ref) returns `sol.global_error`.
 
+[`GlobalDefectCorrection`](@ref) wraps any adaptive solver with global error
+estimation and `gtol`-based control by solving for the correction of the dense
+numerical solution (a nonlinear companion ODE requiring no Jacobian);
+[`global_error_estimate`](@ref)`(prob, alg)` exposes the standalone estimator.
+
 [`GlobalRichardson`](@ref) wraps any fixed-step method in global Richardson
 extrapolation over whole solves, interpreting `abstol` and `reltol` as global
 tolerances. It is the most robust and most expensive option.
@@ -59,4 +64,8 @@ global_error_estimate
 
 ```@docs
 GlobalRichardson
+GlobalDefectCorrection
+GlobalErrorMode
+InterpolatingMode
+SimultaneousMode
 ```
--- a/lib/GlobalDiffEq/Project.toml
+++ b/lib/GlobalDiffEq/Project.toml
@@ -1,7 +1,7 @@
 name = "GlobalDiffEq"
 uuid = "1d72d19b-84cc-4cb7-a099-7cbdb9ccc67c"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com> and contributors"]
-version = "1.4.1"
+version = "1.5.0"
 
 [deps]
 Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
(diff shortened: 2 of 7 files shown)


ChrisRackauckas-Claude:
On the estimation strategy (re: solve `u' = f`, then use `sol(t)` in the companion equation):

That's exactly what this does — a two-pass, **interpolating** scheme. The forward solve is run with `dense = true, save_everystep = true` (forced), the whole dense solution is kept, and the companion error ODE's RHS calls back into it — `u = sol(t)`, `du = sol(t, Val{1})` — to build the dense-output defect `d(t) = f(P(t)) − P'(t)`. It's fully solver-agnostic and simple, but it pins the entire dense solution in memory (`O(n_steps · state · k_stages)`) and runs a second full solve whose every RHS eval does two interpolations plus an `f`.

Two things addressed here:

1. **It now names the strategy and guards its assumption.** Both wrappers gain a `mode` field (`:interpolating`, the only implemented value; `:simultaneous` validated-against with a pointer to the follow-up). And `_validate_estimation_solution` now rejects a forward solve whose interpolation is the trivial `Linear`/`Constant` fallback (or non-dense) — previously it only forced `dense`/`save_everystep` but never checked the solver actually *supplies* a dense first-derivative interpolation, so a non-dense solver would have silently differentiated a low-order interpolant into a wrong defect. Now it fails loudly. Tests cover both (mode rejection + the dense guard).

2. **The "just append and solve it simultaneously" idea is the right efficient path — tracked in #4491.** The catch is that the defect needs the *interpolant's* derivative `P'`, not `f(u)`: in a simultaneous solve `P' ≠ f(u)` (their difference *is* the defect), so naively appending with `P' = f(u)` gives `d ≡ 0`. A simultaneous implementation has to pull `P'` from each step's dense-output polynomial and inject the local defect into a co-evolved `ε` register (the Lang–Verwer / Berzins inline form) — which drops the extra memory to `O(1)` and the second solve to a per-step update, but is no longer black-box (needs per-step stage/defect access). That's the same machinery the GLEE methods (#3954/#4267) already use internally, so it lands as `mode = :simultaneous` on the `savevalues!`/`sol.global_error` interface direction rather than in the black-box wrapper. See #4491.

ChrisRackauckas-Claude:
Superseded by #4492, which unifies the two defect-companion estimators (`GlobalErrorTransport` and `GlobalDefectCorrection`) into a single `GlobalErrorEstimation(alg; equation, mode)` parameterized by orthogonal `GlobalErrorEquation` (DefectCorrection | ErrorTransport) and `GlobalErrorMode` (InterpolatingMode | SimultaneousMode) enums. Closing in favor of that.

Issue #3859, opened 2026-07-09
NonlinearSolveAlg: NonlinearSolve polyalg inner solvers crash (FieldError: PolyAlgorithmCache has no field u)


ChrisRackauckas-Claude:
Found while checking polyalg behavior for the #3823 review. Pre-existing on master (identical there and on the #3819/#3820/#3823 branches).

## Repro

```julia
using OrdinaryDiffEqSDIRK, OrdinaryDiffEqNonlinearSolve, NonlinearSolve, ADTypes
using OrdinaryDiffEqNonlinearSolve: NonlinearSolveAlg
function rober!(du, u, p, t)
    y₁, y₂, y₃ = u; k₁, k₂, k₃ = p
    du[1] = -k₁*y₁ + k₃*y₂*y₃
    du[2] = k₁*y₁ - k₂*y₂^2 - k₃*y₂*y₃
    du[3] = k₂*y₂^2
    nothing
end
prob = ODEProblem(rober!, [1.0,0.0,0.0], (0.0,1e5), [0.04,3e7,1e4])
solve(prob, TRBDF2(nlsolve = NonlinearSolveAlg(RobustMultiNewton(; autodiff=AutoForwardDiff()))); reltol=1e-6, abstol=1e-9)
# ERROR: FieldError: type NonlinearSolveBase.NonlinearSolvePolyAlgorithmCache has no field `u`,
# available fields: `static_length`, `prob`, `caches`, `alg`, `best`, `current`, `nsteps`, `stats`, ...
```

Same for `FastShortcutNonlinearPolyalg()`. Single algorithms (`NewtonRaphson`, `TrustRegion`) work.

## Cause

The NSA glue accesses inner-cache internals by raw field: `initialize!` reads `cache.cache.u` (resize length check) and `cache.cache.stats`; `compute_step!` reads `nlcache.u` after `step!`. `NonlinearSolvePolyAlgorithmCache` stores per-branch `caches` and has no top-level `u`. Beyond the field access, `step!(cache; recompute_jacobian=...)` semantics for a polyalg cache (which branch? switching policy per outer iteration?) are undefined for the current one-`step!`-per-outer-iteration driving mode.

## Fix direction

Short-term: use `NonlinearSolveBase.get_u(cache)`/public accessors instead of `.u`/`.stats` raw fields, and either make polyalgs work with the step!-driving mode or error informatively at `init` time.

[...]

PR #3959, opened 2026-07-19
Fix NonlinearSolveAlg crashing/stalling with polyalgorithm inner solvers

singhharsh1708:
Fixes the `FieldError` crash in #3859.

`NonlinearSolvePolyAlgorithmCache` (RobustMultiNewton, FastShortcutNonlinearPolyalg) keeps `u`/`fu` on its active subcache, not at top level, so every raw `cache.cache.u`-style read in newton.jl threw `FieldError` the moment a polyalgorithm was used as the `NonlinearSolveAlg` inner solver. Its `reinit!` also left `force_stop`/`retcode` set after a branch converged once, so a `step!`-driven loop silently no-oped forever after and reused a stale iterate.

Per review, both problems are fixed at the interface level in NonlinearSolve.jl#1115 (polyalg `get_u`/`get_fu` accessors plus the `force_stop`/`retcode` reset in `reinit!`), merged and released as NonlinearSolveBase 2.39.0. This PR is just the consumer side: every polyalg-unsafe inner-cache read in newton.jl — the resize length check, the stale-`W` rhs rescale from #4028, both `compute_step!` paths, and the nlstep `build_solution` — goes through `NonlinearSolveBase.get_u`/`get_fu`. No local workaround remains. Top-level fields the polyalg cache genuinely has (`retcode`, `stats`, `prob`, `alg`) are read as before, and `not_terminated` only touches `force_stop`/`nsteps`/`maxiters`, which are also top-level. The nlstep `build_solution` no longer attaches the inner trace (no polyalg accessor exists; the transient solution only feeds `nlprobmap`). The compat floor is raised to `"2.39"`, and `get_u`/`get_fu` are added to the QA non-public-import allowlist alongside `not_terminated`.

Regression test `nsa_polyalg_tests.jl` runs TRBDF2 and FBDF with a `RobustMultiNewton` inner solver and checks both against the analytic solution of a stiff linear system to 1e-4, plus retcodes. Against the registered NonlinearSolveBase (pinned to the 2.39.0 floor): 8/8 new, and unchanged elsewhere — newton_tests 6/6, nsa_jacobian_reuse 17/17, nsa_matrixfree 14/14, nsa_smooth_est 10/10, nsa_sparse 10/10, nsa_stats 24/24, QA 16/16, all identical to master.

The #3859 MWE (Robertson, `TRBDF2(nlsolve = NonlinearSolveAlg(RobustMultiNewton(autodiff = AutoForwardDiff())))`, `reltol=1e-6`, `abstol=1e-9`) throws `FieldError` on master and returns `Success` here, with `u(1e5) = [0.0178650, 7.2744e-8, 0.9821349]` against a default-`FBDF` reference of `[0.0178659, 7.2748e-8, 0.9821340]` (5e-5 relative). The active branch is TrustRegion, and the run is step-for-step identical to `NonlinearSolveAlg(TrustRegion())`.

Scope: this fixes the crash, not every polyalgorithm. `FastShortcutNonlinearPolyalg` no longer throws, but its leading branches are Jacobian-free quasi-Newton methods, which do not produce a usable stage step under the one-`step!`-per-outer-iteration driving mode; the integrator reports `Unstable`/`MaxIters` rather than a wrong answer. Polyalgorithms combined with a matrix-free `W` (`linsolve = KrylovJL_GMRES()`) are still refused at `init` by NonlinearSolveBase, identically on master, because `_nlalg_with_linsolve` cannot push a linsolve into a polyalgorithm; the related SciML/NonlinearSolve.jl#1132 is still open. Both are better addressed by the #3817 redesign (drive inner solvers by `solve!` per stage) than by more glue here.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
@@ -140,7 +140,7 @@ function initialize!(
         else
             nlp_params = (tmp, ustep, γ, α, tstep, k, invγdt, method, p, dt, f)
         end
-        if length(cache.cache.u) != length(z)
+        if length(get_u(cache.cache)) != length(z)
             new_prob = if cache.W !== nothing
                 # W-reuse: re-point the inner jacobian at the resized W via the same mapping
                 # used at build time. Only array sizes change, so the jac's concrete type —
@@ -229,7 +229,7 @@ function rescale_stale_W_rhs!(nlcache, nlsolver, integrator, nlstep_data)
     isdae = nlsolve_f(integrator) isa DAEFunction
     γdt = isdae ? nlsolver.α * cache.invγdt : nlsolver.γ * integrator.dt
     W_γdt ≈ γdt && return nothing
-    rmul!(nlcache.fu, 2 / (1 + γdt / W_γdt))
+    rmul!(get_fu(nlcache), 2 / (1 + γdt / W_γdt))
     return nothing
 end
 
@@ -252,11 +252,12 @@ end
         return convert(eltype(z), Inf)
     end
     step!(nlcache; recompute_jacobian)
-    nlsolver.ztmp = nlcache.u
+    active_u = get_u(nlcache)
+    nlsolver.ztmp = active_u
 
     ustep = compute_ustep(tmp, γ, z, method)
     atmp = calculate_residuals(
-        z .- nlcache.u, uprev, ustep, opts.abstol, opts.reltol,
+        z .- active_u, uprev, ustep, opts.abstol, opts.reltol,
         opts.internalnorm, t
     )
     ndz = opts.internalnorm(atmp, t)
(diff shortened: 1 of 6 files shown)


lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
```diff
@@ -1,5 +1,27 @@
...
+
+# `NonlinearSolvePolyAlgorithmCache`'s own `reinit!` resets `current`/`nsteps`/`stats`
+# but not `force_stop`/`retcode`. Once any branch has converged once, those two stay
+# set, so `CommonSolve.step!`'s `not_terminated(cache) || return` guard makes every
+# later `step!` on the same cache silently a no-op (stale `u` reused every step).
+# Reset them explicitly after each `reinit!` until this is fixed upstream.
+@inline function _reset_polyalg_termination!(nlcache)
+    if !hasfield(typeof(nlcache), :u)
+        nlcache.force_stop = false
+        nlcache.retcode = SciMLBase.ReturnCode.Default
```
ChrisRackauckas: It's probably best to improve the interface on the nonlinearsolve.jl side first.

singhharsh1708: done

PR #3975, opened 2026-07-21
Fix ExplicitTaylor fsallast aliasing to state buffer

singhharsh1708:
## Summary

`ExplicitTaylor` and `ExplicitTaylorAdaptiveOrder` build each step from the Taylor jet and never reuse the previous step's last derivative, so they are not FSAL methods. Neither declared `isfsal`, though, so both inherited the `true` default.

That made `ode_determine_initdt` take its FSAL branch, which writes an extra `f` evaluation into `integrator.fsallast`. The mutable caches returned `(cache.u, cache.u)` from `get_fsalfirstlast`, and `cache.u === integrator.u`, so that call was literally `f(u, u, p, t)` — corrupting `u` mid-evaluation for any in-place RHS that reads from `u` after writing to `du`, which is a normal pattern.

Effects on master, before a single step is taken:

- With `u0 = [1.0, 2.0]` and an RHS that writes `du` then reads `u`, `integrator.u` comes back from `init` as `[-16.0, 16.0]`.
- The initial-step estimate sees a `NaN`, warns, and falls back to `tdir * dtmin`. On `prob_ode_pleiades` at `abstol = reltol = 1e-10` that is a first step of `1.0e-323`, and the controller spends hundreds of steps climbing out.

## Fix

Declare the two methods non-FSAL. The heuristic then takes its other branch and uses its own `zero(u0)`, so nothing writes through the state buffer and no cache-owned FSAL storage is needed at all — the caches return `(nothing, nothing)`, as `FunctionMapCache` already does.

Measured on `prob_ode_pleiades` with `ExplicitTaylor(order = Val(6))`:

| setting | before | after |
| --- | --- | --- |
| default tolerances | Success, 386 steps + NaN warning | Success, 59 steps |
| `abstol = reltol = 1e-10` | Success, 905 steps | Success, 587 steps |

Solutions agree to about six significant figures on this chaotic problem.

## Known follow-up, not fixed here

`ExplicitTaylorAdaptiveOrder` had the same problem. Before this fix its auto-dt collapsed to `dtmin` and the solve died loudly with `retcode = MaxIters` after 1,000,001 steps. Afterwards that failure is gone, but a separate pre-existing bug in its per-step order/error controller becomes visible: on Pleiades it returns `Success` after 3 steps with `sol.u[end]` bit-identical to `u0`. That is a silent wrong answer where there used to be a loud failure, filed as #3976 rather than fixed here, and worth weighing when reviewing.

## Tests

`auto-dt must not write through the state buffer` asserts both methods report `isfsal == false` and that `init` leaves `u` unmodified, using a two-line RHS that reads `u` after writing `du`. It fails four times on master and passes here, in about two seconds.

`Auto-dt on Pleiades does not collapse to dtmin` keeps an end-to-end check with a bound that discriminates (`length(sol.t) < 700`; master takes 905, this branch 587).

Existing `OrdinaryDiffEqTaylorSeries` Core group passes unchanged. Runic clean.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_caches.jl
+++ b/lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_caches.jl
@@ -81,7 +81,7 @@ function alg_cache(
     )
 end
 
-get_fsalfirstlast(cache::ExplicitTaylorCache, u) = (cache.u, cache.u)
+get_fsalfirstlast(cache::ExplicitTaylorCache, u) = (nothing, nothing)
 
 struct ExplicitTaylorConstantCache{P, taylorType, uType, tType} <:
     OrdinaryDiffEqConstantCache
@@ -151,7 +151,7 @@ function alg_cache(
     )
 end
 
-get_fsalfirstlast(cache::ExplicitTaylorAdaptiveOrderCache, u) = (cache.u, cache.u)
+get_fsalfirstlast(cache::ExplicitTaylorAdaptiveOrderCache, u) = (nothing, nothing)
 
 struct ExplicitTaylorAdaptiveOrderConstantCache{P, Q, taylorType, uType, tType} <:
     OrdinaryDiffEqConstantCache
--- a/lib/OrdinaryDiffEqTaylorSeries/src/OrdinaryDiffEqTaylorSeries.jl
+++ b/lib/OrdinaryDiffEqTaylorSeries/src/OrdinaryDiffEqTaylorSeries.jl
@@ -8,7 +8,7 @@ import OrdinaryDiffEqCore: alg_stability_size, explicit_rk_docstring,
     _ode_interpolant, _ode_interpolant!,
     OrdinaryDiffEqAlgorithm,
     _ode_addsteps!,
-    get_fsalfirstlast,
+    get_fsalfirstlast, isfsal,
     DerivativeOrderNotPossibleError, unwrap_alg, step_accept_controller!,
     stepsize_controller!, get_current_adaptive_order, get_current_alg_order
 using FastBroadcast: Serial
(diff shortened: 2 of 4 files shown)


lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_caches.jl
```diff
@@ -81,7 +81,14 @@ function alg_cache(
...
 
-get_fsalfirstlast(cache::ExplicitTaylorCache, u) = (cache.u, cache.u)
+# FSAL currently not used. `cache.u` must NOT be reused here: the generic
+# auto-dt heuristic (`ode_determine_initdt`) writes an extra `f` evaluation
+# into whichever buffer `fsallast` points to, so aliasing it to the state
+# itself makes that call `f(u, u, p, t)` and can corrupt `u` mid-evaluation
+# for RHS functions that read from `u` after writing to `du`, which corrupts the
+# state before the first step. Use independent buffers of the `du` type, matching
+# the core `OrdinaryDiffEqConstantCache` default.
+get_fsalfirstlast(cache::ExplicitTaylorCache, u) = (zero(u), zero(u))
```
ChrisRackauckas: where is this actually called? This function shouldn't allocate.

singhharsh1708: Right , switched to a cache buffer, 0 allocs now.

lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_caches.jl
```diff
@@ -75,13 +76,20 @@ function alg_cache(
...
         alg.stage_limiter!, alg.step_limiter!, alg.thread
     )
 end
 
-get_fsalfirstlast(cache::ExplicitTaylorCache, u) = (cache.u, cache.u)
+# FSAL is not used by these methods, but `cache.u` must NOT be handed out as the
+# buffer: the auto-dt heuristic (`ode_determine_initdt`) writes an extra `f`
+# evaluation into whatever `fsallast` points at, so aliasing it to the state makes
+# that call `f(u, u, p, t)` and corrupts `u` before the first step for any RHS that
+# reads `u` after writing `du`. Hand out a dedicated cache buffer instead, the same
```
ChrisRackauckas: But what function needs this from non FSAL methods?

singhharsh1708: they're not FSAL. Declared it, dropped the buffer.

ChrisRackauckas (approved):

PR #3986, opened 2026-07-22
reexport `NoAutoDiff`

oscardssmith:
For circumstances where the user is providing derivatives manually and doesn't want any autodiff to be used (even `FiniteDiff`).


src/OrdinaryDiffEq.jl
Drvi: I think you need to add it here to reexport it

oscardssmith: oops

ChrisRackauckas:
Needs QA handling

PR #1085, opened 2020-03-16
Fix ::Array{Float64,1} use case in QNDF

utkarsh530:
Fixes #1076. @ChrisRackauckas please review.

PR #3989, opened 2026-07-22
Initialize HomotopyProblem init systems by continuation (default_nlsolve → nothing)

ChrisRackauckas-Claude:

## Summary

Makes the DAE-initialization OverrideInit path initialize a `SciMLBase.HomotopyProblem` (and an `SCCNonlinearProblem` that contains homotopy blocks) **by continuation**, by having `default_nlsolve` return `nothing` for them.

## Why

`default_nlsolve` picks the nonlinear algorithm the OverrideInit path solves the init problem with. A `HomotopyProblem` — built by ModelingToolkit when the init system carries Modelica `homotopy(actual, simplified)` operators (SciML/ModelingToolkit.jl#4766) — has a λ-swept residual `f(u, p, λ)`; the operator is the user declaring the system needs continuation. Returning `nothing` routes `solve(initprob, nothing)` to the problem's own default (`NonlinearSolveBase.HomotopyPolyAlgorithm`), which sweeps λ from the `simplified` form to the target.

The alternative — the previous `AbstractNonlinearProblem` default, `FastShortcutNonlinearPolyalg` — under the companion SciML/NonlinearSolve.jl#1085 solves only the **target-λ system** directly (`solve(::HomotopyProblem, ::standard_alg)` fixes λ). That is a real bug, not just slower: on branch-selection it returns a **`Success` retcode on the wrong root**. Measured end-to-end on `homotopy(m² − 1, 2(m−1))` from `m = −3`:

```
BEFORE (FastShortcut / target-λ Newton):  init = -1.0   ✗ Success on the unphysical root
AFTER  (nothing / continuation):          init = +1.0   ✔ physical root
```

An `SCCNonlinearProblem` with homotopy blocks is likewise solved with `nothing`, so each block picks its own default (homotopy blocks continue, plain blocks use the standard nonlinear default); a pure-`NonlinearProblem` SCC keeps `FastShortcutNonlinearPolyalg`.

## Tests

`test/homotopy_default_nlsolve_tests.jl` (registered in `runtests.jl`): `default_nlsolve` returns `nothing` for a `HomotopyProblem` and for an SCC that has a homotopy block, and the usual `FastShortcutNonlinearPolyalg` otherwise. Verified locally (Julia 1.12): 8/8, plus the end-to-end wrong-root/right-root result above.

## Companions

- SciML/NonlinearSolve.jl#1085 — `solve(::HomotopyProblem, ::AbstractNonlinearSolveAlgorithm)` solves the target-λ system (so a standard algorithm here does the obvious thing).
- SciML/ModelingToolkit.jl#4766 — builds the `HomotopyProblem`/homotopy SCC blocks this PR initializes.

Runic-formatted.

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/OrdinaryDiffEqNonlinearSolve.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/OrdinaryDiffEqNonlinearSolve.jl
@@ -16,7 +16,7 @@ import ConstructionBase
 import PreallocationTools: DiffCache, get_tmp
 using SimpleNonlinearSolve: SimpleTrustRegion, SimpleGaussNewton
 using NonlinearSolve: FastShortcutNonlinearPolyalg, FastShortcutNLLSPolyalg, NewtonRaphson,
-    HomotopySweep, step!
+    HomotopySweep, HomotopyPolyAlgorithm, ArcLengthContinuation, step!
 # The operator Jacobian path is implemented in NonlinearSolveBase and needs its own floor.
 import NonlinearSolveBase
 using MuladdMacro: @muladd
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/initialize_dae.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/initialize_dae.jl
@@ -80,6 +80,9 @@ end
 # level `autodiff` knob on `HomotopyPolyAlgorithm` itself cannot do this: it lives in
 # NonlinearSolveBase, which cannot construct the concrete inner solver, so the AD rides on
 # the inner `FastShortcutNonlinearPolyalg` built here.
+#
+# This is exactly `NonlinearSolve.FastShortcutHomotopyPolyalg(; autodiff)`
+# (SciML/NonlinearSolve.jl#1105); once that releases, this body collapses to a call to it.
 function _homotopy_init_alg(u, autodiff, chunksize)
     inner = FastShortcutNonlinearPolyalg(;
         autodiff = autodiff ? _tagged_autodiff(u, chunksize) : AutoFiniteDiff()
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/initialize_dae.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/initialize_dae.jl
@@ -111,11 +111,14 @@ end
 # default. A pure-`NonlinearProblem` SCC keeps the `FastShortcutNonlinearPolyalg` default.
 #
 # NOTE (autodiff): unlike the whole-system `HomotopyProblem` branch above, the `autodiff`
-# signal is NOT threaded here — `SCCNonlinearSolve` applies one algorithm to every block, so
-# there is no single value that both continues the homotopy blocks and Newton-solves the
-# plain ones with a chosen AD backend. Honoring `autodiff` per block needs `SCCNonlinearSolve`
-# to route by block type (Newton for plain blocks, continuation for homotopy blocks, each
-# with the requested AD); until then the blocks fall back to their ForwardDiff defaults.
+# signal is NOT threaded here — with `nothing` each block picks its own default, so the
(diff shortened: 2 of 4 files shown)


ChrisRackauckas-Claude:
Companions: SciML/NonlinearSolve.jl#1085 and SciML/ModelingToolkit.jl#4766. All drafts; ignore until reviewed by @ChrisRackauckas.

ChrisRackauckas-Claude:
Good catch — the `nothing` return dropped the `autodiff` argument. Fixed in 5cd5e7f.

**Whole-system `HomotopyProblem`**: instead of `nothing`, `default_nlsolve` now returns `HomotopyPolyAlgorithm((HomotopySweep(; inner), ArcLengthContinuation(; inner)))` where `inner = FastShortcutNonlinearPolyalg(; autodiff = autodiff ? _tagged_autodiff(u, cs) : AutoFiniteDiff())`. The AD rides the **inner corrector** — that's the differentiation the sweep and the (default secant-predictor) arclength stage actually use, so a finite-diff init no longer dual-numbers a non-ForwardDiff-safe residual.

On the 'high level finitediff option' question: I looked at putting an `autodiff` knob on `HomotopyPolyAlgorithm` itself, but it can't work cleanly — `HomotopyPolyAlgorithm`/`HomotopySweep` live in NonlinearSolveBase, which can't construct the concrete inner (`FastShortcutNonlinearPolyalg` is in NonlinearSolve; the sweep just passes `inner=nothing` and lets `init(prob, nothing)` resolve the default at dispatch time). So the AD has to ride a concrete inner solver, which `default_nlsolve` can build. Proven genuinely threaded by a ForwardDiff-hostile residual (errors on duals): finite-diff continuation solves to the root, ForwardDiff continuation throws. Unit test 12/12.

**SCC caveat**: I can't thread `autodiff` for the `SCCNonlinearProblem` branch — `SCCNonlinearSolve` applies one algorithm to every block, so no single value both continues the homotopy blocks and Newton-solves the plain ones with a chosen AD. Doing it right wants `SCCNonlinearSolve` to route by block type (Newton for plain, continuation for homotopy, each with the requested AD). I left the SCC branch at `nothing` (falls back to ForwardDiff defaults) and recorded this in a code comment — happy to do the `SCCNonlinearSolve` routing as a follow-up if you'd like.

ChrisRackauckas-Claude:
On "this needs a high level finitediff option?" (the `_homotopy_init_alg` / whole-system `HomotopyProblem` methods): no high-level knob is needed, and one on `HomotopyPolyAlgorithm` itself can't do the job. The AD choice (finite-diff included) is threaded through the continuation's **inner corrector** — the `FastShortcutNonlinearPolyalg` built here — which is the differentiation the sweep and the arclength stage actually use. `HomotopyPolyAlgorithm` lives in NonlinearSolveBase, which can't construct a concrete inner solver, so there's nowhere for a high-level `autodiff` field there to turn into an actual backend. The inner-corrector threading is the mechanism; `autodiff=false` → `AutoFiniteDiff()`, `autodiff=true` → tagged `AutoForwardDiff`. Latest push makes those methods honor `autodiff` this way (unit test added: a ForwardDiff-hostile residual solves under finite-diff and throws under forward-diff).

The **SCC** homotopy methods still return `nothing` on purpose — with `nothing` each block picks its own default, so homotopy blocks continue on the correct branch but at the continuation's *default* inner corrector, not the requested AD backend. Threading AD per block needs `SCCNonlinearSolve` to route by block type. That routing is now open as **SciML/NonlinearSolve.jl#1104**: it adds `HomotopyPolyAlgorithm(; inner)` to NonlinearSolveBase and makes `solve_single_scc` continue `HomotopyProblem` blocks with the SCC's algorithm as the inner corrector (plain blocks unchanged). Once #1104 releases, these SCC methods can return `FastShortcutNonlinearPolyalg(; autodiff)` (with a `SCCNonlinearSolve` compat bump) to honor `autodiff` per block. I've left a code NOTE pointing at #1104 rather than flip the return now, because returning a standard algorithm for a homotopy SCC *before* that routing is released would hit the target-λ fallback (#1085) and risk landing on the wrong branch.

Issue #2972, opened 2026-01-06
Downgrade CI tests have been failing since workflow was added


ChrisRackauckas-Claude:
## Description
The Downgrade and Downgrade Sublibraries CI workflows have been consistently failing since they were added in July 2025 (commit c7609b504).

## Error Details
The root cause appears to be `OrdinaryDiffEqNonlinearSolve` failing to precompile when dependencies are downgraded to their minimum compat bounds:

```
Package OrdinaryDiffEqNonlinearSolve does not have SparseArrays in its dependencies
```

This error message is misleading because `OrdinaryDiffEqNonlinearSolve` DOES have SparseArrays in its `[deps]` section. The issue appears to be related to one of the transitive dependencies at its minimum version causing package loading issues.

## Affected Workflows
- `Downgrade.yml` - Tests InterfaceI group with `julia-downgrade-compat@v2`
- `DowngradeSublibraries.yml` - Tests individual sublibraries with downgraded deps

## Failing Sublibraries
Multiple sublibraries fail due to dependency on `OrdinaryDiffEqNonlinearSolve`:
- OrdinaryDiffEqRKN
- OrdinaryDiffEqExponentialRK
- OrdinaryDiffEqDefault
- OrdinaryDiffEqIMEXMultistep
- OrdinaryDiffEqStabilizedIRK
- OrdinaryDiffEqFIRK
- OrdinaryDiffEqAdamsBashforthMoulton
- OrdinaryDiffEqPDIRK
- OrdinaryDiffEqSDIRK
- OrdinaryDiffEqBDF
- OrdinaryDiffEqRosenbrock
- OrdinaryDiffEqLinear
- OrdinaryDiffEqTaylorSeries

## Potential Fixes
Per the CI Health protocol, options include:
1. **Option B: Bump [compat]** - Increase minimum versions of dependencies causing issues
2. **Option B4: Disable downgrade CI** - Temporarily remove the failing tests from the CI matrix

## Investigation Needed
- Identify which dependency at minimum version is causing the SparseArrays loading issue
- Test with incrementally higher minimum versions to find a working combination
- Consider if the minimum compat bounds are actually tested/supported

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3991, opened 2026-07-23
Newton-Krylov integrators default to Hegedüs warm start (require LinearSolve 5.1)

ChrisRackauckas-Claude:
> [!IMPORTANT]
> Draft — please ignore until reviewed by @ChrisRackauckas.

Companion to SciML/LinearSolve.jl#1100 (now **merged and released as LinearSolve v5.1.0**), which added `WarmStart` + the `WarmStart.Auto` default to `KrylovJL`. This makes the integrator-aware default **unconditionally live** on LinearSolve 5.1.

## What

`OrdinaryDiffEqDifferentiation.default_krylov_warm_start(linsolve)` resolves a Krylov solver left at `LinearSolve.WarmStart.Auto` (the `KrylovJL` default) to `WarmStart.Hegedus`, applied on the **Newton nonlinear-solver path** (`NLNewton`/`NonlinearSolveAlg`). Implicit integrators using `KrylovJL_GMRES()`/`KrylovJL_FGMRES()` now reuse the previous solution (Hegedüs-scaled) across the correlated linear solves of a Newton iteration.

- **Rosenbrock / W-methods untouched**: they never call the resolver, so their `Auto` solver stays a cold start (warm starting is unsafe there — no outer Newton iteration absorbs the within-tolerance stage-solve perturbation; benchmarks showed 7–13× slowdowns and step-rejection loops).
- **Explicit choices respected**: `WarmStart.None`/`Previous`/`Hegedus` are never overridden — only `Auto` is resolved.

## Requires LinearSolve 5.1 (no guard, no dormancy)

`WarmStart` is a LinearSolve **5.1** feature. Rather than guard on `isdefined(LinearSolve, :WarmStart)` and be silently dormant on older LinearSolve, the WarmStart-using libraries now **require `LinearSolve = "5.1"`** and the guard is removed, so the default is always live. Every other monorepo sublibrary that declared a LinearSolve compat is moved to `"5.1"` as well: they all depend transitively on OrdinaryDiffEqNonlinearSolve/Differentiation, so they can never resolve with LinearSolve < 5.1 anyway — keeping a 3/4 lower bound would be dead metadata and would re-break the Downgrade CI (which floors deps to their minimum). Test sub-environments updated to match. This is effectively the OrdinaryDiffEq **LinearSolve-5 migration**, done here because the require-5.1 floor depends on it.

Minor version bumps for each sublibrary whose compat changed.

## Verification (local, LinearSolve 5.1.0)

- Fresh resolution pins **LinearSolve 5.1.0** with the two libraries requiring `"5.1"` — the registered NonlinearSolveBase/DiffEqBase/etc. chain already allows LinearSolve 5, so nothing caps it at 4.
- Resolver unit test (`test/warm_start_default_tests.jl`, Differentiation Core group): 7/7 pass — Auto→Hegedus, explicit values preserved, non-Krylov unchanged, other fields survive `remake`.
- `FBDF(linsolve = KrylovJL_GMRES())` and `Rodas5P(linsolve = KrylovJL_GMRES())` both solve `Success`; the resolver returns `Hegedus` for `Auto` with the guard removed.
- Earlier engagement check (ILU Brusselator): FBDF fires warm start (433 engagements), Rodas5P does not (0), explicit `None` on FBDF is cold (0), explicit `Hegedus` on Rodas5P is honored (1528).

CI validates the full umbrella resolution + all sublibrary test groups on LinearSolve 5.1.


ChrisRackauckas-Claude:
CI note: the `downgrade-sublibraries` failures (Differentiation, FIRK) are **pre-existing on master** — the `Downgrade`/`Downgrade Sublibraries` workflows are red across recent master commits (d2577da9, 4b3cafdf, 754b504c), unrelated to this PR. This change keeps the same LinearSolve compat floor (`3.75.0`), so it doesn't affect downgrade resolution, and both the resolver and its test are guarded on `isdefined(LinearSolve, :WarmStart)` so they degrade to a no-op on LinearSolve 3.75. The failure is the known julia-downgrade-compat-vs-path-`[sources]` resolution issue, not a real floor incompatibility. (Triaging the master failure separately.)

The checks that actually exercise this PR — the OrdinaryDiffEqDifferentiation/OrdinaryDiffEqNonlinearSolve sublib test groups and Runic — are the ones to watch.

ChrisRackauckas-Claude:
CI verdict — the checks that exercise this PR are green:
- **Runic** (format + suggestions): pass
- **OrdinaryDiffEqDifferentiation Core** tests (Julia 1 / lts / pre): pass — these run the new `warm_start_default_tests.jl`
- OrdinaryDiffEqNonlinearSolve Core: pending (same class)

Remaining red checks are all **pre-existing on master**, not introduced here:
- **QA** (Differentiation, NonlinearSolve): the SciMLTesting `public API is rendered in docs` check flags pre-existing `public`-but-undocumented internal helpers (`wrapprecs`, `dolinsolve`, `build_J_W`, `update_W!`, …). Confirmed failing on master (sublibrary run 30005108996). The new `default_krylov_warm_start` is deliberately **not** `public`, so it is not in the flagged list and adds nothing to this debt.
- **downgrade-sublibraries**: chronic (never green on master, tracked under #2972/#2977). This PR fixes the two floor conflicts that live in the files it touches (LinearSolve 3.75→4.3, SciMLOperators 1.24.3→1.24.4); any remaining transitive floor conflict is the broader monorepo cleanup.

PR #3996, opened 2026-07-23
Preserve ArrayPartition wrapper in out-of-place stiff steps (#3962)

### LINKED ISSUE #1263: ArrayPartition + Implicit + OOP Fails
```julia
using DiffEqOperators, OrdinaryDiffEq, Plots

ω = 1; k = 1; c = (ω/k)^2
number_of_spatial_cells = 100
dx = 1/(number_of_spatial_cells+1)
ord_spacial_deriv = 2
u_0 = 0.2.*sin.(range(0, 2*pi, length=number_of_spatial_cells))
u_dot_0 = zeros(number_of_spatial_cells)
A_x = CenteredDifference{1}(2, 2, dx, number_of_spatial_cells)
Q = Dirichlet0BC(Float64)

# Inplace works
u_dot_dot_iip(out, du, u, p, t) = out .= c*A_x*Q*u
prob = SecondOrderODEProblem(u_dot_dot_iip, u_dot_0, u_0, (0.0, 1.0))
sol = solve(prob, Rosenbrock23())

# Out of place fails
u_dot_dot(du, u, p, t) = c*A_x*Q*u
prob = SecondOrderODEProblem(u_dot_dot, u_dot_0, u_0, (0.0, 1.0))
sol = solve(prob, Rosenbrock23())
```

### LINKED ISSUE #3962: Out-of-place stiff solvers drop the ArrayPartition wrapper in rosenbrock_perform_step.jl:481 (type Array has no field x)
Found while triaging SciML/DifferentialEquations.jl#301 (verified on OrdinaryDiffEq v7.1.3, Julia 1.11.9).

Out-of-place stiff solvers fail on `SecondOrderODEProblem`/`DynamicalODEFunction` states backed by `ArrayPartition`, because the wrapper is dropped mid-step.

```
out-of-place (vector components):
  Rodas5P      -> ERROR: type Array has no field x
  Rosenbrock23 -> ERROR: type Array has no field x
  FBDF         -> ERROR: MethodError: Cannot convert Vector{Float64} to ArrayPartition{...}

in-place (same problem):
  Rodas5P / Rosenbrock23 / FBDF -> all retcode=Success
```

**Root cause** — `lib/OrdinaryDiffEqRosenbrock/src/rosenbrock_perform_step.jl:481`:

```julia
u = @.. u + A[stage,i]*ks[i]
```

`ks[i]` is a plain `Vector`, so the broadcast yields a plain `Vector` and drops the `ArrayPartition` wrapper. The subsequent `f(u, p, t)` then hits `u.x` and errors.

Confirmed solver-side rather than a bad MWE: `prob.f(u0, p, t)` called directly works and returns an `ArrayPartition`.

Also of note: `alg_hints=[:stiff]` **succeeds**, because auto-selection takes a different path — so this is easy to miss depending on how the algorithm is specified.

Context: the originally reported `DimensionMismatch` in `reshape_jacobian` on DifferentialEquations.jl#301 is fixed (consistent with DiffEqBase.jl#536). This is a distinct, current bug in the same code path, filed separately rather than resurrecting a 2018 / Julia 0.6 thread.

ChrisRackauckas-Claude:

Fixes #3962. Also fixes #1263 (and unblocks the OOP-stiff path noted in SciML/DifferentialEquations.jl#301).

## Problem

Out-of-place stiff solvers failed on `SecondOrderODEProblem` / `DynamicalODEFunction` states backed by `ArrayPartition`:

```
oop (vector components):
  Rodas5P / Rosenbrock23  ->  ERROR: type Array has no field x
  FBDF                    ->  MethodError: Cannot convert Vector{Float64} to ArrayPartition{...}
iip (same problem):
  all -> retcode = Success
```

## Root cause

The out-of-place stages rebuild the state from a flat linear-solve result with

```julia
_reshape(W \ _vec(...), axes(uprev))
```

`reshape` to the (linear) axes of an `ArrayPartition` collapses it to a bare `Vector`. That bare `Vector` then survives the stage broadcasts, and:

- in Rosenbrock/Rodas the next `f(u, p, t)` hits `u.x` → `type Array has no field x`;
- in the shared out-of-place Newton path (FBDF and the other BDF/DIRK methods) the update `z .- dz` produces a `Vector` that can no longer be stored back into the `ArrayPartition`-typed nlsolver fields → `Cannot convert Vector to ArrayPartition`.

## Fix

Use the public API `ArrayInterface.restructure(template, flat)` (documented for ArrayPartition; RAT uses it internally) at every out-of-place reshape site in the Rosenbrock constant caches, stiff dense-output addsteps, and the out-of-place Newton / functional-iteration nlsolver paths.

A thin local wrapper handles `Number` states (`restructure` relies on `similar`, which is not defined for scalars):

```julia
@inline _restructure_state(template, x) = ArrayInterface.restructure(template, x)
@inline _restructure_state(template::Number, x) = oftype(template, x)
```

This mirrors how the in-place paths already keep the wrapper (they fill `_vec(ks[i])` of a preallocated `ArrayPartition` in place). Plain `Vector`, `StaticArray`, matrix-valued, and scalar states go through restructure/`oftype` cleanly.

## Verification (run locally, Julia 1.11.9)

`u'' = -u`, `u(0) = [1,0]`, `u'(0) = [0,1]` as an OOP `SecondOrderODEProblem`; compared against the in-place reference:

| solver | before | after |
|---|---|---|
| Rosenbrock23 | `type Array has no field x` | Success, matches ref |
| Rodas5P | `type Array has no field x` | Success, matches ref |
| FBDF | `Cannot convert Vector to ArrayPartition` | Success, matches ref |

Also verified: #1263 MWE succeeds and stays `ArrayPartition`; scalar and StaticArray OOP stiff paths still succeed after the Number branch.

## Tests

- `lib/OrdinaryDiffEqRosenbrock/test/ode_rosenbrock_tests.jl`: OOP `SecondOrderODEProblem` with Rosenbrock23/Rodas5P/Rodas4/Rodas5 stays an `ArrayPartition` and matches the in-place reference.
- `lib/OrdinaryDiffEqBDF/test/bdf_regression_tests.jl`: same for FBDF.

Patch-version bumps for `OrdinaryDiffEqRosenbrock` (2.6.1) and `OrdinaryDiffEqNonlinearSolve` (2.5.1).

### DIFF
--- a/lib/OrdinaryDiffEqBDF/test/bdf_regression_tests.jl
+++ b/lib/OrdinaryDiffEqBDF/test/bdf_regression_tests.jl
@@ -1,7 +1,8 @@
-using OrdinaryDiffEqBDF, OrdinaryDiffEqCore, ForwardDiff, Test
+using OrdinaryDiffEqBDF, OrdinaryDiffEqCore, ForwardDiff, Test, LinearAlgebra
 using OrdinaryDiffEqCore: DEVerbosity
 import OrdinaryDiffEqCore.SciMLLogging as SciMLLogging
 using OrdinaryDiffEqNonlinearSolve: BrownFullBasicInit, NLNewton
+using RecursiveArrayTools: ArrayPartition
 
 foop = (u, p, t) -> u * p
 proboop = ODEProblem(foop, ones(2), (0.0, 1000.0), 1.0)
@@ -195,4 +196,26 @@ end
     # the solver gives up cleanly (Unstable) instead of overflowing the stack.
     @test sol.retcode != ReturnCode.Success
     @test sol.retcode != ReturnCode.Default
+end
+
+# Regression test for issue #3962: the out-of-place Newton step used
+# `_reshape(W \ _vec(ztmp), axes(ztmp))`, which collapses the ArrayPartition state
+# behind SecondOrderODEProblem/DynamicalODEFunction down to a bare Vector. Storing
+# `z .- dz` back into the ArrayPartition-typed nlsolver fields then failed with
+# `Cannot convert Vector to ArrayPartition`.
+@testset "OOP SecondOrderODEProblem with FBDF preserves ArrayPartition (#3962)" begin
+    # u'' = -u  ⇒  position = [cos t, sin t], velocity = [-sin t, cos t]
+    ho_iip(ddu, du, u, p, t) = (@. ddu = -u)
+    ho_oop(du, u, p, t) = -u
+    u0 = [1.0, 0.0]
+    du0 = [0.0, 1.0]
+    tspan = (0.0, 1.0)
+    prob_iip = SecondOrderODEProblem(ho_iip, du0, u0, tspan)
+    prob_oop = SecondOrderODEProblem(ho_oop, du0, u0, tspan)
+
+    ref = solve(prob_iip, FBDF(), abstol = 1.0e-10, reltol = 1.0e-10)
+    sol = solve(prob_oop, FBDF(), abstol = 1.0e-10, reltol = 1.0e-10)
+    @test sol.retcode == ReturnCode.Success
+    @test sol.u[end] isa ArrayPartition
+    @test norm(sol.u[end] - ref.u[end]) < 1.0e-6
 end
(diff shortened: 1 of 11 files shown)


ChrisRackauckas-Claude:
## Follow-up commit: switched hand-rolled helper → `ArrayInterface.restructure`

Pushed `7f84e043f` on this branch:

- Dropped the `ArrayPartition`-only `_reshape_state` special-case.
- All OOP rebuild sites now go through `ArrayInterface.restructure(template, flat)` (public API; RAT uses it internally for structured containers).
- Thin Number method kept because `restructure` uses `similar`, which fails on scalars.
- Also fixed `stiff_addsteps.jl` OOP dense-output stages, which still used bare `_reshape(... , axes(uprev))`.

Local checks (Julia 1.11.9):
- #3962 / #1263 MWEs: Success, `ArrayPartition` preserved, matches iip reference.
- Scalar + StaticArray + matrix OOP stiff paths: Success.
- FBDF OOP SecondOrderODEProblem: Success.

PR #4000, opened 2026-07-23
Route OOP initdt NaN checks through DiffEqBase.NAN_CHECK (#1404)

### LINKED ISSUE #1404: recursive isnan
We have developed a `Field` type for spectral-element spatial discretizations. By adding definitions to RecursiveArrayTools we have managed to get it to play nicely with OrdinaryDIffEq, except for this line:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/181dcf265351ed3c02437c89a8d2af3f6967fa85/src/initdt.jl#L80

As a result we had to intercept `any` in  way that isn't quite right:
https://github.com/CliMA/ClimateMachineCore.jl/blob/57c9fa1e5642e16fe1a66272c142ba711627f321/src/field/field.jl#L303

Would it be possible to add this as a function to RecursiveArrayTools (e.g. `recursive_isnan`) so that we can overload it appropriately?

ChrisRackauckas-Claude:
## Summary

- OOP `initdt` used `any(x -> any(isnan, x), f₀)`, which custom array/field types cannot overload.
- Replace with `DiffEqBase.NAN_CHECK` (and `isnan(d₁)` for AD partials), aligning with the IIP intent.
- Also early-returns `tdir * dtmin` on NaN (previously only warned and continued).

Fixes #1404

## Test plan

- [x] Local: OOP `f -> [NaN]` exits without MethodError
- [x] Local: finite OOP still succeeds
- [ ] CI: InterfaceI initdt tests

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -145,3 +145,4 @@ Comput = "Comput"  # J. Comput. Phys. (Journal of Computational Physics)
 
 # Julia spelling convention
 inferrable = "inferrable"  # Julia uses "inferrable" not "inferable"
+hom = "hom"  # scc_hom homotopy abbreviation in NonlinearSolve tests (#3989)
--- a/lib/OrdinaryDiffEqCore/src/initdt.jl
+++ b/lib/OrdinaryDiffEqCore/src/initdt.jl
@@ -360,11 +360,15 @@ end
 
     f₀ = f(u0, p, t)
 
-    if any(x -> any(isnan, x), f₀)
+    # Use the overloadable DiffEqBase.NAN_CHECK hook (same intent as the IIP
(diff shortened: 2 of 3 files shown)


ChrisRackauckas-Claude:
## CI babysit (wave3/ode_ci_a)

**PR-caused failures:** none found. Local OOP `NAN_CHECK` test passes (`retcode=Unstable` as expected). Runic already green on previous head.

**Pushed:** `hom` allowlist in `.typos.toml` (master regression from #3989).

**Remaining failures (pre-existing):**
- Downgrade InterfaceI: `ArrayInterface` resolver unsatisfiable (same on sibling PRs)
- Spell: fixed via allowlist

PR #4005, opened 2026-07-23
AutoDePSpecialize: de-specialize non-isbits parameters via OpaqueRef

ChrisRackauckas-Claude:
> [!NOTE]
> Draft PR — please ignore until reviewed by @ChrisRackauckas.

Follow-up to #3692 (which shipped `AutoDePSpecialize` for **isbits** parameters, now on master). This extends the opaque-`p` path to **non-isbits** parameters.

## What changed

`should_opaque_p` now de-specializes any `p` except `NullParameters` and already-packed containers (previously isbits-only). `RespecializeParams.pack_auto`/`opaque_container_type` route the payload: **isbits → `OpaqueParams`** (byte copy, as before), **everything else → `OpaqueRef`** (by reference). The ForwardDiff extension substitutes `opaque_container_type(P)` instead of hardcoding `OpaqueParams`, and an **idempotency guard** stops a re-concretized opaque problem from being re-wrapped (no nested `OpaqueVoid`).

The RespecializeParams machinery (`OpaqueRef`, `OpaqueVoid`'s `OpaqueRef` methods, `pack_auto`) already existed — this just stops gating it out. `DiffEqBase 7.8.0 → 7.9.0`.

## Value + scope

Extends the shared-compilation TTFX benefit to problems whose parameter type is a **non-isbits struct** (e.g. a parameter struct holding arrays — different model types then share one precompiled solve).

It remains a **forward-solve latency tool**. De-specializing makes `sol.prob.p` an opaque container, so:
- **Reverse-mode AD of `p`** (SciMLSensitivity) and **symbolic parameter indexing** (`getp`/`setp`/observed) — which read the concrete parameter — are **not supported**. Recover the payload explicitly with `RespecializeParams.unpack(sol.prob.p, typeof(p))`.
- **Reverse-mode support is a planned follow-up**: the adjoint would unwrap the RHS (via `unwrapped_f`, extended to see through `OpaqueVoid`) and unpack `p`, the same way it already unwraps the plain `AutoSpecialize` function wrapper. (Prototyped; it's a multi-path SciMLSensitivity change, so it's separated out.)
- Forward-mode AD of `p` is unaffected — the existing `isdualtype(u0)` guard skips the opaque path whenever `p` carries duals.

## Verified

- `GROUP=Core` green: **Opaque-p Hook 25/25** (adds OpaqueRef + idempotency + inference cases), ForwardDiff Dual Detection unchanged.
- Full-stack non-isbits solves match `FullSpecialize` across `Tsit5`/`Rodas5P`/`FBDF` (`sol.prob.p isa OpaqueRef`, unpack roundtrips, different non-isbits types share `typeof(prob.f)`).

https://claude.ai/code/session_01UuAbA8hSnrirMrJjFqxwn5

### DIFF
--- a/lib/DiffEqBase/Project.toml
+++ b/lib/DiffEqBase/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqBase"
 uuid = "2b5f629d-d688-5b77-993f-72d75c75574e"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "7.8.0"
+version = "7.9.0"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
--- a/lib/DiffEqBase/ext/DiffEqBaseForwardDiffExt.jl
+++ b/lib/DiffEqBase/ext/DiffEqBaseForwardDiffExt.jl
@@ -133,9 +133,10 @@ function wrapfun_iip(
 end
 
 # Opaque-p variant of the 3-arg wrapfun_iip: same matrix of (du, u, p, t)
-# variants as the regular FD wrapper, but with `OpaqueParams` substituted in
-# place of `T3` (the parameter slot). The user's `ff` is wrapped in
-# `OpaqueVoid(P, ff)` so the unpack-to-`P` happens before `ff` is invoked.
+# variants as the regular FD wrapper, but with the opaque container substituted
+# in place of `T3` (the parameter slot) — `OpaqueParams` for an `isbits` `P`,
+# `OpaqueRef` otherwise. The user's `ff` is wrapped in `OpaqueVoid(P, ff)` so
+# the unpack-to-`P` happens before `ff` is invoked.
 function wrapfun_iip_opaque(
         ff,
         ::Type{P},
@@ -152,7 +153,7 @@ function wrapfun_iip_opaque(
     dualT1_time = ArrayInterface.promote_eltype(T1, dualT_time)
     dualT4_time = dualgen(promote_type(T, T4))
 
-    O = RespecializeParams.OpaqueParams
+    O = RespecializeParams.opaque_container_type(P)
     return _make_fww(
         OpaqueVoid(P, ff),
         Tuple{T1, T2, O, T4},
--- a/lib/DiffEqBase/src/opaque_void.jl
+++ b/lib/DiffEqBase/src/opaque_void.jl
@@ -28,7 +28,10 @@ the wrapper signature carries it in the `p` slot.
 
 Excluded (kept as-is): `SciMLBase.NullParameters` (already a uniform singleton)
 and already-packed containers (idempotency, so a sensitivity adjoint that
-re-concretizes an opaque problem does not nest wrappers).
+re-concretizes an opaque problem does not nest wrappers). Problems carrying a
+symbolic system (`SciMLBase.has_sys(f)`, e.g. ModelingToolkit) are declined at
+the `promote_f` call site rather than here, since that policy needs `f`, not
+just `p`: their `p` must stay concrete for initialization and symbolic indexing.
 
 De-specializing makes `sol.prob.p` an opaque container, so **`AutoDePSpecialize`
 is a forward-solve latency tool, not for parameter-estimation workflows**:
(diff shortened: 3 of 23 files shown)


ChrisRackauckas-Claude:
Added precompilation of the non-isbits (`OpaqueRef`) path — without it the forward-solve speedup this PR enables was not actually being realized on first use.

The existing `AutoDePSpecialize` `@compile_workload` only exercised an isbits `p` (the `lorenz_p_params` NamedTuple → `OpaqueParams`). A non-isbits `p` routes through the distinct `OpaqueRef` wrapper signature, which nothing precompiled, so it compiled fresh on first solve. Added a `lorenz_pref` (`Vector` `p`) variant to the shared setup and each algorithm workload; one precompiled `OpaqueRef` solve is shared across all non-isbits parameter types.

Measured (fresh process, first `Tsit5` solve of a never-seen parameter struct):

| case | before | after |
|---|---|---|
| isbits `p`, FullSpecialize | 2.17s | — |
| isbits `p`, AutoDePSpecialize | 0.39s | 0.40s |
| non-isbits `p`, FullSpecialize | 2.13s | 2.16s |
| **non-isbits `p`, AutoDePSpecialize** | **2.40s** | **0.37s** |

So the non-isbits path now gets the same ~5.8× first-solve TTFX win as the isbits path. Patch-bumped the touched precompile-carrying sublibraries.

ChrisRackauckas (approved):

ChrisRackauckas (approved):

ChrisRackauckas (approved):

ChrisRackauckas-Claude:
Added an MTK-safety guard, because the non-isbits path had a real footgun here.

`MTKParameters` is non-isbits, so the OpaqueRef path swept it in. A ModelingToolkit problem opting into `AutoDePSpecialize` then **errored at solve** — not just symbolic indexing, the solve itself:

```
promote_u0_p(u0::Vector{Float64}, p::OpaqueRef, t0::Float64)   # no method
  in _late_binding_update_u0_p_impl   (MTK initialization)
```

Fix: guard both `promote_f` opaque branches with `!SciMLBase.has_sys(f)`. A function carrying a symbolic system falls back to the normal specialization — a pure no-op for MTK anyway (its params are non-isbits and already type-uniform, so no shared-compilation win exists).

Verified with a real MTK `ODEProblem` after the guard:

| | before guard | after guard |
|---|---|---|
| `prob.p` after opt-in | `OpaqueRef` | `MTKParameters` (declined) |
| `solve` | **errors** | `Success` |
| `sol[x]` | — | ✓ |
| `getp(sys, k)(sol)` | — | ✓ 2.0 |
| `sol(0.5; idxs=x)` | — | ✓ |

Plain non-isbits/isbits problems (no `sys`) still de-specialize exactly as before. Regression test uses a `SymbolCache` sys to reproduce `has_sys` without pulling in ModelingToolkit; DiffEqBase opaque-p suite now 30/30.

ChrisRackauckas-Claude:
Found and fixed a further precompilation gap, this time on the **default solver** — and it was the biggest one.

The workload only swept `solve(prob, solver)`, which covers `solve(prob, DefaultODEAlgorithm())` but **not** the no-algorithm `solve(prob)` entry. That entry runs default algorithm *selection*, a separate code path — and it is the one most user code and MTK tutorials actually write. So the opaque-p precompilation never reached it.

Measured (fresh process, isbits struct `p`, 3-state problem):

| entry point | FullSpecialize | AutoDePSpecialize before | AutoDePSpecialize after |
|---|---|---|---|
| `solve(prob, DefaultODEAlgorithm())` | 7.90s | 0.42s | 0.42s |
| **`solve(prob)`** (no algorithm) | 13.29s | **11.25s** | **0.45s** |

So the most common entry point went from a ~1.2× win to ~30×. Worth noting the polyalgorithm payoff is far larger than for a single method (`Tsit5`: 2.17s → 0.37s), since a uniform `p` slot shares compilation across every solver the polyalgorithm instantiates — more methods, more shared work.

Fix keeps the AutoDePSpecialize problems in their own list so they drive both the `solve(prob, solver)` sweep and a `solve(prob)` pass. Verified by precompiling (the workload solve runs during precompilation, so a clean build confirms it) and timing fresh-process first solves.

PR #4006, opened 2026-07-23
Fix/taylor series implicit

xbtmkmmm5:
## Checklist


lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
oscardssmith, line 154: should `η` be used here? If you look at the equivalent condition (see e.g. the OrdinaryDiffEQFIRK version of this loop).

lib/OrdinaryDiffEqTaylorSeriesImplicit/src/TaylorSeries_perform_step.jl
```diff
@@ -0,0 +1,204 @@
...
+    if is_taylor_pade(alg)
+        tc = t + dt
+        distance = dt
+    else
+        # expansion center
+        tc = t + μ * dt
+        distance = -μ * dt
+    end
+    local ndw
+    # reusing η is causing problems, temporarily initialize it to Inf
```
oscardssmith: is this comment still accurate?

xbtmkmmm5: I don't think so, I can delete this comment. 
I think this this doesn't come from the changes I made. It may come from an older version

PR #4011, opened 2026-07-24
NonlinearSolveAlg: zero-fill the inner Jacobian buffer instead of leaving it undefined

ChrisRackauckas-Claude:
## What

Build `NonlinearSolveAlg`'s inner `jac_prototype` as a zero-filled buffer (`similar` + `fill!`) instead of leaving it uninitialized.

## Why

`similar(::SparseMatrixCSC)` allocates `nzval` with `undef` — the stored values are undefined. That buffer is handed to the inner solver at construction and is **read before `WReuseJac` first fills it** (the linear solve inspects its values while setting up), so this is a genuine uninitialized read, not a theoretical one:

```julia
julia> A = sparse(1:512, 1:512, ones(512));
julia> nz = nonzeros(similar(A));
julia> count(!iszero, nz), maximum(abs, nz)
(473, NaN)
```

It only *tends* to look benign because freshly-mapped pages read as zero; once the allocator reuses dirty memory you get garbage, `NaN` included, and whatever the setup infers from those values becomes nondeterministic.

## Why `fill!` and not `zero`

`zero(::SparseMatrixCSC)` returns a matrix with **no stored entries**, which would discard the sparsity pattern the prototype exists to carry. `fill!(Z, 0)` keeps the pattern — and needs no dispatch, since it covers the dense case too. Starting from `similar` rather than `copy` skips a pointless O(nnz) value copy.

## Requires LinearSolve 5.2

An all-zero prototype is only safe once SciML/LinearSolve.jl#1110 is in. Before it, LinearSolve's `Auto` nonstructural-zeros heuristic read the unfilled buffer as "100% nonstructural zeros", activated the sparse reduction on an all-`false` mask, and then degraded to a per-solve `dropzeros` that dropped nothing:

| inner prototype | LinearSolve 5.1.0 | LinearSolve 5.2.0 |
|---|---|---|
| zero-filled | **11.80 MB/solve** | **0.42 MB/solve** |
| `NLNewton` (reference) | 0.40 MB/solve | 0.40 MB/solve |

(32-state sparse Brusselator, TRBDF2.) With 5.2 the activation decision defers to the first filled solve, so the prototype no longer misleads it. Compat floor raised to `LinearSolve = "5.2"` to make that explicit.

## Note on the earlier versions of this PR

This branch first carried a `nonstructural_zeros = None` workaround for the 30× sparse allocation gap, then `copy(Wr)`. Both were routes around the LinearSolve heuristic; #1110 fixes it at the source, so this PR is now just the safety fix, expressed the natural way.

## Testing

NSA suites pass: smooth_est 10/10, sparse 10/10, matrix-free 14/14, Jacobian-reuse 8/8. Runic clean.

Draft — please ignore until reviewed by @ChrisRackauckas.


lib/OrdinaryDiffEqNonlinearSolve/src/utils.jl
```diff
@@ -278,8 +278,13 @@ end
 function reuse_jac_kwargs(W)
     Wr = W isa WOperator && W.J !== nothing && !(W.J isa AbstractSciMLOperator) ?
         W._concrete_form : W
+    # Zero-filled, not `similar` alone: the buffer is handed to the inner solver at
+    # construction and is read before `WReuseJac` first fills it (the linear solve inspects
+    # its values while setting up), and `similar` leaves them undefined — in practice
+    # reachable garbage, `NaN` included. `fill!` keeps the stored sparsity pattern (unlike
+    # `zero`, which drops it) while making the contents defined.
```
ChrisRackauckas:
```suggestion
```

PR #4009, opened 2026-07-24
Raise Enzyme floor to 0.13.180 for AutoEnzyme Downgrade paths

ChrisRackauckas-Claude:
**This PR should be ignored until reviewed by @ChrisRackauckas.**

## Summary

**Compat floors only.** Raise `Enzyme` `0.13` → **`0.13.180`** in:

- `lib/OrdinaryDiffEqRosenbrock`
- `lib/OrdinaryDiffEqRosenbrockTableaus`
- `lib/OrdinaryDiffEqBDF`
- `test/AD`
- `test/Downstream`

Addresses AutoEnzyme order/step failures under Downgrade Sublibraries (#3987).

## Sibling PR (substantive, no floors)

https://github.com/SciML/OrdinaryDiffEq.jl/pull/4012 — Core SparseArrays extras, initdt `@testset` end, AMF Krylov reference path.

These two PRs are independent and can merge in either order.

## Not in this PR

- SciMLBase 3.39 (already on master)
- Test/project structure fixes (#4012)

PR #4001, opened 2026-07-23
Fix DAE automatic initdt sign on reversed tspan (DE #908)

ChrisRackauckas-Claude:
## Summary

- `ode_determine_initdt` for `AbstractDAEProblem` returned a always-positive dt.
- On reversed tspan (`tdir < 0`) this hit "Automatic dt setting has the wrong sign".
- Multiply the default initial dt by `tdir`. Explicit negative `dt` already worked.

Fixes SciML/DifferentialEquations.jl#908

This PR should be ignored until reviewed by @ChrisRackauckas.

## Test plan

- [x] Local: consistent linear DAE on tspan `(5,0)` succeeds with DImplicitEuler/DABDF2/DFBDF
- [x] Local: final state ≈ exp(0)=1
- [ ] CI: InterfaceI initdt tests

PR #4012, opened 2026-07-25
Downgrade: Core extras, initdt parse fix, AMF Krylov ref (no floors)

ChrisRackauckas-Claude:

## Summary

Substantive fixes for Downgrade CI **without** package version floor bumps.

| Change | Why |
|---|---|
| Core: `SparseArrays` in `[extras]` | Test target was weakdep-only; `Pkg.test` aborts under Downgrade |
| `ode_initdt_tests.jl`: missing `@testset` `end` | ParseError on InterfaceI after #4001/#4000 |
| AMF tests: `ROS34PW1a(linsolve=KrylovJL())` for dense operator-jac refs | SciMLOperators WOperator convert MethodError on AddedOperator/tensor-product graphs (upstream SciMLOperators#411); adjoint FD tol 0.01→0.015 |

## Sibling PR (floors only)

https://github.com/SciML/OrdinaryDiffEq.jl/pull/4009 — Enzyme `0.13` → `0.13.180`.

Independent of this PR; either order is fine.

### DIFF
--- a/lib/OrdinaryDiffEqAMF/test/test_adjoint_fd2d.jl
+++ b/lib/OrdinaryDiffEqAMF/test/test_adjoint_fd2d.jl
@@ -1,5 +1,6 @@
 using OrdinaryDiffEqRosenbrock
 using OrdinaryDiffEqAMF
+using LinearSolve
 using SciMLOperators
 using SciMLSensitivity
 using LinearAlgebra
@@ -122,7 +123,7 @@ end
     end
 
     # Baseline adjoint solve (no AMF)
-    adjoint_sol_baseline = solve(make_baseline_prob(), ROS34PW1a())
+    adjoint_sol_baseline = solve(make_baseline_prob(), ROS34PW1a(linsolve = KrylovJL()))
     @test adjoint_sol_baseline.retcode == SciMLBase.ReturnCode.Success
     dLdu0_baseline = adjoint_sol_baseline.u[end]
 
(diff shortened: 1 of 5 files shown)


ChrisRackauckas-Claude:
Split this into one PR per issue, since the three changes here fail independent CI jobs and are independently reviewable:

- **#4013** — the missing `@testset` `end` in `ode_initdt_tests.jl`. Verified: InterfaceI now passes on julia 1 / lts / pre and `Runic / Runic Format Check` goes green.
- **#4014** — `SparseArrays` in `[extras]` for OrdinaryDiffEqCore. Verified: `downgrade-sublibraries / test (lib/OrdinaryDiffEqCore)` now passes.

The third change here (AMF `linsolve = KrylovJL()` + the `0.01 → 0.015` adjoint tolerance) I did **not** carry over. Root cause is upstream and already has PRs: `WOperator`'s constructor leaves `_concrete_form` typed as an `AddedOperator` when `J` is one, so `convert(AbstractMatrix, W)` cannot assign the materialized matrix back — SciML/SciMLOperators.jl#410 / #411. Minimal repro, no OrdinaryDiffEq involved:

```julia
using SciMLOperators, LinearAlgebra   # v1.24.4
J = MatrixOperator(rand(4,4)) + MatrixOperator(rand(4,4))
W = SciMLOperators.WOperator{true}(I, 1.0, J, rand(4))
convert(AbstractMatrix, W)
# MethodError: Cannot `convert` an object of type Matrix{Float64}
#              to an object of type SciMLOperators.AddedOperator{...}
```

Once that ships, AMF should need only a SciMLOperators floor bump rather than a test-side workaround — worth avoiding the tolerance loosening if possible.

Also found while going through master: `.typos.toml` declares `hom` twice (lines 72 and 149, from #4004 and #4000), which is a TOML duplicate-key error, so `Spell Check with Typos` has been aborting before checking anything. Fixed in **#4016**. And the QA lane's `AutoDePSpecialize` rendered-docs failure is **#4015**.

Issue #3974, opened 2026-07-21
Rodas5P step size collapses (dt ≈ 2e-7, ~1M accepted steps) on a stiff mass-matrix ODE after OrdinaryDiffEqRosenbrock 2.4.2


weigao-123:
## Summary

Upgrading OrdinaryDiffEqRosenbrock 2.4.1 → 2.4.2 makes `Rodas5P` unable to complete a
previously-working integration of a stiff power-system DAE (semi-explicit index-1, lowered
to a mass-matrix `ODEProblem` with a singular diagonal mass matrix, plain
`Vector{Float64}` state, analytic sparse Jacobian). The run ends in `MaxIters`.

The failure is not a rejection loop: the stats show the controller **accepting**
microscopic steps and never growing dt.

```
retcode = MaxIters
naccept = 999960, nreject = 40, njacs = 999960, nf = 8000000
t_end ≈ 1.3477 of tspan (0.0, 3.0)      # average dt ≈ 2.5e-7 over the crawl
```

Raising `maxiters` to 100,000 does not help (it crawls further and still exhausts the
budget), which points at a systematically inflated error estimate rather than a stiffness
or stability problem.

## Isolation

- Pairwise version bisect: with OrdinaryDiffEqCore held at 4.7.1 and all other SciML
  dependencies at current releases, OrdinaryDiffEqRosenbrock **2.4.1 passes** and
  **2.4.2 fails**. Bounding Rosenbrock ≤ 2.4.1 restores the previous behavior with every
  other package current.
- Method differential on the identical problem under 2.4.2:
  - `Rodas4()` — success
  - `Rodas4P()` — success
  - `Rodas5P()` — dt collapse as above
- The problem has two `DiscreteCallback`-driven parameter discontinuities (a bus fault at
  t = 1.0, cleared at t = 1.1, both committed cleanly with `u_modified!`); the crawl
  begins after both, in smooth post-event dynamics, so callback handling itself does not
  appear to be the trigger.

## Suspected change

The only functional commit in 2.4.1 → 2.4.2 is #3935 ("Fuse RosenbrockCache stage
accumulation loops into single-sweep SIMD kernels"), which rewrites the stage, `b`,
`btilde`-error, and dense-output accumulations with a fused path capped at 8 terms.
Rodas5P has 8 stages — exactly at that cap — while the unaffected Rodas4/Rodas4P have
fewer. An error in the 8-term fused accumulation (or its boundary with the fallback
[...]

Issue #3721, opened 2026-06-06
Rosenbrock23 on Float32 mass-matrix problem: JacReuseState Float32/Float64 StaticWOperator convert MethodError


ChrisRackauckas-Claude:
Found while re-enabling DiffEqGPU.jl downgrade CI — but it reproduces on a **plain CPU reference solve** (no GPU/DiffEqGPU involved) at current/latest compat, so it's an OrdinaryDiffEq bug.

**Repro:** a Float32 mass-matrix ROBER problem solved with `Rosenbrock23()` (e.g. DiffEqGPU's `test/gpu_kernel_de/stiff_ode/gpu_ode_mass_matrix.jl` line ~38, the CPU reference `solve(prob, Rosenbrock23(), ...)`).

**Error:**
```
MethodError: Cannot `convert` an object of type
  SciMLOperators.StaticWOperator{true, SMatrix{3,3,Float64,9}, ...} to
  SciMLOperators.StaticWOperator{true, SMatrix{3,3,Float32,9}, ...}
 [1] setproperty!(x::OrdinaryDiffEqRosenbrock.JacReuseState{Float64, SMatrix{3,3,Float32,9}, ...}, f::Symbol, v::StaticWOperator{...Float64...})
 [2] calc_rosenbrock_differentiation(...)
     @ OrdinaryDiffEqDifferentiation .../derivative_utils.jl:922
```

**Diagnosis:** `JacReuseState` is parameterized with the **state** eltype (`Float32`) for its W-operator field, but `calc_rosenbrock_differentiation` builds the W-operator from the **Float64** `dtgamma` and assigns it via `setproperty!` → the `StaticWOperator{…Float64…}` can't convert to the field's `{…Float32…}` type. So a Float32 problem with a Float64-typed gamma path triggers the mismatch.

Likely fix: make `JacReuseState`'s W-operator field eltype follow the W/gamma computation (or convert the W-operator to the state eltype before assignment) so Float32 mass-matrix Rosenbrock solves work. Verified the latest registered `OrdinaryDiffEqRosenbrock` 2.3.0 still has it.

_Filed during the SciML downgrade-CI re-enablement pass; DiffEqGPU's downgrade can't be re-enabled until this is fixed._

PR #4023, opened 2026-07-26
Skip zero-weighted terms in fused Rosenbrock stage kernels (#3974)

niy-ati:
## Summary

- Fixes the actual defect behind #3974: `_madd_terms` (the shared kernel behind `_stage_accum!`, `_stage_rhs!`, `_weighted_sum!`, and `_dense_row!`) multiplied every stage through unconditionally, including exact-zero-weighted ones. For finite inputs `0 * x = 0` is harmless, but `0 * NaN` and `0 * Inf` are `NaN` — so a transient non-finite value in an *unrelated*, zero-weighted stage silently poisons the whole accumulation, even though the mathematically correct result never depends on that term.
- This matters concretely for Rodas5P: its `btilde` is the unit vector `[0,0,0,0,0,0,0,1]` (`Rodas5PTableau`), so the error estimate is supposed to depend on stage 8 alone. The pre-#3935 fallback explicitly skipped zero-weighted terms (`if !iszero(w[i])`); the fused path introduced by #3935 does not. An ill-conditioned early-stage solve (plausible right after a callback-driven parameter jump, on a stiff singular mass-matrix DAE) can produce a transient `NaN`/`Inf` in an early stage; because it's zero-weighted it should never reach the error estimate, but the fused path multiplies it through anyway.
- Fix: `_madd_terms` now skips a term outright when its coefficient is exact zero, matching the fallback's semantics exactly, for all four call sites (stage accumulation, stage RHS, the `b`/`btilde` weighted sum, and dense-output rows) via one shared fix point.

## What this replaces

An earlier version of this PR capped SIMD fusion at 5 prior stages and avoided the fused kernel on an aliased buffer. Neither held up under verification: direct numerical comparison between the fused and fallback paths at every `nks` from 1 to 8 (including the actual `btilde` weighted-sum path) showed only ~1e-16 relative differences (ordinary FMA rounding) uniformly across the board — no anomaly specific to `nks` 6/7, and the `out === base` call site is provably safe as coded (the buffer is `fill!`ed to zero immediately before use, so the "aliased" read is always a known zero). That fix also shipped a regression test that, when actually run against unpatched master, did not reproduce the bug — `Rodas4`/`Rodas5P` naccept counts were normal on both of its synthetic problems, confirmed independently. This PR replaces that diagnosis with the confirmed defect above.

## Test plan

- [x] New regression test in `lib/OrdinaryDiffEqRosenbrock/test/ode_rosenbrock_tests.jl`: directly exercises `_weighted_sum!` with Rodas5P's actual `btilde` and a `NaN`-poisoned zero-weighted stage; confirmed locally that this fails on unpatched code (`NaN` result) and passes with the fix (matches `_weighted_sum_fallback!` exactly, and the mathematically correct `ks[8]`). Also includes an end-to-end Rodas5P solve on a singular-mass-matrix ROBER problem as a sanity check.
- [x] Verified the fix introduces no regression on `nks` 1-8 in the general (non-zero-weight, finite) case: fused vs. fallback stay at the same ~1e-16 rounding-level difference as before the change.
- [x] Ran the related existing suites locally: issue #3631 (DAE dense output), #3721 (static-array mass-matrix OOP JacReuseState), #3962 (OOP SecondOrderODEProblem/ArrayPartition) — all pass unchanged.
- [ ] Full `OrdinaryDiffEqRosenbrock` package test suite in CI.


lib/OrdinaryDiffEqRosenbrock/test/ode_rosenbrock_tests.jl
```diff
@@ -558,3 +558,74 @@ end
...
+        sol5 = solve(
+            prob_i, Rodas5P();
+            abstol = tols[1], reltol = tols[2],
+            initializealg = BrownFullBasicInit(),
+            maxiters = 100_000,
+        )
+        @test SciMLBase.successful_retcode(sol4)
+        @test SciMLBase.successful_retcode(sol5)
+        # Must not crawl: collapse signature was ~1e6 accepts with MaxIters.
+        @test sol5.stats.naccept < 100_000
```
ChrisRackauckas: This test does not fail on master.

Issue #3648, opened 2026-05-18
DelayDiffEq Integrators+Interface tests fail on master after #3641 (SDIRK perform_step regression)


ChrisRackauckas-Claude:
## Summary

\`Sublibrary CI\` is failing on master for both \`DelayDiffEq_Integrators\` and \`DelayDiffEq_Interface\` since the merge of #3641 (\"Fix behavioral regressions in generic SDIRK/IMEX perform_step\", commit \`99187d02ed\`). The failures look like a genuine numerical regression in SDIRK2 / Cash4 / Hairer4 / Hairer42 stepping inside DelayDiffEq's \`MethodOfSteps\`: the SDIRK-based DDE solve terminates after far fewer steps than the scalar baseline (6 steps vs 129 in one case, 11 vs 36 in another), producing a \`DimensionMismatch\` when the test compares trajectories.

## Failing jobs (master, run 25988432548)

- \`test (DelayDiffEq_Integrators, 1, ubuntu-latest, ...)\` (job \`76390120601\`)
- \`test (DelayDiffEq_Interface, 1, ubuntu-latest, ...)\` (same run)

## Bisect

\`Sublibrary CI\` only runs DelayDiffEq tests when DelayDiffEq or its
deps are touched, so the test history is sparse, but the window is
unambiguous:

- #3621 (commit \`c953b8ce1f\`, \"Unify SFSDIRK4-8 and Hairer4/42 into generic PureSDIRK implementation\") -- DelayDiffEq_Integrators **passed** (job 76312934455).
- #3641 (commit \`99187d02ed\`, \"Fix behavioral regressions in generic SDIRK/IMEX perform_step\") -- DelayDiffEq_Integrators **failed** (job 76390120601; PR-run failure also at job 76373414211).

Between those two merges master only saw a version bump (\`92713e7bf0\`) and a SciMLLogging update (\`8ca41c6624\`), neither of which touches solver behavior. So #3641 introduced this.

## Reproduction (Integrators)

\`lib/DelayDiffEq/test/integrators/sdirk.jl\`, lines 30-40:

\`\`\`julia
sol_ip     = solve(prob_dde_constant_1delay_ip,     MethodOfSteps(SDIRK2(nlsolve=NLNewton(fast_convergence_cutoff=0))); dt=0.1)
[...]

PR #4026, opened 2026-07-26
Fix nlsolver.tmp/uprev aliasing on the implicit-first-stage IIP path

niy-ati:
## Summary

- On the IIP stage-1 solve, copy into the nlsolver scratch `tmp` instead of aliasing `uprev`. `get_tmp_cache` for Newton-based algorithms returns `cache.nlsolver.tmp` directly; DelayDiffEq's discontinuity tracking for state-dependent lags (`track_propagated_discontinuities!` / `discontinuity_function`) grabs that buffer via `get_tmp_cache(integrator)` and interpolates into it in place. With `nlsolver.tmp = uprev`, that interpolation silently corrupted `uprev`. `MethodOfSteps(ImplicitEuler())` with `dependent_lags` set drives `dt` below floating-point epsilon and aborts with `ReturnCode.Unstable` on master; it completes with `ReturnCode.Success` with this patch.
- Stage-1 `Predictor.Linear` seeding had been restricted to the ImplicitEuler tableau flavor (`E === :ie_dd2`), so other PureSDIRK/ESDIRK methods with `stage1_extrapolation` (e.g. SDIRK2, ImplicitMidpoint) always started Newton from zero regardless of the requested predictor. Restore Linear seeding for those tableaus; keep the IE/ABDF2 bootstrap (`ie_dd2 && !hasproperty(alg, :predictor)`) unchanged.
- Keep assigning `nlsolver.c = c[1]` every step so a previous stage (e.g. SDIRK2 `c₂ = 0`) cannot leave a stale abscissa for the next step's Newton solve.

Not a fix for #3648 — that was already resolved by #3690 (restoring the missing `nlsolver.c = c[1]` in the IIP path). This PR is a separate, still-open bug found while re-verifying #3648: the `tmp`/`uprev` aliasing above. Cash4 is unaffected by the seed-widening change (`stage1_extrapolation = false`), unlike what an earlier draft of this PR claimed.

## Test plan

- [x] New regression test: `lib/DelayDiffEq/test/regression/dde_sdirk_stage1_tmp_aliasing.jl` — `MethodOfSteps(ImplicitEuler())` + `dependent_lags`; asserts `nlsolver.tmp !== uprev` after a step and that `solve` returns `ReturnCode.Success`. Confirmed locally: fails on unpatched `generic_imex_perform_step.jl` (`nlsolver.tmp === uprev` is `true`, `solve` returns `ReturnCode.Unstable`), passes with the fix.
- [x] `lib/DelayDiffEq/test/integrators/sdirk.jl` — SDIRK2 / Cash4 / Hairer4 / Hairer42 IIP `sol.t` matches scalar
- [x] `lib/DelayDiffEq/test/interface/multi_algorithm.jl` — SDIRK2 Constant delays IIP/scalar grid parity
- [x] Bumped `OrdinaryDiffEqSDIRK` to 2.9.1 for the behavior fix


singhharsh1708:
diagnosis is right, that comment was wrong. SDIRK2/Hairer4/42/ImplicitMidpoint all have `explicit_first_stage = false` so they do reach that branch. the aliasing part came from my #3641 so this one is on us.

the `copyto!` is actually the bigger fix here, `nlsolver.tmp = uprev` leaks `uprev` through `get_tmp_cache` and DelayDiffEq's discontinuity tracking writes into that buffer. `MethodOfSteps(ImplicitEuler())` with `dependent_lags` fails outright on master and works with your patch. that case would make a good regression test since the current test changes are only comments.

one correction, #3648 was already fixed by #3690, both test files you touched pass on master, so drop the "Fixes #3648" and reframe around the aliasing + seed. Cash4 isn't affected btw (`stage1_extrapolation = false`). needs a version bump on OrdinaryDiffEqSDIRK too.

PR #4030, opened 2026-07-27
Route CI jobs to runner pools that exist and can hold them

ChrisRackauckas-Claude:

Infrastructure-only. No test or source code changes — only `test_groups.toml` runner/timeout selection. Three CI lanes have been failing for runner-capacity reasons, and the failures are not code regressions.

## Evidence

Surveyed the last 150 `Sublibrary CI` runs plus the recent `CI` / `Downgrade Sublibraries` runs via the Actions API, tabulating requested `runs-on` labels against the `runner_name` that actually picked the job up:

| requested labels | runner it lands on | count |
|---|---|---|
| `ubuntu-latest` | `self-hosted-4vcpu-8gb-dqgwf-*` | 1275 |
| `ubuntu-latest` | GitHub-hosted (4 vCPU / 16 GB) | 256 |
| `self-hosted,Linux,X64` | `self-hosted-8vcpu-16gb-hh54d-*` | 35 |
| `self-hosted,Linux,X64,gpu` | `…gpu-v100-8vcpu-16gb-*` | 68 |
| `self-hosted,Linux,X64,gpu` | `…gpu-t4-8vcpu-16gb-*` | 7 |
| **`self-hosted,Linux,X64,high-memory`** | **never assigned** | **31** |

### 1. `high-memory` is a label no runner carries

Every job requesting it queues until GitHub expires it — the scan shows 24 h (1440 min) queue times before cancellation. `lib/StochasticDiffEq [NoncommutativeConvergence]` is **not** `local_only`, so it is in nearly every affected-sublibrary matrix, and it hangs the whole `Sublibrary CI` run behind it. Because `SublibraryCI.yml` uses `cancel-in-progress: false` on master, each push then queues behind the hung run and the next push evicts it: **104 of the last 140 master `Sublibrary CI` runs are `cancelled`, only 9 `success`.** The master run created at `2026-07-27T07:29:25Z` had exactly one incomplete job at the time of writing — the `high-memory` one, queued 3.5 h.

Fix: point those groups at `["self-hosted", "Linux", "X64"]` — the 8 vCPU / 16 GB pool, the largest label set the fleet answers (35/35 clean assignments, never lands on the 8 GB or GPU pools). If a `high-memory` scale set is provisioned later, revert.

### 2. `lib/DiffEqDevTools [Core]` OOMs the 8 GB pool

Measured locally: 55:47 wall, **7.62 GiB peak RSS**. `ubuntu-latest` lands on a 4 vCPU / 8 GB self-hosted runner 83 % of the time. Perfect correlation over the last ~15 runs of `downgrade-sublibraries / test (lib/DiffEqDevTools)`:

- 12/12 runs on `self-hosted-4vcpu-8gb-*` → *"The self-hosted runner lost communication with the server"* (and no log is uploaded — the agent died).
- 4/4 runs on a GitHub-hosted runner (16 GB) → **pass**, 44–60 min.

Same story in `Sublibrary CI` on one run: `Julia lts` passed in 47 min while `Julia 1` and `Julia pre` died at 36 / 32 min — i.e. it is marginal at 8 GB and 1.12 uses slightly more than 1.10.

Fix: pin `[Core]` to the 16 GB pool. The peak is driven by the two `1e5`-trajectory ensemble `WorkPrecisionSet`s at the end of `analyticless_stochastic_wp.jl` (that testset alone: 1019 s, 249 GiB allocated). Deliberately **not** reducing `numtraj` — the test asserts `isapprox(err1, err2, atol = 1e-3)` on a Monte-Carlo weak error, so shrinking the sample would mean loosening that tolerance to hide the change.

### 3. The root `GPU` group's 60 min cap is below its runtime

The `gpu` label is answered by both a V100 and a T4 pool. Measuring identical warm-up blocks in both logs, the **T4 is 1.76× slower** (119.3 s vs 67.6 s; 40.9 s vs 23.3 s). The V100 finishes the whole job in 46–48 min, so the T4 needs ≈ 85 min — every T4-scheduled run was cancelled at exactly 60 min with *"The job has exceeded the maximum execution time of 1h0m0s"*.

Fix: `timeout = 120`.

Note for the record: on the V100 the GPU suite **does** complete and reports a genuine failure, so this is not purely infrastructure noise —

```
GPU: dae_tests.jl | 136 pass 1 fail 7 broken | 25m54.3s
Hairer4 [jac=CSC, mass=diag_cu]: Test Failed at test/gpu/dae_tests.jl:467
  Expression: r.status === :pass
   Evaluated: gpu_mismatch === pass
```

That numerics bug is out of scope here; raising the timeout just makes it reportable instead of being masked by cancellations.

## Still broken after this PR — needs a change in `SciML/.github`

`downgrade-sublibraries / test (lib/DiffEqDevTools)` is **not** fixed by this PR. `sublibrary-downgrade.yml` does not read `test_groups.toml`; it hardcodes the runner at [`.github/workflows/sublibrary-downgrade.yml:98`](https://github.com/SciML/.github/blob/v1/.github/workflows/sublibrary-downgrade.yml#L98):

```yaml
runs-on: ${{ (inputs.apt-packages != '' || inputs.container != '') && fromJSON('["ubuntu-24.04"]') || 'ubuntu-latest' }}
```

so every sublibrary's downgrade job goes to the 8 GB pool and `lib/DiffEqDevTools` keeps killing its runner. The minimal change there would be a `runner` input (JSON-encoded label list, defaulting to `'"ubuntu-latest"'`) threaded into `runs-on`, plus an optional per-project override map so this repo could send only `lib/DiffEqDevTools` to the 16 GB pool. Not touching that repo without sign-off since it is shared across all of SciML.

## Verification

Ran the shared matrix generator against this branch:

```
$ julia compute_affected_sublibraries.jl . --root-matrix     # GPU cell -> "timeout":120
$ git ls-files 'lib/*' | julia compute_affected_sublibraries.jl . --projects-matrix
total cells: 283
high-memory cells: 0
```

`lib/DiffEqDevTools [Core]` now emits `runner: ["self-hosted","Linux","X64"], timeout: 120` on all three Julia versions.

### DIFF
--- a/lib/DiffEqDevTools/test/test_groups.toml
+++ b/lib/DiffEqDevTools/test/test_groups.toml
@@ -1,5 +1,11 @@
 [Core]
 versions = ["lts", "1", "pre"]
+# Peaks at ~7.6 GiB RSS (the 1e5-trajectory ensemble WorkPrecisionSets in
+# analyticless_stochastic_wp.jl), which does not fit the 4 vCPU / 8 GB pool that
+# answers the default `ubuntu-latest`; the runner agent gets reclaimed mid-run
+# and the job reports "the self-hosted runner lost communication with the
+# server". Pin the 8 vCPU / 16 GB pool instead.
+runner = ["self-hosted", "Linux", "X64"]
 
 [QA]
 versions = ["lts", "1"]
--- a/lib/StochasticDiffEq/test/test_groups.toml
+++ b/lib/StochasticDiffEq/test/test_groups.toml
@@ -10,9 +10,14 @@ versions = ["lts", "1", "pre"]
 [AlgConvergence]
 versions = ["1"]
 
+# `["self-hosted", "Linux", "X64"]` is the largest pool the SciML fleet actually
+# answers (8 vCPU / 16 GB). No runner carries a `high-memory` label, so asking
+# for one parks the job in the queue until GitHub expires it at 24h -- and this
+# group is not local_only, so it is in nearly every Sublibrary CI matrix and
+# hangs the whole workflow run with it.
 [NoncommutativeConvergence]
 versions = ["1"]
-runner = ["self-hosted", "Linux", "X64", "high-memory"]
+runner = ["self-hosted", "Linux", "X64"]
 timeout = 240
 
 [AlgConvergence2]
@@ -33,13 +38,13 @@ local_only = true
 
 [OOPWeakConvergence]
 versions = ["1"]
-runner = ["self-hosted", "Linux", "X64", "high-memory"]
+runner = ["self-hosted", "Linux", "X64"]
 timeout = 300
 local_only = true
 
 [IIPWeakConvergence]
 versions = ["1"]
-runner = ["self-hosted", "Linux", "X64", "high-memory"]
+runner = ["self-hosted", "Linux", "X64"]
 timeout = 300
 local_only = true
 
--- a/lib/StochasticDiffEqROCK/test/test_groups.toml
+++ b/lib/StochasticDiffEqROCK/test/test_groups.toml
@@ -6,7 +6,7 @@ versions = ["lts", "1", "pre"]
 # compute_affected_sublibraries.jl.
 [SROCKC2WeakConvergence]
 versions = ["1"]
-runner = ["self-hosted", "Linux", "X64", "high-memory"]
+runner = ["self-hosted", "Linux", "X64"]
 timeout = 240
 local_only = true
 
--- a/lib/StochasticDiffEqWeak/test/test_groups.toml
+++ b/lib/StochasticDiffEqWeak/test/test_groups.toml
@@ -13,7 +13,7 @@ local_only = true
 
 [WeakConvergence2]
 versions = ["1"]
-runner = ["self-hosted", "Linux", "X64", "high-memory"]
+runner = ["self-hosted", "Linux", "X64"]
 timeout = 300
 local_only = true
 
@@ -23,7 +23,7 @@ local_only = true
 
 [WeakConvergence3]
 versions = ["1"]
-runner = ["self-hosted", "Linux", "X64", "high-memory"]
+runner = ["self-hosted", "Linux", "X64"]
 timeout = 300
 local_only = true
 
@@ -33,19 +33,19 @@ local_only = true
 
 [WeakConvergence4]
 versions = ["1"]
-runner = ["self-hosted", "Linux", "X64", "high-memory"]
+runner = ["self-hosted", "Linux", "X64"]
 timeout = 420
 local_only = true
 
 [WeakConvergence5]
 versions = ["1"]
-runner = ["self-hosted", "Linux", "X64", "high-memory"]
+runner = ["self-hosted", "Linux", "X64"]
 timeout = 300
 local_only = true
 
 [WeakConvergence6]
 versions = ["1"]
-runner = ["self-hosted", "Linux", "X64", "high-memory"]
+runner = ["self-hosted", "Linux", "X64"]
 timeout = 300
 local_only = true
 
--- a/test/test_groups.toml
+++ b/test/test_groups.toml
@@ -34,4 +34,8 @@ versions = ["lts"]
 [GPU]
 versions = ["1"]
 runner = ["self-hosted", "Linux", "X64", "gpu"]
-timeout = 60
+# The `gpu` label is answered by both a V100 and a T4 pool, and the T4 is ~1.76x
+# slower on this suite (measured on identical warmup blocks). The V100 finishes
+# in ~48 min, so the T4 needs ~85 min: a 60 min cap cancelled every T4-scheduled
+# run and hid whatever the tests actually reported.
+timeout = 120


ChrisRackauckas-Claude:
## Proposed `SciML/.github` change (NOT pushed — needs @ChrisRackauckas sign-off)

`downgrade-sublibraries / test (lib/DiffEqDevTools)` cannot be fixed from this repo. `sublibrary-downgrade.yml` does not read `test/test_groups.toml`, so a per-sublibrary `runner` is not expressible; the runner is fixed at [`.github/workflows/sublibrary-downgrade.yml:98`](https://github.com/SciML/.github/blob/v1/.github/workflows/sublibrary-downgrade.yml#L98).

### `.github/workflows/sublibrary-downgrade.yml`

Add three inputs after `container` (line 49):

```yaml
      runner:
        description: "JSON-encoded runs-on value (string or array of labels) for every sublibrary downgrade job. Empty = ubuntu-latest."
        default: ""
        required: false
        type: string
      heavy-projects:
        description: "Space-separated lib/* paths whose downgrade suite needs `heavy-runner` (memory- or time-heavy)."
        default: ""
        required: false
        type: string
      heavy-runner:
        description: "JSON-encoded runs-on value used for the projects named in `heavy-projects`."
        default: ""
        required: false
        type: string
```

and replace line 98:

```yaml
    runs-on: >-
      ${{ (inputs.apt-packages != '' || inputs.container != '') && fromJSON('["ubuntu-24.04"]')
       || (inputs.heavy-runner != '' && contains(format(' {0} ', inputs.heavy-projects), format(' {0} ', matrix.project)) && fromJSON(inputs.heavy-runner))
       || (inputs.runner != '' && fromJSON(inputs.runner))
       || 'ubuntu-latest' }}
```

(`runs-on` may reference `matrix`, so the per-project override resolves per cell. The `format(' {0} ', …)` wrapping makes `contains` a whole-token match so `lib/Foo` does not match `lib/FooBar`.)

The caller side in this repo then becomes, in `.github/workflows/DowngradeSublibraries.yml`:

```yaml
      heavy-projects: "lib/DiffEqDevTools"
      heavy-runner: '["self-hosted", "Linux", "X64"]'
```

Minimal alternative if the per-project axis is unwanted: add only `runner`, and let callers move *all* their downgrade jobs. Not preferred — this repo has ~50 downgrade jobs and only one is heavy, so that would park the whole fan-out on the 16 GB pool.

### Optional, wider: stop `ubuntu-latest` being ambiguous

The 4 vCPU / 8 GB self-hosted pool squats the `ubuntu-latest` label, so across the last 150 `Sublibrary CI` runs `ubuntu-latest` resolved to that pool 1275 times and to a GitHub-hosted 16 GB runner 256 times (83 % / 17 %) — non-deterministically, for **every** SciML repo. Any job needing more than ~7 GB is a coin flip, and when it loses, the failure surfaces as *"the self-hosted runner lost communication with the server"* with **no uploaded log**, which reads like flake rather than OOM. Either raise that pool to 16 GB or give it its own label instead of `ubuntu-latest`. That is an org/fleet change, not a workflow change.

ChrisRackauckas-Claude:
## Correction from the first CI run on this branch: 16 GB is not enough for the ex-`high-memory` groups

Both groups that previously requested `high-memory` and now land on the 8 vCPU / 16 GB pool **still die**, with the same signature and no uploaded log:

| job | runner | result | wall |
|---|---|---|---|
| `lib/StochasticDiffEq [NoncommutativeConvergence]` | `self-hosted-8vcpu-16gb-hh54d-runner-tcvwn` | *"the self-hosted runner lost communication with the server"* | 28 min |
| `lib/StochasticDiffEq [IIPWeakConvergence]` | `self-hosted-8vcpu-16gb-hh54d-runner-kjrwq` | same | — |

So `high-memory` was not a typo or a stale label — those suites genuinely need more than 16 GB, and there is no pool in the fleet that can hold them. **This PR does not make those two lanes green, and I am not claiming it does.**

What it does change is the failure *mode*, and that is the part worth having:

- **Before:** the job is never assigned, sits in the queue for 24 h, and holds the entire `Sublibrary CI` run open. Because `SublibraryCI.yml` sets `cancel-in-progress: false`, every subsequent master push queues behind it and is then evicted by the push after that — **104 of the last 140 master runs cancelled, 9 green**. No lane in the workflow reports anything.
- **After:** these two jobs fail in ~28 min and every other job in the run — 40+ sublibrary lanes — completes and reports normally.

Trading "everything is cancelled" for "two lanes are red" is the right trade, but it is a stopgap. **The actual fix is fleet-side: provision a runner scale set carrying the `high-memory` label**, at which point these three `test_groups.toml` files should be reverted to requesting it. I am measuring the real peak RSS of `NoncommutativeConvergence` locally to give a sizing number and will post it here.

If you would rather not have two red lanes in the interim, the alternative is to drop those groups from CI entirely until the hardware exists — but I have not done that, since `continue_on_error` / skipping would hide a real resource problem rather than surface it.

ChrisRackauckas-Claude:
## Verified: `lib/DiffEqDevTools [Core]` is fixed

All three Julia versions now pass on the 16 GB pool. Previously `Julia 1` and `Julia pre` died on the 8 GB pool with *"the self-hosted runner lost communication with the server"* and `Julia lts` only squeaked through.

| job | runner | before | after |
|---|---|---|---|
| `DiffEqDevTools / Julia 1 / Core` | `…8vcpu-16gb-hh54d-runner-gclt5` | died at 36 min, no log | **success, 41 min** |
| `DiffEqDevTools / Julia pre / Core` | `…8vcpu-16gb-hh54d-runner-v89xm` | died at 32 min, no log | **success, 42 min** |
| `DiffEqDevTools / Julia lts / Core` | `…8vcpu-16gb-hh54d-runner-crx96` | passed at 47 min (marginal) | **success, 39 min** |

Also faster across the board — 8 vCPU instead of 4, and no GC thrash against the memory ceiling.

Note this only covers the `Sublibrary CI` lane. The same suite in `Downgrade Sublibraries` still runs on the 8 GB pool and will keep dying until `sublibrary-downgrade.yml` gains a runner override — see the patch sketch above.

ChrisRackauckas-Claude:
## GPU lane: timeout change is in, and the underlying failure is now confirmed reproducible

The GPU job on this branch landed on the V100 pool again, so the raised cap was not directly exercised (V100 finishes well inside 60 min either way). What it did confirm is that the failure underneath the timeouts is real and stable:

```
tests / GPU => failure on self-hosted-gpu-v100-8vcpu-16gb-p842s-runner-txt78
11:23:37 -> 12:06:20  (43 min)

Hairer4 [jac=CSC, mass=diag_cu]: Test Failed at test/gpu/dae_tests.jl:467
  Expression: r.status === :pass
   Evaluated: gpu_mismatch === pass
GPU: dae_tests.jl | 136 pass | 1 fail | 7 broken | 27m41.2s
ERROR: Some tests did not pass: 160 passed, 1 failed, 0 errored, 11 broken.
```

Byte-for-byte the same failure as on master (run 30246332301, job 89914192614). So the "GPU DAE numerics question is unanswerable" framing was only true for T4-scheduled runs — on V100 the suite has been completing and reporting a genuine CPU/GPU mismatch all along, and the 60 min cap was hiding it on roughly 1 run in 10.

The timeout bump is still the right change: `dae_tests.jl` alone is 27m41s here, the T4 is ~1.76x slower on identical warmup blocks, and a T4-scheduled run therefore needs ~85 min. Under the old cap every such run was cancelled.

The numerics bug itself is out of scope for this PR and is being investigated separately.

ChrisRackauckas-Claude:
## Confirmed on this PR: the downgrade lane is untouched by these changes

```
downgrade-sublibraries / test (lib/DiffEqDevTools)
  => failure on self-hosted-4vcpu-8gb-dqgwf-runner-v8g8g
  11:19:20 -> 12:19:24
  "The self-hosted runner lost communication with the server."
  (no log uploaded — the agent died before it could)
```

As expected: `sublibrary-downgrade.yml` never consults `test_groups.toml`, so the `runner` pin added here has no effect on that lane and it went to the 8 GB pool again. This is the piece that needs the `SciML/.github` patch sketched earlier, and it is the only one of the three lanes this PR cannot address from inside this repo.

For contrast, the *same* suite in the `Sublibrary CI` lane, where the `runner` pin does apply, passed on all three Julia versions in this run.

ChrisRackauckas-Claude:
## GPU lane follow-up: the failure underneath the timeouts is now root-caused

Tracked separately in **https://github.com/SciML/OrdinaryDiffEq.jl/issues/4034** — first bad commit `56ffb5267cdd3b43cc14190a13b01bf075fffae5` ("Newton-Krylov integrators default to Hegedüs warm start", #3991), last good master `a09809af8`. Nothing to do with this PR; noting it here only because the 60 min cap is what kept it unreadable.

One correction to what I wrote above: I had assumed the failure was V100-specific because that was the only pool whose jobs survived to report. It is not — the mismatch values are identical to 3 s.f. across all 46 post-#3991 GPU jobs on **both** pools. The T4 runs were failing the same way; the cap just cancelled them before they could say so. That strengthens the case for the timeout bump rather than weakening it: under the old cap, roughly 1 GPU run in 10 produced no signal at all.

ChrisRackauckas-Claude:
## Sizing measurement: a `high-memory` runner may not be the answer after all

I measured `NoncommutativeConvergence` locally to work out how big a `high-memory` runner would need to be. The answer is that the question may be wrong — RSS climbs roughly linearly at ~0.5 GiB/min with no plateau, and I stopped it at **51 GiB** without it finishing:

| elapsed | RSS |
|---|---|
| ~25 min | 18 GiB |
| ~47 min | 28 GiB |
| ~63 min | 34 GiB |
| ~85 min | **51 GiB** (stopped here) |

Back-of-envelope, the convergence study retains ~90 MB of solution and noise data, so it is holding roughly 500x what it needs. Filed separately as **https://github.com/SciML/OrdinaryDiffEq.jl/issues/4036**.

The consequence for this PR: **do not size a `high-memory` runner off these two lanes yet.** If the growth is genuinely unbounded, provisioning a bigger box just moves the failure rather than fixing it, and the retention should be understood first. The rest of this PR stands on its own — the `DiffEqDevTools` and GPU-timeout changes are verified, and repointing the dead label still converts a 24 h workflow-wide hang into a fast, visible failure.

### Summary of what this PR does and does not do

| lane | before | after |
|---|---|---|
| `DiffEqDevTools [Core]` (Sublibrary CI) | died at 32-36 min, no log | **passes, 39-42 min** |
| `Sublibrary CI` as a whole | hung 24 h, 104/140 master runs cancelled | completes; all other lanes report |
| `NoncommutativeConvergence`, `IIPWeakConvergence` | never scheduled | fails in ~28 min — needs #4036 |
| root `GPU` group | cancelled at 60 min on T4-scheduled runs | 120 min cap; real failure now visible (#4034) |
| `downgrade (DiffEqDevTools)` | dies on 8 GB pool | **unchanged** — needs the `SciML/.github` patch above |

### One more thing worth deciding separately

Even with the hang gone, `SublibraryCI.yml` sets `cancel-in-progress: false` on master while master pushes arrive with a median gap of 57 min (52% land within 60 min) and a run takes ~50-60 min. So roughly half of master's runs will still be evicted while pending — and the one that survives is the *older* commit, since GitHub cancels the queued run rather than the running one. Flipping to `cancel-in-progress: true` would at least mean the newest commit is the one that gets tested. I have not made that change here since it is a policy call, not a bug fix.

PR #4035, opened 2026-07-27
Refresh FSAL derivatives after solve-level step limiters

svchb:
## Summary

- invalidate cached FSAL derivatives after every non-trivial solve-level `step_limiter`
- dispatch `reset_fsal!` on the algorithm cache so ETD2 can recompute its split derivative while preserving nonlinear history
- document attempted-step, adaptive-estimator, dense-output, and refresh semantics
- add in-place, out-of-place, adaptive TRBDF2/Tsit5, ETD2, callback, and RHS-accounting regressions
- bump OrdinaryDiffEqCore to 4.11.1 and OrdinaryDiffEqExponentialRK to 2.2.1

## Root cause

Solve-level step limiters run centrally after `perform_step!`. FSAL methods have already computed `fsallast` at the unprojected endpoint by then. The accepted state can therefore be projected while the next step starts with a derivative evaluated at a state that was never accepted.

Marking `reeval_fsal` restores the invariant `fsalfirst == f(uprev, p, tprev)` before the next step. ETD2 needs cache-specific handling because its FSAL object stores separate linear/current-nonlinear/previous-nonlinear values; the reset preserves `nlprev` and recomputes only the current split derivative.

## Semantics

The limiter remains outside the embedded adaptive error estimate. It is applied to each attempted endpoint before domain checking and acceptance callbacks. Dense interpolants between accepted endpoints are not projected. The default trivial limiter adds no work; a configured non-trivial limiter adds one RHS refresh between accepted steps, excluding the terminal endpoint where no later step consumes FSAL.

## Verification

- `test/Integrators_I/step_limiter_test.jl`: 263/263 passed
- ExponentialRK functional tests: 134 passed, 4 existing expected-broken
- ExponentialRK QA: 19/19 passed
- ExponentialRK JET: 1/1 passed
- ExponentialRK allocation lane: runtime-allocation regression passed, 17 existing expected-broken
- Core QA: 17/17 passed
- Core JET: 30 passed, 1 existing expected-broken
- Runic and `git diff --check`: passed
- post-merge FSAL probes for adaptive Tsit5, TRBDF2, and ETD2: passed
- numerical probe, adaptive Tsit5 at 1e-8 tolerance: endpoint error improved from 2.96e-7 to 5.87e-11; rejections from 16 to 0; RHS calls from 219 to 107


svchb:
All of the failures seem to be unrelated.

lib/OrdinaryDiffEqCore/src/solve.jl
```diff
@@ -101,7 +101,7 @@ function resolve_stage_step_limiters(alg, stage_limiter, step_limiter, verbose_s
                 "limiters (it does not opt into `OrdinaryDiffEqCore.has_stage_limiter`). This " *
                 "errors by default; set the `stage_limiter_unused` verbosity toggle to " *
                 "`WarnLevel()` or `Silent()` to allow it. Use `step_limiter` for a limiter " *
-                "applied once per accepted step, which every method supports.",
```
ChrisRackauckas, line 105: that seems like a bug if true? If the step is not accepted why put the limiter on it? It doesn't effect the error estimator.

svchb: You are right. I moved the solve-level step_limiter into the adaptive acceptance branch, so rejected attempts no longer invoke it. Fixed-step behavior remains once per accepted step, and FSAL derivatives are still refreshed after a limiter modifies an accepted endpoint.

ChrisRackauckas (approved):

svchb:
@ChrisRackauckas Thanks for your review. Can we get this merged?

ChrisRackauckas:
oh yes thank you for the reminder.

PR #4040, opened 2026-07-27
Default Newton-Krylov Auto warm starts to guarded Hegedüs

### LINKED ISSUE #4034: Hegedüs warm-start default (#3991) amplifies round-off ~1e13x on the Newton-Krylov path (GPU DAE test failure)
## Summary

`test/gpu/dae_tests.jl` fails deterministically on `master` for `Hairer4 [jac=CSC, mass=diag_cu]`:

```
Hairer4 [jac=CSC, mass=diag_cu]: Test Failed at test/gpu/dae_tests.jl:467
  Expression: r.status === :pass
   Evaluated: gpu_mismatch === pass
Summary: pass 136 | cpu_skip 7 | gpu_mismatch 1
```

The GPU/CPU gap is `gpu_abs = 3.62e-03`, `gpu_rel = 7.02e-03` against a gate of `atol = rtol = 1e-3`.

This is **not** a GPU bug, and not flaky hardware. It was introduced by #3991 (`56ffb5267` — *Newton-Krylov integrators default to Hegedüs warm start*), which made every Newton-path Krylov solve amplify last-ulp round-off by ~5 orders of magnitude. The GPU is merely a convenient source of last-ulp differences; the same amplification is reproducible on CPU alone, and I have done so (details below — **I have no GPU, so the GPU half is from CI logs only**).

## First bad commit

**`56ffb5267cdd3b43cc14190a13b01bf075fffae5`** — "Newton-Krylov integrators default to Hegedüs warm start (require LinearSolve 5.1)" (#3991).
Last good master: `a09809af8`.

Bisected through CI history rather than `git bisect` (no GPU available). Two independent lines of evidence pin it exactly:

**1. The GPU column flips at that commit.** The `Hairer4 / CSC` row across GPU CI jobs:

| date | commit | branch | job | cpu_abs | gpu_abs |
|---|---|---|---|---|---|
| 07-23 13:23 | `a09809af8` | master | [89224469104](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/30246332301) | 4.00e-03 | *(gpu_error era)* |
| 07-23 15:19 | `e7c538a19a` | `newton-krylov-warm-start-default` (PR #3991) | [89251799226](https://github.com/SciML/OrdinaryDiffEq.jl/actions/jobs/89251799226) | **4.41e-03** | **3.62e-03** |
| 07-23 15:22 | `56ffb5267c` | **master** | [89280544839](https://github.com/SciML/OrdinaryDiffEq.jl/actions/jobs/89280544839) | **4.41e-03** | **3.62e-03** |

The **CPU-only** column (`cpu_abs`, which needs no GPU) sat at exactly `4.00e-03` for every GPU job from 2026-07-08 through 2026-07-23 13:23 and changed to `4.41e-03` in the very first job containing #3991. The last era in which the GPU column was actually observable for CSC (2026-07-08 … 07-12, before an unrelated all-`gpu_error` outage) recorded `gpu_abs = 4.44e-09` — i.e. CPU and GPU used to agree to ~1e-9.

**2. Only the solvers #3991 touches regressed.** Comparing all 41 `jac=CSC` (Krylov) cases, last-good job `86678375428` vs current `89914192614`:

| solver | gpu_abs before | gpu_abs after | ratio |
|---|---|---|---|
| Hairer42 | 1.33e-15 | 1.38e-03 | 1.0e12 |
| Hairer4 | 4.44e-09 | 3.62e-03 | 8.2e05 |
| ImplicitEuler | 4.44e-16 | 2.46e-08 | 5.5e07 |
| Cash4 | 4.44e-16 | 2.27e-08 | 5.1e07 |
| Trapezoid | 1.33e-15 | 1.58e-08 | 1.2e07 |
| FBDF | 4.35e-14 | 2.56e-09 | 5.9e04 |
| SDIRK2 | 4.44e-16 | 1.80e-15 | 4 |
| *all 30 Rosenbrock/W cases (ROS\*, Rodas\*, GRK4A, Velds4, Scholz4_7, Ros4LStab, …)* | — | — | **~1 (unchanged)** |

That split is exactly the scope of #3991: the resolver is called only on the Newton nonlinear-solver path, and Rosenbrock/W-methods never reach it. Every SDIRK/BDF case lost 4-12 orders of CPU/GPU agreement; not one Rosenbrock case moved.

It is fully deterministic, and **not** V100-specific: the numbers `3.62e-03 / 7.02e-03` are identical to three significant figures across all 46 post-#3991 GPU jobs, on both the V100 and T4 runner pools, over four days.

## Mechanism, reproduced on CPU with no GPU

`build_nlsolver` now routes the Newton-path linear solver through `default_krylov_warm_start`, mapping `WarmStart.Auto → WarmStart.Hegedus`. That makes each Newton linear solve a function not only of `(W, b)` but of the previous solve's `cache.u`, closing a feedback loop across the whole integration.

With LinearSolve 5.2.0 on Julia 1.12, running the exact CPU problem from `test/gpu/dae_tests.jl` and perturbing `u0[1]` by **one ulp** (2.22e-16):

```
=== 1-ulp perturbation of u0[1] -> max change in the solution ===
                                                Hegedus (default)   None (pre-#3991)
  Hairer4                                          3.73e-03            6.64e-08
  Hairer42                                         5.67e-04            1.33e-15
  Cash4                                            4.62e-06            1.33e-15
  ImplicitEuler                                    3.11e-09            1.33e-15
  FBDF                                             1.51e-08            7.71e-09
```

A 2.2e-16 input perturbation moves the Hairer4 solution by **3.7e-03** — an amplification of ~1.7e13, and a quantitative match for the 3.62e-03 CPU-vs-GPU gap CI reports. Under the old cold start the same perturbation moves it by 6.6e-08. The solver ordering also matches the CI table above.

Setting `warm_start = LinearSolve.WarmStart.None` reproduces the pre-#3991 CI value exactly:

```
default (Auto -> Hegedus, post-#3991)   Hairer4 cpu_abs=4.17e-03
warm_start=None (pre-#3991 behaviour)   Hairer4 cpu_abs=4.00e-03   <- matches CI's pre-#3991 4.00e-03
```

Two aggravating factors in the current call path:

1. **The warm-start vector is ill-conditioned exactly where it is used.** Hegedüs picks `ξ = ⟨Ax₀,b⟩ / ⟨Ax₀,Ax₀⟩`. As Newton converges, `dz → 0`, so both inner products go to zero and the *relative* error of `ξ` blows up. The later Newton iterations — the ones that set the converged answer and, through `calculate_residuals!`/`ndz`, the SDIRK error estimate and the accepted step size — are precisely the ones getting a numerically ill-conditioned `x₀`.

2. **`linsolve.u` is aliased to `dz`, and `dz` is rescaled in place after every solve.** In `newton.jl`, `dolinsolve(...; linu = _vec(dz))` aliases the LinearSolve cache's solution to `dz`; the code then does `rmul!(dz, 2 / (1 + γdt / W_γdt))` and `relax!(dz, nlsolver, integrator, f)`. So the "previous solution" the warm start reuses is not the previous linear solution but a relaxation-damped rescaling of it, and the line-search damping factor is fed back into the linear solver's starting point. Hegedüs' optimal scalar absorbs most of the rescaling, which is presumably why this has not shown up as an outright wrong answer — but `WarmStart.Previous` would be actively incorrect on this path, and the aliasing is undocumented.

The tolerance dependence is consistent with this reading (`Hairer4`, 1-ulp perturbation):

```
  reltol=1e-03  Hegedus: 3.73e-03    None: 6.64e-08
  reltol=1e-06  Hegedus: 3.32e-04    None: 3.27e-04
  reltol=1e-10  Hegedus: 2.79e-06    None: 3.27e-07
```

At the default `reltol=1e-3` the integration takes ~21-24 steps and the cold-start path is essentially round-off-reproducible; the warm start destroys that. At tighter tolerances the step count grows and ordinary chaotic amplification dominates either way, so the warm start buys nothing there.

## This is not a tolerance artifact

Filing this rather than widening the gate, per the "default assumption is a real bug" rule. The `1e-3` gate was met with 6-7 orders of margin (`4.44e-09`) before #3991 and is now exceeded; `Hairer42` sits at `1.38e-03`, already over `atol` and passing only because the combined `atol + rtol*|b|` criterion rescues it at large-`|b|` sample points. Widening the tolerance would hide a genuine ~10^13 round-off amplification factor on the Newton-Krylov path that affects every user of `KrylovJL_GMRES()` with an SDIRK/BDF integrator, GPU or not.

## The warm start is not worthless — it is a real tradeoff

It does what it claims performance-wise (default `reltol`, same problem):

| solver | nf (Hegedüs) | nf (None) | change |
|---|---|---|---|
| Cash4 | 775 | 1127 | −31% |
| Hairer42 | 664 | 865 | −23% |
| Hairer4 | 670 | 842 | −20% |
| FBDF | 160 | 191 | −16% |
| ImplicitEuler / SDIRK2 / Trapezoid | — | — | ~0% |

So a blanket revert gives up something real. Possible directions, in rough order of appeal:

1. Give LinearSolve its own warm-start buffer instead of aliasing `dz`, so the relaxed-Newton `rmul!`/`relax!` scaling is not fed back into the linear solver.
2. Fall back to a cold start when the Hegedüs scalar is ill-conditioned (`‖Ax₀‖` small relative to `‖b‖`) — that is the converged-Newton regime, where the warm start saves little anyway.
3. Make the Hegedüs default opt-in rather than on-by-default until 1 and 2 are in place.

## Reproduction

No GPU is needed for the CPU half:

```julia
using OrdinaryDiffEqSDIRK, OrdinaryDiffEqNonlinearSolve, LinearAlgebra, SparseArrays
import LinearSolve

P  = [-1 0 0 0; 1 -0.5 0 0; 1 1 -1 0; -1 1 0 -1]
MM = Diagonal([1, 1, 0, 0])
mkp(u0) = ODEProblem(
    u0, (0.0, 5.0), P; initializealg = BrownFullBasicInit())

u0 = [1.0, 1.0, 0.5, 0.5]
u0p = copy(u0); u0p[1] = nextfloat(u0p[1])   # 1 ulp

for ws in (LinearSolve.WarmStart.Hegedus, LinearSolve.WarmStart.None)
    ls = KrylovJL_GMRES(warm_start = ws)
    a = solve(mkp(u0),  Hairer4(linsolve = ls); maxiters = 10_000)
    b = solve(mkp(u0p), Hairer4(linsolve = ls); maxiters = 10_000)
    @show ws, maximum(maximum(abs.(a(t) .- b(t))) for t in 0.0:0.1:5.0)
end
# Hegedus => 3.73e-03      None => 6.64e-08
```

The GPU half is `GROUP=GPU` on `test/gpu/dae_tests.jl`; CI evidence: run [30246332301](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/30246332301) job [89914192614](https://github.com/SciML/OrdinaryDiffEq.jl/actions/jobs/89914192614) (V100) and run [30260663037](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/30260663037) job [89959708848](https://github.com/SciML/OrdinaryDiffEq.jl/actions/jobs/89959708848).

Local environment for the CPU reproduction: Julia 1.12.4, LinearSolve v5.2.0, `master` at `b9b3d50fa`.

ChrisRackauckas-Claude:

Fixes #4034. **Scope changed after SciML/LinearSolve.jl#1123 merged — please re-read.**

## The bug

#3991 resolved `WarmStart.Auto` to `WarmStart.Hegedus` on the Newton path. The
guess there is the previous Newton *increment*, which converges to zero by
construction — so once Newton converges the guess is pure round-off, while the
Hegedüs scalar `ξ = ⟨Au,b⟩/⟨Au,Au⟩` is ill-conditioned exactly where it is
applied (`ξ` measured over `3.9e-32 … 5.1e+31`; `|cos∠(Au,b)|` down to 2e-16).
GMRES then returns the rescaled noise verbatim at iteration 0 on ~37% of solves,
and it feeds the SDIRK error estimate and step-size control.

Result: a 1-ulp input change moved the Hairer4 solution by 3.7e-03.

## What changed since the first revision

The first revision gated the default behind `isdefined(LinearSolve,
:HEGEDUS_MIN_RESIDUAL_REDUCTION)`. SciML/LinearSolve.jl#1123 has since merged and
released in **5.8.0**, but named the constant `_HEGEDUS_MAX_RESIDUAL_RATIO` —
internal, no `public` declaration. That gate would therefore have been `false`
forever, silently disabling the warm start for good.

Rather than probe someone else's internal, this now states the requirement where
it belongs — a `[compat]` floor:

```toml
LinearSolve = "5.8"
```

`default_krylov_warm_start` resolves `Auto → Hegedus` unconditionally again; the
resolver, not a runtime check, guarantees the guard is present.

## Verification (LinearSolve 5.8.0, the released guard)

1-ulp perturbation of the #4034 DAE, exercising the `Auto` default this PR
resolves — no explicit `warm_start`:

| | max solution change |
|---|---|
| **Auto (this branch)** | **1.33e-15** |
| unguarded Hegedüs (master) | 3.73e-03 |
| cold start | 6.64e-08 |

Better than a cold start, because the guard rejects precisely the guesses that
were injecting noise.

## Performance

The guarded default is also the fastest of the three. 1D Brusselator,
`nsolve` identical within each pair:

| | cold | unguarded | **guarded** |
|---|---|---|---|
| KenCarp4, N=400 | 202,145 | 175,852 | **158,259** (−22%) |
| TRBDF2, N=800 | 273,379 | 396,253 | **288,297** (+5%) |

At identical accuracy — KenCarp4 final-state error against a 1e-13 reference is
3.5894789585e-7 cold vs 3.5894791672e-7 guarded, so the iteration savings are
real work saved, not a weaker stopping test.

TRBDF2's +5% is within noise of a cold start rather than a win; worth knowing
which workloads Hegedüs actually helps, but that is a tuning question now, not a
correctness one.

## Also tried and rejected

Warm-starting only the first Newton iteration of each step (cold-starting the
rest) — the obvious fix if the problem is only converged increments. It reduced
the amplification just 3.73e-03 → 1.69e-03, still ~25,000× worse than cold,
because the first solve of each step inherits the *previous step's* converged
increment. Not carried.

https://claude.ai/code/session_01TPHRh64BLfoXSzQ3AxKNwC


ChrisRackauckas-Claude:
Full local test results (all run after the change, on Julia 1.12.6):

| suite | result |
|---|---|
| `lib/OrdinaryDiffEqDifferentiation` `Pkg.test()` | exit 0 — incl. `Krylov warm_start default 7/7` |
| `lib/OrdinaryDiffEqSDIRK` `Pkg.test()` | exit 0 — Tableau 57/57, Stage predictors 115/115, Convergence 102 pass / 2 pre-existing broken, DAE 8/8 |
| `lib/OrdinaryDiffEqBDF` `Pkg.test()` | exit 0 |
| `lib/OrdinaryDiffEqNonlinearSolve` `Pkg.test()` | exit 0 |
| umbrella `GROUP=Regression_I` | exit 0 — Dense 4104/4104, Special Interp 30/30, Inplace 2/2, Adaptive 46/46, Hard DAE 36/36, **Newton-Krylov Round-off 6/6** |
| LinearSolve `GROUP=Core` (against the patched LinearSolve) | exit 0 — `Krylov warm start 40/40` |

The new regression test was additionally run standalone under both configurations it has to
survive — released LinearSolve 5.2.0 (cold start) and the locally patched LinearSolve
(guarded Hegedüs) — 3/3 for each of `Hairer4` and `Hairer42` in both.

ChrisRackauckas-Claude:
The branch has been rewritten on current clean `master` (`938dae56bf`) after the full comparison. The earlier self-rearming/guard-detection design is superseded.

Final decision:

- keep `WarmStart.Auto` cold in OrdinaryDiffEq;
- pass explicit `None` / `Previous` / `Hegedus` choices through unchanged;
- make explicit Hegedüs safer separately in SciML/LinearSolve.jl#1123.

The reason is performance, not only robustness: on the matrix-free `n=800` Brusselator, guarded Hegedüs still regressed versus cold by +12.7% `nf` / +4.4% time (KenCarp4), +30.0% / +39.3% (TRBDF2), and +7.5% / +11.3% (FBDF). It remains useful on some correlated systems, but is not a sound automatic default.

Exact final local verification on Julia 1.12.6:

| group / check | result |
|---|---|
| umbrella `GROUP=Regression_I` | passed; new one-ulp test 6/6 |
| `OrdinaryDiffEqNonlinearSolve_Core` | passed; Homotopy 25/25, Homotopy init/default 16/16 |
| `OrdinaryDiffEqNonlinearSolve_QA` | passed; JET + Aqua |
| `OrdinaryDiffEqDifferentiation_Core` | passed |
| repository-wide Runic 1.7 `--check` | passed |
| companion LinearSolve Core + QA | passed, including Allocation QA 48/48 and SupernodalLU Allocation QA 8/8 |

The clean-master standalone regression was also run and failed at Hairer4 drift `0.003734637491723536`, while the final cold-Auto branch passed the `<1e-5` regression. An exact explicit guarded-Hegedüs rerun against #1123 produced Hairer4 drift `1.33e-15` (`nf=617`), Hairer42 `1.33e-15` (`nf=625`), and Cash4 `2.77e-9` (`nf=740`).

ChrisRackauckas-Claude:
CI note: the immediate documentation, downstream, and ImplicitDiscreteSolve downgrade failures are the clean-`master` NonlinearSolveBase compatibility conflict fixed independently in #4051. The failing resolver requires `NonlinearSolveBase >=2.40` from ImplicitDiscreteSolve while OrdinaryDiffEqNonlinearSolve currently caps the locally available range at 2.38. These setup failures occur before this PR's tests run and are unrelated to the warm-start diff; #4040 remains intentionally unstacked on the small prerequisite.

Issue #3175, opened 2026-03-18
[SDE] EM solver accepts incompatible tstop behind current time


ChrisRackauckas:
**Migrated from:** https://github.com/SciML/StochasticDiffEq.jl/issues/413

The EM solver accepts a tstop that is behind the current time via `add_tstop!` without error, then produces incorrect behavior. `LambdaEM` correctly rejects this. The non-dtchangeable solvers should validate tstops.

**MWE (confirmed still broken on StochasticDiffEq v6.100.0):**

```julia
using StochasticDiffEq

f(u, p, t) = 1.0
g(u, p, t) = 0.1
p = SDEProblem(f, g, [1.0], (0.0, 1.0))
i = init(p, EM(); dt=0.02)
add_tstop!(i, 0.01)
step!(i)  # Should reject, but doesn't
```

ChrisRackauckas-Claude:
The error this issue was migrated for no longer reproduces on master (`86fb5301`).

StochasticDiffEq.jl#413 reported:

> `Something went wrong. Integrator stepped past tstops but the algorithm was dtchangeable. Please report this error.`

Running the MWE across 8 variants × `EM`/`LambaEM` — float and integer `tspan`, scalar and vector `u0`, rational `dt`, tstop on and off the step grid, tstop equal to `dt`, tstop at `1e-9` — all 16 combinations land on the tstop and finish with `ReturnCode.Success`:

```
=== EM
Float u0, Float tspan          → first stop t = 0.01,   end t = 1.0, Success
Float u0, Int tspan            → first stop t = 0.01,   end t = 1.0, Success
scalar Float u0                → first stop t = 0.01,   end t = 1.0, Success
tstop not on grid, mid-solve   → first stop t = 0.017,  end t = 1.0, Success
tstop just below dt            → first stop t = 0.0199, end t = 1.0, Success
tstop far below dt             → first stop t = 1.0e-9, end t = 1.0, Success
rational dt                    → first stop t = 0.01,   end t = 1.0, Success
tstop == dt exactly            → first stop t = 0.02,   end t = 1.0, Success
```

(identical results for `LambaEM`.)

The validation asked for in the original thread is also in place now — `add_tstop!` with a tstop genuinely behind the current time throws `Tried to add a tstop that is behind the current time. This is strictly forbidden` for both `EM` and `LambaEM`.

Two leftovers, both distinct from what this issue describes:

[...]

PR #4041, opened 2026-07-27
Resync SDE noise when late add_tstop! shortens dt

niy-ati:
## Summary
- After `modify_dt_for_tstops!`, resample the pending noise increment when `W.dt` no longer matches the integrator `dt` (late `add_tstop!` after `init` on fixed-step EM; #3175 / StochasticDiffEq#413).
- When tstops remain, keep the tstop-shortened `dt` instead of letting RSWM expand it via `W.dt`.
- Add EM/LambaEM regression coverage and a past-time `add_tstop!` throw check in StochasticDiffEq `tstops_tests.jl`.

## Test plan
- [ ] `lib/StochasticDiffEq/test/tstops_tests.jl` (new #3175 cases)
- [ ] MWE: `init(SDEProblem(...), EM(); dt=0.02)`, `add_tstop!(i, 0.01)`, `step!(i)` lands at `t == 0.01`
- [ ] Same MWE with `LambaEM()`
- [ ] Past-time `add_tstop!` still throws

### DIFF
--- a/docs/src/devtools/internals/public_api.md
+++ b/docs/src/devtools/internals/public_api.md
@@ -295,6 +295,7 @@ Used by SDE/RODE solver sublibraries; no-ops for pure ODEs.
 ```@docs
 OrdinaryDiffEqCore.accept_noise!
 OrdinaryDiffEqCore.reject_noise!
+OrdinaryDiffEqCore.resync_noise!
 OrdinaryDiffEqCore.save_noise!
 OrdinaryDiffEqCore.reinit_noise!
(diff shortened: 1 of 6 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
thanks for this. It was almost right, but it needed to use a bit more of the RSWM internals to make sure the increment size change gets propagated through the bridge distribution. I finished it up in https://github.com/SciML/OrdinaryDiffEq.jl/pull/4095

PR #3985, opened 2026-07-22
Reuse homotopy caches across implicit ODE stages

ChrisRackauckas-Claude:
Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Dependencies and scope

SciML/NonlinearSolve.jl#1100 is merged and released in NonlinearSolve 4.24 / NonlinearSolveBase 2.38. It provides reusable `HomotopySweep` and `KantorovichHomotopy` `init`/`reinit!`/`solve!` caches through public APIs.

The related ImplicitDiscreteSolve cache work, SciML/OrdinaryDiffEq.jl#4042, and its registered NonlinearSolveBase 2.38 compatibility correction, SciML/OrdinaryDiffEq.jl#4055, are merged.

This PR only changes implicit ODE stage solves through `OrdinaryDiffEqNonlinearSolve.HomotopyNonlinearSolveAlg`.

## What this changes

`HomotopyNonlinearSolveAlg` previously rebuilt a `HomotopyProblem` and used one-shot `solve` for every implicit stage. This PR:

- initializes one continuation cache for `HomotopySweep` and `KantorovichHomotopy`;
- refreshes the stage initial guess, parameter tuple (including stage time), and tolerances with public `reinit!`, then calls public `solve!`;
- rebuilds the continuation cache after integrator resizing;
- retains one-shot `solve` for homotopy algorithms without this reusable cache path, including arc-length continuation;
- imports the continuation algorithms from their declaring public module, `NonlinearSolveBase`;
- documents the cached and fallback behavior.

`IDSolve` should not be replaced by this wrapper. Its controller is a timestep/path-continuation policy that uses the previous accepted state and every Newton contraction ratio, whereas `HomotopyNonlinearSolveAlg` solves an individual implicit ODE stage equation. The merged IDS path now has its own reusable `init`/`reinit!`/`solve!` implementation.

## Solver comparison

On the exact 51-solve IDS recurrence benchmark, using a fresh environment pinned to registered SciMLBase 3.39.1, NonlinearSolve 4.24, and NonlinearSolveBase 2.38, two independent 20-sample runs measured:

- cached IDS: 0.09914-0.10570 ms, 192 bytes, 561 residual calls;
- cached Kantorovich: 0.27951-0.29414 ms, 32 bytes, 2,040 residual calls;
- cached Sweep: 0.30429-0.30484 ms, 32 bytes, 2,346 residual calls.

[...]

PR #3908, opened 2026-07-15
Fix strict Kantorovich step acceptance

ChrisRackauckas-Claude:
> **Please ignore this draft until it has been reviewed by @ChrisRackauckas.**

## Summary

- Fix strict `KantorovichTypeController` acceptance to accept a converged step only when every recorded contraction satisfies `Θk <= Θreject`.
- Keep non-strict mode accepting every converged nonlinear solve.
- Add boundary, rejection, empty-history, and non-strict regression coverage.
- Bump `ImplicitDiscreteSolve` from 2.1.3 to 2.1.4.

## Root cause

The controller documentation says strict mode rejects a step whenever any `Θk > Θreject`. The implementation used `all(Θreject .< Θks)`, which reversed the inequality: normal well-contracted steps were rejected, while a history in which every contraction exceeded the rejection threshold was accepted.

## Why this remains separate from homotopy solving

[NonlinearSolve #1071](https://github.com/SciML/NonlinearSolve.jl/pull/1071) now provides the controller as `KantorovichHomotopy` for a genuine `HomotopyProblem`, and OrdinaryDiffEq's existing `HomotopyNonlinearSolveAlg` adapter can use it for ODE stage equations. That does not make it a drop-in replacement for this controller:

- `HomotopyNonlinearSolveAlg` is an `OrdinaryDiffEqCore.AbstractNLSolverAlgorithm` consumed by the ODE stage-solver machinery, not a NonlinearSolve algorithm accepted by `init(::NonlinearProblem, ...)`. An exact attempted substitution in `IDSolve.nlsolve` exits with `SciMLBase.NonSolverError`.
- A general `ImplicitDiscreteProblem` solves `f(u_next, uprev, p, t) = 0` and replaces `uprev` after every accepted state. It is therefore a sequence of changing nonlinear problems, not one fixed homotopy in `t`.
- An artificial per-step affine homotopy can preserve the endpoint equation for square problems, but it changes the solve strategy, is not a generic construction for the existing rectangular least-squares cases, and does not match current performance.

## Performance check

The closest generic square-problem prototype used

```math
H(x, λ) = (1 - λ)(x - u_{prev}) + λ f(x, u_{prev}, p, t)
```

at every discrete step. The recurrence endpoint matched the current solver to about `1e-12`. Medians below are from 20 warmed Julia 1.12.6 samples:

| Recurrence implementation | Time | Residual calls | Memory | Allocations |
|---|---:|---:|---:|---:|
[...]

PR #1100, opened 2020-04-03
Added Spijker Ferracina 4 stage SDIRK method

kamalojasv181:
Screenshot from 2020-04-03 05-52-07
The convergence plots look fine. Have a look at one of them. The otest value is around 3.96 for this 4th order alg. Please review this PR @ChrisRackauckas @kanav99

PR #4042, opened 2026-07-27
Reuse nonlinear caches for ImplicitDiscreteSolve reinitialization

ChrisRackauckas-Claude:
> [!IMPORTANT]
> Ignore this draft until reviewed by @ChrisRackauckas.

## Summary

- Replace ImplicitDiscreteSolve's hand-unrolled loop over private NonlinearSolve cache fields with the public, allocation-free `NonlinearSolveBase.solve_cache!` driver.
- Preserve the existing per-Newton residual-contraction history through a reusable step observer.
- Reuse the integrator's nonlinear cache during initial-condition solving and `reinit!`, including rectangular nonlinear least-squares problems.
- Pass the accepted constant extrapolant explicitly as the nonlinear initial guess so a rejected or externally modified iterate cannot leak into the next step.
- Run ImplicitDiscrete initialization before `save_start` during integrator reinitialization and preserve `InitialFailure` instead of resetting it afterward.

## Dependencies

This draft depends on [NonlinearSolve.jl #1100](https://github.com/SciML/NonlinearSolve.jl/pull/1100), specifically `NonlinearSolveBase.solve_cache!` in NonlinearSolveBase 2.40. The tested dependency head was `dca80d83c2d74483d1308c70f4315b69a9cbdf85`.

QA also exposed an independent SciMLTesting clean-master harness regression. It was bisected and fixed in [SciMLTesting.jl #38](https://github.com/SciML/SciMLTesting.jl/pull/38); both QA runs below used its exact head `0afa642162a91f8e4df88803690664ded5c52daf`. This is not a runtime dependency of the patch.

The related reusable homotopy-cache integration is [OrdinaryDiffEq.jl #3985](https://github.com/SciML/OrdinaryDiffEq.jl/pull/3985). The final design/benchmark follow-up is also recorded on the merged controller PR [#3908](https://github.com/SciML/OrdinaryDiffEq.jl/pull/3908).

## Root cause

The ImplicitDiscrete initialization added in #2624/#2626 only ran on initial integrator construction. Generic OrdinaryDiffEq `reinit!` saved the raw `u0` before the ImplicitDiscrete consistency solve, and its later retcode reset could erase an initialization failure.

ImplicitDiscreteSolve also rebuilt and solved a fresh `NonlinearProblem` during every initialization. Its time-step path reused a cache, but drove it by reading and mutating private NonlinearSolve fields and helpers. Reinitializing only the parameters left the previous nonlinear iterate in place instead of applying the documented constant extrapolant.

This patch keeps the native IDS Newton/Kantorovich controller rather than replacing it with an artificial per-step `HomotopyProblem`. It shares the new reusable NonlinearSolve cache-driving path with the homotopy solvers while retaining the substantially faster native recurrence solve measured below.

## Correctness and allocation coverage

The new tests cover:

- fresh solve versus `reinit!` + reused solve, with `save_start=true` and `false`;
- nonlinear-cache identity across reinitialization;
- rectangular nonlinear least-squares cache reuse;
- initialization failure and subsequent recovery;
- stale nonlinear iterates being replaced by the accepted constant extrapolant;
- exact zero steady-state allocations for cached `step!` and public integrator `reinit!` on Julia 1.11+.

A separate 20-sample warmed probe on Julia 1.12.6 produced:

- `step!`: first measurement-site warmup 144 B, then 0 B for all 19 samples;
- `reinit!` with a preallocated `u0`: first measurement-site warmup 144 B, then 0 B for all 19 samples;
- `solve!`: 80 B in every sample because it constructs/returns the solution wrapper.

The official regression asserts the two cache operations that are expected to be allocation-free; it does not disguise the solution-construction allocation.

## Performance comparison

Twenty warmed Julia 1.12.6 samples, using `AutoFiniteDiff` consistently and solving the same 50-step implicit recurrence:

| Path | Median time | Bytes | Residual calls | Endpoint |
|---|---:|---:|---:|---|
| fresh IDS | 0.1925435 ms | 8,896 | 565 | `(1.5531006208104776, 0.4090814367409633)` |
| cached IDS | 0.103924 ms | 192 | 561 | `(1.5531006208104776, 0.4090814367409633)` |
| one-shot Kantorovich | 3.3344115 ms | 901,104 | 2,244 | `(1.5531006208114124, 0.4090814367406754)` |
| cached Kantorovich | 0.280863 ms | 32 | 2,040 | `(1.5531006208114124, 0.4090814367406754)` |
| one-shot HomotopySweep | 3.339966 ms | 891,312 | 2,550 | `(1.5531006237647473, 0.4090814358834119)` |
| cached HomotopySweep | 0.294148 ms | 32 | 2,346 | `(1.5531006237647473, 0.4090814358834119)` |

Cache reuse makes IDS 1.85x faster and removes 8,704 B (97.84%) while preserving the endpoint exactly and using four fewer residual evaluations. Cached IDS is 2.70x faster than cached Kantorovich and 2.83x faster than cached HomotopySweep on this recurrence. Between the two homotopy methods, cached Kantorovich is 4.52% faster and uses 13.04% fewer residual calls than Sweep. The Kantorovich and Sweep endpoint deltas from IDS are approximately `(9.35e-13, -2.88e-13)` and `(2.95e-9, -8.58e-10)`, respectively.

These numbers support retaining the IDS-specific solve strategy; they are workload-specific rather than a general ranking of homotopy algorithms.

## Local validation

All commands ran from the exact upstream `master` base `fd8861edd500f8bba76a4f2a865e304d10881fe8` on Julia 1.12:

- `ODEDIFFEQ_TEST_GROUP=Core ... Pkg.test()` for ImplicitDiscreteSolve: 53/53 passed, including cached allocation 2/2, reinitialization 15/15, constant extrapolant 2/2, NLLS 3/3, null state 3/3, and initialization failure 3/3.
- `GROUP=Integrators_I ... Pkg.test()` at the repository root: all active checks passed, including reinitialization 31/31 and integrator interface 36/36. The two existing marked-broken checks remained marked broken.
- ImplicitDiscreteSolve QA against SciMLTesting #38: JET 1/1 and Aqua 20/20 passed.
- OrdinaryDiffEqCore QA against SciMLTesting #38: AllocCheck 1/1, JET 30 passed with one pre-existing marked-broken check, and Aqua 19/19 passed.
- Repository-wide Runic `--inplace` and `--check`: exit 0.
- `git diff --check`: exit 0.

No test was skipped, silenced, deleted, loosened, or converted to a warning.

## Process notes

The work followed the request to evaluate whether IDS should become a homotopy adapter, then narrowed the actual issue to reusable `init`/cache solving. It traced the initialization history, benchmarked fresh and cached IDS against both current homotopy solvers, added the required public cache-driver API upstream instead of depending on private NonlinearSolve internals, implemented the downstream reuse path, reproduced the QA failure on clean master, and validated the independently bisected SciMLTesting fix before publishing this draft.

### DIFF
--- a/lib/ImplicitDiscreteSolve/Project.toml
+++ b/lib/ImplicitDiscreteSolve/Project.toml
@@ -1,7 +1,7 @@
 name = "ImplicitDiscreteSolve"
 uuid = "3263718b-31ed-49cf-8a0f-35a466e8af96"
 authors = ["vyudu <vincent.duyuan@gmail.com>"]
-version = "2.1.4"
+version = "2.1.5"
 
 [deps]
 CommonSolve = "38540f10-b2f7-11e9-35d8-d573e4eb0ff2"
@@ -29,7 +29,7 @@ SciMLTesting = "2.1"
 CommonSolve = "0.2.6"
 Pkg = "1"
 Test = "1.10.0"
-NonlinearSolveBase = "2.31.0"
+NonlinearSolveBase = "2.40"
 OrdinaryDiffEqSDIRK = "2"
 SafeTestsets = "0.1.0"
 SciMLBase = "3.39"
--- a/lib/ImplicitDiscreteSolve/src/ImplicitDiscreteSolve.jl
+++ b/lib/ImplicitDiscreteSolve/src/ImplicitDiscreteSolve.jl
@@ -6,9 +6,10 @@ using SciMLBase: SciMLBase, ImplicitDiscreteProblem, NonlinearFunction,
 import SciMLBase: alg_order, isadaptive
 using NonlinearSolveBase: NonlinearSolveBase
 using NonlinearSolveFirstOrder: NewtonRaphson
-using CommonSolve: init, solve, step!
+using CommonSolve: init
 using DiffEqBase: DefaultInit
 import DiffEqBase: initialize!
+using SymbolicIndexingInterface: state_values
 using OrdinaryDiffEqCore: OrdinaryDiffEqCore, OrdinaryDiffEqAlgorithm,
     OrdinaryDiffEqMutableCache, AbstractController, AbstractControllerCache
 import OrdinaryDiffEqCore: alg_cache, get_fsalfirstlast, isfsal, perform_step!,
--- a/lib/ImplicitDiscreteSolve/src/cache.jl
+++ b/lib/ImplicitDiscreteSolve/src/cache.jl
@@ -4,12 +4,37 @@ struct ImplicitDiscreteState{uType, pType, tType}
     t::tType
 end
 
-mutable struct IDSolveCache{uType, cType, thetaType} <: OrdinaryDiffEqMutableCache
+mutable struct IDSolveStepObserver{T, V}
+    residualnormprev::T
+    Θks::V
+end
+
+function (observer::IDSolveStepObserver)(u, fu, iteration)
+    residualnorm = NonlinearSolveBase.L2_NORM(fu)
+    if iteration > 1
+        if observer.residualnormprev ≈ zero(observer.residualnormprev)
+            push!(observer.Θks, zero(eltype(observer.Θks)))
+        else
+            push!(observer.Θks, residualnorm / observer.residualnormprev)
+        end
+    end
+    observer.residualnormprev = residualnorm
+    return nothing
+end
+
+function reset!(observer::IDSolveStepObserver)
+    empty!(observer.Θks)
+    observer.residualnormprev = zero(observer.residualnormprev)
+    return nothing
+end
+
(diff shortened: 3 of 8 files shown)


ChrisRackauckas-Claude:
Release metadata follow-up: the reusable cache API was ultimately registered as NonlinearSolveBase 2.38.0 after the unreleased version sequence was normalized, while this merged PR still left `ImplicitDiscreteSolve` at the nonexistent `2.40` compat floor. The focused one-line correction is draft PR #4051: https://github.com/SciML/OrdinaryDiffEq.jl/pull/4051. Local functional tests pass 53/53, QA passes JET 1/1 and Aqua 20/20, and whole-repository Runic passes against the registered 2.38.0 release.

Issue #4054, opened 2026-07-29
ImplicitDiscreteSolve pins NonlinearSolveBase 2.40, which is not registered — Pkg.test() fails to resolve monorepo-wide


ChrisRackauckas-Claude:
See analysis: lib/ImplicitDiscreteSolve/Project.toml requires NonlinearSolveBase 2.40 but only 2.38.0 is registered. Bisected to 7eea915d0 (#4042).

ChrisRackauckas-Claude:
Fixed on master by https://github.com/SciML/OrdinaryDiffEq.jl/pull/4055 — `lib/ImplicitDiscreteSolve/Project.toml` now has `NonlinearSolveBase = "2.38"`, which is a registered version (2.38.0 is still the newest in General).

Confirmed the bound is not merely lowered but correct: the API #4042 needs is present in 2.38.0.

```
version: 2.38.0
solve_cache!  true
L2_NORM       true
  solve_cache!(cache::NonlinearSolveBase.AbstractNonlinearSolveCache; step_observer) @ NonlinearSolveBase
```

Both names `ImplicitDiscreteSolve` reaches for — `NonlinearSolveBase.solve_cache!` with the `step_observer` keyword in `src/solve.jl`, and `NonlinearSolveBase.L2_NORM` in `src/cache.jl` — resolve against 2.38.0 with the exact signature used. So nothing was given up by not waiting on a 2.40 release.

Closing.

**`Pkg.test()` is still blocked, but by a different conflict**, so anyone landing here from that symptom wants https://github.com/SciML/OrdinaryDiffEq.jl/pull/4066 rather than this issue. On master `528f8214f`, `lib/StochasticDiffEq` now fails with:

```
ERROR: Unsatisfiable requirements detected for package NonlinearSolveBase [be0214bd]:
 ├─possible versions are: 1.0.0 - 2.38.0 or uninstalled
 ├─restricted to versions 2.35.0 - 2 by OrdinaryDiffEqNonlinearSolve, leaving only versions: 2.35.0 - 2.38.0
 └─restricted by compatibility requirements with SciMLBase [0bca4576] to versions: uninstalled — no versions left
   └─SciMLBase [0bca4576] log:
     ├─restricted to versions 3.39.0 - 3 by StochasticDiffEqLowOrder, leaving only versions: 3.39.0 - 3.40.0
[...]

Issue #4037, opened 2026-07-27
Weak convergence test groups retain millions of full solution objects (~11+ GiB peak)


ChrisRackauckas-Claude:
Split out of SciML/OrdinaryDiffEq.jl#4036, where `IIPWeakConvergence` was assumed to share a root cause with `NoncommutativeConvergence`. It does not — that one is unbounded Lévy-area retention and no runner size fixes it, while this one is bounded, linear in `trajectories`, and *would* fit on a larger box. Filing separately so the two do not get one shared workaround.

## The problem

`DiffEqDevTools.test_convergence` keeps every ensemble solution object alive for the whole study:

https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/DiffEqDevTools/src/convergence.jl#L90-L114

```julia
_solutions = Array{Any}(undef, length(dts))
for i in 1:length(dts)
    sol = solve(ensemble_prob, alg, ensemblealg; dt = dts[i], …, trajectories = Int(trajectories), …)
    _solutions[i] = sol
end
…
solutions = [SciMLBase.calculate_ensemble_errors(sim; …) for sim in _solutions]
```

For a weak-order estimate only `sol.errors` / the final values are needed, but a full `RODESolution` is retained per trajectory — each carrying its own `NoiseProcess` (with `RSWM`, two `ResettableStack`s and an `Xoshiro`), the problem, the algorithm and its `InterpolationData`. `save_everystep = false` trims the saved timeseries but none of that.

The weak groups run this at Monte-Carlo trajectory counts:

| file | largest calls |
|---|---|
| `weak_convergence/iip_weak.jl` | 2 × `trajectories = 5e5`, 1 × `1e5` |
| `weak_convergence/oop_weak.jl` | 1 × `1e6`, 1 × `7e5`, 1 × `2e5`, 1 × `1e5` |

each over 5 `dts`, so 2.5 M and 5 M live solution objects respectively.

## Measurement

`weak_convergence/iip_weak.jl`'s `SimplifiedEM` case, reduced to 20 000 trajectories (Julia 1.12.6, `JULIA_NUM_THREADS=1`, `master @ 0f84b10`):

```
summarysize(sim)                     = 220.5 MiB      (5 dts × 20 000 = 100 000 solutions)
summarysize(one RODESolution)        = 2368 B
raw EnsembleSolution, 20 000 traj    = 55.6 MiB       → 2917 B/trajectory
```

[...]

PR #4056, opened 2026-07-29
Summarise the Monte-Carlo test_convergence ensembles by default (breaking)

ChrisRackauckas-Claude:
> **Please ignore this PR until it has been reviewed by @ChrisRackauckas.**

**BREAKING: DiffEqDevTools 3.1.5 → 4.0.0.**

## The problem

`test_convergence` kept the full `EnsembleSolution` for **every** step size in `_solutions`, reducing them to error summaries only after the loop:

```julia
for i in 1:length(dts)
    sol = solve(ensemble_prob, alg, ensemblealg; dt = dts[i], ...)
    _solutions[i] = sol                       # <- all of them stay alive
end
solutions = [SciMLBase.calculate_ensemble_errors(sim; ...) for sim in _solutions]
```

So the peak is the *sum* over step sizes rather than the largest. At 5e5–1e6 trajectories over 5–8 step sizes that is tens of GiB of solutions retained only to compute a mean, and it is what put `OOPWeakConvergence` and `IIPWeakConvergence` past a 16 GB runner.

## The change

Each step size's ensemble is reduced as soon as it is solved and stripped to one representative trajectory, so only one is alive at a time. **This is now the default.** `retain_solutions = true` restores the old behaviour for callers that want the paths themselves.

## Measured

`JULIA_NUM_THREADS=8`, before/after on non-overlapping checkouts:

| | peak before | peak after | | wall |
|---|---|---|---|---|
| `iip_weak` | 25.10 GiB | **7.93 GiB** | 3.2× | 7:29 → 6:35 |
| `oop_weak` | 22.93 GiB | **7.30 GiB** | 3.1× | 7:33 → 7:00 |

Both now fit the standard self-hosted pool, so the `high-memory` label comes off those two groups — no runner in the fleet carries it, so they have been queueing for 24 h and expiring rather than running (https://github.com/SciML/OrdinaryDiffEq.jl/issues/4030).

## Why this is breaking

`ConvergenceSimulation.solutions` is documented, and the docs said:

> `solutions::Array{<:DESolution}`: Holds **all** the PdeSolutions.

which is exactly the promise being changed. `Base.getindex(sim, i, I...) = sim.solutions[i][I]` forwards into it, so `sim[i][j]` for `j > 1` now errors on the Monte-Carlo method. The field documentation is updated to match, and the compat pin moves `"3"` → `"4"` in 45 `Project.toml` files.

Scope of the break, audited rather than assumed:

- Only the Monte-Carlo method changes. The ODE/DAE methods of `test_convergence` do not take this keyword and are untouched — including the `sim[i].u[2]` usage in `symplectic_convergence.jl`, which is a `DynamicalODEProblem`.
- The one place in this repo that indexes ensemble trajectories is `PL1WM.jl:234` and `:274`, comparing two algorithms path by path. It goes through the `expected_value` path, which takes a reducing `output_func`, builds no `EnsembleTestSolution`, and so never reduces. Unaffected.

## That the statistics are unchanged is checked, not argued

A 20000-trajectory SROCK2 study run both ways:

```
retain_solutions=true   retained= 0.3957 GiB   O_weak_final=1.8376692858
retain_solutions=false  retained= 0.0008 GiB   O_weak_final=1.8376692858

𝒪est values bitwise identical:   true
errors values bitwise identical: true
memory ratio: 489.1x smaller
```

`test/retain_solutions_tests.jl` covers statistics preservation across `𝒪est`, `errors`, `weak_errors`, `error_means` and the per-trajectory `errors` vectors; pins the new default; checks the trajectory count actually drops; and checks the `expected_value` path is left alone. 39 assertions, all passing.

## What this does *not* fix

**`SROCKC2WeakConvergence` is not helped.** Measured with the reduction in place it still peaks at **34.15 GiB**, because its cost is a single 3e6-trajectory ensemble rather than retention across step sizes. Its `high-memory` label stays. It needs either a lower trajectory count or the reducing-`output_func` pattern the StochasticDiffEqWeak studies already use (8 bytes/trajectory instead of KiB).

**That group also fails its assertions.** `𝒪est[:weak_final]` came out 1.567 and 1.563 against `abs(𝒪 - 2) < 0.4`, i.e. just under the 1.6 floor, at both 3e6 and 1e6 trajectories. This has gone unseen because the group has never once been scheduled in the CI history I looked at. Fixing its memory would turn "never runs" into "runs and fails", so it wants its own investigation into whether the tolerance or the method is at fault.

## Caveat on how this was measured

`Pkg.test()` for `lib/StochasticDiffEq` does not resolve on current master — `ImplicitDiscreteSolve` 2.1.5 requires `NonlinearSolveBase` ≥ 2.40.0 while the registry tops out at 2.38.0 — so the two weak files were run directly in a purpose-built environment rather than through the group harness. Same shape as the `NonlinearSolve` 4.24.0 floor that was breaking the GPU jobs, and it is why the `lib/StochasticDiffEq` and `lib/DiffEqDevTools` groups are currently red on master.

---

### Links

- Related: https://github.com/SciML/OrdinaryDiffEq.jl/issues/4030, https://github.com/SciML/OrdinaryDiffEq.jl/issues/4036
- Memory survey these numbers come from: https://github.com/SciML/OrdinaryDiffEq.jl/issues/4036#issuecomment-5092118693
- Earlier merged pieces: https://github.com/SciML/OrdinaryDiffEq.jl/pull/4050, https://github.com/SciML/OrdinaryDiffEq.jl/pull/4053

https://claude.ai/code/session_013WW2jGeoWZVZu7AiscUNSz


ChrisRackauckas-Claude:
Heads-up: I opened https://github.com/SciML/OrdinaryDiffEq.jl/pull/4060 for the same problem before seeing this. Two sessions converged on #4037 independently. They should not both merge — flagging the differences so one can be picked rather than reconciled later.

**Where the reduction happens.** This PR reduces each ensemble *after* `solve` returns, which bounds the sum across step sizes but leaves the peak at one full ensemble. #4060 reduces *during* the solve, via an `output_func` on the `EnsembleProblem`, so the full solutions never coexist at all and the peak is bounded too.

That difference is exactly what this PR identifies as its own limitation:

> **`SROCKC2WeakConvergence` is not helped.** Measured with the reduction in place it still peaks at **34.15 GiB**, because its cost is a single 3e6-trajectory ensemble rather than retention across step sizes. Its `high-memory` label stays. It needs either a lower trajectory count or the reducing-`output_func` pattern the StochasticDiffEqWeak studies already use.

#4060 is that `output_func` pattern. Measured in a 16 GiB cgroup at `JULIA_NUM_THREADS=2` (the self-hosted `num-threads: auto` default):

| file | on `master` | with #4060 |
|---|---|---|
| `weak_srockc2.jl` | OOM 15.61 GiB @ 3m29s | **exit 0**, 24m31s, **4.69 GiB** |
| `oop_weak.jl` | OOM 15.62 GiB @ 23m50s | **exit 0**, 12m48s, **1.58 GiB** |
| `additive_weak.jl` | (same group) | **exit 0**, 3m10s, **1.03 GiB** |
| `iip_weak.jl` | OOM 15.61 GiB @ 21m24s | runs fully, 10m13s, **2.22 GiB** |

so the `high-memory` label comes off all eight groups rather than two.

**Breaking vs not.** This PR changes the default and moves DiffEqDevTools to 4.0.0, touching the compat pin in 45 `Project.toml` files. #4060 adds an opt-in `save_solutions` keyword, leaves the default alone, and is 3.1.5 → 3.2.0 with no compat churn. The tradeoff is real in both directions — reduce-by-default protects future callers from the footgun, opt-in avoids a major bump — worth a deliberate choice.

**One measurement to double-check.** This PR reports `SROCKC2WeakConvergence` failing its assertions with `𝒪est[:weak_final]` of 1.567 and 1.563 against `abs(𝒪 - 2) < 0.4`. My run of `weak_srockc2.jl` **passed** (exit 0). The likely difference is thread count: these numbers were taken at `JULIA_NUM_THREADS=8`, while CI resolves `auto` to 2 on self-hosted runners and I ran at 2. `EnsembleThreads` chunks trajectories per thread, so the thread count changes which noise realisation each trajectory gets. If that is the explanation, the assertion is thread-count sensitive, which is worth knowing independently of either PR.

I did hit the same `NonlinearSolveBase` 2.40 resolver breakage; filed as https://github.com/SciML/OrdinaryDiffEq.jl/issues/4054 with a bisect to https://github.com/SciML/OrdinaryDiffEq.jl/pull/4042.

ChrisRackauckas-Claude:
Closing in favour of https://github.com/SciML/OrdinaryDiffEq.jl/pull/4060, which solves the same issue by reducing each trajectory **during** the solve through the ensemble's `output_func` rather than stripping the `EnsembleTestSolution` afterwards. That bounds the peak rather than only the sum across step sizes, which is what this PR identified as its own remaining gap:

> **`SROCKC2WeakConvergence` is not helped.** Measured with the reduction in place it still peaks at **34.15 GiB** … It needs either a lower trajectory count or the reducing-`output_func` pattern the StochasticDiffEqWeak studies already use.

#4060 takes `weak_srockc2.jl` to 4.69 GiB and exit 0, so the `high-memory` label comes off all eight groups.

Two things from here are being carried over rather than lost:

- **`retain_solutions` and the reduce-by-default decision.** #4060 originally used an opt-in keyword to stay non-breaking. The argument here — that the footgun should be closed for every caller, not just the ones who remember a keyword — is the better one, so #4060 now adopts this PR's name, its `false` default, and the 4.0.0 bump with the compat pin moved across the dependent `Project.toml` files.
- **`drop_trajectories`.** Reducing at `output_func` bounds the peak but still leaves one small object per trajectory in the returned `ConvergenceSimulation`; stripping to a single representative after the errors are computed removes that too. #4060 now does both.

Also worth not losing: this PR's observation that `SROCKC2WeakConvergence` fails its own assertions (`𝒪est[:weak_final]` of 1.567 and 1.563 against `abs(𝒪 - 2) < 0.4`). My run of `weak_srockc2.jl` passed, and the likely difference is thread count — these were taken at `JULIA_NUM_THREADS=8` while CI resolves `auto` to 2 on self-hosted runners. If that is the explanation the assertion is thread-count sensitive, which is worth its own issue independent of either PR.

Thanks — the analysis here was sound and the parts of it that were better than mine are in #4060.

PR #4064, opened 2026-07-29
Run Runic on the disco-optimizations files

ChrisRackauckas-Claude:
Please ignore until reviewed by @ChrisRackauckas.

## Motivation

The DISCO optimization merge (#3720) left three files that Runic rejects, so
`format-check` fails on current `master` and on every PR based on it.

The aggregate whole-tree failure begins at `408a7a30fe`. Per-file bisects place
`disco.jl` and `disco_tests.jl` at that commit and
`integrators/controllers.jl` at `80d7e74c82`; the relevant files at parent
`c538565b25` all pass.

## What this does

Runs Runic on exactly the three affected files:

- `lib/OrdinaryDiffEqCore/src/disco.jl`
- `lib/OrdinaryDiffEqCore/src/integrators/controllers.jl`
- `test/Integrators_I/disco_tests.jl`

The diff contains only Runic's whitespace, line-wrapping, numeric-literal, and
explicit-return normalization. There is no behavioral or public-API change.
Formatting exact clean master produces the same 38-insertion/34-deletion tree
as this PR.

## Local validation

At rebased head `5c05be4605` on upstream `b22e2e58b2`:

- Runic 1.7.0, whole tree: pass (`--check .`, exit 0)
- `git diff --check`: pass
- `git range-diff` against the prior PR commit: exact (`=`)
- official `GROUP=Integrators_I` on the coherent `#4064 + #4066` validation
  stack: Step Limiter 251/251, DISCO 28/28, package tests passed in 2882.56 s

The #4066 compatibility commit is validation-only context and is not included
in this branch. An independent `#3985 + #4066 + #4064` validation also passed
homotopy tests 38/38, default nonlinear-solve tests 16/16, whole-tree Runic,
and `git diff --check`.

## Hosted CI

At the one-hour terminal audit, hosted Runic, Runic suggestions, and Typos were
green. The remaining 241 failures all stopped before tests: 127 sublibrary
setup jobs, 44 root setup jobs, 12 downstream resolvers, documentation, and 57
downgrade resolvers. Representative setup logs report the separate SciMLBase
3.39/3.40 boundary fixed by #4066; the downgrade jobs stop inside their
resolver. No hosted failure is in this formatting diff.

No version bump is needed for a formatting-only change.

## Checklist

- [x] No new test is required for a formatting-only change
- [x] No public API is changed
- [x] Whole-tree Runic and `git diff --check` pass
- [x] The affected official test group passes on the coherent local-source stack

PR #4065, opened 2026-07-29
Restore DelayDiffEq 6.1 DEOptions compatibility

ChrisRackauckas-Claude:

## Summary

- Restore the positional `DEOptions` constructor layout used by registered DelayDiffEq 6.1.0, supplying `trivial_limiter!` for the later-added stage and step limiter fields.
- Add a pinned downstream regression that exercises the public `init`/`solve!` path and verifies both limiter fields.
- Add the already-public, docstringed `DEOptions` type to the rendered OrdinaryDiffEqCore API page.

## Root cause

`c017e6c452883848ad86255de5ca3a1d033ebe83` added two limiter fields to `DEOptions`. DelayDiffEq 6.1.0 still constructs the intermediate 21-type-parameter/41-argument layout, while the pre-existing compatibility constructor only covers an older 21-type-parameter/46-argument layout. The released package therefore hits a `MethodError` during `init`.

The added constructor is backward-compatible and does not introduce or remove a public name.

## Validation

- Bisected the released DelayDiffEq `init` failure from good `fde6e1e6bcd93d3e4c40c16038af9593c69e6ece` to first bad `c017e6c452883848ad86255de5ca3a1d033ebe83`.
- Ran the exact registered DelayDiffEq 6.1.0 regression locally: 4/4 assertions passed.
- Ran `GROUP=Downstream julia +1.12 --project=. -e 'using Pkg; Pkg.test()'`: the new DelayDiffEq compatibility set passed 4/4 and the existing time-derivative set passed 202/202. Resolution used SciMLBase 3.39.1 because the current 3.40 floor is not registered; no temporary package-floor edits are included here.
- The remainder of that Downstream run reaches an independent DynamicQuantities 1.13 error that also reproduces on unmodified `fd3f95bb99` master. It is not skipped or silenced and is being investigated separately.
- Runic passes on every changed Julia file. A whole-tree Runic check only reports pre-existing DISCO formatting differences, which are being handled separately.
- `git diff --check` passes.


ChrisRackauckas-Claude:
Hosted-CI checkpoint on amended head `813afcb824`:

Representative root, documentation, and sublibrary jobs stop during dependency resolution because the current OrdinaryDiffEq master copy of `OrdinaryDiffEqNonlinearSolve` requires unregistered SciMLBase 3.40 while General ends at 3.39.1. That clean-master floor is isolated in #4066. Runic reports the separately isolated clean-master DISCO formatting files from #4064. The quick SciMLSensitivity and PositiveIntegrators downstream failures are the independent compatibility boundaries recorded on #4066.

These setup failures occur before the DelayDiffEq compatibility regression is loaded. The exact registered DelayDiffEq 6.1.0 `init`/`solve!` test on this head passed 4/4 locally, and no resolver workaround or unrelated prerequisite is being added to this focused PR. The remaining hosted jobs are still being monitored.

ChrisRackauckas:
Instead we should retrocap

ChrisRackauckas-Claude:
Agreed. I am replacing this compatibility-constructor approach with a registry retrocap. The locally identified boundary is OrdinaryDiffEqCore 4.8.1: registered DelayDiffEq 6.0.0–6.1.0 works with Core 4.8.0, while the new DEOptions layout starts in 4.8.1. I am validating a General compat split of `DelayDiffEq 6–6.1.0` against `OrdinaryDiffEqCore 4–4.8.0`; once that resolver/solve check passes and the registry draft is open, I will close this PR rather than rebase it.

ChrisRackauckas-Claude:
Superseded by the registry corrections:

- JuliaRegistries/General#162780 gives the exact OrdinaryDiffEqDefault 2.4.1 release its required OrdinaryDiffEqCore 4.11 floor.
- JuliaRegistries/General#162781 caps DelayDiffEq 6.0.0–6.1.0 at OrdinaryDiffEqCore 4.8.0, before the incompatible `DEOptions` layout.

The final stacked registry resolution selected DelayDiffEq 6.1.0, OrdinaryDiffEqCore 4.8.0, and OrdinaryDiffEqDefault 2.4.0. The representative delayed problem then passed all 4 targeted checks through the public `init` and `solve!` path.

Closing this source-level compatibility constructor in favor of the reviewed retrocap direction.

PR #4066, opened 2026-07-29
Restore SciMLBase 3.39 compatibility for owner imports

ChrisRackauckas-Claude:
Ignore until reviewed by @ChrisRackauckas.

## Summary

- restore the SciMLBase compatibility floor from 3.40 to 3.39 for DiffEqBase, OrdinaryDiffEqCore, OrdinaryDiffEqDifferentiation, and OrdinaryDiffEqNonlinearSolve
- import `MatrixOperator` and `update_coefficients!` from their declaring public package, SciMLOperators, in the affected tests
- declare SciMLOperators in the separate differentiation sparse-test environment

## Why

The first bad commit is d78527d4d7992f69215b1526003dd9b33aaf2230 (#4059). It switched the tests from SciMLBase reexports to direct SciMLOperators ownership, but also raised the four SciMLBase floors to 3.40. Current General has SciMLBase 3.39.1 as its newest registered release, so a fresh checkout of OrdinaryDiffEq master cannot resolve OrdinaryDiffEqCore 4.12.

The used SciMLOperators interfaces are already public and available with SciMLBase 3.39. These source versions (DiffEqBase 7.10.0, OrdinaryDiffEqCore 4.12.0, OrdinaryDiffEqDifferentiation 3.6.0, and OrdinaryDiffEqNonlinearSolve 2.6.0) are newer than their latest registered versions and already contain the owner-import fixes. Lowering their source floors is therefore independently safe: it controls the future registration of the fixed versions and does not alter compatibility metadata for older releases.

The pending General registry PR JuliaRegistries/General#162635 separately retro-caps affected older releases before SciMLBase 3.40 becomes available. It is not yet effective and is not a prerequisite for the fixed source versions to support both SciMLBase 3.39 and 3.40.

## Investigation

- reproduced the resolver failure from a fresh Julia 1.12 depot on clean master
- bisected the first failure to d78527d4d7992f69215b1526003dd9b33aaf2230
- verified the parent commit resolves with registered SciMLBase 3.39.1
- tested this exact patch with registered SciMLBase 3.39.1 and local SciMLBase 3.40.1

## Tests

- [x] `GROUP=Core` for `OrdinaryDiffEqNonlinearSolve` with registered SciMLBase 3.39.1 — passed
- [x] `GROUP=Core` for `OrdinaryDiffEqNonlinearSolve` with local SciMLBase 3.40.1 — all affected suites passed; the run then reached the independent `LinearSolveForwardDiffExt` `issquare` error described below
[...]

PR #4068, opened 2026-07-29
Develop local sources in Julia 1.10 root tests

ChrisRackauckas-Claude:
> Ignore this draft until reviewed by @ChrisRackauckas.

Depends on #4066. This branch intentionally contains only the test-harness and regression-test changes; it does not duplicate the compatibility-floor changes from #4066.

## Summary

- Call SciMLTesting's public `develop_sources!` helper before root-package group dispatch so Julia 1.10 tests the monorepo's local sublibraries. The helper is a no-op on Julia 1.11 and later, where `[sources]` is native.
- Assert that the limiter integration test loads OrdinaryDiffEqCore, OrdinaryDiffEqLowOrderRK, and OrdinaryDiffEqSSPRK from this checkout.

## Problem

Julia 1.10 ignores the root project's `[sources]` section when `Pkg.test()` creates its temporary environment. A clean `GROUP=Integrators_I Pkg.test()` at `fd3f95bb99` therefore selected registered OrdinaryDiffEqCore 4.11.0, OrdinaryDiffEqLowOrderRK 2.2.1, and OrdinaryDiffEqSSPRK 2.2.0 instead of the local Core 4.12.0, LowOrderRK 2.2.2, and SSPRK 2.2.3. The Step Limiter test then reported 170 passes, 2 failures, and 1 error.

The hosted Julia 1.10 jobs for #3835 and #4003 passed because the reusable workflow separately developed in-repo sources before testing. The package's direct test harness should establish the same coherent source graph itself.

This is a test-environment issue, not a regression in the recent stage-limiter trait. The immediate parent of #4003 also fails with registered siblings. Older checkouts are not stable bisect endpoints against today's moving registry: the pre-#3835 checkout fails all 129 old constructor-level limiter assertions because newer registered siblings call those callbacks twice. Commit `fe90a44dd8` is relevant only as the structural introduction of the root `[sources]` assumption that Julia 1.10 cannot honor natively.

## Local verification

- Julia 1.10.11, `GROUP=Integrators_I`, on #4066 plus this focused diff:
  - Step Limiter Tests: 254 passed / 254 total
  - Discontinuity Detection Tests: 28 passed / 28 total
  - `Testing OrdinaryDiffEq tests passed`
- Julia 1.12.6, `GROUP=Integrators_I`, on #4066 plus this focused diff:
  - Step Limiter Tests: 254 passed / 254 total
  - Discontinuity Detection Tests: 28 passed / 28 total
  - `Testing OrdinaryDiffEq tests passed`
- Runic passes both changed files.
- `git diff --check` passes.

A repository-wide Runic check identifies only the pre-existing `test/Integrators_I/disco_tests.jl` formatting failure handled independently by #4064.

### DIFF
--- a/test/Integrators_I/step_limiter_test.jl
+++ b/test/Integrators_I/step_limiter_test.jl
@@ -12,6 +12,13 @@ using OrdinaryDiffEqExtrapolation
 using OrdinaryDiffEqFIRK: AdaptiveRadau, RadauIIA9, RadauIIA5, RadauIIA3
 using LinearAlgebra
 
+@testset "Tests use local monorepo sources" begin
+    repo = normpath(joinpath(@__DIR__, "..", ".."))
+    @test pkgdir(OrdinaryDiffEqCore) == joinpath(repo, "lib", "OrdinaryDiffEqCore")
+    @test pkgdir(OrdinaryDiffEqLowOrderRK) == joinpath(repo, "lib", "OrdinaryDiffEqLowOrderRK")
+    @test pkgdir(OrdinaryDiffEqSSPRK) == joinpath(repo, "lib", "OrdinaryDiffEqSSPRK")
+end
+
 # define the counting variable
 const STEP_LIMITER_VAR = Ref(0)
 # define the step_limiter! function which just counts the number of step_limiter calls
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -279,6 +279,9 @@ end
             Pkg.test(base_group, julia_args = ["--check-bounds=auto", "--compiled-modules=yes", "--depwarn=yes"], force_latest_compatible_version = false, allow_reresolve = true)
         end
     else
+        # Julia 1.10 does not resolve root [sources], so develop local sublibraries explicitly.
+        develop_sources!(dirname(@__DIR__))
+
         # Root-package group dispatch. `run_tests` owns the All / Interface /
         # Integrators / Regression / QA / functional-group routing that the old
         # hand-written `if GROUP == ...` ladder expressed. The group bodies


ChrisRackauckas-Claude:
Implementation/verification scratchpad:

1. Reproduced the clean Julia 1.10 failure with the package's `GROUP=Integrators_I` harness and recorded the registered/local version mismatch.
2. Confirmed the failure predates #4003, so this is not attributed to the stage-limiter trait implementation.
3. Used SciMLTesting's public `develop_sources!` backport and added local-path assertions for the three packages implicated by the limiter failures.
4. Kept the branch to two test files; #4066 remains an explicit prerequisite rather than being copied into this branch.
5. Ran the full group locally on Julia 1.10.11 and Julia 1.12.6 atop #4066. Both passed Step Limiter 254/254 and Discontinuity Detection 28/28, with the package-level test result passing.
6. Ran Runic on both changed files and `git diff --check`. The only repository-wide Runic failure is the independent `disco_tests.jl` formatting handled by #4064.

ChrisRackauckas-Claude:
Early CI dependency checkpoint: the Julia 1.10 `Integrators_I` job stops during package resolution before loading this test-harness change. General ends at SciMLBase 3.39.1, while the current master copy of `OrdinaryDiffEqNonlinearSolve 2.6.0` still requires 3.40, producing the exact 3.39-vs-3.40 unsatisfiable requirement. This is the declared #4066 prerequisite. The local #4066 + #4068 stack completed the same official group on Julia 1.10 and 1.12 (Step Limiter 254/254, DISCO 28/28, package passed). Runic independently reports the clean-master DISCO formatting handled by #4064. No resolver workaround is being added here.

PR #4072, opened 2026-07-29
Preserve structured Jacobian cache storage during initialization

ChrisRackauckas-Claude:

## Summary

- initialize only stored entries for constrained structured Jacobian/W prototypes instead of calling `fill!` on the logical matrix
- retain the existing sparse and dense initialization paths
- cover `Diagonal`, `Bidiagonal`, `Tridiagonal`, `SymTridiagonal`, and `UpperTriangular` prototypes, including `init`/`solve!` coverage for diagonal and tridiagonal caches

## Root cause

A clean-master bisect identified `d8cbb9e4446427305484e25eb4c81c9f85c09277` (`Initialize sparse Jacobian caches with stored values (#3833)`) as the first bad commit. That change correctly avoided initializing sparse structural entries to zero, but routed every non-sparse prototype through `fill!`. `fill!(::Diagonal, one(...))` attempts to write structural zeros outside the stored diagonal and throws.

The implementation uses public structured-array interfaces and the public backing storage of structured wrappers. `SymTridiagonal` requires a constructor/`copyto!` specialization because its public `setindex!` rejects off-diagonal writes even at stored positions.

## Local verification

- focused include of `default_linsolve_structure.jl` on Julia 1.12.6: exit 0
- official `GROUP=Downstream` target testset: `Default linsolve with structure | Pass 28 | Broken 2 | Total 30`
- full Downstream process: exit 1 later in unrelated `callback_detection.jl` ordering assertions (lines 53 and 58); kept out of this branch for a separate clean-master investigation
- Runic on all four changed files: exit 0
- `git diff --check`: exit 0
- whole-repository Runic scan: the changed files pass; exit 1 only for unrelated existing DISCO formatting in `lib/OrdinaryDiffEqCore/src/disco.jl`, `lib/OrdinaryDiffEqCore/src/integrators/controllers.jl`, and `test/Integrators_I/disco_tests.jl`


ChrisRackauckas-Claude:
CI audit after completion:

- The 179 failing root, sublibrary, downgrade, downstream, and documentation jobs all stopped before executing a test body. Stage counts were: root tests 15 failures in in-repo source development and 29 in `julia-buildpkg`; sublibrary CI 1 in source development and 65 in `julia-buildpkg`; 57 downgrade-resolution failures; 12 downstream-loading failures; and the documentation dependency-install failure. Representative logs in each lane show the SciMLBase 3.39/3.40 incompatibility handled by the focused prerequisite #4066.
- Runic failed only on the pre-existing `lib/OrdinaryDiffEqCore/src/disco.jl`, `lib/OrdinaryDiffEqCore/src/integrators/controllers.jl`, and `test/Integrators_I/disco_tests.jl` formatting debt isolated in #4064. CI marked all three changed Julia files in this PR as formatted: `default_linsolve_structure.jl`, `OrdinaryDiffEqDifferentiation.jl`, and `derivative_utils.jl`.

Local execution on the exact PR patch did reach the relevant tests: the focused structured-Jacobian initialization regression passed, and the official DiffEqBase Downstream target `Default linsolve with structure` passed 28 assertions with the 2 existing broken assertions unchanged. A broader Downstream run advanced past this target and stopped at the independently reproduced callback-ordering regression on clean master; that regression is under a separate bisect/fix and is not folded into this PR. Changed-file Runic and the full repository Runic stack with #4064 passed locally.

PR #4101, opened 2026-08-02
Preserve AD wrappers in Rosenbrock stages

ChrisRackauckas-Claude:
## Summary

- restore AD storage wrappers after restructuring flat Rosenbrock linear-solve results
- preserve existing state containers, including `ArrayPartition`, unless the struct-of-arrays conversion matches the original container type
- add a ReverseDiff regression that exercises the public `init` / `solve!` path

## Root cause

The first commit of #3996 (`100fe236`) replaced the former reshape helper with `ArrayInterface.restructure`. For an out-of-place ReverseDiff state, that changes a `TrackedArray` template into a plain `Vector{TrackedReal}`, and the Rosenbrock stage update later attempts to convert the tracked scalars back to `Float64`.

The focused reproducer passes at the parent commit (`5e3c607`) and fails at `100fe236`. Converting the restructured result with the public `ArrayInterface.aos_to_soa` API restores the tracked wrapper. The type guard avoids changing other state-container representations.

## Validation

- `GROUP=Core julia +1.10 --project=lib/OrdinaryDiffEqRosenbrock -e 'using Pkg; Pkg.test()'`
  - DAE Rosenbrock AD Tests: 16 / 16
  - Rosenbrock AD Tests: 64 / 64
  - Rosenbrock Convergence Tests: 192 / 192
- whole-repository Runic check passed

### DIFF
--- a/lib/OrdinaryDiffEqRosenbrock/Project.toml
+++ b/lib/OrdinaryDiffEqRosenbrock/Project.toml
@@ -58,6 +58,7 @@ Preferences = "1.5.0"
 Random = "<0.0.1, 1"
 RecursiveArrayTools = "4.2.0"
 Reexport = "1.2.2"
+ReverseDiff = "1"
 SafeTestsets = "0.1"
 SciMLBase = "3.39"
 StaticArrays = "1.9.18"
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Doesn't https://github.com/JuliaArrays/ArrayInterface.jl/pull/498 cover this?

Issue #3474, opened 2026-04-19
HHT and Generalized-Alpha type methods for Structural Problems


Sreeram-Shankar:
What kind of problems is it mostly used for? Please describe.

mainly used in structural dynamics problems and similar second order systems that come from finite element methods. These methods are especially useful simulations where high-frequency numerical noise or stiff elastic modes would otherwise contaminate or destabilize the solution. The generalized-alpha method and the older HHT-alpha method are designed to introduce controllable numerical dissipation, particularly damping out unwanted high-frequency behavior while still preserving accuracy in the lower-frequency modes that are usually of physical interest. This is the same general class of problems where Newmark methods are used, which are already available in OrdinaryDiffEq through the Newmark subpackage.

Describe the algorithm you’d like:

The generalized alpha method as a solver alongside Newmark-beta, ideally living close to the existing Newmark implementation in the same subpackage or wherever it fits best. The main interface should be based on choosing a rho_inf, which then determines the usual alpha_m, alpha_f, gamma, and beta parameters automatically . Users could still override the full parameter set through optional direct keyword arguments, but setting the spectral radius is the default. 

HHT-alpha serves a similar purpose of introducing numerical dissipation in structural time integration, it is a special case where alpha_m = 0.

The Newmark case happens when alpha_f = alpha_m = 0

Other implementations to know about

There is already a fixed timestep Julia implementation in StructuralVibration.jl that could serve as a reference for both behavior. Outside of Julia, these methods are standard in essentially all structural, multibody, and finite element dynamics software such as Abaqus and NASTRAN. 

References:

J. Chung and G. M. Hulbert, “A time integration algorithm for structural dynamics with improved numerical dissipation: the generalized-alpha method,” Journal of Applied Mechanics, 60(2), 371–375, 1993.

H. M. Hilber, T. J. Hughes, and R. L. Taylor, “Improved numerical dissipation for time integration algorithms in structural dynamics,” Earthquake Engineering & Structural Dynamics, 5(3), 283–292, 1977.

Sreeram-Shankar:
I’d be happy to take a first pass at implementing this. Before getting started, I wanted to ask about where this should live architecturally. Since generalized-alpha / Newmark are closely related, would you prefer:

- adding this alongside the existing Newmark implementation (e.g. in OrdinaryDiffEqNewmark),
- grouping these under a more general structural dynamics submodule,
-  or keeping the current structure and just extending it?

[...]

Issue #1570, opened 2022-01-11
Nonlinear Solver Interface


termi-official:
Some nonlinear problems can be very efficiently solved with non-standard methods, possibly abusing the specific structure of the problem. To allow users to efficiently utilize such methods in tandem with existing time solvers it might be helpful to expand the nonlinear solver interface (in the long term). A nice first approach to a nonlinear solver interface is given in [6]. Let me give 4 examples for possible applications of such an interface to boost the time step performance in specific applications to study possible requirements on such an interface (in addition to the existing ones, e.g. W-transformation, Jacobian reuse, ...). This list is not ment to be exhaustive, as I do not have a full overview on the state of the art on nonlinear solvers:

## Multilevel-Newton

Nonlinear problems arising from space-time discretizations of PDEs, PDAEs and similar, with solution variables that are not directly dependent on spatial derivatives can be efficiently solved with a multilevel strategy. These solution variables are often referred to as *internal variables*, occurring regularly in electronics and mechanics. Now, denoting the internal variables with `q` and the remaining variables with `u`, we can write the problem in mass matrix form as
```
Mu' = f(u,s,t)
Ns' = g(u,s,t)
```
where `'`denotes the time derivative and `M` is a positive semi-definite matrix and `N` is a positive semi-definite diagonal matrix. Now we can algorithmically decouple the solve for s and u by an alternating Newton scheme, where we solve for `s` on the "local" level and `u` on the "global" level. This way we obtain a large number of small local problems and one large problem on global level, however if dim(s) >> dim(u) we can greatly improve the efficiency of the scheme. A detailed derivation with FEM in conjunction with Runge-Kutta schemes is for example in [1].

## Quasi-newton methods

An alternative approach to efficient Jacobian reuse, which still requires the solution of equation systems, can be a quasi-Newton strategy. This method class utilizes (updated) approximations of the Jacobian in each iterative solve (see e.g. [4]) - however it is often observed that the quadratic convergence of the classical Newton-Raphon is lost. Nonetheless these methods can be highly efficient as e.g. studied in [5].

## Globalization strategies

Newton-Raphon does not perform well if no good initial guess for the nonlinear solve is available. In these scenarios so-called globalization strategies can help to some extend, usually by limiting the length of the tangent appropriately. Some overview is given in the methods section of [2] and the references therein, as well as the Eisenstat-Walker paper [3].

## Nonlinear preconditioning

[...]

PR #4112, opened 2026-08-04
Fix Newmark crash with polyalgorithm nonlinear solvers

singhharsh1708:
`nlsolve` is a public keyword argument on both `NewmarkBeta` and `GeneralizedAlpha`.
The default `NewtonRaphson` works, and so does any other algorithm whose cache steps
in place, such as `TrustRegion`, `Broyden` and `Klement`. Everything else crashed on the first
step. `RobustMultiNewton` and `FastShortcutNonlinearPolyalg` threw `type
NonlinearSolvePolyAlgorithmCache has no field p`; every SimpleNonlinearSolve
algorithm threw `type NonlinearSolveNoInitCache has no field p`. Same bug class as
#3859, fixed for `NonlinearSolveAlg` in #3959, in a different sublibrary.

The integrator cache was using the inner nonlinear cache as its own storage.
`perform_step!` constructed a fresh `NewmarkDiscretizationCache` on every step, and
recovered the three `DiffCache` AD buffers it needed by reaching into `nlcache.p`.
A cache that keeps its parameter object somewhere other than a `p` field has nothing
to reach into. The discretization state now lives on the integrator cache and is
refreshed in place per step, so nothing is read back out of the inner cache.

That alone is not sufficient. `NonlinearSolveBase`'s `@generated CommonSolve.solve!`
specialization for `NonlinearSolvePolyAlgorithmCache` returns the winning
sub-cache's solution and never writes back the parent cache's `retcode`, which stays
at `ReturnCode.Default`; the `NonlinearSolveNoInitCache` specialization delegates to
a one-shot `solve` and likewise leaves the cache's own `retcode` untouched. A caller
that drives these caches through `solve!` therefore has to read the returned
solution rather than the cache. This is specific to `solve!`. `InternalAPI.step!`
does assign `cache.retcode` (NonlinearSolveBase/src/polyalg.jl:316, 323, 359), which
is why the `newton.jl` path merged in #3959 drives with `step!` and is unaffected.

Both halves are load-bearing. With only the storage defect fixed, the guard reads
`nlcache.retcode`, sees `Default`, sets `force_stepfail` on every step, and the solve
returns `ReturnCode.ConvergenceFailure` with the state still at `u0` and
`t[end] == 0.0`, a crash traded for a silent stall.

The two in-place `perform_step!` bodies and the two `initialize!` bodies differed
only in which `αm`/`αf` they passed into the discretization cache, so they are now
shared `Union` methods. Newmark-β is generalized-α with `αm = αf = 0`, and those two
parameters enter only through the residual the inner solver evaluates. Neither the
state update nor the Zienkiewicz-Xie error estimate references them, so the merge is
exact, not an approximation. They no longer need to live on `GeneralizedAlphaCache`
at all, since they are read once when the discretization cache is built. Output was
checked bit-identical against master across 285 configurations. Across the changed
source files the diff removes more lines than it adds.

Result matrix, `u'' = -u` as a `SecondOrderODEProblem`, `dt = 0.1`,
`adaptive = false`, identical for both schemes:

| inner solver | before | after |
| --- | --- | --- |
| `NewtonRaphson` (default), `TrustRegion`, `Broyden`, `Klement` | Success | Success, unchanged to the last bit |
| `RobustMultiNewton`, `FastShortcutNonlinearPolyalg` | throws, no field `p` | Success |
| `SimpleNewtonRaphson`, `SimpleTrustRegion`, `SimpleBroyden`, `SimpleKlement` | throws, no field `p` | Success |

All twenty configurations agree with the reference to 1e-8 relative and sit at the
expected second-order error against the analytic solution (7.0e-4 for Newmark-β,
1.11e-3 for generalized-α at `rho_inf = 0.8`).

Not rebuilding a cache per step also drops allocations: on a two-DOF harmonic
oscillator at `dt = 0.01`, `perform_step!` goes from 2512 to 2336 bytes for
`NewmarkBeta` and from 2528 to 2336 for `GeneralizedAlpha`. AllocCheck's static site
count goes from 518 to 517; that test is still `broken` on both sides.

The new `test/nlsolve_cache_tests.jl` adds 98 assertions, taking the sublibrary's
Core group from 16 to 114. The six pre-existing testsets are unchanged at 16 of 16 on
both trees. The new file pins both state components of both schemes against a
closed-form step map rather than against each other, so an error common to every
inner solver still shows; it steps across a `tstop` that clips `dt` to 0.05, so a
discretization cache refreshed only at init cannot pass; it rebinds `integrator.p`
from a callback and checks the result against two chained constant-parameter solves,
so the parameters actually reaching the residual are pinned; and it drives a stub
inner solver that reports `ConvergenceFailure` alongside a finite iterate, so the
step-failure branch is exercised rather than assumed. Against master the file gives
24 passed and 24 errored.

Mutation testing confirms it discriminates. Deleting `evalcache.dt = dt`, deleting
`evalcache.t = t`, deleting `evalcache.p = p`, deleting the retcode guard, reverting
`nlsol.u` to `nlcache.u`, reverting `nlsol.retcode` to `nlcache.retcode`, and
building the generalized-α cache with `αm = αf = 0` each leave the suite red.

`NonlinearSolve` enters the sublibrary as a test-only dependency; 4.20.3 is the
lowest floor that resolves against the rest of the downgrade floor set, and the new
tests pass there.

Every number above is an `adaptive = false` number. Adaptive Newmark is separately
broken on master: `solve(prob, NewmarkBeta(), dt = 0.1)` with no further keyword
arguments throws `DimensionMismatch` from the error-estimate broadcast, on master and
on this branch alike. That is unrelated to the inner solver and is fixed in a
separate PR.

This does not route Newmark through `build_nlsolver`/`NonlinearSolveAlg`. That path's
unknown is the full state under a single scalar `γ`, while Newmark solves for the
acceleration alone, which enters position with `dt²β` and velocity with `dt·γ`.
Mapping it onto the DIRK form pins `β = γ²`, i.e. Trapezoid. `NewmarkBeta()` and
`Trapezoid()` already agree to 3e-16 on `u'' = -u`. The `rho_inf`
parameterization of generalized-α gives `β = γ²/4` identically, so no dissipative
member is representable. Reaching the shared path needs a second-order residual mode
and a second-order Jacobian interface for `(1-αm)M - (1-αf)(dt²β ∂fᵤ + dt·γ ∂fᵥ)`,
which is #1570 / #3474 territory. The two changes here are correct either way: both
replace reaching into a foreign cache's internals with the public API.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqNewmark/Project.toml
+++ b/lib/OrdinaryDiffEqNewmark/Project.toml
@@ -35,6 +35,7 @@ JET = "0.9, 0.11"
 LinearAlgebra = "<0.0.1, 1"
 MacroTools = "0.5.13"
 MuladdMacro = "0.2.4"
+NonlinearSolve = "4.20.3"
 NonlinearSolveFirstOrder = "2.1.1"
 OrdinaryDiffEqCore = "4"
 Pkg = "1"
@@ -51,6 +52,7 @@ julia = "1.10"
 [extras]
 DiffEqDevTools = "f3b72e0c-5b89-59e1-b016-84e28bfd966d"
 JET = "c3a54625-cd67-489e-a8e7-0a5a0ff4e31b"
+NonlinearSolve = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
@@ -58,4 +60,4 @@ SciMLTesting = "09d9d899-5365-40a9-917a-5f67fddea283"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["DiffEqDevTools", "Random", "SafeTestsets", "Test", "Pkg", "SciMLTesting"]
+test = ["DiffEqDevTools", "NonlinearSolve", "Random", "SafeTestsets", "Test", "Pkg", "SciMLTesting"]
--- a/lib/OrdinaryDiffEqNewmark/src/newmark_caches.jl
+++ b/lib/OrdinaryDiffEqNewmark/src/newmark_caches.jl
@@ -1,4 +1,4 @@
-@cache struct NewmarkBetaCache{uType, rateType, parameterType, N, Thread} <:
+@cache struct NewmarkBetaCache{uType, rateType, parameterType, N, EC, Thread} <:
     OrdinaryDiffEqMutableCache
     u::uType # Current solution
     uprev::uType # Previous solution
@@ -6,6 +6,7 @@
     β::parameterType # newmark parameter 1
     γ::parameterType # newmark parameter 2
     nlcache::N # Inner solver
+    evalcache::EC # Discretization state passed to the inner solver as its parameter
     tmp::uType # temporary, because it is required.
     atmp::uType
     thread::Thread
@@ -38,7 +39,7 @@ function alg_cache(
 
     tmp = zero(u)
     atmp = zero(u)
-    return NewmarkBetaCache(u, uprev, fsalfirst, β, γ, nlcache, tmp, atmp, thread)
+    return NewmarkBetaCache(u, uprev, fsalfirst, β, γ, nlcache, evalcache, tmp, atmp, thread)
 end
 
 @cache struct NewmarkBetaConstantCache{uType, rateType, parameterType, N, Thread} <:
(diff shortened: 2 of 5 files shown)


ChrisRackauckas:
This should just use NonlinearSolveAlg?

singhharsh1708:
Can't. NLSolver's γ is scalar over the whole state, so Newmark's two coefficients collapse to β = γ², which is just Trapezoid. Generalized-α gives β = γ²/4 for every rho_inf, so nothing dissipative survives either. termi-official hit this in #2187 and left it.

singhharsh1708:
NSA also still trips the same #3859 class: SimpleNewtonRaphson throws on stats, Broyden diverges after step 1. Those stats reads are mine from #3959, I have to fix them.

singhharsh1708:
we can do that but it needs the second-order residual and Jacobian interface first. This PR is just the crash.

PR #4115, opened 2026-08-04
Add MREIL extrapolated multirate linearly implicit Euler

### LINKED ISSUE #884: Multirate Extrapolation
https://www.mcs.anl.gov/~emconsta/Repository/Papers/MRExtrap_Preprint.pdf
https://www.mcs.anl.gov/~emconsta/Repository/Papers/MRExtrap_TR.pdf

singhharsh1708:
MREIL is the linearly implicit member of the extrapolated multirate family of
Constantinescu and Sandu, "Extrapolated Multirate Methods for Differential Equations
with Multiple Time Scales", J. Sci. Comput. 2013, 56:28-44. That paper builds
extrapolated multirate methods on two bases: multirate forward Euler, already here as
`MREEF`, and multirate linearly implicit Euler, which is this. Closes #884.

Relative to `MREEF` only the inner substep changes. The subdivision sequence, the slow
rate frozen across a macro interval, the Aitken-Neville extrapolation and the error
estimate are identical. Where `MREEF` takes an explicit Euler substep, MREIL solves

    (M - h_fast*J) * delta = h_fast * (k_slow + k_fast)

with `J` the Jacobian of the fast component `f1`, which is where the stiffness lives.

## Why it is worth having

Fixed step, order 4, on a problem with a stiff fast block and a mildly nonlinear slow
part. `MREEF` is stability limited and MREIL is not: at `dt = 0.05`, `lambda = 1e4`,
`MREEF` returns `Unstable` with `NaN` while MREIL returns `Success`.

Adaptive at tolerance 1e-6, `m = 4`, `order = 4`, accepted steps:

| lambda | MREEF | MREIL |
| --- | --- | --- |
| 1e2 | 92 | 71 |
| 1e3 | 179 | 72 |
| 1e4 | 978 | 36 |
| 1e6 | 97461 | 34 |

Errors are around 1e-8 on both sides. `MREEF`'s step count scales linearly with the
fast time constant; MREIL's is flat.

The advantage is specific to a dissipative fast subsystem. MREIL is A-stable and
L-stable at `order = 2` only. At orders 3, 4 and 5 the supremum of `|R(z)|` over the
closed left half plane is 1.000594, 1.001502 and 1.002282, attained on the imaginary
axis, so it is not A-stable there, though `R(inf) = 0` at every order and decays as
`|z|^(-m)`. `MREEF` by contrast is unbounded off the real axis: `|R(1000i)|` is 5.8e29
for `MREEF` against 4.3e-11 for MREIL. That bounds the per step amplification but does
not make an oscillation-dominated fast block a good fit, because the excursion
compounds. At fixed `dt` an under-resolved oscillation is either damped away at large
`omega*dt` or amplified at roughly 1.0015 per step near `omega*dt` around 2.56, and
`Success` is returned either way. Adaptive stepping resolves the transient and returns
the right answer in both regimes.

Extrapolated linearly implicit Euler is a W-method, so the Richardson order survives an
arbitrary frozen `J`: measured 2.00, 2.97 and 4.24 for orders 2, 3 and 4 with `J = 0`,
and 2.07, 3.04 and 4.03 with a deliberately wrong `J`, against 2.01, 3.00 and 3.99 with
the true one. What a bad `J` costs is stability, not order. With `J = 0` the substep is
explicit Euler and MREIL reduces to `MREEF`.

## Convergence

`test_convergence` with `m = 10`, `dts = 1//2^(6:-1:2)` on a non-stiff split problem:
2.0286, 3.0422 and 4.0320 for orders 2, 3 and 4 harmonic, and 4.0354 for order 4
romberg. The out-of-place scalar path gives the same numbers. A second problem with a
non-constant fast Jacobian and a time-dependent slow part, which exercises both the
frozen Jacobian and the frozen slow rate, also attains its order.

## Jacobian and W

`J` is taken once per macro step at `(uprev, t)` and shared by every column of the
extrapolation table. A per column or per substep Jacobian would give each column a
different asymptotic expansion in `h` and Aitken-Neville would stop cancelling the
leading terms. This is what single rate extrapolated linearly implicit Euler does.

`h_fast = dt/(ns[j]*m)` depends only on the column index, so `W = J - M/h_fast` is
constant for a whole column. It is assembled and factorized once per column and reused
across every substep in it. Verified by the stats identities `njacs == naccept + nreject`
and `nw == order * (naccept + nreject)`, with `nsolve` growing linearly in `m` while
`nw` stays fixed, and by wall time staying flat as substeps grow 16 fold.

The Jacobian is also reused across a rejected step, since `uprev` and `t` are unchanged
on a retry. Measured `njacs/(naccept + nreject)` of 0.9600, 0.9985 and 0.9345 on three
problems, against 0.965 for a `Rodas5P` control. Reuse is tied to the step attempt, so
it cannot survive `reinit!`, `remake`, a parameter change or a callback.

Mass matrices work: `Diagonal([2, 3])` gives 1.56e-10 against the closed form, and a
scalar state with `mass_matrix = 2.0I` gives 1.82e-14 at `dt = 0.0125`. Matrix-free
`linsolve` and an operator valued `f1` are both supported. `linsolve` applies to the
in-place path; a scalar or out-of-place problem always uses a direct factorization, as
Rosenbrock's constant cache does.

A dense all-zero `jac_prototype` on `f1` is rejected with an `ArgumentError`. The
right way to request a dense Jacobian is no prototype at all, but SciMLBase reads an
all-zero one as a sparsity pattern with no nonzeros, so `J` comes back identically
zero when it is passed anyway. That silently degrades
every implicit solver in the repo, not just this one, and MREIL is the case where it
matters most because the method stops being implicit rather than merely losing an
accelerator. Filed separately.

## Dense output

Both `perform_step!` methods fill `k[2]` from `f1 + f2` at `(u, t + dt)`, so the Hermite
interpolant is fed real derivatives. Step point error 1.54e-12 against interpolated
error 1.73e-10 at tolerance 1e-10, and dense output converges at 3.95, 3.98, 3.99 and
3.99. The cost is exactly one extra `f1` and one extra `f2` evaluation per attempted
step, no extra Jacobians, W builds or linear solves, which is 2.5 percent on `f1` and 10
percent on `f2` at the default `m = 4`, `order = 4`.

`MREEF`, `MRAB`, both MRI-GARK families and `MIS` have the same omission and are still
order 1 through `saveat` and `sol(t)`. That is filed separately rather than fixed here.

## Tests

Core group 235 of 235: `MREEF` 19, MREIL 136, `MRAB` 16, MRI-GARK 57, `MIS` 7. Against
current master the same group is 99 of 99, so nothing pre-existing moved. Solution
hashes over the full `(t, u)` byte stream are identical to master for all 60
combinations of the 12 existing algorithm configurations across in-place, out-of-place,
fixed and adaptive stepping, including retcodes and step counts.

AllocCheck reports 534 static sites for MREIL against 531 for `Rosenbrock23` and 533 for
`Rodas5P`, so it sits inside the band of every in-place implicit method here. Roughly
435 of those are LinearSolve's default polyalgorithm dispatch, shared by all of them.

Two notes for the standards audit, both deliberate. `seq::Symbol` matches `MREEF`'s
existing declaration in the same file; changing only MREIL would give the two multirate
methods different APIs for the same option. The `length(...) == 0` it flags is an
allocation test assertion, not a shape check used to pick behaviour.

## AI Disclosure


ChrisRackauckas:
Methods which are implicit should probably move to a separate MultirateImplicit set, since they are adding a lot of dependencies the basic ones don't need.

singhharsh1708:
done, moved MREIL plus the three implicit MRI-GARK methods to a new OrdinaryDiffEqMultirateImplicit sublib. base Multirate drops Differentiation/NonlinearSolve/LinearSolve/SciMLOperators/SparseArrays entirely now, bumped to 3.0.0 since it loses those exports. test counts unchanged after the split (explicit 70, implicit 165, MRI-GARK 28+29 matches the old 57).

singhharsh1708:
self-review pass before you get to it: the W-failure and linsolve-failure early returns only set EEst, which fixed-dt stepping ignores, so a singular W returned the initial condition with Success (u'=40u at dt=0.1 makes W exactly 0 and reproduced it). those paths now set force_stepfail. also fixed: a dead issuccess_W guard on the iip path, the retcode filter exempting Default, user jac_prototype getting overwritten in the sparse reset, Int division freezing the Neville weights at Float64 for BigFloat states, dead m>=1 validation for scalar problems, a wasted f2 eval per step in the oop MRI-GARK path, spzeros passing the zero-pattern guard, nw over-counting on operator paths, and stray qa test deps. 21 new tests, each verified to fail with its fix reverted. suites 70/70 and 186/186.

ChrisRackauckas:
conflicts.

singhharsh1708:
yes

PR #4129, opened 2026-08-05
fix: save all simultaneous events in a VCC

AayushSabharwal:
The current behavior only saves the first event, which leads to incorrect discrete timeseries in the solution.


ChrisRackauckas:
DiffEqBase MTK test fails

AayushSabharwal:
Makes no sense why it fails, I can run it successfully in a fresh REPL. I'm probably missing something obvious.

singhharsh1708:
ran this against the exact package set CI resolves (MTKBase 1.60.0, RAT 4.3.6): `c_dn` never registers as a timeseries parameter since `parse_callbacks_for_discretes!` only scans `affect`, and that is what surfaces as the `Indexing with parameters is deprecated` error in the CI log. your SciML/ModelingToolkit.jl#4876 fixes that half, so locally you presumably have it dev'd.

but with 4876 patched in, the test still fails differently: `sol2[c_dn]` comes out right ([0, 0, 1, 1, 2]) while `sol2[c_up]` stays [0.0], its affect never executes because only the winning index of the two identical conditions fires. so saving all simultaneous events isn't enough here, the affects need to run for all of them too. (also 4876 has a typo, `even.affect` in the else branch.)

AayushSabharwal:
🤦🏻 of course, I lost track of the branches I had deved. Thank you for investigating!

PR #4137, opened 2026-08-05
Fix NonlinearSolveAlg crash with SimpleNonlinearSolve inner solvers

singhharsh1708:
`NonlinearSolveAlg` crashes with every SimpleNonlinearSolve inner algorithm. Those
define no `__init`, so NonlinearSolveBase hands back a `NonlinearSolveNoInitCache`,
which carries only `prob, alg, args, kwargs, initializealg, retcode, verbose`. The
first raw field read then throws:

```julia
solve(prob, Trapezoid(nlsolve = NonlinearSolveAlg(SimpleNewtonRaphson())))
ERROR: type NonlinearSolveNoInitCache has no field stats
```

This is the last of the #3859 bug class. #3959 routed `u` and `fu` through
`get_u`/`get_fu` but left the raw `stats` reads, which I said I would fix when it came
up on #4112. Extracted from #3798 so the crash fix lands independently of the W-reuse
feature still under discussion there; #3798 rebases onto this once it lands.

Every guard dispatches on the cache type, not on a list of algorithm types, so a newly
exported SimpleNonlinearSolve algorithm is picked up automatically. The changes: the
stats copies in both `initialize!` methods skip a cache that has no counters;
`compute_step!` drives a no-init cache with `solve!` (it has no `step!`) and reads the
returned solution; `_rejected_trial_step` skips it; the builds and the post-`resize!`
rebuild keep the solver's own tolerances instead of the zeroed ones the stepped path
uses, which would send every inner solve to `MaxIters`; and `initialize!` hands the
cache a copy of `z` rather than `z` itself, since `reinit!` on this cache stores the
array it is given and a later in-place `resize!` of `z` otherwise defeats the
length-mismatch rebuild and leaves a stale, wrongly sized Jacobian buffer behind.

Stats on this path are honest rather than invented. SimpleNonlinearSolve builds every
solution with `stats === nothing`, so the residual and Jacobian work its solvers do
cannot be attributed to the integrator; nothing is counted for `nf` and `nsolve`, and
`njacs`/`nw` count only the W assemblies performed on this side of the boundary.

Verified on a matrix of three problems (scalar decay, stiff in-place Van der Pol,
singular-mass-matrix Robertson) times three methods (Trapezoid, TRBDF2, FBDF): every
SimpleNonlinearSolve cell crashes on master and succeeds on this branch with worst
error 1.9e-6 at reltol 1e-8, and every already-working cell (NewtonRaphson,
TrustRegion, RobustMultiNewton, FastShortcutNonlinearPolyalg and the rest) produces
byte-for-byte identical trajectories to master.

The new `test/nsa_noinit_tests.jl` passes 102 assertions on this branch; against
master's source it errors out on the exact crash above. Mutation testing was run guard
by guard when this change was first assembled: deleting any of the guards leaves the
suite red, including the tolerance guards, which needed a singular-mass-matrix stage
to discriminate because an easy problem's residual can hit exact zero and terminate
`Success` under `MaxIters` tolerances. The full Core group is 707 passing with no
regressions.

Two upstream findings from the same investigation, filed separately rather than
patched here: `Broyden` and `Klement` diverge on the second solve after `reinit!`
because their update-rule caches never reset the stored previous residual, and
`NonlinearSolveAlg(LevenbergMarquardt())` on the stepped path returns `Success` with a
materially wrong answer on a scalar problem.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
@@ -41,7 +41,11 @@ function initialize!(
     cache.tstep = integrator.t + nlsolver.c * dt
 
     (; ustep, tstep, k, invγdt) = cache
-    if SciMLBase.has_stats(integrator)
+    # A no-init cache has no `stats` (SimpleNonlinearSolve builds every solution with
+    # `stats === nothing`), and its `reinit!` only remakes the problem without evaluating
+    # the residual, so neither the copies below nor the `+1` apply: nothing is counted and
+    # nothing is invented.
+    if SciMLBase.has_stats(integrator) && !(cache.cache isa NonlinearSolveNoInitCache)
         # The `reinit!` below evaluates the residual at the new `z` and *then* zeroes the
         # inner cache's counters, so that evaluation is never visible in
         # `cache.cache.stats.nf`. Left uncounted it loses one `f` call per stage.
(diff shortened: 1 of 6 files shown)


lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
```diff
@@ -152,21 +161,39 @@ function initialize!(
...
-            )
+            if cache.cache isa NonlinearSolveNoInitCache
+                # A no-init cache is `solve!`-driven, so it must keep terminating on its own
+                # (nonzero, default) tolerances here exactly as on the build path: rebuilding
+                # it with the zeroed tolerances below would leave every complete inner solve
+                # returning `MaxIters`.
+                cache.cache = init(
+                    new_prob, cache.cache.alg;
+                    verbose = integrator.opts.verbose.nonlinear_verbosity
+                )
```
ChrisRackauckas, line 172: What do you actually get from this? If it's a noinit cache, there isn't anything?

ChrisRackauckas: oh I forgot we made a noinitcache object.

Issue #4156, opened 2026-08-08
FIRK solvers do not support matrix-free Jacobians / Krylov linsolve


ChrisRackauckas-Claude:
## Summary

The FIRK solvers reject any non-concrete (matrix-free) Jacobian, so they cannot be combined with a Krylov `linsolve`:

```julia
solve(prob, RadauIIA5(linsolve = KrylovJL_GMRES()))
# ERROR: Non-concrete Jacobian not yet supported by RadauIIA5.
```

Thrown from [`lib/OrdinaryDiffEqFIRK/src/firk_caches.jl`](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/OrdinaryDiffEqFIRK/src/firk_caches.jl), which has four such sites:

```
firk_caches.jl:262   error("Non-concrete Jacobian not yet supported by RadauIIA5.")
firk_caches.jl:466   error("Non-concrete Jacobian not yet supported by RadauIIA5.")
firk_caches.jl:689   error("Non-concrete Jacobian not yet supported by AdaptiveRadau.")
firk_caches.jl:838   error("Non-concrete Jacobian not yet supported by GaussLegendre.")
```

## Why it matters

FIRK + Krylov is the natural combination for large stiff systems at tight tolerances — exactly the regime where Radau's high order pays for itself and where forming a dense Jacobian is not an option. BDF and Rosenbrock already support matrix-free `linsolve`; FIRK is the gap. The W-transform structure is Krylov-friendly, since it decouples into one real shifted solve and complex-conjugate-pair solves, each of which is a shifted operator `(c/h)I - J` backed by the same JVP.

The main implementation work is the complex-shifted case: a lazy operator representing `(α + βi)/h · I - J` acting on complex vectors while the underlying JVP is real.

## Secondary: the error message

Independent of whether support lands, the message could name what *is* supported instead of only refusing. As written a user has no indication whether to switch solver, switch `linsolve`, or supply `concrete_jac = true`.

## Environment

```
Julia 1.12.6
OrdinaryDiffEq v7.3.0
```

Filing to track. Someone is looking at implementing this; if it turns out to be large, a scoped-down first pass (e.g. `RadauIIA5` only) would still be worth having.

ChrisRackauckas-Claude:
[...]

PR #4157, opened 2026-08-08
Support matrix-free Krylov linear solves in RadauIIA5

ChrisRackauckas-Claude:

Part of https://github.com/SciML/OrdinaryDiffEq.jl/issues/4156.

## What changed and why

`RadauIIA5(linsolve = KrylovJL_GMRES())` errored in `alg_cache` with `Non-concrete Jacobian not yet supported by RadauIIA5.`, so the one FIRK/Krylov combination people actually want for large stiff systems at tight tolerances was unavailable. The real-shifted stage system now reuses `SciMLOperators.WOperator` (the same object `build_J_W` already returns on the matrix-free path, exactly as BDF/Rosenbrock use it), and a new internal `ComplexShiftOperator` backs the complex-shifted system `J - (α + βi)/h·M` by applying the real matrix-free Jacobian-vector product separately to the real and imaginary parts of the argument. The W-form update in `perform_step!`/`_ode_addsteps!` is now a two-method `firk_update_W!` so the operator and dense paths dispatch instead of branching inline.

**Scope was cut deliberately: this is `RadauIIA5` only.** `RadauIIA3`, `RadauIIA9`, `AdaptiveRadau` and `GaussLegendre` still require a materializable Jacobian. Extending to them is mechanical for `RadauIIA9` (one more complex pair) and less so for `AdaptiveRadau` (a stage-count-dependent vector of complex operators plus the threaded stage loop), and `AdaptiveRadau` cache construction is currently under separate investigation, so I kept it out. Their error message now names what *is* supported rather than only refusing, and the duplicated `"…not yet supported by RadauIIA5."` message that actually guarded the `RadauIIA9` cache is corrected.

`ComplexShiftOperator` is internal (not exported, not `public`), so no public API is added; the version bump to 2.7.0 is a minor bump for the new capability.

## Verification

Julia 1.12.6, branch based on `c11ff74bb4`.

**New test failing before the fix** (`git stash push -- lib/OrdinaryDiffEqFIRK/src`, then run the new test file against unpatched sources):

```
RadauIIA5 matrix-free Krylov: Error During Test at .../test/ode_firk_krylov_tests.jl:48
  Non-concrete Jacobian not yet supported by RadauIIA5.
RadauIIA5 matrix-free Krylov with mass matrix: Error During Test at .../test/ode_firk_krylov_tests.jl:69
  Non-concrete Jacobian not yet supported by RadauIIA5.
Unsupported FIRK methods report what is supported: Test Failed at .../test/ode_firk_krylov_tests.jl:101
     Message: "Non-concrete Jacobian not yet supported by AdaptiveRadau."
Unsupported FIRK methods report what is supported: Test Failed at .../test/ode_firk_krylov_tests.jl:101
     Message: "Non-concrete Jacobian not yet supported by GaussLegendre."
Test Summary:                                       | Pass  Fail  Error  Total     Time
FIRK Krylov                                         |    1     3      2      6  1m02.1s
  RadauIIA5 matrix-free Krylov                      |                 1      1     3.3s
  RadauIIA5 matrix-free Krylov with mass matrix     |                 1      1    17.4s
  Unsupported FIRK methods report what is supported |    1     3             4     5.6s
ERROR: Some tests did not pass: 1 passed, 3 failed, 2 errored, 0 broken.
```

**FIRK test group with the fix** — `cd lib/OrdinaryDiffEqFIRK && GROUP=Core julia --project -e 'using Pkg; Pkg.test()'`:

```
FIRK Tests    |  105    105  4m07.5s
FIRK Krylov Tests |   12     12  1m09.0s
```

**Accuracy** — the new tests solve a 2D Brusselator (8x8x2 = 128 unknowns) and a mass-matrix nonlinear heat problem with `KrylovJL_GMRES` / `KrylovJL_BICGSTAB` at `abstol = reltol = 1e-10` and compare against `RadauIIA5()` with the default dense `linsolve` at `1e-12`, `rtol = 1e-7`. They also assert `cache.J isa AbstractSciMLOperator` and `cache.W2 isa ComplexShiftOperator`, so a regression that silently fell back to a dense Jacobian would fail rather than pass. Additional manual spot checks (not in the suite) on a 32-point nonlinear heat problem:

```
RadauIIA5 KrylovJL concrete_jac=true    ad=AutoForwardDiff retcode=Success err=1.32e-10
RadauIIA5 KrylovJL concrete_jac=nothing ad=AutoForwardDiff retcode=Success err=1.08e-10
RadauIIA5 KrylovJL concrete_jac=nothing ad=AutoFiniteDiff  retcode=Success err=1.15e-10
```

**Formatting / spelling** — `Runic` v1.5.1 `--check` over the touched files exits 0; `typos` over `lib/OrdinaryDiffEqFIRK` and `docs/src/implicit/FIRK.md` exits 0.

## Pre-existing failures I did NOT introduce (verified on the unmodified base)

Two checks fail identically with and without this change, on a clean `c11ff74bb4` tree:

1. `GROUP=QA` — `public API has docstrings: Evaluated: isempty([:SciMLBase])`. Reproduced with my changes stashed, and it is red on master's own CI. Root cause is a SciMLTesting regression, not this PR: SciML/SciMLTesting.jl#41 (v2.6.2, registered 2026-08-05) rewrote `_has_docstring` from a `occursin("No documentation found", ...)` string match to `Base.Docs.hasdoc`. The old check silently passed *every* undocumented module binding, because an undocumented module renders as `"No docstring found for public module …"`. A module self-exports its own name, so `@reexport using SciMLBase` pulls `:SciMLBase` into the public-API set. `run_api_docs`'s *rendered* check already exempts external reexports (SciML/SciMLTesting.jl#30); its *docstrings* check does not. Pinning SciMLTesting 2.6.1 or adding `ignore = (:SciMLBase,)` both make it green. 11 sublibrary QA lanes are red on master for this reason.
2. `julia --project=docs docs/make.jl` — `Error: Cannot resolve @ref for md"[`DummyControllerCache`](@ref)" in docs/src/devtools/internals/public_api.md` → `makedocs encountered an error [:cross_references]`. Reproduced with my changes stashed; both builds emit exactly one `Error`, and my `docs/src/implicit/FIRK.md` addition produces no new Documenter messages. Introduced by #4118, which rendered `DummyController`'s docstring; that docstring `@ref`s `DummyControllerCache`, whose `@docs` entry was removed in #3832. `DummyControllerCache` is deliberately non-public, so the fix is to drop the `@ref` in `lib/OrdinaryDiffEqCore/src/integrators/controllers.jl:498`.

Both belong in their own PRs rather than being smuggled in here.

## What I did NOT verify

- GPU paths, downstream packages, and the allow-failure CI jobs (`julia pre`, GPU, Downstream) — not run locally.
- Julia 1.10/1.11; only 1.12.6 was exercised.
- Out-of-place (`Val{false}`) FIRK caches — those still `lu` a dense matrix and are unchanged; matrix-free out-of-place is unsupported before and after.
- Preconditioners with the complex operator. `ComplexShiftOperator` supports `mul!` only; a left/right preconditioner supplied through `linsolve` is untested here.

## Things a reviewer should push back on

- **`nf` accounting for the complex solve is incomplete.** `OrdinaryDiffEqDifferentiation.dolinsolve` increments `stats.nf` by the Krylov iteration count only when `linsolve.A isa WOperator`, so the real solve is counted and the complex one is not. Fixing it properly belongs in that helper rather than here, so `stats.nf` under-reports on this path. Say the word and I will do it in a follow-up (or here).
- **The Jacobian is not truly frozen across a W-reuse step.** `JVPCache` holds a reference to `uprev`, so on a step where `do_newW` returns `false` the JVP silently tracks the current `uprev` rather than the one from when W was built. This matches the existing BDF/Rosenbrock matrix-free behaviour and only affects Newton convergence rate, not the fixed point, but it is worth a second opinion.
- **`ComplexShiftOperator` lives in `OrdinaryDiffEqFIRK` rather than `OrdinaryDiffEqDifferentiation`** next to `WOperator`/`JVPCache`. I kept it local to avoid a cross-package version bump for a single-consumer type; if you would rather it be shared infrastructure it should move.
- The `firk_update_W!` dispatch pair changes `_ode_addsteps!` for `RadauIIA5Cache` as well as `perform_step!`. The dense branch is byte-identical to what was inlined there before, but it is a behaviour-adjacent refactor in a path the new tests do not reach.

https://claude.ai/code/session_01F75XsVeZ94QH3PmCuq6QUF

### DIFF
--- a/docs/src/implicit/FIRK.md
+++ b/docs/src/implicit/FIRK.md
@@ -65,6 +65,21 @@ Gauss–Legendre methods use the roots of the Legendre polynomial directly as co
   - **Systems < 200**: FIRK methods are competitive due to better multithreading
   - **Systems > 200**: Consider SDIRK or BDF methods instead
 
+### Matrix-free (Krylov) linear solves
+
+`RadauIIA5` supports matrix-free linear solves, so large stiff systems can be handled
+without ever forming the Jacobian:
+
+```julia
+using LinearSolve
+solve(prob, RadauIIA5(linsolve = KrylovJL_GMRES()))
+```
+
+The W-transform splits the stage system into one real-shifted and one complex-shifted
+linear system per step, and both are applied lazily through Jacobian-vector products.
+`RadauIIA3`, `RadauIIA9`, `AdaptiveRadau` and `GaussLegendre` still require a Jacobian that
+can be materialized, and error if given a matrix-free `linsolve`.
+
 ## Performance Guidelines
 
   - **Best for tolerances ≤ 1e-9** where high accuracy justifies the cost
--- a/lib/OrdinaryDiffEqFIRK/Project.toml
+++ b/lib/OrdinaryDiffEqFIRK/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqFIRK"
 uuid = "5960d6e9-dd7a-4743-88e7-cf307b64f125"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "2.6.1"
+version = "2.7.0"
 
 [deps]
 FastBroadcast = "7034ab61-46d4-4ed7-9d0f-46aef9175898"
--- a/lib/OrdinaryDiffEqFIRK/src/OrdinaryDiffEqFIRK.jl
+++ b/lib/OrdinaryDiffEqFIRK/src/OrdinaryDiffEqFIRK.jl
@@ -36,8 +36,9 @@ using OrdinaryDiffEqDifferentiation: build_J_W, build_jac_config,
     calc_J!, dolinsolve, calc_J,
     islinearfunction
 import ADTypes: AutoForwardDiff
+import LinearAlgebra
 import SciMLOperators
-import SciMLOperators: AbstractSciMLOperator
+import SciMLOperators: AbstractSciMLOperator, WOperator, update_coefficients!
 # Load-bearing runtime dependency: provides the nonlinear-solver machinery the FIRK
 # integrators dispatch into. The convergence-state names FIRK uses are owned by
 # OrdinaryDiffEqCore (imported above), so this is a module-only import to keep the
@@ -63,6 +64,7 @@ using Reexport: @reexport
 include("algorithms.jl")
 include("alg_utils.jl")
 include("controllers.jl")
+include("firk_operators.jl")
 include("firk_caches.jl")
 include("firk_tableaus.jl")
 include("firk_perform_step.jl")
--- a/lib/OrdinaryDiffEqFIRK/src/firk_addsteps.jl
+++ b/lib/OrdinaryDiffEqFIRK/src/firk_addsteps.jl
@@ -454,10 +454,7 @@ function _ode_addsteps!(integrator, cache::RadauIIA5Cache, repeat_step = false)
     γdt, αdt, βdt = γ / dt, α / dt, β / dt
     new_jac = false
     if (new_W = do_newW(integrator, alg, new_jac, cache.W_γdt))
-        @inbounds for II in CartesianIndices(J)
-            W1[II] = -γdt * mass_matrix[Tuple(II)...] + J[II]
-            W2[II] = -(αdt + βdt * im) * mass_matrix[Tuple(II)...] + J[II]
-        end
+        firk_update_W!(W1, W2, J, mass_matrix, γdt, αdt, βdt, uprev, p, t)
         integrator.stats.nw += 1
     end
 
--- a/lib/OrdinaryDiffEqFIRK/src/firk_caches.jl
+++ b/lib/OrdinaryDiffEqFIRK/src/firk_caches.jl
@@ -107,6 +107,9 @@ function alg_cache(
     jac_config = build_jac_config(alg, f, uf, du1, uprev, u, tmp, dw12)
 
     J, W1 = build_J_W(alg, u, uprev, p, t, dt, f, jac_config, uEltypeNoUnits, Val(true))
+    if J isa AbstractSciMLOperator || W1 isa WOperator
+        firk_matrixfree_unsupported("RadauIIA3")
+    end
     W1 = similar(J, Complex{eltype(W1)})
     recursivefill!(W1, false)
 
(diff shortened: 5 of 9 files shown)


ChrisRackauckas-Claude:
## CI status

The jobs this PR can affect are green:

| check | result |
|---|---|
| `sublibrary-ci / lib/OrdinaryDiffEqFIRK / Julia 1 / Tests - Core` | pass |
| `sublibrary-ci / lib/OrdinaryDiffEqFIRK / Julia lts / Tests - Core` | pass |
| `sublibrary-ci / lib/OrdinaryDiffEqFIRK / Julia pre / Tests - Core` | pass |
| `downgrade-sublibraries / test (lib/OrdinaryDiffEqFIRK)` | pass |
| `Runic / Runic Format Check` | pass |
| `Spell Check with Typos` | pass |

Every red check on this PR is also red on the base commit c11ff74bb4:

- `sublibrary-ci / lib/OrdinaryDiffEqFIRK [QA]` — fails **only** on `public API has docstrings: Evaluated: isempty([:SciMLBase])`, nothing else in the log. Ten other sublibrary QA lanes (Rosenbrock, StabilizedIRK, LowOrderRK, StochasticDiffEq{Implicit,Weak,Leaping,LowOrder}, TaylorSeries, …) are red on master's own CI for the same SciMLTesting v2.6.2 regression described in the PR body; FIRK's QA lane just was not triggered on that master push. I reproduced it locally with this PR's `src/` changes stashed.
- `tests / {AD, Integrators_II, InterfaceII, InterfaceV, QA}` on Julia 1/lts/pre — all present in master c11ff74bb4's failure list, byte-for-byte the same job names.
- `Documentation / Build and Deploy Documentation` — the `DummyControllerCache` `@ref` break from #4118; reproduced locally on an unmodified tree.
- `downgrade-sublibraries / test (lib/{OrdinaryDiffEqExtrapolation, OrdinaryDiffEqIMEXMultistep, OrdinaryDiffEqNonlinearSolve, OrdinaryDiffEqSIMDRK, StochasticDiffEq})` — sublibraries this PR does not touch; the FIRK downgrade lane passes.
- Downstream (`PositiveIntegrators.jl`, `ProbNumDiffEq.jl`) and GPU are allow-failure lanes I did not verify locally.

No red check is attributable to this PR.

ChrisRackauckas-Claude:
Superseded by #4166, which merged and implements the same design more broadly — a dedicated `firk_operators.jl` for the complex-shifted operator, reusing `WOperator` for the real-shifted system. #4166 covers the FIRK methods generally where this PR was deliberately scoped to `RadauIIA5` in-place only.

Closing as a duplicate. Two things from this branch worth a second look against the merged version, offered as questions rather than claims:

1. This PR also touched `firk_addsteps.jl` (`_ode_addsteps!`), which #4166 does not appear to modify — its `firk_perform_step.jl` changes may already cover that path, but the dense-output/`addsteps!` route is worth confirming under a matrix-free `linsolve`, since it was a separate W-form update site here.
2. `stats.nf` under-reports the complex Krylov solve, because `dolinsolve` only counts when `A isa WOperator`. That was flagged as a known gap here rather than fixed, and it is not specific to either implementation — worth its own issue if #4166 inherited it.

Tracking issue #4156 remains the right home for anything still outstanding on FIRK + Krylov.

PR #4166, opened 2026-08-08
Matrix-free Jacobians / Krylov linsolve for the Radau FIRK methods

### LINKED ISSUE #4156: FIRK solvers do not support matrix-free Jacobians / Krylov linsolve

ChrisRackauckas-Claude:
Closes https://github.com/SciML/OrdinaryDiffEq.jl/issues/4156.

> **Stacked on https://github.com/SciML/OrdinaryDiffEq.jl/pull/4160** (`AdaptiveRadau` reads uninitialized stage-value slots). That fix has to land first or the `AdaptiveRadau` cases here are intermittently unstable for reasons unrelated to this change. Review only the second commit; the diff against master will shrink once #4160 merges.

## What changed

`RadauIIA3`, `RadauIIA5`, `RadauIIA9` and `AdaptiveRadau` now work with a matrix-free `linsolve`. The W-transform already decouples a step into one real shifted solve plus one shifted solve per complex-conjugate stage pair; each is `-γ M + J` for a different `γ`, so each can be applied lazily through the same real Jacobian-vector product instead of being assembled entrywise.

The real stage is `SciMLOperators.WOperator`, which already exists and already does this. The new piece is the complex stage: `ComplexWOperator` in `lib/OrdinaryDiffEqFIRK/src/firk_operators.jl`, a complex-`gamma` counterpart of `WOperator` that assembles the complex product from two real Jacobian-vector products (`J(a + bi) = Ja + i Jb`). That is what lets a real JVP operator — which cannot be applied to a complex vector at all — back the complex-conjugate pairs.

Everything else is plumbing at the four `alg_cache`/`perform_step!` sites: pick the lazy or the concrete stage matrix, and set `γ` on the lazy one instead of writing `n²` entries.

Two combinations that were broken independently of Krylov also start working, because `build_J_W` hands back a `WOperator` whenever a JVP exists and FIRK could not index into it:

  - `RadauIIA5(concrete_jac = true)` (and the other three) — previously `MethodError: no method matching setindex!(::WOperator, ...)`. `firk_real_W` now falls back to a concrete matrix when the linear solver wants one.
  - `RadauIIA3(linsolve = KrylovJL_GMRES())` — previously `MethodError: no method matching similar(::JVPCache, ::Type{ComplexF64})`; `RadauIIA3` had no guard at all.

`GaussLegendre` is unchanged: it solves a single coupled `num_stages * length(u)` system rather than decoupled shifted ones, so there is nothing to apply lazily without a new block operator. Its error message now says that, and says what to use instead.

## Failing before, passing after

The issue's reproducer, extended to the neighbouring combinations, on a 40-state 1D Brusselator at `reltol = abstol = 1e-10`.

Before (at the parent commit):

```
RadauIIA3(linsolve = KrylovJL_GMRES())                         ERROR: MethodError: no method matching similar(::OrdinaryDiffEqDifferentiation.JVPCache{Float64}, ::Type{ComplexF64})
RadauIIA5(linsolve = KrylovJL_GMRES())                         ERROR: Non-concrete Jacobian not yet supported by RadauIIA5.
RadauIIA9(linsolve = KrylovJL_GMRES())                         ERROR: Non-concrete Jacobian not yet supported by RadauIIA5.
AdaptiveRadau(linsolve = KrylovJL_GMRES())                     ERROR: Non-concrete Jacobian not yet supported by AdaptiveRadau.
RadauIIA5(linsolve = KrylovJL_GMRES(), concrete_jac = true)    ERROR: MethodError: no method matching setindex!(::SciMLOperators.WOperator{true, Float64, ...
RadauIIA5(concrete_jac = true)                                 ERROR: MethodError: no method matching setindex!(::SciMLOperators.WOperator{true, Float64, ...
GaussLegendre(num_stages = 3, linsolve = KrylovJL_GMRES())     ERROR: Non-concrete Jacobian not yet supported by GaussLegendre.
```

After:

```
RadauIIA3(linsolve = KrylovJL_GMRES())                         Success  nf=14149
RadauIIA5(linsolve = KrylovJL_GMRES())                         Success  nf=830
RadauIIA9(linsolve = KrylovJL_GMRES())                         Success  nf=588
AdaptiveRadau(linsolve = KrylovJL_GMRES())                     Success  nf=781
RadauIIA5(linsolve = KrylovJL_GMRES(), concrete_jac = true)    Success  nf=585
RadauIIA5(concrete_jac = true)                                 Success  nf=289
GaussLegendre(num_stages = 3, linsolve = KrylovJL_GMRES())     ERROR: GaussLegendre does not support a matrix-free Jacobian: unlike the Radau methods it solves a single coupled 120-by-120 system rather than decoupled shifted systems, ...
```

The new test file cannot run at all before this commit — it references `ComplexWOperator`, which does not exist there — so the reproducer above is the before/after evidence rather than the test suite.

## Accuracy against a tight reference

Same problem, `reltol = abstol = 1e-10`, reference from `RadauIIA5` at `1e-14`. `cj-kry` is `concrete_jac = true` with GMRES, `cj-lu` is `concrete_jac = true` with the default factorization. `massmat` repeats it with a non-identity diagonal mass matrix.

```
plain    RadauIIA3      dense 8.44e-12 | krylov 8.44e-12 | cj-kry 8.44e-12 | cj-lu 8.44e-12
plain    RadauIIA5      dense 5.10e-09 | krylov 2.15e-10 | cj-kry 2.15e-10 | cj-lu 5.10e-09
plain    RadauIIA9      dense 1.34e-09 | krylov 4.76e-09 | cj-kry 4.76e-09 | cj-lu 1.34e-09
plain    AdaptiveRadau  dense 1.99e-09 | krylov 4.75e-09 | cj-kry 4.75e-09 | cj-lu 1.99e-09
massmat  RadauIIA3      dense 2.54e-11 | krylov 2.54e-11 | cj-kry 2.54e-11 | cj-lu 2.54e-11
massmat  RadauIIA5      dense 1.45e-08 | krylov 5.42e-09 | cj-kry 5.42e-09 | cj-lu 1.45e-08
massmat  RadauIIA9      dense 4.42e-08 | krylov 4.44e-07 | cj-kry 4.44e-07 | cj-lu 4.42e-08
massmat  AdaptiveRadau  dense 1.08e-08 | krylov 6.12e-08 | cj-kry 6.12e-08 | cj-lu 1.08e-08
```

`cj-lu` reproduces the dense-Jacobian result exactly, as it should — that path assembles and factorizes the same `W`. The matrix-free results sit within an order of magnitude of the dense ones, in both directions; the residual difference is the usual inexact-Newton effect of solving each stage system to `reltol` instead of exactly, and it shrinks with the tolerance (`massmat RadauIIA9` goes 4.4e-7 → 2.4e-7 on tightening `1e-9 → 1e-11`).

Separately, `ComplexWOperator` is unit-tested against its dense form for `I`, `λI`, a diagonal, and a full mass matrix.

## `nf` accounting

Each complex stage solve spends two real Jacobian-vector products per Krylov iteration, and `dolinsolve` cannot see them — it only accounts for `WOperator`. The operator counts its own products and drains them into the stats after each Newton iteration (a counter on the operator rather than one inferred from `linres.iters`, so `AdaptiveRadau`'s threaded stage solves each tally into their own operator instead of racing on `integrator.stats`).

Counting the RHS calls directly and comparing against `sol.stats.nf`, on a 12-state linear stiff problem at `1e-9`:

```
RadauIIA3      dense   actual nf  39827   reported  39827   ratio 1.000
RadauIIA3      krylov  actual nf  18669   reported  18669   ratio 1.000
RadauIIA3      kry+cj  actual nf  50505   reported  50505   ratio 1.000
RadauIIA3      fd-kry  actual nf  29335   reported  29335   ratio 1.000
RadauIIA5      dense   actual nf    414   reported    414   ratio 1.000
RadauIIA5      krylov  actual nf   1407   reported   1407   ratio 1.000
RadauIIA5      kry+cj  actual nf   1419   reported    900   ratio 0.634
RadauIIA5      fd-kry  actual nf   2492   reported   2492   ratio 1.000
RadauIIA9      dense   actual nf    143   reported    143   ratio 1.000
RadauIIA9      krylov  actual nf    779   reported    779   ratio 1.000
RadauIIA9      kry+cj  actual nf    791   reported    596   ratio 0.753
RadauIIA9      fd-kry  actual nf   1432   reported   1432   ratio 1.000
AdaptiveRadau  dense   actual nf    181   reported    181   ratio 1.000
AdaptiveRadau  krylov  actual nf    881   reported    881   ratio 1.000
AdaptiveRadau  kry+cj  actual nf    893   reported    665   ratio 0.745
AdaptiveRadau  fd-kry  actual nf   1641   reported   1641   ratio 1.000
```

Exact for the fully matrix-free path with both AD backends; the test suite asserts `sol.stats.nf == actual` for all four methods × `AutoForwardDiff`/`AutoFiniteDiff`.

The `kry+cj` undercount is **pre-existing and not FIRK-specific**: `dolinsolve` gates its accounting on `linsolve.A isa WOperator && linsolve.A.J isa AbstractSciMLOperator`, and with `concrete_jac = true` the `J` is a concrete matrix even though the products go through `jacvec`. That affects every Newton-Krylov integrator (SDIRK, BDF, Rosenbrock) and belongs in a separate PR against `OrdinaryDiffEqDifferentiation`. `RadauIIA3` reads 1.000 there only because it has no real stage solve, so all of its products go through the new operator.

## Verification

```
$ GROUP=Core julia --project=lib/OrdinaryDiffEqFIRK -e 'using Pkg; Pkg.test()'
FIRK Tests    |  107    107  3m45.4s
FIRK Krylov Tests |   93     93  5m43.2s
core exit=0
```

```
$ GROUP=QA julia --project=lib/OrdinaryDiffEqFIRK -e 'using Pkg; Pkg.test()'
Test Summary:                                  | Pass  Fail  Total     Time
Aqua                                           |   19     1     20  1m17.1s
    ExplicitImports                            |    6            6    29.3s
    Public API documentation                   |          1      1     2.8s
      public API has docstrings                |          1      1     2.8s   <-- isempty([:SciMLBase])
```

The single QA failure is **pre-existing**. The same run on this branch's parent gives the identical `19 passed, 1 failed` with the same `:SciMLBase` docstring check, and `lib/OrdinaryDiffEqFIRK [QA]` is already red on master upstream: https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/31198062683.

An earlier revision of this PR did add three real `ExplicitImports` failures (`alg_autodiff` imported from a non-owner, `SciMLOperators.getops`, `LinearSolve.needs_concrete_A`). Two are gone — `alg_autodiff` is replaced by the algorithm's own public `autodiff` field, and the speculative `getops` method is dropped. `needs_concrete_A` is allow-listed in `test/qa/qa.jl` with a comment: it is the only way to ask whether a linear solver can consume an operator `A`, `build_J_W` in `OrdinaryDiffEqDifferentiation` reaches for the same name and ignores it the same way, and it is a make-public candidate upstream in LinearSolve.

```
$ julia -e 'using Runic; exit(Runic.main(["--check","--diff","lib/OrdinaryDiffEqFIRK/"]))'   # clean
$ typos lib/OrdinaryDiffEqFIRK/ docs/src/implicit/FIRK.md                                    # clean
```

## What I did not verify

- **Docs build fails on this branch and on its parent, identically** — `Cannot resolve @ref for md"[`DummyControllerCache`](@ref)"`, then `makedocs encountered an error [:cross_references]`. Already red on master upstream for the same reason (https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/31198061733). Not introduced here; https://github.com/SciML/OrdinaryDiffEq.jl/pull/4165 fixes it, and with that one-line change applied the full build succeeds, so the `docs/src/implicit/FIRK.md` addition here does render.
- Preconditioners. `wrapprecs` is commented out in the FIRK caches and stays that way; `concrete_jac = true` gives you the concrete `J` to build one from, but nothing wires it up.
- GPU and downstream lanes.
- Out-of-place (`Val{false}`) caches. They build a dense `J` directly and are untouched.
- `lib/OrdinaryDiffEqFIRK/src/firk_addsteps.jl` carries a near-duplicate of the same W assembly and is **not** updated. Those methods take `(integrator, cache[, repeat_step])`, while every `_ode_addsteps!` call site in `OrdinaryDiffEqCore` passes 11 arguments — I checked every call site's arity and none is ≤ 4, so they are unreachable. Rather than ship an untestable parallel change I left them; if they are ever wired up they will need the same treatment (or, better, to stop being a copy of `perform_step!`).

## Things worth pushing back on

1. **The lazy `W` is re-formed every step, ignoring the `do_newW` reuse heuristic.** Setting `γ` on an operator is O(1) and no factorization follows it, so there is nothing to amortize and a current `W` is a strictly better Newton matrix. It also sidesteps a staleness hazard specific to `AdaptiveRadau`: `do_newW` looks only at `dt`, so when the controller raises `num_stages` the newly used `W2[i]` would otherwise still hold whatever shift it was last given — for a slot never used before, that is the construction-time placeholder. The concrete path keeps the original behaviour untouched.
2. **`integrator.stats.nw` therefore increments every step on the matrix-free path**, where the concrete path increments only when `do_newW` says so. That is accurate — `W` really is re-formed each step — but it does make `nw` mean something slightly different between the two paths.
3. **`ComplexWOperator` is internal**, not exported and not `public`. It is reachable only through `alg_cache`, and I would rather not commit to it as API before someone has used FIRK+Krylov in anger.
4. **The `massmat RadauIIA9` accuracy gap (4.4e-7 vs 4.4e-8 dense)** is the largest divergence in the table. I attribute it to inexact Newton and checked that it converges with tolerance, but a reviewer who wants the linear solves driven tighter than `integrator.opts.reltol` should say so.

---

https://claude.ai/code/session_01Gstik9KAfxtNAz53ejBPwN

### DIFF
--- a/docs/src/implicit/FIRK.md
+++ b/docs/src/implicit/FIRK.md
@@ -63,7 +63,25 @@ Gauss–Legendre methods use the roots of the Legendre polynomial directly as co
 ### System size considerations
 
   - **Systems < 200**: FIRK methods are competitive due to better multithreading
-  - **Systems > 200**: Consider SDIRK or BDF methods instead
+  - **Systems > 200**: Consider SDIRK or BDF methods instead, or keep a Radau method and
+    switch to a matrix-free linear solver (see below)
+
+## Matrix-free (Newton-Krylov) usage
+
+The RadauIIA methods and `AdaptiveRadau` accept a matrix-free `linsolve`, which avoids
+forming a Jacobian at all:
+
+```julia
+solve(prob, RadauIIA5(linsolve = KrylovJL_GMRES()))
+```
+
+The W-transform decouples each step into one real shifted solve and one shifted solve per
+complex-conjugate stage pair, and every one of them is applied through the same real
+Jacobian-vector product. `concrete_jac = true` keeps a concrete Jacobian alongside the
+matrix-free application, which is what a preconditioner needs.
+
+`GaussLegendre` has no matrix-free path: it solves a single coupled
+`num_stages * length(u)` block system rather than decoupled shifted systems.
 
 ## Performance Guidelines
 
--- a/lib/OrdinaryDiffEqFIRK/src/OrdinaryDiffEqFIRK.jl
+++ b/lib/OrdinaryDiffEqFIRK/src/OrdinaryDiffEqFIRK.jl
@@ -25,6 +25,7 @@ import DiffEqBase: initialize!, calculate_residuals, calculate_residuals!
 using MuladdMacro: @muladd
 using RecursiveArrayTools: recursivefill!
 import Polyester
+import LinearAlgebra
 using LinearAlgebra: I, UniformScaling, mul!, lu, dot
 import LinearSolve
 import FastBroadcast: @..
@@ -35,6 +36,7 @@ import FastPower: fastpower
 using OrdinaryDiffEqDifferentiation: build_J_W, build_jac_config,
     calc_J!, dolinsolve, calc_J,
     islinearfunction
+import ADTypes
 import ADTypes: AutoForwardDiff
 import SciMLOperators
 import SciMLOperators: AbstractSciMLOperator
@@ -64,6 +66,7 @@ include("algorithms.jl")
 include("alg_utils.jl")
 include("controllers.jl")
 include("firk_caches.jl")
+include("firk_operators.jl")
 include("firk_tableaus.jl")
 include("firk_perform_step.jl")
 include("firk_interpolants.jl")
--- a/lib/OrdinaryDiffEqFIRK/src/firk_caches.jl
+++ b/lib/OrdinaryDiffEqFIRK/src/firk_caches.jl
@@ -107,8 +107,8 @@ function alg_cache(
     jac_config = build_jac_config(alg, f, uf, du1, uprev, u, tmp, dw12)
 
     J, W1 = build_J_W(alg, u, uprev, p, t, dt, f, jac_config, uEltypeNoUnits, Val(true))
-    W1 = similar(J, Complex{eltype(W1)})
-    recursivefill!(W1, false)
+    W1 = firk_real_W(alg, J, W1)
+    W1 = build_complex_W(alg, f, u, J, W1)
 
     linprob = LinearProblem(W1, _vec(cubuff), (nothing, u, p, t); u0 = _vec(dw12))
     linsolve = init(
@@ -258,11 +258,8 @@ function alg_cache(
     jac_config = build_jac_config(alg, f, uf, du1, uprev, u, tmp, dw1)
 
     J, W1 = build_J_W(alg, u, uprev, p, t, dt, f, jac_config, uEltypeNoUnits, Val(true))
-    if J isa AbstractSciMLOperator
-        error("Non-concrete Jacobian not yet supported by RadauIIA5.")
-    end
-    W2 = similar(J, Complex{eltype(W1)})
-    recursivefill!(W2, false)
+    W1 = firk_real_W(alg, J, W1)
+    W2 = build_complex_W(alg, f, u, J, W1)
 
     linprob = LinearProblem(W1, _vec(ubuff), (nothing, u, p, t); u0 = _vec(dw1))
     linsolve1 = init(
@@ -462,13 +459,9 @@ function alg_cache(
     jac_config = build_jac_config(alg, f, uf, du1, uprev, u, tmp, dw1)
 
     J, W1 = build_J_W(alg, u, uprev, p, t, dt, f, jac_config, uEltypeNoUnits, Val(true))
-    if J isa AbstractSciMLOperator
-        error("Non-concrete Jacobian not yet supported by RadauIIA5.")
-    end
-    W2 = similar(J, Complex{eltype(W1)})
-    W3 = similar(J, Complex{eltype(W1)})
-    recursivefill!(W2, false)
-    recursivefill!(W3, false)
+    W1 = firk_real_W(alg, J, W1)
+    W2 = build_complex_W(alg, f, u, J, W1)
+    W3 = build_complex_W(alg, f, u, J, W1)
 
     linprob = LinearProblem(W1, _vec(ubuff), (nothing, u, p, t); u0 = _vec(dw1))
     linsolve1 = init(
@@ -685,12 +678,8 @@ function alg_cache(
     jac_config = build_jac_config(alg, f, uf, du1, uprev, u, zero(u), dw1)
 
     J, W1 = build_J_W(alg, u, uprev, p, t, dt, f, jac_config, uEltypeNoUnits, Val(true))
-    if J isa AbstractSciMLOperator
-        error("Non-concrete Jacobian not yet supported by AdaptiveRadau.")
-    end
-
-    W2 = [similar(J, Complex{eltype(W1)}) for _ in 1:((max_stages - 1) ÷ 2)]
-    recursivefill!.(W2, false)
+    W1 = firk_real_W(alg, J, W1)
+    W2 = [build_complex_W(alg, f, u, J, W1) for _ in 1:((max_stages - 1) ÷ 2)]
 
     linprob = LinearProblem(W1, _vec(ubuff), (nothing, u, p, t); u0 = _vec(dw1))
     linsolve1 = init(
@@ -842,7 +831,14 @@ function alg_cache(
 
     J, _ = build_J_W(alg, u, uprev, p, t, dt, f, jac_config, uEltypeNoUnits, Val(true))
     if J isa AbstractSciMLOperator
-        error("Non-concrete Jacobian not yet supported by GaussLegendre.")
+        error(
+            "GaussLegendre does not support a matrix-free Jacobian: unlike the Radau " *
+                "methods it solves a single coupled $(num_stages * n)-by-$(num_stages * n) " *
+                "system rather than decoupled shifted systems, and no operator form of that " *
+                "system is implemented. Use `RadauIIA3`, `RadauIIA5`, `RadauIIA9` or " *
+                "`AdaptiveRadau` with a Krylov `linsolve`, or set `concrete_jac = true` " *
+                "together with a factorization `linsolve`."
+        )
     end
 
     W = similar(J, num_stages * n, num_stages * n)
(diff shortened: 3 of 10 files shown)


ChrisRackauckas (approved):

termi-official:
How do we inform the used preconditioner about the special block structure for the block-stages? (xref https://arxiv.org/abs/2001.00711 or https://arxiv.org/pdf/2101.00512 section 3)

ChrisRackauckas:
There are no block stages for Radau, so nothing required there, as long as the preconditioner works on the complex form.

PR #4181, opened 2026-08-09
Test DDE discontinuity handling at tspan[1]

### LINKED ISSUE #4180: `initialize_d_discontinuities` does not filter out `t0`, stalling DDEs at the first propagated discontinuity
`initialize_d_discontinuities` keeps entries at `t0`, unlike `initialize_tstops`. For DDEs such an entry is never popped, and it blocks every later discontinuity, so `MethodOfSteps` stalls with `dt == 0` at the first propagated discontinuity.

## Reproducer

```julia
using DelayDiffEq  # v5.69.0, OrdinaryDiffEqCore v3.25.0

f(du, u, h, p, t) = (du[1] = -h(p, t - 10.0)[1]; nothing)
hist(p, t) = ones(1)
prob = DDEProblem(f, [1.0], hist, (0.0, 50.0); constant_lags = (10.0,))

solve(prob, MethodOfSteps(Tsit5()); d_discontinuities = Float64[]).retcode  # Success
solve(prob, MethodOfSteps(Tsit5()); d_discontinuities = [0.0]).retcode      # DtLessThanMin
```

The second solve stops at exactly `t = 10.0`, i.e. `t0 + lag`:

```
┌ Warning: Verbosity toggle: dt_min_unstable
│  dt(0.0 <= dtmin(2.220446049250313e-16), at t=10.0, and step error estimate = 1.3876246002703048e-14. Aborting.
```

`dt` is exactly zero, not merely small.

Isolating variables, on the same problem:

| `d_discontinuities` | `constant_lags` | retcode | `sol.t[end]` |
| --- | --- | --- | --- |
| `Float64[]` | `(10.0,)` | `Success` | 50.0 |
| `[0.0]` | `(10.0,)` | `DtLessThanMin` | 10.0 |
| `[0.0, 0.0]` | `(10.0,)` | `DtLessThanMin` | 10.0 |
| `[5.0]` | `(10.0,)` | `Success` | 50.0 |
| `[5.0, 5.0]` | `(10.0,)` | `Success` | 50.0 |
| `[0.0]` | `nothing` | `Success` | 50.0 |

So it is specifically an entry at `t0` combined with a propagated discontinuity. Duplicate entries are handled fine, and a single entry is enough to trigger it.

## Mechanism

1. `initialize_tstops` (`lib/OrdinaryDiffEqCore/src/solve.jl:922`) filters user times to the open interval `tdir * t0 < tdir * t < tdir * tf`. `initialize_d_discontinuities` (`lib/OrdinaryDiffEqCore/src/solve.jl:1019`) applies no filter — it destructures `t0, tf = tspan` and then only uses `tf` via `tdir`. So a discontinuity at exactly `t0` enters the heap as `Discontinuity(t0, 0)`.

2. `loopheader!` skips `apply_step!` while `iter == 0` (`integrator_utils.jl:31`), and `handle_discontinuities!` is only reachable through `apply_step! → update_fsal!` (`integrator_utils.jl:131-133`). The entry at `t0` is therefore never popped, and it stays at the head of the min-heap for the whole solve.

3. That head-of-heap entry is load-bearing in DelayDiffEq. `handle_discontinuities!` (`DelayDiffEq/src/integrators/utils.jl:191-232`) is the only code path that removes *propagated* tstops which fall strictly between two user `tstops`: `tstops_propagated` is a separate heap from `opts.tstops`, and `solve!`'s outer `handle_tstop!` is only reached once `tdir * t >= first(opts.tstops)` (`DelayDiffEq/src/solve.jl:593-613`).

4. At `t = 10.0` the guard in `update_fsal!` evaluates `first_discontinuity(integrator) == tdir * t`, i.e. `0.0 == 10.0` — false, because the stuck `t0` entry is still the minimum. `handle_discontinuities!` does not run, so the propagated tstop at `10.0` is never popped. `modify_dt_for_tstops!` (`integrator_utils.jl:175`) then computes `distance_to_tstop = 0` on every subsequent iteration, and `check_error!` returns `DtLessThanMin`.

For plain ODEs the same stuck entry is harmless: `handle_discontinuities!` is a bare `pop_discontinuity!` there and nothing else consults the heap.

## Suggested fix

Apply the same filter in `initialize_d_discontinuities` and `reinit_d_discontinuities!` that `initialize_tstops` already uses:

```julia
for t in d_discontinuities
    tdir_t = tdir * t
    tdir_t0 < tdir_t < tdir_tf && push!(d_discontinuities_internal, tdir_t)
end
```

Nothing is lost. For plain ODEs an entry at `t0` is dead weight, and for DDEs the discontinuity at `t0` is already carried by `order_discontinuity_t0`. The current code keeps entries beyond `tf` as well, which the same filter removes.

DelayDiffEq is only where this becomes visible, so a filter on the DelayDiffEq side would also work — but the asymmetry between the two `initialize_*` functions in OrdinaryDiffEqCore looks like the actual defect, especially since `t0` is destructured there and then unused.

Unrelated minor observation while tracing this: `DelayDiffEq.initialize_tstops_d_discontinuities_propagated` (`DelayDiffEq/src/solve.jl:683`) takes `tstops` and `d_discontinuities` parameters that its body never reads.

## Versions

Reproduced on Julia 1.12.6 with DelayDiffEq v5.69.0 and OrdinaryDiffEqCore v3.25.0. `initialize_d_discontinuities` is unchanged in v3.23.0, v3.31.0 and v3.32.0.

Found while debugging a downstream package that passes covariate times as `d_discontinuities`, where the first covariate time coincides with `t0`.

ChrisRackauckas-Claude:
> Ignore this PR until reviewed by @ChrisRackauckas.

## What changed and why

This adds a DelayDiffEq regression test for a user derivative discontinuity exactly at `tspan[1]` with a constant lag. The test verifies that initialization handles the discontinuity instead of filtering it: the integrator advances to the post-discontinuity side, Tsit5's FSAL derivative is restarted there, and the propagated discontinuity no longer stalls the solve.

The behavior is already implemented on `master` by the starting-time handler merged in https://github.com/SciML/OrdinaryDiffEq.jl/pull/3394; this PR adds the missing DDE coverage requested in https://github.com/SciML/OrdinaryDiffEq.jl/issues/4180.

Closes https://github.com/SciML/OrdinaryDiffEq.jl/issues/4180

### Failing before

I temporarily removed the existing `OrdinaryDiffEqCore.handle_starting_time_discontinuity!(integrator)` call from DelayDiffEq initialization and ran the new test in isolation:

```sh
julia --project=lib/DelayDiffEq -e 'push!(LOAD_PATH, abspath("lib/OrdinaryDiffEqTsit5")); using DelayDiffEq, OrdinaryDiffEqTsit5, SciMLBase, Test; @testset "discontinuity at tspan[1]" begin history(p,t)=zeros(1); function f(du,u,h,p,t); du[1]=t>0.0 ? 1.0 : 0.0; nothing; end; prob=DDEProblem(f,[0.0],history,(0.0,2.0);constant_lags=(1.0,)); integrator=init(prob,MethodOfSteps(Tsit5());d_discontinuities=[prob.tspan[1]]); @test integrator.t == nextfloat(prob.tspan[1]); @test integrator.fsalfirst == [1.0]; sol=solve!(integrator); @test sol.retcode == ReturnCode.Success; @test sol.t[end] == prob.tspan[2]; @test sol.u[end] ≈ [2.0]; end'
```

```text
integrator.t: 0.0 == 5.0e-324 failed
integrator.fsalfirst: [0.0] == [1.0] failed
dt(0.0) <= dtmin(...) at t=1.0
sol.t[end]: 1.0 == 2.0 failed
Test Summary:             | Fail  Total  Time
discontinuity at tspan[1] |    5      5  7.1s
```

### Passing after

After restoring the handler, the same command produced:

```text
Test Summary:             | Pass  Total  Time
discontinuity at tspan[1] |    5      5  4.4s
```

The full relevant group passed:

```sh
GROUP=Interface julia --project=lib/DelayDiffEq -e 'using Pkg; Pkg.test()'
```

```text
Test Summary:       | Pass  Total   Time
Discontinuity Tests |   53     53  24.8s
Multi-Algorithm Tests |   28     28  4m01.9s
Testing DelayDiffEq tests passed
```

Formatting and spelling passed:

```sh
julia --project=@runic -m Runic --check lib/DelayDiffEq/test/interface/discontinuities.jl
git diff --unified=0 -- lib/DelayDiffEq/test/interface/discontinuities.jl | typos -
```

Both commands exited 0 with no output.

The QA command was also run:

```sh
GROUP=QA julia --project=lib/DelayDiffEq -e 'using Pkg; Pkg.test()'
```

It exposed unrelated failures already present on the synced base: three `has_tstop`/`first_tstop`/`pop_tstop!` qualified accesses through `OrdinaryDiffEqCore` instead of their `SciMLBase` owner, plus the `SciMLBase` public API documentation check. A clean `upstream/master` reproduction with `GROUP=DelayDiffEq_QA ~/.juliaup/bin/julia +1.10 --project=. -e 'using Pkg; Pkg.test()'` confirmed the same ExplicitImports failures (QA summary: 14 passed, 2 failed, 1 errored). The separate investigation at https://github.com/SciML/OrdinaryDiffEq.jl/issues/4182 traces the wrong qualifiers to commit `e0fb409ab96`; the later `run_qa` conversion exposed them.

## Not verified

I did not run `GROUP=Everything`, GPU jobs, downstream jobs, or a docs build. This change touches only a test and adds no dependency or public API. The direct `integrator.fsalfirst` assertion is intentional: it ensures initialization actually restarts FSAL rather than only removing the queue entry.

### DIFF
--- a/lib/DelayDiffEq/test/interface/discontinuities.jl
+++ b/lib/DelayDiffEq/test/interface/discontinuities.jl
@@ -104,6 +104,26 @@ end
         [Discontinuity(0.3, 4), Discontinuity(0.6, 5)]
 end
 
+@testset "discontinuity at tspan[1]" begin
+    history(p, t) = zeros(1)
+    function f(du, u, h, p, t)
+        du[1] = t > 0.0 ? 1.0 : 0.0
+        return nothing
+    end
+    prob = DDEProblem(f, [0.0], history, (0.0, 2.0); constant_lags = (1.0,))
+    integrator = init(
+        prob, MethodOfSteps(Tsit5()); d_discontinuities = [prob.tspan[1]]
+    )
+
+    @test integrator.t == nextfloat(prob.tspan[1])
+    @test integrator.fsalfirst == [1.0]
+
+    sol = solve!(integrator)
+    @test sol.retcode == ReturnCode.Success
+    @test sol.t[end] == prob.tspan[2]
+    @test sol.u[end] ≈ [2.0]
+end
+
 # discontinuities induced by callbacks
 @testset "#190" begin
     cb = ContinuousCallback((u, t, integrator) -> 1, integrator -> nothing)


ChrisRackauckas:
Yeah @devmotion I was confused by the issue because I thought I already handled this. It seems this test passes? So the issue is moot.

devmotion:
Possibly Claude was using old versions? It definitely caused problems downstream I could fix by filtering `d_discontinuities` before handing it to DelayDiffEq - but I'm pinned to an older version.

devmotion:
Confirmed — this is fixed on master, and the suggestion in the issue to filter `t0` like `tstops` was the wrong prescription. Handling the discontinuity is correct.

The version skew was real. I reported against `DelayDiffEq v5.69.0` / `OrdinaryDiffEqCore v3.25.0`, which is an exact-equality pin in the downstream package I was debugging. Running the issue's reproducer unchanged across releases:

| DelayDiffEq | OrdinaryDiffEqCore | `d_discontinuities = [0.0]` | `sol.t[end]` |
| --- | --- | --- | --- |
| 5.69.0 | 3.25.0 | `DtLessThanMin` | 10.0 |
| 5.73.0 | 3.30.0 | `DtLessThanMin` | 10.0 |
| 5.74.0 | 3.33.1 | `Success` | 50.0 |

So the first fixed releases are **DelayDiffEq v5.74.0** and **OrdinaryDiffEqCore v3.31.0**, and the fix is #3394 (`e45efdbef`, merged 2026-04-12) — four months before I filed.

Where my report went wrong: I checked whether `initialize_d_discontinuities` had changed across v3.23–v3.32, found it had not, and concluded the defect was still live. The fix is a new function, `handle_starting_time_discontinuity!`, elsewhere in the same file, so that check missed it. Its comment — "The regular `update_fsal!` mechanism only fires at iter > 0, so handle the iter==0 case explicitly here" — is the same mechanism the issue describes, so the diagnosis and the existing fix agree.

Two notes for anyone arriving here from a pinned environment:

- Both halves are needed. `lib/DelayDiffEq/src/solve.jl` makes its own call to `OrdinaryDiffEqCore.handle_starting_time_discontinuity!`, so bumping only `OrdinaryDiffEqCore` does not fix DDEs.
- The shift past `t0` is visible on `integrator.t` right after `init` (`5.0e-324`), as the new test asserts, but the initial save happens before it — `sol.t[1]` is still `0.0`. The saved solution is unaffected.

Nothing outstanding upstream from my side.

PR #4208, opened 2026-08-11
Add serial spectral deferred correction (SDC) solver family

singhharsh1708:
Adds `OrdinaryDiffEqSDC`, a sublibrary with a serial spectral deferred correction solver. SDC sweeps the collocation problem on each step with a cheap preconditioner instead of solving it directly, so accuracy is a runtime parameter: `num_nodes` sets the ceiling, `num_sweeps` how far toward it, and one method covers order 1 up to the collocation order (`2M` Gauss, `2M-1` Radau, `2M-2` Lobatto) without new coefficients.

`SDC(; num_nodes = 3, node_type = SDCNodes.Legendre, quad_type = SDCQuadrature.RadauRight, num_sweeps = 3, sweeper = SDCSweeper.BE, step_update = SDCStepUpdate.Quadrature, nlsolve = NLNewton(), linsolve, autodiff, concrete_jac)`; `SDCSweeper.LU` is the preconditioner for stiff problems. Fixed step only for now, so `adaptive = false` and a `dt` are required. Not re-exported from the umbrella until the package is registered.

Two design points worth a look: one `NLSolver` per node, because each node has its own `gamma = QDelta[m,m]` and `do_newJW` keys `W` reuse on `J_t` without comparing `gamma` (the `PDIRK44` precedent); and coefficients depend only on the node set, so they are built once in `alg_cache`, nodes from Jacobi polynomial roots and `Q` assembled at extended precision before rounding.

Coefficients match qmat, the reference SDC implementation, to `atol = 1e-13` across 6 node configurations times 7 sweepers, with the generator checked in at `test/gen_reference.py`. The convergence gate mirrors qmat's `test_timestepping.py` and 470 combinations pass, edges included (`M = 4` Radau-right with `K = 6` measures 6.79 against a ceiling of 7). 2230 assertions in total, with Aqua, JET and ExplicitImports.

## Benchmarks

Fixed-step work-precision on two problems, run locally on Julia 1.10.10, Apple M1,
single threaded, minimum of 3 samples per point. Reference solution from `Vern9` at
`abstol = reltol = 1e-14`; error is relative L2 at the final time. Every method runs
with `adaptive = false`, so this compares method quality rather than step control.

The natural comparator is `RadauIIA5`. SDC on Radau-Right nodes converges to exactly
that collocation solution, so it is the fixed point serial SDC iterates toward.

Work-precision diagrams. Each series is truncated where it reaches the roundoff floor,
so a line ending is the method running out of achievable accuracy, not diverging.

| | |
|---|---|
| Lotka-Volterra, error vs time | Lotka-Volterra, error vs f evals |
| Van der Pol, error vs time | Van der Pol, error vs f evals |

On Lotka-Volterra the SDC(M=5,K=9) line crosses below RadauIIA5 at the tight end. On
Van der Pol RadauIIA5 stays lowest and leftmost at every accuracy. The f-eval panels show
where the cost goes: the SDC curves sit far to the right of the RK ones at equal error.

Fastest method reaching each accuracy target:

**Lotka-Volterra (non-stiff, t = 10)**

| target | fastest | time (ms) | runner-up | time (ms) |
|---|---|---|---|---|
| 1e-6 | RadauIIA5 | 0.435 | KenCarp4 | 0.821 |
| 1e-9 | RadauIIA5 | 1.649 | KenCarp4 | 3.313 |
| 1e-12 | SDC(M=5,K=9) | 4.351 | RadauIIA5 | 6.237 |
| 1e-14 | SDC(M=5,K=9) | 6.526 | SDC(M=3,K=5) | 47.150 |

At 1e-14 no non-SDC method in the set reaches the target anywhere in the dt range tested.

**Van der Pol mu=10 (stiff, t = 3)**

| target | fastest | time (ms) | best SDC | time (ms) |
|---|---|---|---|---|
| 1e-6 | RadauIIA5 | 0.029 | SDC(M=3,K=3) | 0.679 |
| 1e-9 | RadauIIA5 | 0.132 | SDC(M=3,K=5) | 1.067 |
| 1e-12 | RadauIIA5 | 0.497 | SDC(M=3,K=5) | 2.335 |
| 1e-14 | RadauIIA5 | 1.012 | SDC(M=5,K=9) | 3.495 |

Serial SDC wins in the tight-tolerance non-stiff regime, where high order pays for the
extra implicit solves. On the stiff problem RadauIIA5 leads at every target, though the
gap narrows from 23x at 1e-6 to 3.5x at 1e-14.

That is the expected shape for vanilla serial SDC, and the cost driver shows up in the
f-eval counts: SDC does `num_nodes * num_sweeps` implicit solves per step, so
SDC(M=3,K=5) spends 20412 f evals where RadauIIA5 spends 4612 for the same 2.5e-13 error
on Van der Pol. Parallel-across-the-nodes and IMEX are what buy that back, and both are
follow-up work rather than part of this PR.

Convergence orders were checked separately against `alg_order`: SDC(M=2,K=2) observed
3.0 and SDC(M=3,K=3) observed 4.0 on a scalar linear problem. Higher configurations
reach the roundoff floor before an order can be measured.

##


singhharsh1708:
added wpd.

lib/OrdinaryDiffEqSDC/src/sdc_perform_step.jl
```diff
@@ -0,0 +1,141 @@
...
+                f(k, ubuf, p, t + nodes[m] * dt)
+                OrdinaryDiffEqCore.increment_nf!(stats, 1)
+                @.. broadcast = false zk1[m] = dt * k
+            else
+                nls = nlsolvers[index]
+                @.. broadcast = false nls.tmp = tmp
+                @.. broadcast = false nls.z = zk[m]
+                nls.γ = QΔ[m, m]
+                nls.c = nodes[m]
+                markfirststage!(nls)
```
ChrisRackauckas, line 61: why is it marking it as the first stage each time?

lib/OrdinaryDiffEqSDC/src/sdc_perform_step.jl
```diff
@@ -0,0 +1,141 @@
...
+            if iszero(index)
+                ulast = tmp
+                zk1[m] = dt * f(ulast, p, t + nodes[m] * dt)
+                OrdinaryDiffEqCore.increment_nf!(stats, 1)
+            else
+                nls = nlsolvers[index]
+                nls.tmp = tmp
+                nls.z = zk[m]
+                nls.γ = QΔ[m, m]
+                nls.c = nodes[m]
```
ChrisRackauckas, line 119: What justifies this stage predictor?

lib/OrdinaryDiffEqSDC/src/sdc_tableaus.jl
```diff
@@ -0,0 +1,325 @@
...
+
+Written in the monomial basis: the Lagrange coefficients are `V⁻¹` for the
+Vandermonde matrix `V[k, i] = τₖ^{i-1}` (since `ℓⱼ(τₖ) = δⱼₖ`), and the monomial
+integrals are `C[m, i] = uppers[m]^i / i`, so the result is `C V⁻¹`.
+"""
+function _lagrange_integrals(τ::Vector{BigFloat}, uppers::Vector{BigFloat})
+    M = length(τ)
+    V = [τ[k]^(i - 1) for k in 1:M, i in 1:M]
+    C = [uppers[m]^i / i for m in 1:length(uppers), i in 1:M]
+    return C / V
```
ChrisRackauckas, line 226: Using bigfloats is going to be pretty slow, does it need to be higher precision?

lib/OrdinaryDiffEqSDC/test/gen_reference.py
```diff
@@ -0,0 +1,104 @@
...
+from qmat.utils.sdc import getOrderSDC, errorDahlquistSDC
+from qmat.utils.num import numericalOrder
+
+NODE_TYPES = ["LEGENDRE", "EQUID"]
+QUAD_TYPES = ["GAUSS", "RADAU-LEFT", "RADAU-RIGHT", "LOBATTO"]
+SWEEPERS = ["BE", "FE", "TRAP", "LU", "PIC", "BEPAR", "MIN-SR-NS"]
+
+out = {"coeffs": [], "orders": [], "measured": []}
+
+for nodeType in NODE_TYPES:
```
ChrisRackauckas, line 20: What's this? Why not make it Julia?

ChrisRackauckas:
It seems correct, so I'll merge, but follow up with improvements based on the comments I gave.

lib/OrdinaryDiffEqSDC/src/sdc_perform_step.jl
singhharsh1708: Each node has its own gamma = QDelta[m,m], and W reuse is keyed on `isfirststage`. Reusing W across sweeps within a node would be a real win, can follow up on that.

lib/OrdinaryDiffEqSDC/src/sdc_perform_step.jl
singhharsh1708: `zk[m]` is the previous sweep iterate at that node, so it is the natural initial guess; a cold start costs about one extra Newton iteration per node.

lib/OrdinaryDiffEqSDC/src/sdc_tableaus.jl
singhharsh1708: Only at init, 2 to 12 ms for M = 3 to 8; the Vandermonde solve is ill-conditioned in Float64, hence BigFloat. Can memoize like `RadauIIATableauCache` if that matters.

lib/OrdinaryDiffEqSDC/test/gen_reference.py
singhharsh1708: It regenerates the checked-in JSON from qmat so the coefficients are checked against something independent of this code; the tests only read the JSON.

lib/OrdinaryDiffEqSDC/src/sdc_perform_step.jl
singhharsh1708: Measured: cold start doubles the Newton iterations per solve on ROBER (5.6 vs 2.8), and VdP mu=100 takes 70 steps against 23, error 8.4e-4 against 5.6e-9.

lib/OrdinaryDiffEqSDC/test/gen_reference.py
singhharsh1708: Porting it would check our coefficients against our own code; qmat is the independent source and the tests only read its JSON.

PR #4245, opened 2026-08-13
Fix ESDIRK order loss under a mass matrix

### LINKED ISSUE #4173: ESDIRK methods handle mass matrices incorrectly
**Describe the bug 🐞**

All ESDIRK methods converge with order 1 on my end when solving a simple heat problem (M du = K u).

Bots explanation follows:

```
**The whole ESDIRK family loses its order to the mass matrix.** An ESDIRK
  method has an explicit first stage (`a₁₁ = 0`), which needs `M⁻¹f`; with a
  non-identity `M` it silently degrades to order 1. Measured on the linear heat
  problem, as order with `M ≠ I` vs order with `M = I`:

  | method | nominal | `M ≠ I` | `M = I` |
  |:---|---:|---:|---:|
  | TRBDF2 | 2 | **1.009** | 2.001 |
  | Kvaerno3 | 3 | **1.027** | 2.984 |
  | KenCarp4 | 4 | **1.009** | 4.003 |
  | SDIRK22 | 2 | 1.999 | 2.000 |
  | SSPSDIRK2 | 2 | 2.000 | 2.000 |
  | Hairer4 | 4 | 4.025 | — |
```

**Expected behavior**

Method should either converge with the correct order or throw an error.

singhharsh1708:
An ESDIRK's explicit first stage builds `zs[1]` from `integrator.fsalfirst`, which under a mass matrix holds the residual `f` rather than the rate `M \ f`, so the first step carries an O(dt) error that never refines away and every ESDIRK drops to first order when `M != I`.
A convergence test under a mass matrix for TRBDF2, Kvaerno3, KenCarp3 and KenCarp4, with `M = I` alongside as the control, fails on master for all four and passes here; TRBDF2 goes from order 0.966 to 1.985. ROBER still solves to `Success` with TRBDF2, KenCarp4 and Kvaerno3, so behaviour with a consistent first step is unchanged. Full `OrdinaryDiffEqSDIRK` suite passes.
`M` is never inverted anywhere else, only folded into `W`, so the proper fix is stages carrying `f(Y)`; @oscardssmith said a better version is coming and this stands until it does.
The rate is computed with `solve(LinearProblem(M, f), alg.linsolve)` on the steps whose `fsalfirst` holds the residual, so it uses whatever linear solver the user chose rather than a fixed factorization; that makes LinearSolve a direct SDIRK dependency. Only an `AbstractMatrix` mass matrix is converted; `UniformScaling` and operator mass matrices keep the residual as before.
Fixes #4173.


lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
```diff
@@ -1,3 +1,23 @@
...
+
+Only an `AbstractMatrix` is converted. A singular one (an index-1 DAE such as ROBER) has
+no unique derivative to recover, so the factorization is checked and the residual kept
+when it fails. Everything else, `UniformScaling` and the SciMLOperators mass matrices,
+keeps the residual too: they have no factorization to ask for, and converting only some
+of them would make `ScalarOperator(λ)` and `λ * I` disagree on the same problem.
+"""
+_explicit_stage1_rate(mass_matrix, fsalfirst) = fsalfirst
+function _explicit_stage1_rate(mass_matrix::AbstractMatrix, fsalfirst)
+    F = lu(mass_matrix, check = false)
```
termi-official: Why hardcode an lu here?

singhharsh1708: switched to backslash.Singular case now uses integrator.isdae instead.

lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
```diff
@@ -1,3 +1,26 @@
...
+
+`isdae` is the integrator's own flag, set once from `ArrayInterface.issingular`. A
+singular `M` has no unique derivative to recover, so the residual is kept, which is what
+the solver did before. `UniformScaling` and the SciMLOperators mass matrices keep it too:
+converting only some of them would make `ScalarOperator(λ)` and `λ * I` disagree.
+"""
+_explicit_stage1_rate(mass_matrix, fsalfirst, isdae) = fsalfirst
+function _explicit_stage1_rate(mass_matrix::AbstractMatrix, fsalfirst, isdae)
+    isdae && return fsalfirst
+    return mass_matrix \ fsalfirst
```
ChrisRackauckas, line 21: no never backslash

singhharsh1708:
Makes sense, M isn't inverted anywhere else, it's always folded into W. To do this properly the stages would need to carry f(Y) instead of dt*Ydot so the W solve absorbs it, which is a fair bit bigger than this PR. Should I do that here, or just error for now when there's a non-identity mass matrix with an explicit first stage?

oscardssmith:
I think I have a better version of this. pr incoming

singhharsh1708:
Sounds good. The convergence test in here should apply to whatever the fix ends up being, happy to split it out or you can just lift it.

singhharsh1708:
The red Integrators_II lts lane here is #4425, LinearSolve's QR fallback marking cache.tmp shared on Julia 1.10, not this change.

lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
singhharsh1708: Replaced by `solve(LinearProblem(M, f), alg.linsolve)`, so it lands on whatever the user's linear solver picks; LinearSolve is now a direct SDIRK dependency.

lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
singhharsh1708: Gone: the solve goes through LinearSolve with the algorithm's `linsolve` now.

singhharsh1708:
@ChrisRackauckas ready for another look, the backslash is gone and it goes through `alg.linsolve` now.

ChrisRackauckas:
Letting Oscar take this.

PR #4253, opened 2026-08-14
Guard the SciMLBase hook and restore released compat

ChrisRackauckas-Claude:
> Ignore this PR until reviewed by @ChrisRackauckas.

## What changed and why

DiffEqBase 7.15.1 currently extends `SciMLBase.report_integrator_failure`, but no released SciMLBase defines that hook. The normal resolver selects SciMLBase 3.47.0 and DiffEqBase fails while loading. The same change raised the SciMLBase compat floor to nonexistent version `3.46.2`, which makes minimum-version resolution unsatisfiable.

This guards the method extension until SciMLBase provides the hook, restores the last released floor `SciMLBase = "3.46"`, and adds a Core regression test for both dependency states. Released SciMLBase versions continue using their existing `check_error` diagnostics, while the DiffEqBase method is installed when the pending hook exists. There is no public API change.

Git bisect identified https://github.com/SciML/OrdinaryDiffEq.jl/commit/0ed2615804a38c915846358d39987f689f2511e6 as the first bad commit. The migration merged in https://github.com/SciML/OrdinaryDiffEq.jl/pull/4153 ahead of its still-open prerequisite https://github.com/SciML/SciMLBase.jl/pull/1504. The failure is also tracked at https://github.com/SciML/OrdinaryDiffEq.jl/issues/4249.

### Failing before

On unmodified upstream master with SciMLBase 3.47.0:

```text
$ /home/crackauc/.juliaup/bin/julia +1.12 --project=. --startup-file=no --compiled-modules=no lib/DiffEqBase/test/check_error.jl
ERROR: LoadError: UndefVarError: `report_integrator_failure` not defined in `SciMLBase`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
  [1] getproperty(x::Module, f::Symbol)
    @ Base ./Base_compiler.jl:47
  [2] top-level scope
    @ lib/DiffEqBase/src/check_error.jl:6
...
in expression starting at lib/DiffEqBase/test/check_error.jl:1
```

The unmodified root `GROUP=AD` run independently resolved:

```text
[2b5f629d] + DiffEqBase v7.15.1 `lib/DiffEqBase`
[0bca4576] + SciMLBase v3.47.0
...
✗ DiffEqBase
```

The exact `julia-downgrade-compat@v2` resolver passes on parent `9d92d1c722` and fails on `0ed2615804`. Its current root CI failure is https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/31768146910/job/94668206969.

[...]

PR #4254, opened 2026-08-14
Clamp initial dt probe to first tstop

### LINKED ISSUE #4240: `ode_determine_initdt` evaluates the RHS beyond the first `tstop`/`d_discontinuity`
`ode_determine_initdt` evaluates the RHS at `t0 + dt₀` with `dt₀` clamped only by `dtmax`, which defaults to the full `tspan` width. When a `tstop` or `d_discontinuity` sits close to `t0`, the probe therefore crosses it and the initial step size is estimated from an ODE that does not apply on the first step.

The actual stepping is fine — `modify_dt_for_tstops!` clamps `dt` to the first `tstop` — so this only affects the initial step size estimate, and any user code that observes the times at which the RHS is called.

## Reproducer

```julia
using OrdinaryDiffEqTsit5

# 1. `f` is evaluated beyond the first tstop / discontinuity before any step is taken
const evaluated = Float64[]
f(u, p, t) = (push!(evaluated, t); -u)

prob = ODEProblem(f, [1.0], (0.0, 10.0))
init(prob, Tsit5(); tstops = [1e-3], d_discontinuities = [1e-3])

println("first tstop            : ", 1e-3)
println("f called at            : ", evaluated)
println("beyond the first tstop : ", filter(>(1e-3), evaluated))

# 2. The consequence: the initial `dt` is estimated from the wrong side of the discontinuity
stiff_after(u, p, t) = t < 1e-3 ? -u : -1.0e6 * u
mild_only(u, p, t) = -u

for (name, rhs) in (("piecewise (stiff after 1e-3)", stiff_after), ("mild everywhere", mild_only))
  integ = init(
    ODEProblem(rhs, [1.0], (0.0, 10.0)),
    Tsit5();
    tstops = [1e-3],
    d_discontinuities = [1e-3],
  )
  println("initial dt, $name: ", integ.dt)
end
```

Output:

```
first tstop            : 0.001
f called at            : [0.0, 0.0, 0.01]
beyond the first tstop : [0.01]
initial dt, piecewise (stiff after 1e-3): 0.00251744426529339
initial dt, mild everywhere: 0.10001999200479662
```

The third evaluation is at `t = 0.01`, ten times beyond the discontinuity at `1e-3`.

In the second part the two problems are identical on `[0, 1e-3]`, which is the entire span the first step can cover, yet the initial `dt` differs by a factor of 40 because the probe samples the stiff branch that only takes effect after the discontinuity.

## Where

- `OrdinaryDiffEqCore/src/initdt.jl:216` — `dt₀ = min(dt₀, dtmax_tdir)`, then `:241` `f(f₁, u₁, p, t + dt₀_tdir)` (in-place); the same pair at `:423` and `:427` for the out-of-place path
- `OrdinaryDiffEqCore/src/solve.jl:153` — `dtmax = (prob.tspan[end] - prob.tspan[1])`
- `OrdinaryDiffEqCore/src/integrators/integrator_utils.jl:268` — `modify_dt_for_tstops!`, which does clamp the actual step

## Suggested fix

Clamp the `dtmax` handed to `ode_determine_initdt` (or `dt₀` inside it) by the distance from `t0` to the first `tstop`, mirroring what `modify_dt_for_tstops!` already does for the step itself. `_determine_initdt` (`integrators/integrator_interface.jl:644`) has the integrator available, so `first_tstop(integrator)` is reachable there.

## Versions

```
Julia 1.12.6
OrdinaryDiffEqCore 4.14.2
OrdinaryDiffEqTsit5 2.1.3
DiffEqBase 7.14.0
SciMLBase 3.46.0
```

---

This issue was written with the assistance of generative AI (Claude Code).

ChrisRackauckas-Claude:

## What changed and why

Bound the `dtmax` passed to `ode_determine_initdt` by the directional distance from the integrator's current time to its first `tstop`. This prevents the initial step-size probe from evaluating the RHS after a user `tstop` or `d_discontinuity`; the existing stop heap also makes the bound apply correctly to backward integration.

Added a looped regression test covering forward and backward integration, in-place and out-of-place RHS functions, and independent `tstops` and `d_discontinuities` inputs.

Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/4240

## Root cause

`_determine_initdt` passed the configured `dtmax` directly to `ode_determine_initdt`. The default is the full signed `tspan` width, while `modify_dt_for_tstops!` only clamps the actual step afterward, so the estimator's second RHS evaluation could already have crossed the first stop.

## Verification

Failing before the implementation change:

```text
$ GROUP=InterfaceI julia +1.12 --project -e 'using Pkg; Pkg.test()'
Test Summary:                                | Pass  Fail  Total     Time
Initdt Tests                                 |   29     8     37  1m55.4s
  Initial dt probe respects the first tstop  |          8      8     7.6s
ERROR: Some tests did not pass: 29 passed, 8 failed, 0 errored, 0 broken.
```

Passing with the implementation change:

```text
$ GROUP=InterfaceI julia +1.12 --project -e 'using Pkg; Pkg.test()'
Initdt Tests  |   37     37  1m54.7s
...
Testing OrdinaryDiffEq tests passed
```

The final regression assertion was rerun after formatting:

```text
Test Summary:                             | Pass  Total  Time
Initial dt probe respects the first tstop |    8      8  4.4s
```

Additional checks:

```text
$ GROUP=QA julia +1.12 --project -e 'using Pkg; Pkg.test()'
Test Summary:           | Pass  Total   Time
Quality Assurance Tests |   90     90  15.3s
Testing OrdinaryDiffEq tests passed

$ julia +1.12 -m Runic --check lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl test/InterfaceI/ode_initdt_tests.jl
# exit 0, no output

$ typos lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl test/InterfaceI/ode_initdt_tests.jl
# exit 0, no output
```

The issue reproducer now reports probes at the stop and bounded initial steps:

```text
oop-forward: dt=0.001, evaluated=[0.0, 0.0, 0.001]
iip-forward: dt=0.001, evaluated=[0.0, 0.0, 0.001]
oop-backward: dt=-0.0009999999999994458, evaluated=[10.0, 10.0, 9.999]
iip-backward: dt=-0.0009999999999994458, evaluated=[10.0, 10.0, 9.999]
```

The registered SciMLBase 3.47.0 did not yet contain `report_integrator_failure`, which current OrdinaryDiffEq `master` expects. Local tests therefore used the pending SciMLBase source containing that hook through an ignored Manifest entry; this workaround is not part of the diff.

## Not verified

- `GROUP=Everything`, downstream, GPU, Julia pre-release, and Julia LTS lanes were not run locally.
- Docs were not built because the change does not touch docstrings, rendered docs, or public API.

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/integrator_interface.jl
@@ -642,9 +642,13 @@ Convenience wrapper that calls [`ode_determine_initdt`](@ref) with the fields of
 `integrator` (state, tolerances, norm, problem).
 """
 function _determine_initdt(integrator)
+    tdir = integrator.tdir
+    dtmax = tdir * min(
+        abs(integrator.opts.dtmax), abs(first_tstop(integrator) - tdir * integrator.t)
+    )
     return ode_determine_initdt(
         integrator.u, integrator.t,
-        integrator.tdir, integrator.opts.dtmax,
+        tdir, dtmax,
         integrator.opts.abstol, integrator.opts.reltol,
         integrator.opts.internalnorm, integrator.sol.prob,
         integrator
--- a/test/InterfaceI/ode_initdt_tests.jl
+++ b/test/InterfaceI/ode_initdt_tests.jl
@@ -130,3 +130,23 @@ end
     sol_ok = solve(ODEProblem(f_ok, [1.0], (0.0, 1.0)), Tsit5())
     @test SciMLBase.successful_retcode(sol_ok)
 end
+
+@testset "Initial dt probe respects the first tstop" begin
+    for (tspan, tstop) in (((0.0, 10.0), 1.0e-3), ((10.0, 0.0), 9.999))
+        tdir = sign(tspan[2] - tspan[1])
+        for stop_keyword in (:tstops, :d_discontinuities)
+            options = NamedTuple{(stop_keyword,)}(([tstop],))
+            for isinplace in (false, true)
+                evaluated = Float64[]
+                f = if isinplace
+                    (du, u, p, t) -> (push!(p, t); du .= -u)
+                else
+                    (u, p, t) -> (push!(p, t); -u)
+                end
+                prob = ODEProblem(f, [1.0], tspan, evaluated)
+                init(prob, Tsit5(); options...)
+                @test maximum(tdir .* (evaluated .- tstop)) <= 0
+            end
+        end
+    end
+end


ChrisRackauckas-Claude:
Plan/status scratchpad:

1. Reproduce the initial-step estimator crossing the first tstops/d_discontinuities entry in both time directions and for in-place/out-of-place functions.
2. Add a regression test that fails on the unmodified implementation.
3. Bound the estimator probe by the first scheduled stop.
4. Run the targeted InterfaceI group, QA, Runic, and typos before pushing.

All four steps are complete; exact failing-before and passing-after output is in the PR body.

CI triage: PositiveIntegrators.jl/Downstream/1.10 currently fails at https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/31771416297/job/94677970135 and was reproduced locally on clean upstream master b1c4c7fc675b67f7c3df6ac4c339b725a3d3aad3. The DiffEqBase downgrade failure at https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/31771416514/job/94678023625 also occurs in the master run at https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/31768146846. Both are being investigated separately; I will not mix unrelated compatibility or CI changes into this PR.

ChrisRackauckas-Claude:
Base-failure investigation result:

The downgrade failures reproduce locally on clean master b1c4c7fc675b67f7c3df6ac4c339b725a3d3aad3 for both lib/DiffEqBase with --min=@alldeps and the root project with --min=@deps. The parent 9d92d1c722 passes both, so the first bad commit is https://github.com/SciML/OrdinaryDiffEq.jl/commit/0ed2615804a38c915846358d39987f689f2511e6.

That commit changed DiffEqBase SciMLBase compat from 3.46 to 3.46.2, but no SciMLBase 3.46.2 was registered; released SciMLBase 3.47.0 also lacks report_integrator_failure. The hook is still pending in https://github.com/SciML/SciMLBase.jl/pull/1504. Draft compatibility guard https://github.com/SciML/OrdinaryDiffEq.jl/pull/4253 separately addresses the package-load side.

A compat rollback alone is not valid because it makes resolution pass while DiffEqBase still fails to load. PR 4254 remains intentionally focused on issue 4240.

ChrisRackauckas-Claude:
Final relevant-CI triage:

All three InterfaceI jobs fail before executing tests because DiffEqBase cannot precompile against released SciMLBase 3.47.0:

- Julia release: https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/31771416686/job/94678497944
- Julia LTS: https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/31771416686/job/94678497553
- Julia pre: https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/31771416686/job/94678497722

Each log reports UndefVarError: report_integrator_failure not defined in SciMLBase. This is the same clean-master regression documented in https://github.com/SciML/OrdinaryDiffEq.jl/pull/4253 and occurs before PR 4254 code or tests load. The local InterfaceI group passed with the released dependency gap handled as documented in the PR body.

PR #4266, opened 2026-08-15
Keep W split when the linear solver is LHLFactorization

ChrisRackauckas-Claude:
>
> **Blocked on SciML/LinearSolve.jl#1215.** This bumps `OrdinaryDiffEqDifferentiation`'s `LinearSolve` compat to `5.11`, which does not exist until that PR is registered, so CI here will fail to resolve until then. Nothing to fix on this side.

## What and why

`calc_W!` rebuilds `W = J - M/(dt·γ)` far more often than it recomputes `J` — on a dense Brusselator, 112 W rebuilds against 2 Jacobians in one run. With a concrete `W`, each of those rebuilds is followed by an `O(n³)` refactorization to absorb a scalar.

LinearSolve's new `LHLFactorization` reduces `J` once to Hessenberg form and absorbs a new `dtgamma` in `O(n²)` — but only if `W` reaches it unassembled. So `build_J_W` now hands it a `ShiftedJacobian` (a lazy `AbstractMatrix` equal to `α·J + β·I`) instead of a matrix, `jacobian2W!` writes only the two scalars, and `calc_W!` calls `mark_jacobian_updated!` when it recomputes `J`.

That is the whole change, ~50 lines, and it is not solver-specific: every Newton-based integrator picks it up. `NordsieckBDF`, `FBDF` and `QNDF` are covered by the tests.

## Verification

`lib/OrdinaryDiffEqBDF/test/lhl_factorization_tests.jl`, run against the branch of LinearSolve.jl#1215 and SciML/LHL.jl:

```
W is kept split                                |  3   3   2.7s
same trajectory as an assembled W              |  6   6  16.2s
the reduction is reused across dtgamma changes |  3   3   0.0s
unsupported combinations are rejected          |  2   2   2.0s
```

The trajectories are not merely close — at most tolerances they are step-for-step identical to the assembled-W path, so the entire difference is linear algebra.

### Work-precision, Brusselator, dense Jacobian, single-threaded BLAS

`NordsieckBDF`, wall time in seconds, best of 3, at matched tolerance. `r0` is `LHLFactorization(refine = 0)`:

| n | tol | LU | LHL r=1 | LHL r=0 | speedup (r=0) | nJac | nW | nsolve |
|---|---|---|---|---|---|---|---|---|
| 128 | 1e-6 | 0.0163 | 0.0191 | 0.0095 | 1.72× | 1 | 71 | 562 |
| 128 | 1e-10 | 0.0311 | 0.0521 | 0.0278 | 1.12× | 2 | 118 | 1696 |
| 288 | 1e-6 | 0.0947 | 0.0696 | 0.0361 | 2.62× | 1 | 82 | 576 |
| 512 | 1e-6 | 0.4248 | 0.2439 | 0.1365 | 3.11× | 1 | 85 | 664 |
| 512 | 1e-10 | 0.7425 | 0.6757 | 0.3348 | 2.22× | 2 | 137 | 1871 |
| 800 | 1e-5 | 0.8555 | 0.4520 | 0.2344 | 3.65× | 1 | 60 | 465 |
| 800 | 1e-8 | 1.6156 | 0.9538 | 0.5097 | 3.17× | 2 | 112 | 1085 |
| 800 | 1e-10 | 2.0385 | 1.5506 | 0.7821 | 2.61× | 2 | 131 | 1979 |

Relative error matches LU's to 3–4 digits at every point. `FBDF` at `n = 800` is 1.4–2.9× over the same tolerance sweep with no solver-specific work.

## The thing a reviewer should push back on

**The ratio that decides this is solves per γ, and it is ~14** — not the 1–3 a naive Newton count suggests. `W` is rebuilt every ~13 steps and each `W` then serves ~14 solves. That inverts the sensible-looking default: `LHLFactorization()` ships with `refine = 1` (one step of iterative refinement, which is what makes its backward error match LU's on hostile Jacobians), but refinement roughly triples a solve and so is charged 14× per `γ` saved. At `n = 128`, `refine = 1` is **slower** than LU while `refine = 0` is faster.

A Newton iteration is itself a correction loop whose convergence test governs the answer, so an inexact linear solve costs at most an extra iteration — the measured step counts confirm this. So the useful invocation here is `NordsieckBDF(linsolve = LHLFactorization(refine = 0))`, and a user who writes the bare `LHLFactorization()` gets the slower variant. That is a real usability trap. Options, in order of my preference: document it prominently (done in the LinearSolve docstring), flip LinearSolve's default to `refine = 0`, or have this package override `refine` when the caller did not set it explicitly. I did not take the third — silently rewriting a user-specified algorithm seems worse than the trap.

The single measured loss was a run at tol `1e-4` that needed 8 Jacobians for 57 W rebuilds — seven γ per reduction, below break-even. Everywhere else `J` is reused for 30–200 steps.

## Guard rails

Combinations the reduction cannot serve raise rather than degrade silently:

- a sparse `jac_prototype` — the Hessenberg reduction fills in, so sparsity buys nothing and densifying behind the user's back would be a memory trap;
- a non-scalar mass matrix — that needs a Hessenberg–triangular reduction of the pencil, which is not implemented.

Out-of-place problems and DAEs take the ordinary assembled-W path.

## Not verified

The full BDF test group, other integrator families' groups, GPU paths, downstream packages, the docs build.

https://claude.ai/code/session_014CrTAi2YLkosEXXbudXyKf


ChrisRackauckas-Claude:
**Restructured**: the split `W` is now a `SciMLOperators.WOperator` rather than a bespoke `LinearSolve.ShiftedJacobian`. `build_J_W` already knew how to build one, so this is a smaller and less novel change than the first push.

What that means concretely:

- `build_J_W` constructs `WOperator{IIP}(f.mass_matrix, dtgamma, J, _vec(u))` with a concrete dense `J` for the `LHLFactorization` case — no new type.
- The hooks moved into `calc_W!`'s **existing** `W isa WOperator` branch rather than the concrete-matrix one: `mark_jacobian_updated!(W)` on `new_jac`, and the `jacobian2W!(W._concrete_form, …)` assembly is skipped when the linear solver consumes the split, since it never reads it. That skip is the other half of the win — it is `O(n²)` per step of pure waste otherwise.
- `jacobian2W!(::WOperator, …)` is now just `update_coefficients!(W; gamma)`; `J` is aliased, so `calc_J!` has already refreshed it.

Now depends on SciML/SciMLOperators.jl#419 (compat bumped to `SciMLOperators = "1.27"`), which adds the Jacobian generation counter this needs — without it there is no way to tell an in-place write to `J` from a change of `gamma`.

Merge order: SciMLOperators#419 → SciML/LHLFactorization.jl registration → SciML/LinearSolve.jl#1215 → this.

Re-measured after the restructure (n=800, single-threaded BLAS, best of 3) — the numbers held, and the new test asserts `jacobian_version(W) == stats.njacs`, i.e. the counter tracks Jacobian recomputations exactly:

| n | tol | LU | LHL r=0 | speedup |
|---|---|---|---|---|
| 128 | 1e-6 | 0.0158 | 0.0110 | 1.44× |
| 128 | 1e-10 | 0.0317 | 0.0301 | 1.05× |
| 800 | 1e-6 | 1.2964 | 0.3887 | 3.33× |
| 800 | 1e-8 | 1.7478 | 0.5397 | 3.24× |
| 800 | 1e-10 | 2.0271 | 0.7843 | 2.58× |

Integration tests green (15 assertions): W is kept split, NordsieckBDF/FBDF/QNDF match the assembled-W trajectory, the reduction is reused across dtgamma changes, sparse and non-scalar-mass-matrix combinations raise.

ChrisRackauckas-Claude:
Re-measured on the current HEAD, since the earlier tables were taken before the counter → `jac_stale` flag change. Brusselator, `NordsieckBDF`, single-threaded BLAS, best of 3:

| n | tol | LU | LHL r=1 | LHL r=0 | speedup (r=0) |
|---|---|---|---|---|---|
| 128 | 1e-6 | 0.0159 | 0.0194 | 0.0110 | 1.45× |
| 128 | 1e-8 | 0.0228 | 0.0305 | 0.0158 | 1.44× |
| 128 | 1e-10 | 0.0314 | 0.0545 | 0.0311 | 1.01× |
| 800 | 1e-6 | 1.2730 | 0.6985 | 0.3916 | **3.25×** |
| 800 | 1e-8 | 1.6325 | 1.0018 | 0.5338 | **3.06×** |
| 800 | 1e-10 | 2.0435 | 1.7732 | 0.7870 | **2.60×** |

Within run-to-run noise of the previous figures (3.33 / 3.24 / 2.58), as expected — the flag change is control flow only. Final state matches LU in every row.

ChrisRackauckas-Claude:
**Correcting the performance numbers in this PR.** They had accumulated three vintages — NordsieckBDF on the original kernels, FBDF on the current ones, and a "up to 9.5×" summary that was the single most favourable cell of ten. Here is one table, both integrators, current kernels, so the comparison is like-for-like.

Brusselator and a scaled POLLUTION-style stiff kinetics problem; single-threaded BLAS; best of 3; speedup against `LUFactorization` at matched tolerance:

| problem | tol | NordsieckBDF r=1 | r=0 | FBDF r=1 | r=0 |
|---|---|---|---|---|---|
| Brusselator n=32 | 1e-6 | 1.13× | 1.47× | 1.17× | 1.43× |
| Brusselator n=32 | 1e-9 | 1.01× | 1.15× | 1.08× | 1.73× |
| Brusselator n=128 | 1e-6 | 1.96× | 3.45× | 2.37× | 3.89× |
| Brusselator n=128 | 1e-9 | 1.37× | 2.37× | 1.48× | 2.72× |
| Brusselator n=800 | 1e-6 | 2.65× | 4.65× | 4.93× | 9.52× |
| Brusselator n=800 | 1e-9 | 2.10× | 4.48× | 2.80× | 7.21× |
| Pollution n=40 | 1e-6 | **0.98×** | 1.26× | 1.12× | 1.66× |
| Pollution n=40 | 1e-9 | 1.24× | 1.99× | 1.08× | 1.19× |
| Pollution n=200 | 1e-6 | 1.17× | 2.01× | 1.54× | 2.53× |
| Pollution n=200 | 1e-9 | 1.10× | 1.96× | 1.24× | 2.34× |

`r=1` is the shipped default; `r=0` is `LHLFactorization(refine = 0)`, which is the setting an integrator should prefer since Newton is already a correction loop.

### What to take from it

- **2–5× is the honest headline**, not 9.5×. The default `refine = 1` spans 0.98×–4.93×; `refine = 0` spans 1.15×–9.52×.
- **The 9.52× cell deserves an asterisk.** It is FBDF, `refine = 0`, the loosest tolerance and the largest size — and on that run LHL took ~20% less work than LU (179 W rebuilds / 943 solves against 205 / 1168), so corrected for equal work it is nearer 8×. It is also the one row where the two trajectories differ; both land ~2e-4 from a 1e-12 reference, with LHL marginally *more* accurate (1.966e-4 vs 2.151e-4). Every NordsieckBDF row above is step-for-step identical to LU.
- **`Pollution n=40` at `refine = 1` is 0.98×** — a slight loss, at a size above the `n ≥ 32` cutoff. The cutoff came from a per-operation model assuming ~14 solves per γ and ~55 γ per Jacobian; that problem does not hit those ratios. The docstring now says so: the cutoff is where the algorithm stops losing, not where it starts winning, and it only pulls clear above n ≈ 100.

The win scales with how often `W` is rebuilt per Jacobian, which is why FBDF beats NordsieckBDF here (nW 205 vs 88 at n=800) and why RadauIIA would beat both if it could reuse the reduction — 279 rebuilds against 3 Jacobians, see #4281.

ChrisRackauckas-Claude:
The 136 red checks had a single cause, now fixed in 5a54ef4.

`LinearSolve 5.11.0` was registered upstream **without** `LHLFactorization` in it. This branch had `LinearSolve = "5.11"` in OrdinaryDiffEqDifferentiation (and the old `"5.1"` in FIRK/BDF), so CI resolved a LinearSolve that has no such name and `LinearSolve.LHLFactorization` became an `UndefVarError` at load time. OrdinaryDiffEqDifferentiation is a dependency of nearly every sublibrary, which is why one bad bound took out the whole downgrade matrix and the docs build alongside the three sublibraries that actually touch LHL.

Raised to `LinearSolve = "5.12"` in the three sublibraries that name it, and bumped OrdinaryDiffEqFIRK `2.7.0 → 2.8.0` (new `ArgumentError` path) and OrdinaryDiffEqBDF `2.4.3 → 2.4.4` (test-extra bound only).

**This PR is now explicitly blocked on LinearSolve 5.12.0 being registered**, which needs https://github.com/SciML/LinearSolve.jl/pull/1215 merged first. Until then CI here will fail at resolution rather than at load — that is the honest state and is the point of the bound. Do not read the next red run as a code regression; check whether 5.12.0 is in the General registry first.

Not verified locally: the full downgrade matrix and the docs build, both of which cannot resolve until 5.12.0 exists.

ChrisRackauckas-Claude:
Rebased onto current master (17 commits behind, clean rebase, all 5 branch commits mine — force-pushed with `--force-with-lease`).

https://github.com/SciML/LinearSolve.jl/pull/1215 is merged as 47cdfb2 and LinearSolve 5.12.0 is registering in https://github.com/JuliaRegistries/General/pull/164991. Once that auto-merges, the `LinearSolve = "5.12"` bounds here resolve and CI becomes meaningful. Until then the failures are resolution failures, not code.

Version bumps confirmed still correct against master after the rebase:

| package | master | this branch |
|---|---|---|
| OrdinaryDiffEqDifferentiation | 3.9.0 | 3.10.0 |
| OrdinaryDiffEqFIRK | 2.7.0 | 2.8.0 |
| OrdinaryDiffEqBDF | 2.4.3 | 2.4.4 |

## Scope note from the LinearSolve side

`defaultalg` now declines the split form whenever `W.J` is an operator rather than a plain dense matrix. That was a correctness fix, not a tuning choice: `calc_W!` guards `mark_jacobian_updated!` behind `!(W.J isa AbstractSciMLOperator)`, so on the `islin` path `update_coefficients!` moves `J` in place with nothing marking it stale, and the cached reduction answered with the previous Jacobian — relative residual 1.34, no exception. Filed as https://github.com/SciML/OrdinaryDiffEq.jl/issues/4303.

Practical consequence for this PR: `_uses_split_W` already restricted itself to `jac_prototype === nothing || jac_prototype isa Matrix`, so nothing here changes. But the `islin` path — a linear ODE with a dense operator, arguably the best case for reducing `J` once and re-shifting per step — stays excluded until #4303 is fixed.

Also filed while checking downstream impact: https://github.com/SciML/OrdinaryDiffEq.jl/issues/4302, a matrix-free `jac_prototype` throwing in `prepare_user_sparsity` before any solve. Independent of this PR — it reproduces on released OrdinaryDiffEq v7.6.0 with an explicit `KrylovJL_GMRES()` too.

Still open for @ChrisRackauckas: the FIRK guard currently **throws** `ArgumentError` for `RadauIIA5(linsolve = LHLFactorization())` rather than warning and falling back. Throwing is honest — it would be 5.2× slower — but it hard-fails a combination a user might reasonably write. Happy to switch it to a warn-and-fall-back if you prefer.

Not verified locally: the full downstream and downgrade matrices and the docs build, none of which can resolve until 5.12.0 is in the registry.

ChrisRackauckas-Claude:
LinearSolve 5.12.0 is registered, so this resolves now. Validated locally.

## Resolution

```
LinearSolve v5.12.0
SciMLOperators v1.28.0
LHLFactorization v2.0.0
```

## LHL tests, exit 0

```
Test Summary:   | Pass  Total  Time
W is kept split |    3      3  2.8s
Test Summary:                     | Pass  Total   Time
same trajectory as an assembled W |    6      6  15.5s
Test Summary:                                  | Pass  Total  Time
the reduction is reused across dtgamma changes |    5      5  0.0s
Test Summary:                            | Pass  Total  Time
FIRK refuses a reduction it cannot reuse |    2      2  4.0s
Test Summary:                         | Pass  Total  Time
unsupported combinations are rejected |    2      2  2.1s
```

## One pre-existing failure in the same group, not from this PR

`GROUP=Core` in `lib/OrdinaryDiffEqBDF` errors before it reaches the LHL file, because `dae_convergence_tests.jl` is first in the list and a `@safetestset` failure aborts the block. Checked against a clean `master` worktree (202f5fd):

| | result |
|---|---|
| `master` | 45 passed, 0 failed, 1 errored |
| this branch | 45 passed, 0 failed, 1 errored |

Byte-identical — same testset (`DAE Solver Convergence Tests (out-of-place, custom jac)`), same `MethodError: no method matching lu(::Nothing; check::Bool)` from `lu_instance(::Nothing)` in `build_J_W`. Filed as https://github.com/SciML/OrdinaryDiffEq.jl/issues/4304.

Worth flagging that this means **CI has not been exercising the LHL tests in this group at all** — they sit after the aborting file. I ran them in a separate environment to get the numbers above. Once #4304 is fixed the group will actually reach them.

## Not verified locally

The full downgrade matrix, the docs build, the downstream suites, and GPU paths. CI needs a re-run now that 5.12.0 exists — the 127 failures currently shown all predate registration.

ChrisRackauckas-Claude:
Radau now consumes the split form too — 76333b8 replaces the `ArgumentError` guard with support for the complex stage pair.

## What changed

- `firk_real_W` keeps the real stage split (the `WOperator` `build_J_W` made) instead of concretizing and refusing.
- `build_complex_W` builds each complex-conjugate pair as a **`WOperator` with a complex `gamma` slot over the same real Jacobian object** — LinearSolve then keeps one real O(n³) reduction per stage matrix and only the shift goes complex (`ws.factors::Matrix{Float64}`, `ws.σ::ComplexF64`), which is the LHLFactorization 2.0 shape built for exactly this.
- `firk_new_J!` marks **every** stage matrix stale when it rewrites `J` in place. This is the correctness keystone: reverting only the marking produces 6 test failures — failed retcodes and wrong trajectories — because the pair's cached reduction keeps answering for the previous Jacobian.

Covers RadauIIA3 (complex-only), RadauIIA5 (`W2`), RadauIIA9 (`W2`,`W3`), and AdaptiveRadau (vector of pairs).

## The #4281 measurement, re-run

RadauIIA5, 128-unknown Brusselator, tol 1e-8, single-threaded BLAS, best of 5:

| linsolve | before (guard rationale) | now |
|---|---|---|
| `LUFactorization()` | 0.169 s | 0.2308 s (this machine) |
| `LHLFactorization()` | 0.883 s — **5.2× slower** | **0.0891 s — 2.6× faster** |
| `LHLFactorization(refine = 0)` | — | **0.0754 s — 3.1× faster** |

```
LU          0.2308 s   err=5.024e-08   njacs=3  nw=279  nsolve=1668
LHL         0.0891 s   err=1.351e-07   njacs=1  nw=274  nsolve=1687
LHL r=0     0.0754 s   err=1.351e-07   njacs=1  nw=274  nsolve=1687
```

## Verification

`lib/OrdinaryDiffEqBDF/test/lhl_factorization_tests.jl`, exit 0 — the old `@test_throws` flipped to failing the moment support landed (the discriminating "before"), and was replaced by:

```
Test Summary:                         | Pass  Total   Time
Radau keeps both stage matrices split |   14     14  13.9s
Radau re-reduces when the Jacobian moves |    2      2   0.0s
```

including RadauIIA3/5/9 trajectory ≈ LU at rtol 1e-6 on a stiff dense problem with `njacs > 1`. FIRK's own `ode_firk_tests.jl` (LU/Krylov/GaussLegendre/AdaptiveRadau paths): exit 0, no failures. Runic and typos clean.

## Not verified / out of scope

- AdaptiveRadau with LHL has the marking path wired (`W2` vector) but no dedicated trajectory test — RadauIIA9 covers the multi-pair shape.
- Sharing literally *one* reduction across the real and complex LinearSolve caches (the last line of https://github.com/SciML/OrdinaryDiffEq.jl/issues/4281): not done. Each stage matrix keeps its own reduction, re-taken only when J moves — with ~90 shifts amortizing each, the additional saving is a small constant, and it needs a cross-cache interface that doesn't exist. Left on #4281.
- GaussLegendre never took the split path before and still doesn't.

PR #4299, opened 2026-08-18
Remove duplicate naming in keyword arguments

abhro:
## Checklist

- [x] The new code follows the [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and [COLPRAC](https://github.com/SciML/COLPRAC).
  
## Additional context

If a local variable shares the same name as a keyword argument to a function call, the local variable can be passed as a keyword argument without having to duplicate the names in the call.


ChrisRackauckas:
Most of the changes don't match the PR description

abhro:
They... should? Also for the struct fields, the field names get mapped to the kwarg name, so that might be a bit non-obvious. And I think there's some line wrap stuff in there because they were in the vicinity. But other than that, I'm not sure which changes you're referring to?

ChrisRackauckas:
There's 165 files changed, lots of formatting.

abhro:
I was looking over the changes, and yeah, a lot more formatting changes went through then I'd expected. I've amended the commits to include only the kwargs stuff exclusively :+1:

PR #4192, opened 2026-08-09
Apply the integrator tolerance to the Krylov linear solve

ChrisRackauckas-Claude:
Closes #4161.

**Please ignore until reviewed by @ChrisRackauckas.**

## What changed and why

`dolinsolve` forwarded the integrator's tolerance as `solve!(linsolve; reltol = integrator.opts.reltol)`, but `SciMLBase.solve!(::LinearCache, ::KrylovJL; kwargs...)` takes its tolerances off the cache and ignores the keywords. The argument was dead, so every Krylov solve on the Newton path ran at LinearSolve's `sqrt(eps) = 1.49e-8` default regardless of what the user asked for. Same pattern at the one other site, `firk_perform_step.jl` (GaussLegendre's stage solve).

The fix applies the tolerance through `LinearSolve.update_tolerances!`, which is the entry point that takes effect, and stops passing the dead keyword. It is factored into `set_linear_reltol!` so the two call sites cannot drift.

I made the tolerance live rather than deleting the argument. Two reasons, both checked:

1. It is what the rest of the repo already does. Rosenbrock (`rosenbrock_caches.jl:257`) and RadauIIA3 (`firk_caches.jl:116`) pass `abstol = reltol, reltol = reltol` to `LinearSolve.init`, so their Krylov solves have always run at the integrator tolerance. The Newton path is the outlier: `build_nlsolver` never receives `reltol`, so it cannot set it at `init`, and `dolinsolve` was the only place left that could. This is not a new regime, it is the shipped behaviour of the most-used stiff family.
2. Measured below: fewer RHS evaluations, same answers.

`set_linear_reltol!` skips solvers that need a concrete `A` (a factorization has no tolerance and `update_tolerances!` throws for one) and skips non-scalar `reltol`, since a `LinearCache` holds one scalar tolerance. Rosenbrock and RadauIIA3 caches are already at the right value, so the guard makes them a no-op and their behaviour is bit-identical.

## Failing before / passing after

All evidence below was re-run on the rebased commit (`4f6dedcf5`, on top of master `b9889d3ae`), not carried over from the pre-rebase branch.

The new test asserts the integrator's tolerance reaches the linear cache. To show it discriminates, the two behavioural call sites were reverted to `solve!(linsolve; reltol)` / `LinearSolve.solve!(linsolve; reltol = integrator.opts.reltol)` — i.e. exactly master's behaviour — with `set_linear_reltol!` and the committed test file left untouched:

```
[...]

PR #4301, opened 2026-08-18
Floor the Newton direction's linear tolerance at sqrt(eps)

### LINKED ISSUE #4293: Hairer4/Hairer42 with KrylovJL_GMRES fail on an index-1 DAE (Regression_I)
`Hairer4` and `Hairer42` with `KrylovJL_GMRES` fail on the index-1 DAE that
`test/Regression_I/newton_krylov_roundoff.jl` uses. Reproduced on master (523c650cb6):

```julia
using OrdinaryDiffEqSDIRK, OrdinaryDiffEqNonlinearSolve
using OrdinaryDiffEqNonlinearSolve: BrownFullBasicInit
using LinearAlgebra, SparseArrays

P = [-1.0 0 0 0; 1.0 -0.5 0 0; 1.0 1.0 -1.0 0; -1.0 1.0 0 -1.0]
MM = Diagonal([1.0, 1.0, 0.0, 0.0])
prob = ODEProblem(
    [1.0, 1.0, 0.5, 0.5], (0.0, 5.0), P; initializealg = BrownFullBasicInit(),
)

solve(prob, Hairer4(linsolve = KrylovJL_GMRES()); maxiters = 10_000).retcode  # Unstable
solve(prob, Hairer4(); maxiters = 10_000).retcode                             # Success
```

| alg | linsolve | retcode |
| --- | --- | --- |
| Hairer4 | KrylovJL_GMRES | Unstable |
| Hairer4 | default | Success |
| Hairer42 | KrylovJL_GMRES | Unstable |
| Hairer42 | default | Success |

So it's specific to the Krylov solver on this singular mass matrix, not to the methods.

The test that covers this was added on Aug 9 in #4040 and asserts `successful_retcode`, so it
passed then. It now fails on Julia 1, lts and pre in the `Regression_I` group, which is three of
the red jobs on master.

I tried to bisect the 44 commits since but old checkouts pick up current registry deps and die
on an unrelated `_find_large_jac_entries!` MethodError, so I don't have the culprit commit.
Anything in the Krylov tolerance area since Aug 9 look like the likely candidate to you?

singhharsh1708:
Fixes #4293.

`compute_step!` for `NLNewton` hands `integrator.opts.reltol` straight to the linear solve.
A Newton correction has to be more accurate than the tolerance it is being used to reach, so
at the default `reltol = 1e-3` the Krylov solve returns a direction good only to 1e-3, the
stage never converges to 1e-3, `dt` collapses and the step controller gives up.

`test/Regression_I/newton_krylov_roundoff.jl` is the thing that broke because it calls `solve`
with no tolerances, so the linear solve inherits the 1e-3 default. On `upstream/master` at
202f5fda90:

| | retcode | accepted | rejected | drift |
| --- | --- | --- | --- | --- |
| `Hairer4` master | Unstable | 3 | 31 | n/a |
| `Hairer42` master | Unstable | 5 | 31 | n/a |
| `Hairer4` here | Success | 21 | 1 | 4.44e-16 |
| `Hairer42` here | Success | 21 | 1 | 6.66e-16 |

The gate is `drift < 1e-5` and that file's own comment puts a round-off reproducible run in
the 1e-15 to 1e-7 band.

The floor is `sqrt(eps)` because that is LinearSolve's own default, which is where this sat
before #4192 made the linear tolerance integrator-driven. It is not a constant fitted to this
reproducer. A run looser than that behaves as it did before #4192, a tighter one still gets
the tighter linear solve.

## Why here and not in `set_linear_reltol!`

Clamping inside `set_linear_reltol!` also fixes the reproducer, and I measured it, but it
floors every caller including the Rosenbrock path. A Rosenbrock stage is a single linear solve
whose accuracy caps the stage directly rather than feeding an iteration, so `reltol` is the
right target there and flooring it is just over-solving: at `reltol = 1e-3` on Robertson it
moves `Rodas5P` from 1.1e-3 error to 2.1e-6 while doing more work for accuracy nobody asked
for. It also fails four assertions across two files rather than one.

I also tried extending the same floor to `set_inner_linear_reltol!`, on the theory that
`NonlinearSolveAlg`'s inner Krylov descent is a Newton direction too. That fails two more
assertions in `nsa_krylov_tolerance_tests.jl`, and that file's third testset already measures
the loose tolerance there as costing no accuracy and no `nnonlinconvfail`, so there is nothing
to fix on that path. Left alone.

## Test change, please look at this one

`lib/OrdinaryDiffEqDifferentiation/test/krylov_linear_tolerance_tests.jl` asserted
`newton_cache(integrator).reltol == tol` for `tol = 1e-6`, which is the behaviour that breaks
#4293, so it had to move to `min(tol, sqrt(eps(Float64)))`. The `tol = 1e-10` case in the same
loop is unchanged and still checks that a tighter tolerance passes through. The other three
assertions in that file, and all three testsets of `nsa_krylov_tolerance_tests.jl`, are
untouched and pass.

## Testing

Ran against this branch: `krylov_linear_tolerance_tests.jl`, `nf_accounting_tests.jl`,
`warm_start_default_tests.jl`, `stale_w_linear_operator_tests.jl`, `newton_krylov_roundoff.jl`,
`linear_nonlinear_tests.jl`, `nsa_krylov_tolerance_tests.jl`, `nsa_residual_convergence_tests.jl`,
`nsa_conditioning_tests.jl`, `mass_matrix_tests.jl`. All pass. `newton_tests.jl` I could not run
locally, `DiffEqDevTools` fails to precompile on my Julia 1.10 for unrelated reasons.

Also swept `Hairer4`, `Hairer42`, `TRBDF2`, `KenCarp4`, `FBDF`, `QNDF`, `Rodas5P`,
`Rosenbrock23` and the `NonlinearSolveAlg` variants of `TRBDF2` and `KenCarp4` with
`KrylovJL_GMRES` at default tolerances over an index-1 DAE, Allen-Cahn and Robertson. Every
case that succeeded on master still succeeds, and `KenCarp4` and `FBDF` on Robertson improve
from 2.7e-4 and 5.7e-4 error to 1.4e-6 and 3.1e-5.

One thing that sweep turned up and that this PR does not address: `Rosenbrock23` with
`KrylovJL_GMRES` on a mass-matrix problem with a `jac_prototype` errors with
`MethodError: no method matching jacobian!(::SciMLOperators.NullOperator, ...)`. That happens
on unmodified master too, so it is a separate bug. I can file it if it is not already known.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/test/krylov_linear_tolerance_tests.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/test/krylov_linear_tolerance_tests.jl
@@ -21,12 +21,15 @@ newton_cache(integrator) = integrator.cache.nlsolver.cache.linsolve
 @testset "integrator tolerance reaches the Krylov cache" begin
     # Newton path: `build_nlsolver` builds the `LinearCache` without tolerances, so
     # `dolinsolve` is the only thing that can put the integrator's `reltol` there.
+    # A Newton direction is floored at `sqrt(eps)` however loose the integrator is, because
+    # a correction solved only to `reltol` cannot drive the stage to `reltol` (#4293), so a
+    # looser tolerance clamps and a tighter one passes through unchanged.
     for tol in (1.0e-6, 1.0e-10)
         integrator = init(prob, FBDF(linsolve = KrylovJL_GMRES()); reltol = tol, abstol = tol)
         for _ in 1:5
             step!(integrator)
         end
-        @test newton_cache(integrator).reltol == tol
+        @test newton_cache(integrator).reltol == min(tol, sqrt(eps(Float64)))
     end
 
     integrator = init(prob, KenCarp4(linsolve = KrylovJL_GMRES()); reltol = 1.0e-7, abstol = 1.0e-7)
--- a/lib/OrdinaryDiffEqNonlinearSolve/Project.toml
+++ b/lib/OrdinaryDiffEqNonlinearSolve/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqNonlinearSolve"
 uuid = "127b3ac7-2247-4354-8eb6-78cf4e7c58e8"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
-version = "2.10.0"
+version = "2.10.1"
 
 [deps]
 CommonSolve = "38540f10-b2f7-11e9-35d8-d573e4eb0ff2"
--- a/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
+++ b/lib/OrdinaryDiffEqNonlinearSolve/src/newton.jl
@@ -616,6 +616,30 @@ end
 end
 
 """
+    newton_linear_reltol(reltol, uT) -> reltol
+
+Relative tolerance for the iterative linear solve that produces the modified Newton
+direction.
+
+The correction has to be more accurate than the tolerance it is being used to reach.
+Solving the stage system only to the integrator's own `reltol` gives a direction good to
+`reltol`, so the stage never converges there, `dt` collapses and the step controller gives
+up. That is SciML/OrdinaryDiffEq.jl#4293: the `Hairer4` and `Hairer42` runs in
+`test/Regression_I/newton_krylov_roundoff.jl` call `solve` with no tolerances, so the
+Krylov solve inherits the default `reltol` of 1e-3 and both return `Unstable`.
+
+The floor is `sqrt(eps)`, which is where LinearSolve's own default sat before the linear
+tolerance became integrator-driven, so a looser integrator behaves as it did before and a
+tighter one still gets the tighter linear solve.
+
+This is about Newton directions specifically. A Rosenbrock stage is a single linear solve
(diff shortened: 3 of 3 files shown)


ChrisRackauckas-Claude:
I reproduced the NonlinearSolve downstream Hairer regression locally and verified this patch discriminates it.

Using the same focused Hairer regression file with NonlinearSolve 4.27, NonlinearSolveBase 2.46, and SciMLBase 3.49:

```text
unfixed 4bb84954086103d6a3a472673cee84db18b7c816:
  Hairer4: two retcode failures; drift = 4096 (required < 1e-5)

this PR (02ac18e7d687756864b94e70a0909fd0b86ddef0):
  Hairer4  | 3/3 pass
  Hairer42 | 3/3 pass
```

The branch also cherry-picks cleanly onto current OrdinaryDiffEq master and `git diff --check` exits 0. The current Regression_I CI lanes on this PR are green.

For tracking, the failure surfaced from https://github.com/SciML/OrdinaryDiffEq.jl/commit/4bb84954086103d6a3a472673cee84db18b7c816. Current master has separate inherited QA/Core failures being investigated independently; they are not caused by this focused tolerance-floor change.

ChrisRackauckas:
No just update the test to use appropriate defaults.

ChrisRackauckas:
I mean not defaults but tolerances.

PR #4305, opened 2026-08-19
ExpRK: cache ishermitian implementation

KeshavVenkatesh:
## Checklist

- [ Yes] Appropriate tests were added
- [ No] Any code changes were done in a way that does not break public API
- [ N/A] All documentation related to code changes were updated
- [ Yes] The new code follows the
- [ N/A] Any new documentation only uses public API
  
## Additional context
While reviewing the Benchmark results of Exponential Utilties came across a way to improve the performance of ETDRK methods by caching the ishermitian call results. 

These are the results of the time savings with the implementation for per arnoldi call.

 -- per Arnoldi build (ms, min of 100 calls) --
  operator                                                ishermitian(A)   arnoldi! before   arnoldi!-after   saving
  Brusselator N=32 (symmetric L)               0.02463           0.24900          0.22187          10.9%
  upwind 1-D n=2048 (non-symmetric L)          0.00025           0.22587    0.22350          1.1%

### DIFF
--- a/lib/OrdinaryDiffEqExponentialRK/src/exponential_rk_caches.jl
+++ b/lib/OrdinaryDiffEqExponentialRK/src/exponential_rk_caches.jl
@@ -112,8 +112,23 @@ Symmetry of the ExpRK linear operator, or `nothing` when it cannot safely be cac
 `ETDRK4` step. For a sparse symmetric operator that check is a full `O(nnz)` scan (it can only
 exit early on finding an asymmetry), costing roughly 10% of a Krylov build at `m = 15`.
 
-A `SplitFunction`'s linear part is fixed for the whole solve, so the answer can be computed
-once. Everything else returns `nothing`, and `arnoldi!` derives the flag itself as before.
+Caching a *value* -- "are these entries symmetric?" -- is only accurate while the entries cannot
+change, so two conditions must hold. Anything else returns `nothing` and `arnoldi!` derives the
+flag itself on every call, exactly as before:
+
+  * `f isa SplitFunction`, so `A` is the fixed linear part rather than a Jacobian that
+    `calc_J!` rebuilds every step; and
+  * `isconstant(A)`, so the operator carries no `update_func` that `update_coefficients[!]`
+    could fire to replace its entries part-way through the solve.
+
+The second condition matters because the error is asymmetric. A stale `false` merely costs
+performance -- `arnoldi!` runs full Arnoldi. A stale `true` sends it to `lanczos!`, whose
+three-term recurrence is valid only for symmetric operators, and the result is *silently wrong*.
+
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
I'm a bit scared that it's possible to have a matrix that is weird right at the start. Hermitian at t=0 because the initial point is "odd" (ball on the ground, etc.), but then not for any t>0. It would be safer to periodically re-check.

KeshavVenkatesh:
> I'm a bit scared that it's possible to have a matrix that is weird right at the start. Hermitian at t=0 because the initial point is "odd" (ball on the ground, etc.), but then not for any t>0. It would be safer to periodically re-check.

@ChrisRackauckas : 
 Ontop of the current implementation, I have added a check to see if A is constant or if it can be modified mid-way through the solve.  Rather than re-check periodically, I've used `isconstant(A)` Is there any other scenario that we should handle?  Added a test case to verify this.
 
The issue with periodic re-check could still leave a window where we might return a stale true , which is not just a slow but a error condition.

lib/OrdinaryDiffEqExponentialRK/src/exponential_rk_caches.jl
```diff
@@ -112,15 +112,31 @@ Symmetry of the ExpRK linear operator, or `nothing` when it cannot safely be cac
...
 exit early on finding an asymmetry), costing roughly 10% of a Krylov build at `m = 15`.
 
-A `SplitFunction`'s linear part is fixed for the whole solve, so the answer can be computed
-once. Everything else returns `nothing`, and `arnoldi!` derives the flag itself as before.
+Caching a *value* -- "are these entries symmetric?" -- is only accurate while the entries cannot
+change, so two conditions must hold. Anything else returns `nothing` and `arnoldi!` derives the
+flag itself on every call, exactly as before:
+
+  * `f isa SplitFunction`, so `A` is the fixed linear part rather than a Jacobian that
+    `calc_J!` rebuilds every step; and
```
ChrisRackauckas, line 120: It's not necessarily a fixed linear part in a split function, only if the operator A is constant.

ChrisRackauckas:
Rebased and finished in https://github.com/SciML/OrdinaryDiffEq.jl/pull/4497. Kept your commits for authorship.

lib/OrdinaryDiffEqExponentialRK/src/exponential_rk_caches.jl
KeshavVenkatesh: As you pointed out the code has the right implementation. This first section of the comment was trying to check for the same logic we have in perform_step which has A = isa(f, SplitFunction) ? f.f1.f : (calc_J!(J, integrator, cache); J).

In the subsequent step we are handling if isConstant(A).

Issue #4281, opened 2026-08-16
RadauIIA: reuse one Jacobian reduction across the real and complex stage matrices


ChrisRackauckas-Claude:
A FIRK step forms a real stage matrix and a complex-conjugate pair from **one** Jacobian, then assembles and factorizes each separately. That is the ideal workload for a factorization that can be reused across shifts, and today none of it is reused.

### The measurement

Brusselator, n = 128, `RadauIIA5`, tol 1e-8, single-threaded BLAS:

| linsolve | time | error | njacs | nW |
|---|---|---|---|---|
| `LUFactorization()` | 0.169 s | 9.59e-09 | 3 | 279 |
| `LHLFactorization()` | 0.883 s (**5.2× slower**) | 9.59e-09 | 3 | 279 |

**279 stage-matrix updates against 3 Jacobians** — 93 shifts per Jacobian, a better ratio than the ~55 that makes this pay for BDF.

### Why none of it is reused

`firk_real_W` discards the split form whenever the linear solver needs a concrete `A`:

```julia
function firk_real_W(alg, J, W)
    (is_lazy_W(W) && !(J isa AbstractSciMLOperator)) || return W
    (alg.linsolve === nothing || LinearSolve.needs_concrete_A(alg.linsolve)) || return W
    return recursivefill!(similar(J), false)   # ← split thrown away
end
```

So `LHLFactorization` receives a plain matrix and re-reduces from scratch for every stage matrix — hence 5.2× slower rather than faster. SciML/LinearSolve.jl#1215 now raises rather than silently taking that path.

### What the fix looks like

The pieces already exist. LHLFactorization.jl reduces `J` once and serves any number of shifts off it, including a **complex shift on a real reduction** — verified at 1e-15, with one real reduction serving a real and a complex shift simultaneously:

```julia
ws  = lhl(J)                       # real, O(n³), once
shr = LHLShift{Float64}(ws);    lhl_shift!(shr, ws, -inv(γ_real), 1.0)
shc = LHLShift{ComplexF64}(ws); lhl_shift!(shc, ws, -inv(γ_complex), 1.0 + 0im)
```

That is exactly `W1` and `W2`. What is missing is a cache shape that expresses it: LinearSolve's model is one cache per matrix, so two stage matrices means two caches and two reductions. FIRK would need either a dedicated cache holding one `LHLWorkspace` and two `LHLShift`s, or a LinearSolve-level multi-shift cache.

[...]

PR #4333, opened 2026-08-21
Serve Radau's complex stage pair from the split W

ChrisRackauckas-Claude:
> [!NOTE]
> Please ignore this PR until it has been reviewed by @ChrisRackauckas.

Closes the headline of https://github.com/SciML/OrdinaryDiffEq.jl/issues/4281. Follows https://github.com/SciML/OrdinaryDiffEq.jl/pull/4266, which was squash-merged from its reviewed tip before this commit was pushed to that branch — hence the separate PR.

## What changed and why

A FIRK step forms a real stage matrix and complex-conjugate pairs from **one** Jacobian (279 stage-matrix updates against 3 Jacobians on the Brusselator), and previously discarded the split form, so `LHLFactorization` re-reduced from scratch per update — measured 5.2× slower than LU, which is why #4266 shipped an `ArgumentError` guard. Both stages now stay split: the real stage keeps the `WOperator` `build_J_W` made, and each complex pair becomes a `WOperator` whose `gamma` slot is complex while its Jacobian — the *same object* as the real stage's — stays real. Inside LinearSolve one real O(n³) reduction per stage matrix then serves every `dtgamma`, with only the shift going complex (`ws.factors::Matrix{Float64}`, `ws.σ::ComplexF64` — asserted in the tests). Covers RadauIIA3 (complex-only), RadauIIA5, RadauIIA9, and AdaptiveRadau (vector of pairs).

The correctness keystone is `firk_new_J!` marking **every** stage matrix stale when it rewrites `J` in place; the guard and `_firk_reject_reduction_reuse` are removed.

## Benchmark (the #4281 measurement)

RadauIIA5, 128-unknown Brusselator, tol 1e-8, single-threaded BLAS, best of 5:

```
LU          0.2308 s   err=5.024e-08   njacs=3  nw=279  nsolve=1668
LHL         0.0891 s   err=1.351e-07   njacs=1  nw=274  nsolve=1687   (2.6× faster)
LHL r=0     0.0754 s   err=1.351e-07                                  (3.1× faster)
```

against **5.2× slower** before. Caveat: the LHL run took `njacs=1` vs LU's 3 (different Newton histories), so the timing is not perfectly controlled; the correctness tests exercise `njacs > 1` explicitly.

## Failing-before / passing-after

Before this change the new tests cannot pass: `RadauIIA5(linsolve = LHLFactorization())` throws `ArgumentError` (the #4266 guard) — the old `@test_throws` asserting that is flipped by this PR into trajectory tests. The staleness marking is separately shown to be load-bearing, both ways:

- Reverting **all** marking: **6 failures** — failed retcodes and trajectories diverging from LU, because the pair's cached reduction keeps answering for the previous Jacobian.
- Reverting **only the vector path** (`_mark_J_moved!(::AbstractVector) = nothing`): **2 failures**, exactly the AdaptiveRadau assertions.

With the change, `lib/OrdinaryDiffEqBDF/test/lhl_factorization_tests.jl` exit 0:

```
Test Summary:                            | Pass  Total   Time
W is kept split                          |    3      3   3.5s
same trajectory as an assembled W        |    6      6  18.5s
the reduction is reused across dtgamma changes |  5   5   0.0s
Radau keeps both stage matrices split    |   20     20  23.7s
Radau re-reduces when the Jacobian moves |    2      2   0.0s
unsupported combinations are rejected    |    2      2   2.7s
```

including RadauIIA3/5/9/AdaptiveRadau trajectory ≈ LU at rtol 1e-6 on a stiff dense problem with `njacs > 1`, and AdaptiveRadau structure assertions (every element of the `W2` vector is a split complex-γ `WOperator` over the shared `J`).

`lib/OrdinaryDiffEqFIRK/test/ode_firk_tests.jl` (LU/Krylov/GaussLegendre/AdaptiveRadau paths, no LHL): exit 0, no failures. Runic and typos clean.

## Version

No bump: rides OrdinaryDiffEqFIRK 2.8.0, which #4266 bumped and which is not yet registered (registry has 2.7.0).

## Not verified / out of scope

- The full downgrade matrix, docs build, and downstream suites (CI).
- Sharing literally one reduction across the real and complex LinearSolve caches — each stage matrix keeps its own, re-taken only when `J` moves; with ~90 shifts amortizing each, the residual saving is a small constant and needs a cross-cache interface. Stays on #4281.
- GaussLegendre never took the split path and still does not.

https://claude.ai/code/session_014CrTAi2YLkosEXXbudXyKf


ChrisRackauckas:
Follow-up in bb955cd7d:

- Moved the Radau LHL factorization tests from the BDF Core suite into the FIRK Core suite.
- Removed the BDF test dependency on OrdinaryDiffEqFIRK.
- Declared the BDF test dependencies directly, including SciMLOperators compat 1.27 and the missing SparseArrays compat.

Local verification:

- OrdinaryDiffEqFIRK GROUP=Core: passed, including 22/22 FIRK LHL factorization tests.
- OrdinaryDiffEqBDF GROUP=Core: passed, including 16/16 BDF LHL factorization tests.
- Full repository Runic check: all 1078 Julia files passed.
- Typos and TOML/dependency metadata checks passed.

The BDF QA dependency-resolution blockers are fixed. That suite then reached its allocation checks locally and reported the existing 160-byte warmed FBDF and DFBDF allocations; those solver allocation checks are unrelated to this test-ownership change.

PR #4338, opened 2026-08-22
Let IDSolve drive polyalgorithm and SimpleNonlinearSolve caches

### LINKED ISSUE #4138: IDSolve crashes with any polyalgorithm or SimpleNonlinearSolve nlsolve
`nlsolve` is a public keyword on the exported `IDSolve` constructor. The default `NewtonRaphson` works, every NonlinearSolve polyalgorithm throws, and every SimpleNonlinearSolve algorithm is rejected.

```julia
using ImplicitDiscreteSolve, NonlinearSolve

f!(resid, u_next, u, p, t) = (resid[1] = u_next[1] - u[1] / 2; nothing)
prob = ImplicitDiscreteProblem(f!, [1.0], (0, 5), nothing)

solve(prob, IDSolve())                                  # Success, u(end) = [0.0625]
solve(prob, IDSolve(nlsolve = RobustMultiNewton()))     # throws
solve(prob, IDSolve(nlsolve = SimpleNewtonRaphson()))   # throws
```

observed:

```
RobustMultiNewton              ErrorException: type NonlinearSolvePolyAlgorithmCache has no field termination_cache
FastShortcutNonlinearPolyalg   same
SimpleNewtonRaphson            ArgumentError: `solve_cache!` requires an algorithm that supports the nonlinear solver iterator interface.
```

expected: all of them integrate the same trajectory.

cause: `_solve_nonlinear!` (lib/ImplicitDiscreteSolve/src/solve.jl:34) drives the cache with `NonlinearSolveBase.solve_cache!`, which reads `cache.termination_cache` unconditionally. A polyalg cache keeps that on its subcaches and a no-init cache does not have it at all. same class as #3859, fixed for `NonlinearSolveAlg` in #3959/#4137 and for Newmark in #4112, by driving `solve!` and reading the returned solution instead of cache internals.

singhharsh1708:
`IDSolve` now drives a `NonlinearSolveNoInitCache` (what SimpleNonlinearSolve algorithms return) and a polyalgorithm cache through `solve!` instead of `solve_cache!`, and reads the result off the returned solution. Before, every polyalgorithm threw `has no field termination_cache`, and falling back to `solve!` while still reading `state_values` afterwards turned that into a silent wrong answer: `SimpleNewtonRaphson` returned `u(end) = 1.0`, the initial condition, with `Success`.

The polyalgorithm stays on `solve!` even though NonlinearSolve.jl#1189 now lets `solve_cache!` accept it: observing its steps hands the Kantorovich controller the subsolver's rates and `dt` is driven below eps on the first step.

Four inner solvers go from 0 of 4 to 4 of 4 passing and match the default's `t` and `u[end]`; the full `ImplicitDiscreteSolve` suite passes. Noticed and not touched: `TrustRegion()` returns 2.9458767598199065e-13 rather than 0.0625 on this problem, bit-identical before and after, so pre-existing and separate.

Fixes #4138.

### DIFF
--- a/lib/ImplicitDiscreteSolve/src/solve.jl
+++ b/lib/ImplicitDiscreteSolve/src/solve.jl
@@ -20,19 +20,40 @@ function perform_step!(integrator, cache::IDSolveCache, repeat_step = false)
 
     # nonlinear solve step
     SciMLBase.reinit!(nlcache, cache.z; p = state)
-    if !_solve_nonlinear!(nlcache, observer)
+    converged, znew = _solve_nonlinear!(nlcache, observer)
+    if !converged
         integrator.force_stepfail = true
         return
     end
 
     # Accept step
-    return u .= state_values(nlcache)
+    return u .= znew
 end
+
+"""
+    _can_observe_steps(nlcache) -> Bool
+
+Whether the nonlinear solve is driven step by step through `NonlinearSolveBase.solve_cache!`
+with the step observer, or run to completion with `solve!`.
+
+`NonlinearSolveNoInitCache` is NonlinearSolveBase's public marker for a solver without the
+iterator interface, so it takes the `solve!` path. A polyalgorithm cache does too, even
+though it can be stepped: observing it feeds every subsolver's iterations, including the
+ones that fail before the next subsolver takes over, to the step controller, which drives
+`dt` to zero on the first step.
+"""
+@inline _can_observe_steps(nlcache) = true
+@inline _can_observe_steps(::NonlinearSolveBase.NonlinearSolveNoInitCache) = false
+@inline _can_observe_steps(::NonlinearSolveBase.NonlinearSolvePolyAlgorithmCache) = false
 
 function _solve_nonlinear!(nlcache, observer)
     reset!(observer)
-    retcode = NonlinearSolveBase.solve_cache!(nlcache; step_observer = observer)
-    return retcode == ReturnCode.Success
+    if _can_observe_steps(nlcache)
+        retcode = NonlinearSolveBase.solve_cache!(nlcache; step_observer = observer)
+        return retcode == ReturnCode.Success, state_values(nlcache)
+    end
+    sol = solve!(nlcache)
+    return sol.retcode == ReturnCode.Success, sol.u
 end
 
 function initialize!(integrator, cache::IDSolveCache)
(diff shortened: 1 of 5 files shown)


lib/ImplicitDiscreteSolve/src/solve.jl
```diff
@@ -20,19 +20,48 @@ function perform_step!(integrator, cache::IDSolveCache, repeat_step = false)
...
+
+`solve_cache!` runs the nonlinear solver iterator interface and then reads
+`cache.termination_cache` directly. A `NonlinearSolvePolyAlgorithmCache` keeps its
+termination cache on whichever subcache it selected, and a SimpleNonlinearSolve cache
+implements no `step!` at all, so neither can be driven that way
+(SciML/OrdinaryDiffEq.jl#4138). Both checks are on concrete types, so this folds away.
+"""
+@inline function _can_observe_steps(nlcache)
+    return applicable(step!, nlcache) && hasfield(typeof(nlcache), :termination_cache)
+end
```
ChrisRackauckas, line 45: this is a bit of a hack: it should be using an interface in NonlinearSolve

singhharsh1708:
Fair. The SimpleNonlinearSolve half now uses the `NonlinearSolveNoInitCache` branch from your
iterator-interface docs.

The polyalg half turned out to be an upstream bug rather than something this package should know
about: `solve_cache!` accepts a polyalg cache, then reads `cache.termination_cache` and
`cache.trace`, which a polyalgorithm keeps on its active subsolver. SciML/NonlinearSolve.jl#1189
gives both the forwarding hook `get_u` already has. Once that releases, this is just the
documented cache check.

singhharsh1708:
NonlinearSolve#1189 shipped in NonlinearSolveBase 2.47, so I went back to route the polyalgorithm through `solve_cache!` as I said above. It does not work and I was wrong about the reason it was excluded: `solve_cache!` on the polyalg cache itself is fine now, but observing it hands the Kantorovich controller the active subsolver's rates and dt is forced below eps on the first step (`Unstable` at t = 0 on the implicit Euler test). NewtonRaphson and SimpleNewtonRaphson through the same observer path are unaffected.

So the polyalg stays on `solve!` regardless of the upstream fix, and the docstring now says that instead of the earlier story. Rebased onto master.

lib/ImplicitDiscreteSolve/src/solve.jl
singhharsh1708: Now dispatches on the public `NonlinearSolveNoInitCache`, which NonlinearSolveBase documents as the marker for a solver without the iterator interface (and what OrdinaryDiffEqNonlinearSolve checks too), instead of sniffing fields. The polyalgorithm case stays a dispatch on its cache type: observing one feeds the failed subsolvers' iterations to the step controller and drives dt to zero.

PR #4339, opened 2026-08-22
Skip the explicit stage machinery for purely implicit SDIRK methods

### LINKED ISSUE #4149: TRBDF2 on SplitODEProblem: BoundsError from 0x0 tableau in generic_imex_perform_step
TRBDF2 on a SplitODEProblem crashes with a BoundsError from a 0x0 tableau.

```julia
using OrdinaryDiffEqSDIRK, SciMLBase

f1 = (du, u, p, t) -> (du .= -u)
f2 = (du, u, p, t) -> (du .= 0.1 .* u)
prob = SplitODEProblem(f1, f2, [1.0], (0.0, 1.0))
solve(prob, TRBDF2())   # BoundsError at generic_imex_perform_step.jl:144
```

OrdinaryDiffEqSDIRK v2.8.2. Surfaced while benchmarking multirate methods against single rate references on split problems. KenCarp4 handles the same problem, so the IMEX path exists; the pure-SDIRK tableau route through the generic perform_step gets a 0x0 explicit tableau instead of treating f1+f2 as the implicit part.

Expected: either solve the combined RHS implicitly the way a non-split TRBDF2 solve would, or reject SplitODEProblem with a clear error naming the IMEX-capable alternatives.

singhharsh1708:
Fixes #4149.

A purely implicit tableau has no explicit part: `Ae`, `be` and `split_guess` are empty. The
generic perform_step gated all of the explicit machinery on `integrator.f isa SplitFunction`
alone, so handing any of those methods a `SplitODEProblem` walked `tmp += Ae[i, j] * ks[j]`,
the `be[i] * ks[i]` output sum and `split_guess[i] > 0` and indexed empty arrays.

The gate is now `is_imex = integrator.f isa SplitFunction && issplit(alg)`, which is what the
seven sites that were already correct used. Nothing runs the explicit path unless there is an
explicit part to run it with, and the `k_i` buffers are no longer allocated for methods that
have no explicit stages.

`f2` is not lost. It reaches the step through `fsalfirst` and through the implicit solve, which
sees the whole `SplitFunction` rather than `f.f1`, so the method integrates `f1 + f2` the way a
non-split solve of the same method does. That is the first of the two behaviours the issue
offered.

The issue reported `TRBDF2`. It is twelve methods:

```
TRBDF2  ImplicitEuler  ImplicitMidpoint  SDIRK2  Cash4  Hairer4  Hairer42
SSPSDIRK2  SFSDIRK4  SFSDIRK5  SFSDIRK6  SFSDIRK7  SFSDIRK8
```

`Trapezoid`, `SDIRK22`, `Kvaerno3`, `Kvaerno4` and the `KenCarp` family were already fine, and
`Kvaerno4` is the reason this is the right shape of fix: it is not IMEX either, it already
solved a `SplitODEProblem` correctly, and it did so without any explicit contribution.

Checked against the same method run on `f1 + f2` directly, with `f1 = -u`, `f2 = 0.1u` and fixed
`dt = 0.01` so the comparison does not depend on adaptivity. `ImplicitEuler`, `Trapezoid`,
`Cash4`, `SDIRK2`, `SSPSDIRK2`, `SFSDIRK5`, `Kvaerno4` and `ImplicitMidpoint` come out bit
identical; `TRBDF2`, `Hairer4` and `SFSDIRK8` differ by one ulp, 1.11e-16 and 5.55e-17, which is
`f1 + f2` accumulating through the split function's cache rather than a single fused
right-hand side.

## Testing

A testset in `tableau_consistency_tests.jl` over the `all_algs` list already there: every
non-IMEX method solves the `SplitODEProblem` and matches the combined solve to `rtol = 1e-12`.
IMEX methods are skipped by `issplit`, since treating the split differently is the point of
those.

Verified red to green against `upstream/master`: 13 of the 33 sub-testsets error without the
change, all pass with it. `tableau_consistency_tests.jl`, `predictor_tests.jl`,
`sdirk_convergence_tests.jl` and `dae_esdirk_test.jl` all pass.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqSDIRK/Project.toml
+++ b/lib/OrdinaryDiffEqSDIRK/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqSDIRK"
 uuid = "2d112036-d095-4a1e-ab9a-08536f3ecdbf"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "2.9.0"
+version = "2.9.1"
 
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
--- a/lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
+++ b/lib/OrdinaryDiffEqSDIRK/src/generic_imex_perform_step.jl
@@ -67,12 +67,17 @@ end
     reuse_W_at_stage2 = tab.reuse_W_at_stage2
     split_guess = tab.split_guess
     alg = unwrap_alg(integrator, true)
+    # Only an IMEX tableau has an explicit part to put `f2` in. A `SplitFunction` handed to a
+    # purely implicit method has `Ae`, `be` and `split_guess` empty, so none of the explicit
+    # machinery below may run: `f1 + f2` goes to the implicit solve instead, which is what a
+    # non-split solve of the same method does (#4149).
+    is_imex = integrator.f isa SplitFunction && issplit(alg)
     predictor = _predictor(alg)
     s = tab.s
     γ = Ai[s, s]
 
     f2 = nothing
-    if integrator.f isa SplitFunction
+    if is_imex
         f_impl = integrator.f.f1
         f2 = integrator.f.f2
     else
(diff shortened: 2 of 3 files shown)


lib/OrdinaryDiffEqSDIRK/src/imex_tableaus.jl
```diff
@@ -58,9 +58,21 @@ function ESDIRKIMEXTableau(
     # ce defaults to c — preserves ESDIRK behaviour (implicit and explicit abscissae
     # are the same). IMEX-SSP / BHR methods supply a distinct ce.
     ce_vec = isnothing(ce) ? copy(c) : ce
+    # A purely implicit tableau carries no explicit part and passes an empty `Ae`/`be`.
+    # The generic perform_step still walks the `tmp += Ae[i,j]*ks[j]` accumulator and the
+    # `be[i]*ks[i]` output sum whenever `f` is a `SplitFunction`, whatever the algorithm,
+    # so an empty one indexes out of bounds rather than contributing nothing: twelve
+    # methods threw `BoundsError` on a `SplitODEProblem` (#4149). Sizing them to zeros
```
ChrisRackauckas: why does it still walk it?

Issue #4140, opened 2026-08-05
Dense all-zero jac_prototype silently zeroes the AD Jacobian for every implicit solver


singhharsh1708:
`jac_prototype = zeros(n, n)` silently zeroes the AD Jacobian for every implicit solver. `ODEFunction` defaults `sparsity` to `jac_prototype`, `prepare_user_sparsity` wraps a dense matrix in a `KnownJacobianSparsityDetector`, and an all-zero dense matrix declares "no nonzeros", so the coloring is empty and AD writes nothing.

The right way to get a dense buffer is passing nothing, which is the default. The issue is the failure mode when someone passes this anyway: nothing errors, and the solve degrades or silently mis-solves.

```julia
using OrdinaryDiffEqRosenbrock, SciMLBase

f!(du, u, p, t) = (du[1] = -1000u[1] + 999u[2]; du[2] = -u[2])
prob = ODEProblem(ODEFunction(f!; jac_prototype = zeros(2, 2)), [1.0, 1.0], (0.0, 1.0))
integ = init(prob, Rosenbrock23())
step!(integ); integ.cache.J        # [0.0 0.0; 0.0 0.0]
```

under adaptive control the answer stays right and the cost explodes: Rodas5P at 1e-10 takes 79932 steps against 182 without the prototype. at fixed dt it is a silent wrong answer: dt = 0.005 gives Rosenbrock23 and Rodas5P `Success` with relative error around 7e9, ImplicitEuler and TRBDF2 `ConvergenceFailure`. `spzeros(n, n)` and any sparse pattern that is a strict subset of the true Jacobian fail the same way, and the subset case is a plausible mistake rather than a strange input.

MREIL (#4115) rejects the dense all-zero case with an `ArgumentError` because for a linearly implicit method a zero J changes the method rather than just the convergence rate. Suggesting the same loud rejection generically, or a docs note, whichever fits.

ChrisRackauckas:
> zeros(n, n) is a natural way to spell "give me a dense n by n buffer".

No, the natural way is just `nothing`, which is the default.

singhharsh1708:
fair, nothing is the right way, reworded the issue around the failure mode instead: nothing errors when a degenerate or undercovering pattern is passed, fixed dt returns Success at relerr ~7e9, and the strict-subset sparse case fails the same way. loud rejection or a docs note either way, your call.

singhharsh1708:
Two corrections to my own text above, both measured.

[...]

PR #4342, opened 2026-08-22
Keep the sparsity pattern when building the diagnostic Jacobian

singhharsh1708:
`get_fresh_jacobian` builds its scratch Jacobian with

```julia
Jfresh = zero(cache.J)
```

`zero` on a `SparseMatrixCSC` returns a matrix with no stored entries. That is the pattern
`calc_J!` colours against, so the next line throws:

```
DimensionMismatch: `A` and `bg.S2` must have the same sparsity pattern.
```

This is the numerical-instability diagnostic path, reached from `instability_jacobian` for the
caches that carry `J` themselves. So a sparse solve that goes unstable raises a confusing
`DimensionMismatch` from the reporting code instead of returning `retcode = Unstable` with the
Jacobian it was trying to show.

It needs nothing unusual to hit. A correct tridiagonal prototype for a heat equation is enough:

```julia
n = 12
tri = spdiagm(-1 => ones(n - 1), 0 => ones(n), 1 => ones(n - 1))
prob = ODEProblem(ODEFunction(heat!; jac_prototype = tri), u0, (0.0, 1.0))
integ = init(prob, Rosenbrock23(); dt = 0.01, adaptive = false)
step!(integ); step!(integ)
OrdinaryDiffEqCore.get_fresh_jacobian(integ, integ.cache)   # throws on master
```

`nnz(cache.J)` is 34 there and `nnz(zero(cache.J))` is 0. `RadauIIA5` behaves the same.

`fill!(similar(cache.J), zero(eltype(cache.J)))` keeps the pattern and still hands `calc_J!` a
zeroed buffer, and is the same thing as `zero` for a dense `J`. It needs no `SparseArrays`
dependency in this file.

## Testing

Added to `nojac_tests.jl`, which already imports `get_fresh_jacobian`: with both a sparse and a
dense prototype, on `Rosenbrock23` and `Rodas5P`, the returned Jacobian must match the analytic
one and keep `nnz`.

Verified red to green against `upstream/master`: without the change the new testset errors with
the `DimensionMismatch` above; with it the file passes 19 of 19. The recovered Jacobian is exact,
`maxabsdiff` against the analytic Jacobian is 0 for sparse and dense on both algorithms.

Found while working #4140, but it is independent of that: this fires with a fully correct
sparsity pattern, not only a degenerate one.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/Project.toml
+++ b/lib/OrdinaryDiffEqDifferentiation/Project.toml
@@ -1,5 +1,5 @@
 name = "OrdinaryDiffEqDifferentiation"
-version = "3.10.0"
+version = "3.10.1"
 uuid = "4302a76b-040a-498a-8c04-15b101fed76b"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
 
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Supersceded

PR #4344, opened 2026-08-22
Reject a jac_prototype that declares no structural nonzeros

singhharsh1708:
`prepare_alg` now throws when a sparse `jac_prototype` stores no entries and there is no analytic `jac` or `Wfact`, because an empty pattern gives the coloring nothing to write and the solve then runs on whatever the mass matrix diagonal patch leaves behind. On #4140's 2x2 problem that is a diagonal `J` with the coupling missing, and `Rosenbrock23` at dt = 0.005 returns `Success` with relative error 8.4e168 while `Rodas5P` at 1e-10 takes 494 steps against 27.
The test is structural, so it is the SparseArrays extension's `nnz(S) == 0` and nothing is converted to sparse to count; a sparse prototype whose stored entries are all zero still stores them and is accepted. It runs before the mass matrix patch writes into the prototype.
The dense all-zero prototype in #4140's title is not covered here. It leaves `J == 0` and a silent wrong answer for the same reason, but rejecting it means reading a dense buffer's values as structure, which is a separate call to make.

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/ext/OrdinaryDiffEqDifferentiationSparseArraysExt.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/ext/OrdinaryDiffEqDifferentiationSparseArraysExt.jl
@@ -2,11 +2,15 @@ module OrdinaryDiffEqDifferentiationSparseArraysExt
 
 using OrdinaryDiffEqDifferentiation
 import SparseArrays
-import SparseArrays: nonzeros, spzeros, SparseMatrixCSC, AbstractSparseMatrix
+import SparseArrays: nonzeros, spzeros, SparseMatrixCSC, AbstractSparseMatrix, nnz
 
 # Override the sparse checking functions
 OrdinaryDiffEqDifferentiation.is_sparse(::AbstractSparseMatrix) = true
 OrdinaryDiffEqDifferentiation.is_sparse_csc(::SparseMatrixCSC) = true
+
+function OrdinaryDiffEqDifferentiation._declares_no_nonzeros(S::AbstractSparseMatrix)
+    return nnz(S) == 0
+end
 
 # Override the sparse array manipulation functions
 OrdinaryDiffEqDifferentiation.nonzeros(A::AbstractSparseMatrix) = nonzeros(A)
--- a/lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
@@ -59,6 +59,18 @@ import DiffEqBase: OrdinaryDiffEqTag
 is_sparse(::Any) = false
 is_sparse_csc(::Any) = false
 
+"""
+    _declares_no_nonzeros(S) -> Bool
+
+Whether `S`, used as a `jac_prototype`/`sparsity`, stores no entries at all, so the
+coloring is empty and AD writes nothing into the Jacobian.
+
+Only structural emptiness counts: a `SparseMatrixCSC` whose stored entries are all
+`0.0` still stores them. The test needs `nnz`, so it lives in the SparseArrays
+extension alongside `is_sparse_csc`.
+"""
+_declares_no_nonzeros(::Any) = false
+
 # Seeding a sparsity pattern from the mass matrix uses `findall`/`getindex`, which a
 # SciMLOperator does not support. `convert` reads the operator's currently-cached array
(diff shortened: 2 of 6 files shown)


lib/OrdinaryDiffEqDifferentiation/ext/OrdinaryDiffEqDifferentiationSparseArraysExt.jl
```diff
@@ -2,12 +2,19 @@ module OrdinaryDiffEqDifferentiationSparseArraysExt
...
 
 # Override the sparse checking functions
 OrdinaryDiffEqDifferentiation.is_sparse(::AbstractSparseMatrix) = true
 OrdinaryDiffEqDifferentiation.is_sparse_csc(::SparseMatrixCSC) = true
 
+function OrdinaryDiffEqDifferentiation._declares_no_nonzeros(S::AbstractSparseMatrix)
+    return nnz(S) == 0
+end
+function OrdinaryDiffEqDifferentiation._declares_no_nonzeros(S::AbstractMatrix)
+    return nnz(sparse(S)) == 0
```
ChrisRackauckas: this can be a big long allocation, not a good idea.

lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
```diff
@@ -59,6 +59,24 @@ import DiffEqBase: OrdinaryDiffEqTag
...
+The quantity that matters is what the colouring consumes, the count of *stored* entries,
+not whether the values are zero. A `SparseMatrixCSC` whose stored entries are all `0.0`
+still declares the right structure, and a `Tridiagonal` of zeros has real structure too;
+only a pattern with nothing stored leaves AD nothing to write. The structured and sparse
+cases need `nnz`, so they live in the SparseArrays extension alongside `is_sparse_csc`.
+
+Matrix prototypes only; an operator-valued prototype is not classified here.
+"""
+_declares_no_nonzeros(::Any) = false
+_declares_no_nonzeros(S::StridedMatrix) = all(iszero, S)
```
ChrisRackauckas: why do the values matter?

lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
```diff
@@ -59,6 +59,24 @@ import DiffEqBase: OrdinaryDiffEqTag
...
+_declares_no_nonzeros(S::AbstractArray{Bool}) = !any(S)
```
ChrisRackauckas: ? no

lib/OrdinaryDiffEqDifferentiation/ext/OrdinaryDiffEqDifferentiationSparseArraysExt.jl
singhharsh1708: Removed; no dense conversion anywhere now.

lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
singhharsh1708: They do not; only stored entries do. Dense and structured prototypes are never rejected now, only a sparse matrix with nothing stored.

lib/OrdinaryDiffEqDifferentiation/src/OrdinaryDiffEqDifferentiation.jl
singhharsh1708: Dropped.

ChrisRackauckas:
I don't see why this is necessary at all.

singhharsh1708:
Correction to my numbers above: `spzeros(2,2)` does not leave `J == 0`, it leaves a wrong diagonal `J` once the mass matrix patch runs, and that returns Success with relative error 8.4e168 at dt = 0.005 (494 steps against 27 adaptive at 1e-10). The zero-J case in #4140's title is the dense one and this PR does not fix it. Still happy to drop this if the sparse case alone is not worth an error.

ChrisRackauckas:
The problem in that issue is that the Jacobian is wrong. If it's actually a pure zero Jacobian, then the solution is constant, and it does that correctly.

singhharsh1708:
You are right. And worse than unnecessary: `spzeros(n,n)` is a truthful declaration when J really is zero, and those cases already solve correctly, so this check would reject working code. Closing.

PR #4345, opened 2026-08-22
Fix crashes when a cache holds a vector of nonlinear solvers

singhharsh1708:
`instability_jacobian` and `get_tmp_cache` assumed `cache.nlsolver` is one `NLSolver`; `PDIRK44` keeps one per parallel stage in a `Vector`, so both threw `type Array has no field ...` when reached. The diagnostic now checks the property first and `PDIRK44` gets its own `get_tmp_cache`.

Both fail on master with those errors and pass after; covered in `instability_diagnostics_tests.jl` and `core_interface_tests.jl`. #4351 bumps the same Core version line, so whichever lands second needs it refreshed.


ChrisRackauckas:
rebasee

singhharsh1708:
Rebased onto master. Core and PDIRK bumps moved to 4.15.3 and 2.5.2 since master had taken the earlier numbers.

PR #4346, opened 2026-08-23
Add adaptive step size control to SDC

singhharsh1708:
Makes `SDC` adaptive: the embedded solution is the previous sweep's step update, already in the cache, so the estimate costs a few `axpy`s and no extra right-hand side evaluations; the algorithm moves to `OrdinaryDiffEqNewtonAdaptiveAlgorithm` with `alg_adaptive_order` one below the order. Fixed steps remain available with `adaptive = false` and a `dt`.

On the reference problem at `M = 3`, `K = 3` the adaptive solve takes 272 steps with 188x fewer Jacobian evaluations than the fixed-step run at the same accuracy; on Prothero-Robinson at `lambda = -1e4` and tolerance `1e-8`, `K = 3` takes 22768 steps and 206203 nonlinear solves for an error of 4.1e-9.

The base type change alters results slightly even under `adaptive = false`; I bisected that rather than assuming, and the pre-existing numbers return exactly when the base type is reverted. The nonlinear tests now pick step sizes by expected order, as the Dahlquist gate already did, with a regression residual gate of 0.1 rather than qmat's 0.05.

The change here is the single "Add adaptive step size control to SDC" commit. Full suite passes.

### DIFF
--- a/lib/OrdinaryDiffEqSDC/Project.toml
+++ b/lib/OrdinaryDiffEqSDC/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqSDC"
 uuid = "aadddb15-010a-4799-93bd-652a57b90e34"
 authors = ["singhharsh1708 <hs1663531@gmail.com>"]
-version = "1.0.0"
+version = "1.1.0"
 
 [deps]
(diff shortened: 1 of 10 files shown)


singhharsh1708:
@ChrisRackauckas ready for review, base of the stack (#4346, then #4347, then #4348).

PR #4347, opened 2026-08-23
Add MIN-SR-S and MIN-SR-FLEX diagonal sweepers to SDC

singhharsh1708:
Adds the MIN-SR-S and MIN-SR-FLEX diagonal sweepers from Čaklović, Lunet, Götschel and Ruprecht (SISC 47, 2025); a diagonal `QΔ` removes the only node-to-node coupling inside a sweep, which is what makes parallel-across-the-nodes SDC worth doing. They run serially here.

The coefficients are tabulated for 2 to 10 nodes across both node distributions and all four quadrature types, matching how the library ships tableau data; asking for more nodes raises an error naming the limit. The residual spectral radius grows with `M`, from about 1e-8 at `M = 2` to 1e-1 at `M = 10`, because the paper gives no guarantee that an exact solution exists. `MIN_SR_FLEX` varies the preconditioner per sweep, so the tableau holds `Vector{Matrix}` read through `sdc_qdelta_for(tab, k)`.

On Prothero-Robinson at `lambda = -1e4`, `M = 3`, `dt = 0.1`, serial `LU` reaches 3.6e-3, 8.2e-4, 2.9e-7 and 1.2e-9 at `K = 3, 4, 6, 8`; the diagonal sweepers are worse than `LU` at equal sweeps, which is the cost parallelism has to buy back. Coefficients are checked against qmat in `sdc_diagonal_sweeper_tests.jl`, and both sweepers are now in the convergence-order gate in `sdc_convergence_tests.jl`.


singhharsh1708:
@ChrisRackauckas ready for review, stacked on #4346.

singhharsh1708:
@ChrisRackauckas rebased on master now that #4346 is in, ready for review.

PR #4348, opened 2026-08-23
Run SDC sweeps in parallel across the collocation nodes

singhharsh1708:
With a diagonal `QΔ` the `M` node solves in an SDC sweep are independent, so `threading = true` runs them concurrently on the in-place path; `BaseThreads()` and `PolyesterThreads()` from Core work too, and threading with a non-diagonal sweeper is rejected at construction because it would silently drop the coupling. This is parallel across the method in Speck's sense.

Stiff 1-D reaction-diffusion with `M = 4` nodes on 4 threads: 64 unknowns 0.0093 s to 0.0072 s (1.3x), 256 unknowns 0.0905 s to 0.0460 s (2.0x), 1024 unknowns 2.42 s to 1.42 s (1.7x). Results match the serial sweep; the full suite passes on one, two and four threads.

The threading tests run as their own `Threaded` group at two threads, because the default sublibrary matrix runs Core at one thread and the comparison against the serial sweep proves nothing there.

Threaded runs undercount `sol.stats`: `nlsolve!` and `calc_W!` add to `integrator.stats` with a plain `+=` from whichever thread runs the node, so the threading test compares solutions and not statistics. Counting exactly needs `nlsolve!` to take its own counters.


singhharsh1708:
@ChrisRackauckas ready for review, stacked on #4347.

singhharsh1708:
@ChrisRackauckas restacked on #4347 after the rebase, ready for review once that one is in.

PR #4351, opened 2026-08-23
Buffer solver statistics per nonlinear solve to fix threaded counts

singhharsh1708:
Nonlinear solves of a threaded algorithm now accumulate `nf`, `nw`, `nsolve`, `njacs`, `nnonliniter` and `nnonlinconvfail` in a per-solve `StatsDelta` and fold them into `integrator.stats` once per solve, so solvers that run several solves concurrently (`PDIRK44`) stop losing increments and report the same statistics on every run. `jacobian!` and `jacobian` take a `stats_sink` keyword and `calc_J!` passes the solver cache's, since counts charged while building a Jacobian were a second source of drift.
NonlinearSolve and the QA projects of PDIRK, Extrapolation and IMEXMultistep get the `OrdinaryDiffEqCore` path source SDIRK and Differentiation already have, because those environments were resolving Core from the registry and precompiling the new code against a Core without these names. `OrdinaryDiffEqCore` goes to 4.17.0 rather than a patch, since it gains ten public names.
The Core compat floors of Differentiation and NonlinearSolve still say 4.12 and 4.13 while their new code needs 4.17, so a registry resolution can pair them with a Core that lacks these names. Raising the floors cannot be done here: it makes every environment that takes Core from the registry unsatisfiable until 4.17.0 is released, so it belongs in a follow-up after this lands.
`stats_delta_tests.jl` covers the buffering.
Cost on the single-threaded path: the delta is only allocated for an algorithm whose `threading` is on (PDIRK today), every other nonlinear solve charges straight into `integrator.stats` as before, and the drain folds without the lock when Julia has one thread. Measured with BenchmarkTools on Robertson at 1e-8, this branch against master: KenCarp4 9.91 vs 9.90 ms at one thread and 10.07 vs 9.93 at four, TRBDF2 6.97 vs 6.97 and 7.07 vs 6.97, FBDF and Rodas5P unchanged; taking the lock on every solve had cost 13% on KenCarp4 at four threads, which is what this avoids.


ChrisRackauckas:
Does this impact the timings for non-multithreaded solves? Seems like the locks could have an overhead.

singhharsh1708:
Measured it: with the lock on every solve, KenCarp4 on Robertson lost 13% at four Julia threads (nothing at one). Now the delta only exists for threaded algorithms and the drain skips the lock on a single thread; timings are within noise of master at one and four threads (numbers in the body).

ChrisRackauckas-Claude:
I independently checked the overhead at base `e8290c0b` versus PR head `cd510384`. The two benchmark environments used the same registry dependency versions; only the three local packages version-bumped by this PR differed. Local measurements used Julia 1.10.12 on an AMD EPYC 7502 with one BLAS thread.

The clearest result is the stats-path microbenchmark on the PR code (10,000 BenchmarkTools samples, zero allocations in every case):

| operation | old/direct median | PR median | added time |
|---|---:|---:|---:|
| one non-threaded counter update | 4.04 ns | 6.01 ns | 1.97 ns |
| representative `nf` + `nsolve` + `nnonliniter` updates and postamble drain | 4.84 ns | 11.72 ns | 6.88 ns |
| same representative sequence using a buffered delta and the uncontended lock (4 Julia threads) | — | 87.28 ns | — |

The end-to-end BenchmarkTools runs showed identical allocation counts and bytes on both revisions for Robertson with KenCarp4/TRBDF2/FBDF/Rodas5P, for 1,000-step fixed-step KenCarp4/TRBDF2 stress cases, and for serial PDIRK44. Threaded PDIRK44 added 2 allocations / 128 bytes per `solve` (the two per-solve `StatsDelta`s): 5,604 allocations / 883,904 bytes on base versus 5,606 / 884,032 bytes on the PR.

The repository's CI benchmark jobs provide the cleaner end-to-end timing comparison because this shared host was heavily loaded during my run:

| runner | benchmark | master | PR | master / PR |
|---|---|---:|---:|---:|
| Julia LTS | Robertson/FBDF | 0.734 ± 0.027 ms | 0.816 ± 0.097 ms | 0.899 ± 0.11 |
| Julia LTS | Robertson/Rodas5P | 0.219 ± 0.0037 ms | 0.227 ± 0.011 ms | 0.965 ± 0.051 |
| Julia latest | Robertson/FBDF | 0.961 ± 0.057 ms | 0.959 ± 0.063 ms | 1.00 ± 0.089 |
| Julia latest | Robertson/Rodas5P | 0.260 ± 0.035 ms | 0.249 ± 0.032 ms | 1.04 ± 0.19 |

All four CI ratio intervals include 1. Together with the direct measurement of roughly 2 ns per non-threaded counter update, I do not see evidence of a measurable non-threaded regression. The threaded path does have the expected fixed cost: about 87 ns for a representative buffered update/drain sequence and 128 bytes per PDIRK44 solve in this case.

Commands used for the local measurements were the equivalent of:

```bash
taskset -c 47 env JULIA_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 julia +1.10.12 --startup-file=no --project=<pr-benchmark-env> micro_stats.jl
taskset -c 44-47 env JULIA_NUM_THREADS=4 OPENBLAS_NUM_THREADS=1 julia +1.10.12 --startup-file=no --project=<pr-benchmark-env> micro_stats.jl
```

CI details: [Julia LTS benchmark job](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/33415451529/job/99564877541), [Julia latest benchmark job](https://github.com/SciML/OrdinaryDiffEq.jl/actions/runs/33415451529/job/99564877816).

🤖 Benchmarked and drafted by an AI agent using Codex (harness version: unknown; model ID: unknown; session: local session, no shareable URL exposed).

singhharsh1708:
Matches what I measured: the fixed cost stays on the threaded path and the non-threaded lanes sit inside noise. CI is green on the rebase now.

ChrisRackauckas:
This isn't the right solution, https://github.com/SciML/OrdinaryDiffEq.jl/pull/4465 is better, though I'm not sure we actually want to do this quite yet...

PR #4357, opened 2026-08-23
Give SSPRK932 its own fsalfirst buffer

singhharsh1708:
`SSPRK932` is third order with dense output on and first order with it off.

```julia
f!(du, u, p, t) = (@. du = -0.7u + 0.3sin(t) + 0.1u^2; nothing)
prob = ODEProblem(f!, [1.0], (0.0, 1.0))
```

Measured against a fine `SSPRK104` reference, over `dt = 2^-4 … 2^-8`:

```
dense = true                       err 3.8e-11   orders [3.0, 3.0, 3.0, 2.99]
save_everystep = false, dense = false   err 1.1e-4    orders [1.01, 1.0, 1.0, 1.0]
```

## Cause

With no dense output, no intermediate saves, no callbacks and no `saveat`, the
integrator sets `calck = false`, and the in-place SSPRK caches then alias
`fsalfirst` to `k` to save an allocation
(`lib/OrdinaryDiffEqSSPRK/src/ssprk_caches.jl`):

```julia
k = zero(rate_prototype)
if calck
    fsalfirst = zero(k)
else
    fsalfirst = k
end
```

Seventeen SSPRK caches do this, and for sixteen of them it is fine: they write
`fsalfirst` once at the top of the step and never read it again. `SSPRK932` is
the exception. Its `u6*` stage reads the value back out:

```julia
f(fsalfirst, uprev, p, t)
...                                  # six `f(k, u, ...)` calls, k === fsalfirst
@.. u = (3 * uprev + dt_2 * integrator.fsalfirst + 2 * u) / 5
```

By the time that line runs, `fsalfirst` holds `f(u5)` or `f(u6)` instead of
`f(uprev)`. `isfsal(::SSPRK932) == false`, so this is not an FSAL value being
carried between steps — it is a scratch buffer the method needs to keep for the
whole step, and it cannot share storage with `k`.

## Fix

Always allocate `fsalfirst` for `SSPRK932`. The other sixteen caches keep the
alias.

This costs one array. `SSPRK932` needs `f(uprev)` alive for the whole step while
`k` is reused by every stage, so there is no way for it to be both correct and
share the buffer. The storage assertions in the `SSPRK932` block go from
`<= 6`/`<= 5` to `<= 7`/`<= 6` (measured: 7 and 6). No other method's storage
changes -- `SSPRK33` is 4/3, `SSPRK43` and `SSPRK104` are 6/5, unchanged.

## Testing

Swept all sixteen exported SSPRK methods for a `dense=true` vs
`save_everystep=false, dense=false` order gap. `SSPRK932` was the only one
affected, and it is 2.99 in both columns after this change.

`lib/OrdinaryDiffEqSSPRK/test/ode_ssprk_tests.jl` gains a testset that runs the
existing convergence problems with `save_everystep = false, dense = false` for
`SSPRK932` plus `SSPRK33`/`SSPRK43` as controls. Every convergence test in that
file currently runs with the default `calck = true`, which is why this was not
caught. The new testset fails on master and passes here.

Bumps `OrdinaryDiffEqSSPRK` one patch from master.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqSSPRK/src/ssprk_caches.jl
+++ b/lib/OrdinaryDiffEqSSPRK/src/ssprk_caches.jl
@@ -1047,9 +1047,6 @@ function alg_cache(
         ::Val{true}, verbose
     ) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
     k = zero(rate_prototype)
-    # The `u6*` stage reads `f(uprev)` back out of `fsalfirst`, after six `f(k, ...)`
-    # calls have overwritten `k`. Aliasing the two when `calck` is false -- as the
-    # other SSPRK caches do -- silently drops this method to first order.
     fsalfirst = zero(k)
     utilde = zero(u)
     atmp = similar(u, uEltypeNoUnits)
--- a/lib/OrdinaryDiffEqSSPRK/src/ssprk_caches.jl
+++ b/lib/OrdinaryDiffEqSSPRK/src/ssprk_caches.jl
@@ -1047,11 +1047,10 @@ function alg_cache(
         ::Val{true}, verbose
     ) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
     k = zero(rate_prototype)
-    if calck
-        fsalfirst = zero(k)
-    else
-        fsalfirst = k
-    end
+    # The `u6*` stage reads `f(uprev)` back out of `fsalfirst`, after six `f(k, ...)`
+    # calls have overwritten `k`. Aliasing the two when `calck` is false -- as the
+    # other SSPRK caches do -- silently drops this method to first order.
+    fsalfirst = zero(k)
     utilde = zero(u)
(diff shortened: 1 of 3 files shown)


lib/OrdinaryDiffEqSSPRK/src/ssprk_caches.jl
```diff
@@ -1047,11 +1047,10 @@ function alg_cache(
...
     ) where {uEltypeNoUnits, uBottomEltypeNoUnits, tTypeNoUnits}
     k = zero(rate_prototype)
-    if calck
-        fsalfirst = zero(k)
-    else
-        fsalfirst = k
-    end
+    # The `u6*` stage reads `f(uprev)` back out of `fsalfirst`, after six `f(k, ...)`
+    # calls have overwritten `k`. Aliasing the two when `calck` is false -- as the
+    # other SSPRK caches do -- silently drops this method to first order.
```
ChrisRackauckas:
```suggestion
```

PR #4359, opened 2026-08-23
Fill k in ExplicitRK's addsteps! when there is no B_interp

singhharsh1708:
`sol(t0, Val{1})` on an `ExplicitRK` solution reads past the end of `k` before the first step:

```julia
f!(du, u, p, t) = (du .= -u; nothing)
integ = init(ODEProblem(f!, [1.0, 2.0], (0.0, 1.0)), ExplicitRK(); dt = 0.1)
integ.sol(0.0, Val{1})   # BoundsError under --check-bounds=yes, silent OOB read otherwise
```

`hermite_interpolant!` reads `k[2][i]` under `@inbounds @simd ivdep`, so on the default `--check-bounds=auto` this is an unchecked read rather than an error.

## Cause

`_ode_addsteps!` for `ExplicitRK` returns without doing anything when there is no `B_interp`:

```julia
# Without B_interp, the default Hermite addsteps suffices.
isnothing(cache.B_interp) && return nothing
```

The comment states the intent, but defining the method is what prevents it: the generic `_ode_addsteps!` in `generic_dense.jl` is what fills `k[1]`/`k[2]`, and this specialisation shadows it. So `k` stays at length 1.

Instrumenting which path runs:

```
sol(t0)          addsteps! not called   (i₋ == i₊ && deriv === Val{0} short-circuits, #4214)
sol(t0, Val{1})  addsteps! called, B_interp === nothing, length(k) == 1, returns without filling
```

## Fix

The no-`B_interp` branch now does the two evaluations the generic method would, so `k[1] == f(uprev)` and `k[2] == f(u)`.

`sol(t0, Val{1})` then returns `NaN`, because the Hermite derivative divides by `dt == 0`. On erroring instead: that would make `ExplicitRK` the only method that does, since the rest of the Hermite-fallback set returns `NaN` here too.

```
ExplicitRK      [NaN, NaN]
RK4             [NaN, NaN]
Midpoint        [NaN, NaN]
BS3             [NaN, NaN]
ImplicitEuler   [NaN, NaN]
Tsit5           [-1.0, -2.0]
Vern7           [-1.0, -2.0]
```

Worth noting the methods with their own interpolants do return the correct `f(u0)`, so the Hermite fallback giving `NaN` at `dt == 0` is a real gap, just a wider one than this PR. Happy to raise it separately; changing it here would only move `ExplicitRK` out of step with the other five.

## Testing

The new testset asserts the `k` values rather than the interpolant output, so it pins the fix rather than the symptom. It errors on master with the `BoundsError` and passes here. Full `interpolation_tests.jl` passes under `--check-bounds=yes`.

Bumps `OrdinaryDiffEqExplicitRK` one patch from master.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqExplicitRK/src/interpolants.jl
+++ b/lib/OrdinaryDiffEqExplicitRK/src/interpolants.jl
@@ -16,9 +16,13 @@ const ExplicitRKCacheTypes = Union{ExplicitRKCache, ExplicitRKConstantCache}
         always_calc_begin = false, allow_calc_end = true,
         force_calc_end = false
     )
-    # Only compute all stage vectors when B_interp is available for generic RK
-    # interpolation. Without B_interp, the default Hermite addsteps suffices.
-    isnothing(cache.B_interp) && return nothing
+    if isnothing(cache.B_interp)
+        if length(k) < 2 || always_calc_begin
+            copyat_or_push!(k, 1, f(uprev, p, t))
+            copyat_or_push!(k, 2, f(u, p, t + dt))
+        end
+        return nothing
+    end
 
     (; A, c, stages) = cache
 
@@ -41,9 +45,15 @@ end
         always_calc_begin = false, allow_calc_end = true,
         force_calc_end = false
     )
-    # Only compute all stage vectors when B_interp is available for generic RK
-    # interpolation. Without B_interp, the default Hermite addsteps suffices.
-    isnothing(cache.tab.B_interp) && return nothing
+    if isnothing(cache.tab.B_interp)
+        if length(k) < 2 || always_calc_begin
+            f(cache.kk[1], uprev, p, t)
+            copyat_or_push!(k, 1, cache.kk[1])
+            f(cache.kk[1], u, p, t + dt)
+            copyat_or_push!(k, 2, cache.kk[1])
+        end
+        return nothing
+    end
 
     (; kk, tmp, tab) = cache
     (; A, c, stages) = tab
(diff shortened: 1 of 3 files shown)


lib/OrdinaryDiffEqExplicitRK/src/interpolants.jl
```diff
@@ -112,7 +112,13 @@ for order in 0:3
...
             B_interp = get_B_interp(cache)
             if isnothing(B_interp)
-                # Fall back to Hermite interpolation (the generic default)
+                # Fall back to Hermite interpolation (the generic default).
+                # `hermite_interpolant` reads `k[2]` under `@inbounds`, so the short-`k`
+                # guard the generic `_ode_interpolant` applies has to be repeated here --
+                # `sol(t0)` right after `init` has only `k[1]`.
+                if isempty(k) || length(k) < 2
+                    return OrdinaryDiffEqCore.linear_interpolant(Θ, dt, y₀, y₁, idxs, T)
+                end
```
ChrisRackauckas: We should instead error, but how are we getting here without getting these derivatives? There is a function call that always prepares derivatives that is supposed to be done before interpolation each time, is it not called?

ChrisRackauckas: Or rather, is `addsteps!` not setup on ExpiclitRK correctly? That is more likely the issue here, it doesn't have the right fallback.

singhharsh1708:
You were right, it is `addsteps!`. The `isnothing(B_interp)` early return shadows the generic method that fills `k`, so `k` stays length 1 and the interpolant reads `k[2]`. Fixed there and dropped the guard.

On erroring: `RK4`, `Midpoint`, `BS3` and `ImplicitEuler` all return NaN at `dt == 0` too, so erroring would single out `ExplicitRK`. `Tsit5`/`Vern7` return the correct `f(u0)` via their own interpolants, so the Hermite fallback is worth fixing properly, just wider than this PR.

PR #4361, opened 2026-08-23
Keep ExplicitTaylor2's Taylor seed out of the output slot

singhharsh1708:
`ExplicitTaylor2` is second order or first order depending on how the right-hand
side is spelled. Same ODE, `u' = -u`, in place:

```julia
direct!(du, u, p, t)     = (@. du = -u; nothing)
accumulate!(du, u, p, t) = (fill!(du, 0); @. du = du - u; nothing)
```

```
@. du = -u              err 9.38e-7   orders [2.03, 2.02, 2.01, 2.0]
fill!(du,0); du -= u    err 7.20e-4   orders [1.02, 1.01, 1.00, 1.0]
```

Zero-then-accumulate is the normal shape for a PDE semi-discretisation and for
ModelingToolkit-generated code, so this is not an exotic spelling.

## Cause

`lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_perform_step.jl`:

```julia
f(k1, uprev, p, t)
u1 = make_taylor(uprev, k1)     # k1 is u1's first-order partial
t1 = TaylorScalar{1}(t, one(t))
out1 = make_taylor(k1, k2)      # k1 is *also* the output value slot
f(out1, u1, p, t1)
@.. u = uprev + dt * k1 + dt^2 / 2 * k2
```

`out1.value === k1 === u1.partials[1]`. Writing the output therefore writes the
input seed. It survives only if the user's `f` assigns every element of `du` exactly
once; the moment `f` zeroes `du` first, the seed becomes zero, `k2` comes out zero,
and the `dt^2/2 * k2` term disappears. No error, no warning -- the method just
silently becomes Euler.

## Fix

Use `k3` as the value slot. `k3` is already allocated in the cache and wired to
`integrator.k[3]`, but nothing writes or reads it -- the `ExplicitTaylor2`
interpolant uses only `k[1]` and `k[2]` (`interpolants.jl`). So the step stays fully
non-allocating and `k1` is left intact. `out1.value` receives `f(uprev)`, which `k1`
already holds, so nothing is lost by discarding it.

## Testing

The existing convergence tests use `prob_ode_linear`/`prob_ode_2Dlinear`, whose RHS
assign `du` directly, which is why this was never caught.

Added a testset that runs the convergence check with both spellings and asserts the
two give the same answer. It measures order 1.02 for the accumulating RHS on master
and 2.03 with this change.

Bumps `OrdinaryDiffEqTaylorSeries` one patch from master.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_perform_step.jl
+++ b/lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_perform_step.jl
@@ -58,11 +58,6 @@ end
     f(k1, uprev, p, t)
     u1 = make_taylor(uprev, k1)
     t1 = TaylorScalar{1}(t, one(t))
-    # The value slot must not be `k1`: that is `u1`'s first-order partial, and an RHS
-    # that zeroes `du` before accumulating into it (`fill!(du, 0)`, or any `du .= 0`
-    # followed by `+=`) would wipe the Taylor seed it is being evaluated on, leaving
-    # `k2 == 0` and silently dropping the method to first order. `k3` is unused --
-    # the interpolant reads only `k[1]` and `k[2]` -- so this stays non-allocating.
     out1 = make_taylor(k3, k2)
     f(out1, u1, p, t1)
     @.. u = uprev + dt * k1 + dt^2 / 2 * k2
--- a/lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_perform_step.jl
+++ b/lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_perform_step.jl
@@ -58,7 +58,12 @@ end
     f(k1, uprev, p, t)
     u1 = make_taylor(uprev, k1)
     t1 = TaylorScalar{1}(t, one(t))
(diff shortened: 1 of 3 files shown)


lib/OrdinaryDiffEqTaylorSeries/src/TaylorSeries_perform_step.jl
```diff
@@ -58,7 +58,12 @@ end
     f(k1, uprev, p, t)
     u1 = make_taylor(uprev, k1)
     t1 = TaylorScalar{1}(t, one(t))
-    out1 = make_taylor(k1, k2)
+    # The value slot must not be `k1`: that is `u1`'s first-order partial, and an RHS
+    # that zeroes `du` before accumulating into it (`fill!(du, 0)`, or any `du .= 0`
+    # followed by `+=`) would wipe the Taylor seed it is being evaluated on, leaving
+    # `k2 == 0` and silently dropping the method to first order. `k3` is unused --
+    # the interpolant reads only `k[1]` and `k[2]` -- so this stays non-allocating.
```
ChrisRackauckas:
```suggestion
```

PR #4363, opened 2026-08-23
Copy into IRKC's f1ⱼ₋₂ buffer instead of rebinding onto du₁

singhharsh1708:
`IRKC` is order 1 in place and order 2 out of place.

Split linear problem, `f1 = diagm([-100, -50])` (stiff, implicit), `f2 = [0 1; -1 0]`
(non-stiff, explicit), `u0 = [1, 1]`, `t ∈ (0, 0.1)`, against `exp((A1+A2)t)u0`:

```
in-place       err 4.98e-3   orders [0.82, 0.93, 0.97, 0.99]
out-of-place   err 1.31e-3   orders [2.29, 2.06, 1.92, 1.78]
```

## Cause

`lib/OrdinaryDiffEqStabilizedIRK/src/irkc_perform_step.jl`, stage 1 of the in-place
method:

```julia
f1ⱼ₋₂ = du₁
```

`f1ⱼ₋₂` is a cache field (`irkc_caches.jl:17`) and is destructured a few lines above,
so this rebinds the name onto `du₁` and orphans the buffer. The stage shift inside
the loop is an in-place broadcast:

```julia
@.. broadcast = false f1ⱼ₋₂ = f1ⱼ₋₁
```

which therefore writes into `du₁`. `du₁` holds `f1(uprev)` and is read on every stage
-- the `nlsolver.tmp` expression uses `du₁` and `f1ⱼ₋₂` as two separate terms, so
after the first iteration both read the same array -- and again after the loop in
`@.. f1ⱼ₋₁ = du₁`.

The out-of-place method has the identical two lines and is fine, because there both
are value assignments and `du₁` is never mutated. That is why only the in-place path
loses an order.

## Fix

Copy instead of rebinding, so the cache's own buffer is the target:

```julia
@.. broadcast = false f1ⱼ₋₂ = du₁
```

Afterwards the two paths agree to better than 1e-10 at every step size tested, and
in-place measures the same orders as out-of-place.

## Testing

`lib/OrdinaryDiffEqStabilizedIRK/test/irkc_tests.jl` covered the eigenvalue estimator
and `@test_nowarn solve(prob, alg)`. Nothing checked convergence order, and nothing
compared the two paths, which is why an order loss in one of them went unnoticed.

Added a testset asserting the two paths agree and that the in-place order is above
1.5. Both assertions fail on master and pass here. The existing eigenvalue testset is
unchanged and still passes.

Bumps `OrdinaryDiffEqStabilizedIRK` one patch from master.

## AI Disclosure

### DIFF
--- a/lib/OrdinaryDiffEqStabilizedIRK/src/irkc_perform_step.jl
+++ b/lib/OrdinaryDiffEqStabilizedIRK/src/irkc_perform_step.jl
@@ -169,11 +169,6 @@ function perform_step!(integrator, cache::IRKCCache, repeat_step = false)
     Bⱼ₋₁ = 1 / ω₀
 
     #stage-1
-    # Copy, not rebind: the cache's own `f1ⱼ₋₂` buffer is the target. Rebinding the
-    # name to `du₁` makes the in-place `@.. f1ⱼ₋₂ = f1ⱼ₋₁` below write through to
-    # `du₁`, which has to hold `f1(uprev)` for the whole step -- it is read again on
-    # every stage and after the loop. The out-of-place method rebinds safely because
-    # there the same assignment is a value copy.
     @.. broadcast = false f1ⱼ₋₂ = du₁
     @.. broadcast = false gprev2 = uprev
     μs = ω₁ * Bⱼ₋₁
--- a/lib/OrdinaryDiffEqStabilizedIRK/src/irkc_perform_step.jl
+++ b/lib/OrdinaryDiffEqStabilizedIRK/src/irkc_perform_step.jl
@@ -169,7 +169,12 @@ function perform_step!(integrator, cache::IRKCCache, repeat_step = false)
     Bⱼ₋₁ = 1 / ω₀
 
     #stage-1
-    f1ⱼ₋₂ = du₁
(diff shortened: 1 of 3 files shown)


lib/OrdinaryDiffEqStabilizedIRK/src/irkc_perform_step.jl
```diff
@@ -169,7 +169,12 @@ function perform_step!(integrator, cache::IRKCCache, repeat_step = false)
     Bⱼ₋₁ = 1 / ω₀
 
     #stage-1
-    f1ⱼ₋₂ = du₁
+    # Copy, not rebind: the cache's own `f1ⱼ₋₂` buffer is the target. Rebinding the
+    # name to `du₁` makes the in-place `@.. f1ⱼ₋₂ = f1ⱼ₋₁` below write through to
+    # `du₁`, which has to hold `f1(uprev)` for the whole step -- it is read again on
+    # every stage and after the loop. The out-of-place method rebinds safely because
+    # there the same assignment is a value copy.
```
ChrisRackauckas:
```suggestion
```

PR #4365, opened 2026-08-24
Resolve current sub-cache in PredictiveControllerCache for composite algorithms (#4364)

### LINKED ISSUE #4364: AutoSwitch composite with a FIRK stiff member errors: "type CompositeCache has no field iter" in stepsize_controller!
## What happens

Any `CompositeAlgorithm` (e.g. the `AutoSwitch` wrappers `AutoVern7`, `AutoTsit5`, ...) whose
stiff member is a **FIRK** method (`RadauIIA3`, `RadauIIA5`, `RadauIIA9`) throws

```
type CompositeCache has no field iter
```

the first time the composite takes a step on the FIRK branch. A bare `RadauIIA5(...)` on the
same problem solves fine.

The offending line is in `OrdinaryDiffEqCore/src/integrators/controllers.jl`, in the
`PredictiveControllerCache` method of `stepsize_controller!`
([v4.14.3, line 1201-1206](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib/OrdinaryDiffEqCore/src/integrators/controllers.jl)):

```julia
if isfirk(alg)
    (; iter) = integrator.cache          # <-- assumes integrator.cache is the FIRK cache
    (; maxiters) = alg
else
    (; iter, maxiters) = integrator.cache.nlsolver
end
```

The composite dispatch just above it,

```julia
@inline function stepsize_controller!(integrator, cache::CompositeControllerCache, alg::CompositeAlgorithm)
    current_idx = integrator.cache.current
    return stepsize_controller!(integrator, @inbounds(cache.caches[current_idx]), @inbounds(alg.algs[current_idx]))
end
```

correctly forwards the *controller* cache and the *sub*-algorithm, but `integrator.cache` is
still the `CompositeCache`. So when `alg` is the FIRK member, `isfirk(alg)` is `true` and the
destructuring of `integrator.cache` fails. The `else` branch does not have this problem because
`integrator.cache.nlsolver` resolves through the `CompositeCache`'s `getproperty`.

`fac_default_gamma(RadauIIA5(...)) == false` and `isfirk(RadauIIA5(...)) == true`, so this path
is always taken once the composite switches to the FIRK branch — it is not tolerance- or
problem-dependent.

## Minimal reproducer

Only `OrdinaryDiffEq`, `OrdinaryDiffEqFIRK` and `ADTypes` are needed. `stiffalgfirst = true` is
used purely to make the switch to the FIRK branch happen on step 1 and keep the script
deterministic; the same error is what fires when the switch happens on its own (see "Where it
was hit").

```julia
using OrdinaryDiffEq, OrdinaryDiffEqFIRK, ADTypes

function rober!(du, u, p, t)
    y1, y2, y3 = u
    du[1] = -0.04*y1 + 1e4*y2*y3
    du[2] =  0.04*y1 - 1e4*y2*y3 - 3e7*y2^2
    du[3] =  3e7*y2^2
end
prob = ODEProblem(rober!, [1.0, 0.0, 0.0], (0.0, 1e5))

solve(prob, RadauIIA5(autodiff = AutoFiniteDiff()))                                     # works
solve(prob, AutoVern7(RadauIIA5(autodiff = AutoFiniteDiff()); stiffalgfirst = true))    # errors
solve(prob, AutoVern7(Rodas5P(); stiffalgfirst = true))                                 # works (not FIRK)
solve(prob, AutoTsit5(RadauIIA5(autodiff = AutoFiniteDiff()); stiffalgfirst = true))    # errors
```

## Observed output

```
### 1. plain RadauIIA5(autodiff = AutoFiniteDiff())
retcode = Success, steps = 44

### 2. AutoVern7(RadauIIA5(autodiff = AutoFiniteDiff())), stiff branch first
type CompositeCache has no field iter
  [1] getproperty  at Base.jl:49
  [2] stepsize_controller!  at controllers.jl:1202
  [3] stepsize_controller!  at controllers.jl:1291
  [4] stepsize_controller!  at controllers.jl:253
  [5] _loopfooter!  at integrator_utils.jl:610
  [6] loopfooter!  at integrator_utils.jl:595
  [7] solve!  at solve.jl:922
  [8] #__solve#97  at solve.jl:10

### 3. AutoVern7(Rodas5P()), stiff branch first  (non-FIRK control)
retcode = Success, steps = 73

### 4. AutoTsit5(RadauIIA5(autodiff = AutoFiniteDiff())), stiff branch first
type CompositeCache has no field iter
  [1] getproperty  at Base.jl:49
  [2] stepsize_controller!  at controllers.jl:1202
  [3] stepsize_controller!  at controllers.jl:1291
  [4] stepsize_controller!  at controllers.jl:253
  [5] _loopfooter!  at integrator_utils.jl:610
  [6] loopfooter!  at integrator_utils.jl:595
  [7] solve!  at solve.jl:922
  [8] #__solve#97  at solve.jl:10
```

It is not specific to `AutoFiniteDiff`, to `Vern7`, or to `RadauIIA5` — same script, same
problem, one call each:

```
AutoVern7(RadauIIA5())  [default AD]     -> type CompositeCache has no field iter
AutoVern7(RadauIIA3())                   -> type CompositeCache has no field iter
AutoVern7(RadauIIA9())                   -> type CompositeCache has no field iter
AutoVern7(FBDF())                        -> retcode = Success, steps = 136
AutoVern7(KenCarp4())                    -> retcode = Success, steps = 82
```

i.e. every FIRK stiff member fails and every non-FIRK stiff member works.

## Expected

`AutoVern7(RadauIIA5(...))` should integrate the problem, as it did before the OrdinaryDiffEq
v7 migration (OrdinaryDiffEq 6.101 era, where the same benchmark line read
`AutoVern7(RadauIIA5(autodiff = false))`).

The fix looks like it just needs the FIRK branch to reach the *current* sub-cache rather than
`integrator.cache`, the same way the `nlsolver` branch does.

## Version info

```
  [47edcb42] ADTypes v1.23.0
  [1dea7af3] OrdinaryDiffEq v7.6.0
  [bbf590c4] OrdinaryDiffEqCore v4.14.3
  [5960d6e9] OrdinaryDiffEqFIRK v2.7.0
  [43230ef6] OrdinaryDiffEqRosenbrock v2.6.6
  [2d112036] OrdinaryDiffEqSDIRK v2.8.2
  [79d7bb75] OrdinaryDiffEqVerner v2.3.0
  [0bca4576] SciMLBase v3.49.1
```

```
Julia Version 1.11.8
Commit cf1da5e20e3 (2025-11-06 17:49 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin24.0.0)
  CPU: 12 × Apple M2 Max
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, apple-m2)
Threads: 1 default, 0 interactive, 1 GC (on 8 virtual cores)
```

I also checked `OrdinaryDiffEqCore` v4.14.1: `controllers.jl` there is byte-identical in this
function, so 4.14.1 through 4.14.3 are all affected.

## Where it was hit

`benchmarks/SimpleHandwrittenPDE` in SciMLBenchmarks.jl. Four files
(`kdv_fdm_wpd`, `kdv_spectral_wpd`, `nls_fdm_wpd`, `nls_spectral_wpd`) use
`AutoVern7(RadauIIA5(...))` for their reference solve. Since the 2026-05-04 v7 migration the
reference chunk throws this error, every later chunk then dies on an undefined `test_sol`, and
the files have produced no output at all. Worked around (by switching the reference solves to
`AutoVern7(Rodas5P(...))`, which is not FIRK) in
https://github.com/SciML/SciMLBenchmarks.jl/pull/1672 — the workaround should be reverted once
this is fixed.

There, the switch to the FIRK branch happens on its own during the reference solve; the
`stiffalgfirst = true` above is only to make a 10-line script deterministic. One caveat worth
recording: on the small stiff test problems I tried (ROBER, van der Pol) `AutoVern7(RadauIIA5())`
never selected the stiff branch at all within `maxiters`, while `AutoVern7(Rodas5P())` on the same
problem did switch. That may be a second, separate issue with stiffness detection for FIRK
members, but I have not chased it and am not claiming it here.

niy-ati:
## Summary

- `PredictiveControllerCache`'s `stepsize_controller!` reads `integrator.cache` directly to get `iter` (FIRK branch) or `integrator.cache.nlsolver` (non-FIRK branch). For a `CompositeAlgorithm`, `integrator.cache` is the `CompositeCache`, not the active branch's own cache — so `(; iter) = integrator.cache` throws `type CompositeCache has no field iter` the moment an `AutoSwitch` composite's stiff branch is a FIRK method (`RadauIIA3/5/9`).
- The `nlsolver` branch has the same latent bug for any non-`fac_default_gamma`, non-FIRK algorithm used as a composite member — it just isn't exercised by the reporter's repro because the non-FIRK stiff algorithms they tried (`FBDF`, `KenCarp4`) have `fac_default_gamma(alg) == true`, which skips this code path entirely, not because `integrator.cache.nlsolver` actually resolves correctly for a composite cache (there is no `getproperty` override anywhere in this codebase that would make that work).
- Fix: resolve the currently active sub-cache the same way `CompositeControllerCache`'s own `stepsize_controller!` dispatch (one call frame up) already selected `cache` and `alg` for this call — `integrator.cache.caches[integrator.cache.current]` when `integrator.cache isa CompositeCache`, else `integrator.cache` unchanged. Applied to both branches.

Fixes #4364.

## Test plan

- [x] New regression testset in `test/InterfaceI/composite_algorithm_test.jl`: `AutoVern7(RadauIIA5(autodiff = AutoFiniteDiff()); stiffalgfirst = true)` and `AutoTsit5(RadauIIA5(autodiff = AutoFiniteDiff()); stiffalgfirst = true)` on the issue's ROBER reproducer — both crashed on unpatched master, both `retcode == ReturnCode.Success` with this patch.
- [x] Verified locally against the exact reproducer in the issue body (all 4 cases, including the non-FIRK `AutoVern7(Rodas5P())` control): unpatched master crashes on the two FIRK-stiff-branch cases with `type CompositeCache has no field iter`; patched branch completes all four (`Success`, 57 / 73 / 54 steps respectively).

### DIFF
--- a/lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
@@ -1206,15 +1206,19 @@ end
             # not this branch's own cache (which is what `iter`/`nlsolver` live
             # on) -- resolve the currently active sub-cache the same way the
             # CompositeControllerCache dispatch above already selected `cache`
-            # and `alg` for this call.
-            alg_cache = integrator.cache isa CompositeCache ?
+            # and `alg` for this call. DefaultCache (DefaultODEAlgorithm) is
+            # deliberately not handled here: it holds cache1..cache6 rather
+            # than caches/current, but none of its members are FIRK or set
+            # fac_default_gamma(alg) == false, so this branch is unreachable
+            # for it today.
+            current_cache = integrator.cache isa CompositeCache ?
                 @inbounds(integrator.cache.caches[integrator.cache.current]) :
                 integrator.cache
             if isfirk(alg)
-                (; iter) = alg_cache
+                (; iter) = current_cache
                 (; maxiters) = alg
             else
-                (; iter, maxiters) = alg_cache.nlsolver
+                (; iter, maxiters) = current_cache.nlsolver
             end
             fac = min(gamma, (1 + 2 * maxiters) * gamma / (iter + 2 * maxiters))
         end
--- a/lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
+++ b/lib/OrdinaryDiffEqCore/src/integrators/controllers.jl
@@ -1202,11 +1202,19 @@ end
         if fac_default_gamma(alg)
             fac = gamma
         else
+            # For a CompositeAlgorithm, integrator.cache is the CompositeCache,
+            # not this branch's own cache (which is what `iter`/`nlsolver` live
+            # on) -- resolve the currently active sub-cache the same way the
+            # CompositeControllerCache dispatch above already selected `cache`
+            # and `alg` for this call.
+            alg_cache = integrator.cache isa CompositeCache ?
+                @inbounds(integrator.cache.caches[integrator.cache.current]) :
+                integrator.cache
             if isfirk(alg)
-                (; iter) = integrator.cache
+                (; iter) = alg_cache
                 (; maxiters) = alg
             else
-                (; iter, maxiters) = integrator.cache.nlsolver
+                (; iter, maxiters) = alg_cache.nlsolver
             end
             fac = min(gamma, (1 + 2 * maxiters) * gamma / (iter + 2 * maxiters))
         end
--- a/test/InterfaceI/composite_algorithm_test.jl
+++ b/test/InterfaceI/composite_algorithm_test.jl
@@ -133,7 +133,9 @@ sol = solve(prob_mm, DefaultODEAlgorithm(), callback = cb)
     prob_stiff = ODEProblem(rober_stiff!, [1.0, 0.0, 0.0], (0.0, 1.0e5))
 
     for alg in (
+            AutoVern7(RadauIIA3(autodiff = AutoFiniteDiff()); stiffalgfirst = true),
             AutoVern7(RadauIIA5(autodiff = AutoFiniteDiff()); stiffalgfirst = true),
+            AutoVern7(RadauIIA9(autodiff = AutoFiniteDiff()); stiffalgfirst = true),
             AutoTsit5(RadauIIA5(autodiff = AutoFiniteDiff()); stiffalgfirst = true),
         )
         sol = solve(prob_stiff, alg)
--- a/test/InterfaceI/composite_algorithm_test.jl
+++ b/test/InterfaceI/composite_algorithm_test.jl
@@ -2,6 +2,7 @@ using OrdinaryDiffEq, OrdinaryDiffEqCore, Test, LinearAlgebra
 import ODEProblemLibrary: prob_ode_linear, prob_ode_2Dlinear
 using DiffEqDevTools, ADTypes
 using OrdinaryDiffEqAdamsBashforthMoulton, OrdinaryDiffEqExplicitRK, OrdinaryDiffEqRosenbrock
+using OrdinaryDiffEqFIRK
 import OrdinaryDiffEqExplicitTableaus
 using OrdinaryDiffEqCore: CompositeAlgorithm
 
@@ -114,3 +115,28 @@ cb = DiscreteCallback(
     (u, t, integrator) -> true, (integrator) -> derivative_discontinuity!(integrator, true)
 )
 sol = solve(prob_mm, DefaultODEAlgorithm(), callback = cb)
+
+# https://github.com/SciML/OrdinaryDiffEq.jl/issues/4364
+# PredictiveControllerCache's `isfirk(alg)` branch destructured `integrator.cache`
+# directly for `iter`, which is the CompositeCache when the FIRK method is a
+# composite member (not the FIRK sub-cache the destructuring assumes) --
+# "type CompositeCache has no field iter". Every FIRK stiff member of an
+# AutoSwitch composite hit this once the switch reached the stiff branch.
+@testset "AutoSwitch composite with a FIRK stiff member (#4364)" begin
+    function rober_stiff!(du, u, p, t)
+        y1, y2, y3 = u
+        du[1] = -0.04 * y1 + 1.0e4 * y2 * y3
+        du[2] = 0.04 * y1 - 1.0e4 * y2 * y3 - 3.0e7 * y2^2
+        du[3] = 3.0e7 * y2^2
+        return nothing
+    end
+    prob_stiff = ODEProblem(rober_stiff!, [1.0, 0.0, 0.0], (0.0, 1.0e5))
+
+    for alg in (
+            AutoVern7(RadauIIA5(autodiff = AutoFiniteDiff()); stiffalgfirst = true),
+            AutoTsit5(RadauIIA5(autodiff = AutoFiniteDiff()); stiffalgfirst = true),
+        )
+        sol = solve(prob_stiff, alg)
+        @test sol.retcode == ReturnCode.Success
+    end
+end


singhharsh1708 (commented):
Reproduced on master, then re-ran with only the `lib/OrdinaryDiffEqCore` half of the patch applied:

```
                       master                                    patched
AutoVern7(RadauIIA5)   THREW "CompositeCache has no field iter"   Success, 57 steps
AutoTsit5(RadauIIA5)   THREW "CompositeCache has no field iter"   Success, 54 steps
AutoVern7(Rodas5P)     Success, 73 steps                          Success, 73 steps
```

Diagnosis looks right. The dispatch at `controllers.jl:1293` unwraps the controller cache and the algorithm but leaves `integrator` alone, so `integrator.cache` is still the `CompositeCache` inside. Indexing it with `integrator.cache.current` matches what line 1294 already does, and `@inbounds(cache.caches[cache.current])` matches the idiom in `composite_perform_step.jl`.

The "latent" claim on the `nlsolver` branch holds up. `fac_default_gamma` defaults to false, and the only `PredictiveController` users are FIRK plus the StabilizedRK family, which all set `fac_default_gamma = true` and short-circuit before those lines. FIRK is the only live path today.

The two `get_current_adaptive_order(alg, integrator.cache)` calls left untouched in the same function are fine: the generic method ignores `cache` and reads `alg_adaptive_order(alg)`, and `alg` arrives already unwrapped.

Two non-blocking notes:

1. The local `alg_cache` shadows `OrdinaryDiffEqCore.alg_cache`, which is declared public and referenced across ~112 files. No functional issue here, but `current_cache` or `sub_cache` avoids leaving a trap for the next edit.

2. `DefaultCache` is not a `CompositeCache` and has `cache1..cache6` rather than `caches`, so it falls through the `isa` check into the same failure with a different message. Not reachable today since `DefaultODEAlgorithm`'s members are Tsit5/Vern7/Rosenbrock23/Rodas5P/FBDF/KrylovFBDF, none FIRK. A one-line comment saying `DefaultCache` is deliberately out of scope would save the next reader that check.

singhharsh1708:
Checked IIA3 and IIA9 as well, both go from throwing to `Success`, so everything in #4364's reproducer is covered.

Worth adding those two to the test loop since the issue lists them. Also, the issue says the `nlsolver` branch is safe because `getproperty` resolves it, but there is no such override anywhere; it is just unreachable, since every non-FIRK `PredictiveController` user sets `fac_default_gamma = true`. Fixing it anyway is right.

Unrelated to this PR: `AdaptiveRadau` dies earlier on `integrator.alg.max_order` in FIRK's `initialize!`. Filed as #4367.

singhharsh1708:
Correction to my first comment: those `get_current_adaptive_order` calls are not safe for `AdaptiveRadau`. It defines `get_current_adaptive_order(alg::AdaptiveRadau, cache) = cache.num_stages` (`OrdinaryDiffEqFIRK/src/alg_utils.jl:23`), so `controllers.jl:1221` hands it the `CompositeCache` and it dies there instead.

Nothing in #4364's reproducer reaches it, so optional scope here. Sorry for the bad steer.

niy-ati:
Good catch — confirmed at `alg_utils.jl:22-23`, `get_current_adaptive_order(AdaptiveRadau, cache)` does read `cache.num_stages` and breaks the same way once `integrator.cache` is a `CompositeCache`.

One thing worth flagging: fixing it isn't quite reusing the local as-is — `current_cache` is scoped inside the `else` of `fac_default_gamma(alg)`, and both crash sites sit outside that block, so a direct swap would throw `UndefVarError` for the `fac_default_gamma == true` algorithms. It'd need hoisting the resolution above the `if/else`.

Since it's unreachable without #4367's `max_order` fix landing first, and doesn't touch #4364's reproducer, I'd rather keep this PR scoped to what's verified here and pick it up alongside #4367 instead. Thanks for tracing it through.

singhharsh1708:
Update looks good. Diffed the revisions with comments stripped, the code change is just the rename, and IIA3/IIA9 still pass.

One nit: the `DefaultCache` comment says none of its members set `fac_default_gamma(alg) == false`, but they all are false, that is the fallback. The reason it is unreachable is that none of them select `PredictiveController` at all.

singhharsh1708 (approved):
Verified independently: reproduces on master, fixed by this patch, and all four methods #4364 reports (RadauIIA3/5/9 under both wrappers) go from throwing to `Success` with the non-FIRK control unchanged. The update is a pure rename plus the two extra test cases, confirmed by diffing the revisions with comments stripped.

The remaining `get_current_adaptive_order` point is optional scope and cannot affect this issue's reproducer, and the `DefaultCache` comment wording is a nit. Neither blocks.

LGTM.

singhharsh1708 (approved):

singhharsh1708:
Opened #4420 with this same change rebased onto current master, since this has been sitting for four days and #4364 still reproduces. The fix is yours, I only added the `OrdinaryDiffEqCore` version bump and dropped the comment blocks. Happy to close mine if you would rather rebase this one instead.

For what it is worth the 42 red lanes here are not your patch. The ones I opened are Aqua `persistent_tasks` failures, and a two line change in `stepsize_controller!` cannot break `OrdinaryDiffEqTaylorSeries [QA]`.

PR #4366, opened 2026-08-24
Fix SBDF order ramp, history-shift timing, and dt-change reset

### LINKED ISSUE #4329: SBDF solvers do not achieve expected convergence order
From Claude Opus 5 🤖 :

## Summary

`SBDF2`, `SBDF3` and `SBDF4` do not reach their nominal convergence order. Three
separate defects cause this:

1. The order ramp advances one step too fast, so a `k`-point formula runs before `k` points exist.
2. The history shift is conditional on `cnt`, so `k₂` and `k₃` still hold their initial values when a higher-order formula first reads them.
3. The fixed BDF coefficients are applied after a change of step size, which injects an error of the size of one old step.

Defect 3 is the most consequential: it fires whenever the final step is clipped to
land on `tspan[2]`, which happens for most step sizes, and it silently reduces
SBDF2/3/4 to first order.

`IMEXEuler()` (which is `SBDF(1)`, see `src/algorithms.jl:602`) is correct and no
defect affects it.

There is no convergence test for SBDF in the test suite, which is why this is not caught.

## Environment

- Julia 1.12.7
- OrdinaryDiffEqBDF v2.4.3
- OrdinaryDiffEqNonlinearSolve, OrdinaryDiffEqRosenbrock (for reference solutions)

---

## Defect 1: the order ramp is one step too fast

`src/bdf_perform_step.jl:211` (out-of-place) and `:294` (in-place):

```julia
cnt = cache.cnt = min(alg.order, integrator.iter + 1)
integrator.iter == 1 && !integrator.derivative_discontinuity && (cnt = cache.cnt = 1)
```

Only the first step is forced down to 1. At step 2, `cnt == 3` for `SBDF3`, but only
two past points exist (`u_1` and `u_0`). The three-point formula therefore reads
`uprev3` and `k₂`, which no step has written.

A step trace for `SBDF3` on `u' = -u`, `u(0) = 1`, `dt = 0.1`:

```
after step 1: iter=1 cnt=1  u=0.90909091  exact=0.90483742
after step 2: iter=2 cnt=3  u=0.85893417  exact=0.81873075    <- 4% error
after step 3: iter=3 cnt=3  u=0.79991352  exact=0.74081822
```

The error at step 2 is about 4%. The local truncation error at this step size is about `1e-3`.

## Defect 2: the history shift is gated on `cnt`

Same function, at the end of the step:

```julia
cnt == 4 && (cache.uprev4 = uprev3; cache.k₃ = k₂)
cnt >= 3 && (cache.uprev3 = uprev2; cache.k₂ = k₁)
           (cache.uprev2 = uprev;  cache.k₁ = du₂)
```

With a corrected ramp (defect 1), step 3 is the first step that reads `uprev3` and
`k₂`, and no earlier step writes them, because those steps ran at `cnt <= 2`. They
still hold the values set at cache construction. Dumping the cache after `init` for
`SBDF3` on the problem above, where `f2 ≡ 0`:

```
uprev=1.0  uprev2=1.0  uprev3=1.0  uprev4=1.0
k₁=1.0  k₂=1.0  k₃=1.0  du₁=-1.0  du₂=0.0
```

`k₁`, `k₂` and `k₃` hold past values of `f2`, which is identically zero here
(`du₂ = 0.0` is correct). The value `1.0` in all three slots is wrong.

## Defect 3: fixed coefficients applied after a change of step size

The BDF part of `tmp` is not scaled by `dt`:

```julia
tmp = γ * (2 * uprev - 1//2 * uprev2 + dt * (2 * du₂ - k₁))    # cnt == 2
```

When the final step is clipped to land exactly on `tspan[2]`, the formula runs with
the new small `h` while its history is still spaced at the old `h`. Substituting
`uprev2 ≈ uprev - h_old * u'` and letting `h → 0`:

```
u_new ≈ (2/3) * (2*u_n - (1/2)*(u_n - h_old*u')) = u_n + (1/3) * h_old * u'
```

The solution advances by one third of a full old step while time advances by
almost nothing.

`SBDF(1)` is immune because `tmp = uprev + dt*du₂` correctly collapses to `uprev`
as `h → 0`. There is no unscaled BDF combination.

This defect fires on any substantial change of step size: the clipped final step,
`tstops`, or callbacks. A near-full-size final step is harmless.

---

## Reproducer 1 (defects 1 and 2)

```julia
using OrdinaryDiffEqBDF, OrdinaryDiffEqNonlinearSolve

exact(t) = exp(-t)
prob = SplitODEProblem((u, p, t) -> -u, (u, p, t) -> 0.0, 1.0, (0.0, 1.0))

for alg in (SBDF2(), SBDF3(), SBDF4())
    errs = [abs(solve(prob, alg; dt, adaptive = false).u[end] - exact(1.0))
            for dt in (1e-2, 1e-3, 1e-4)]
    @show errs
end
```

Note that `f2 = 0` here, so `SBDF(k)` reduces to plain `BDF(k)`.

**Expected:** order 2, 3, 4.
**Actual** (order taken between the first two step sizes):

| method | errors at dt = 1e-2, 1e-3, 1e-4 | order |
|---|---|---|
| `SBDF2` | 1.54e-5, 1.53e-7, 1.23e-5 | 2.0 |
| `SBDF3` | 2.53e-3, 2.46e-4, 7.81e-6 | **1.01** |
| `SBDF4` | 2.32e-5, 2.75e-6, 1.89e-5 | **0.93** |

The third column is also polluted by defect 3.

Cache state for defect 2:

```julia
integ = init(prob, SBDF3(); dt = 0.1)
@show integ.cache.uprev3, integ.cache.k₂, integ.cache.du₂
# (1.0, 1.0, 0.0)   <- k₂ must be a past value of f2, which is 0
```

## Reproducer 2 (defect 3)

Two nearly equal step sizes give errors that differ by a factor of about 4200:

```julia
using OrdinaryDiffEqBDF, OrdinaryDiffEqNonlinearSolve

prob = SplitODEProblem((u, p, t) -> -u, (u, p, t) -> 0.0, 1.0, (0.0, 2.0))

for dt in (1e-3, 2.0^-10)
    s = solve(prob, SBDF2(); dt, adaptive = false)
    h = diff(s.t)
    println("dt=$dt nsteps=$(length(h)) min_step=$(minimum(h)) err=$(abs(s.u[end]-exp(-2.0)))")
end
```

**Actual:**

```
dt=0.001            nsteps=2001  min_step=1.0946799022804043e-13  err=4.512306532589161e-5
dt=0.0009765625     nsteps=2048  min_step=0.0009765625            err=1.0755522417138508e-8
```

The first run takes a final step of `1.09e-13`. The second step size is exact in
binary, so no micro step occurs, and the error is 4200x smaller for a 2% larger step.

The polluted error matches the predicted magnitude. On a second test problem the
derivative at the end time is `0.015469`, and `(1/3) * dt * u'` predicts:

| dt | predicted | measured |
|---|---|---|
| 1e-3 | 5.156e-6 | 5.111e-6 |
| 5e-4 | 2.578e-6 | 2.568e-6 |
| 1e-4 | 5.156e-7 | 5.157e-7 |

Note that `SBDF` is fixed-step: passing `reltol`/`abstol` or `adaptive = false`
gives bit-identical results, and `length(sol.t)` equals the requested step count.
So none of the above is a tolerance artifact.

---

## Proposed fixes

```diff
- cnt = cache.cnt = min(alg.order, integrator.iter + 1)
+ cnt = cache.cnt = min(alg.order, integrator.iter)
```

```diff
- cnt == 4 && (cache.uprev4 = uprev3; cache.k₃ = k₂)
- cnt >= 3 && (cache.uprev3 = uprev2; cache.k₂ = k₁)
-            (cache.uprev2 = uprev;  cache.k₁ = du₂)
+ cache.uprev4 = uprev3; cache.k₃ = k₂
+ cache.uprev3 = uprev2; cache.k₂ = k₁
+ cache.uprev2 = uprev;  cache.k₁ = du₂
```

The in-place version needs temporaries, because `.=` overwrites buffers that the
later lines still read.

For defect 3, the minimal correct fix is to restart the ramp (`cnt = 1`) on any
change of `dt`. A single order-1 step of size `1e-13` is harmless. A full fix for
variable steps needs variable-coefficient BDF formulas plus variable-step
extrapolation weights, which `FBDF` and `QNDF` already have.

## Results with the fixes applied

Patched at runtime with `Core.eval` and re-measured.

Scalar test, after fixes 1 and 2:

| method | before | after (`f2 = 0`) | after (half/half split) |
|---|---|---|---|
| `SBDF3` | 1.01 | 2.0 | 2.99 |
| `SBDF4` | 0.93 | 2.0 | 3.02 |

The remaining cap is expected multistep start-up order reduction, not a defect: the
first step runs at order 1, its local error is O(dt²), and that limits the global
order to 2. On the half/half split the first step is coincidentally the trapezoidal
rule (O(dt³)), so the cap is 3. Seeding the first `k-1` steps with exact values
confirms this — `SBDF3` then gives orders 2.96, 2.98, 2.99 and `SBDF4` gives 3.94,
3.97, 3.99. A production start-up needs a one-step IMEX method of order `k` or higher.

Fix 3 on a semi-explicit DAE (row-wise split, `M = Diagonal([1.0, 0.0])`), at step
sizes 2e-3 down to 1e-4:

| | errors | order |
|---|---|---|
| without fix 3 | -2.246e-7, **+5.111e-6**, +2.568e-6, +1.030e-6, +5.157e-7 | 1 |
| with fix 3 | -1.632e-7, -5.606e-8, -1.401e-8, -2.240e-9, -5.600e-10 | 2 |

## Missing test coverage

`test/bdf_convergence_tests.jl` covers `ABDF2`, `QBDF1`, `QBDF2`, `QNDF1`, `QNDF2`,
`QNDF`, `QBDF`, `FBDF` and `MEBDF2`. `SBDF` appears only in `test/qa/jet.jl` and
`test/qa/allocation_tests.jl`, which call `init` and `step!` and never measure
accuracy. Reproducer 1 would make a suitable convergence test.

## Related observations

Two points that are not defects in SBDF, but that came up while measuring the fixes.

**The low-order first steps cap the global order.** A k-step BDF formula needs k
past points, but only the initial condition exists at the start, so the first `k-1`
steps come from the lower-order members of the family (`cnt` ramps 1, 2, 3). The
first step runs at order 1 and commits a local error of O(dt^2) once. That term
survives in the accumulated global error, so the total error cannot fall below
O(dt^2) for any `k`, and the whole run measures order 2.

This ceiling costs more than anything else measured here. With all three defects
fixed, but with the order-1 first step still in place, `SBDF2`, `SBDF3` and `SBDF4`
all measure order 2 on a 26-state semi-explicit DAE. They need 513, 257 and 257
steps to reach an error of `1e-6` (7.04 ms, 3.26 ms, 3.28 ms). When the first
`k-1` steps are seeded from a trusted reference, `SBDF3` reaches order 3 (129 steps,
1.68 ms) and `SBDF4` needs 65 steps (0.84 ms). `FBDF` needs 53 steps (0.36 ms) at
the same accuracy. A real start-up procedure is therefore worth about 8x on this
problem.

**The Jacobian is computed at full system size.** For a split problem the Newton
differentiates only the implicit part, because `nlsolve_f` returns `f.f1`. That
Jacobian can be far smaller in content than the full one: in a row-wise
differential/algebraic split it has nonzeros only in the algebraic rows, which was
1 row out of 26 in the model above. None of that is exploited. The AD pass
differentiates `f1` across all n components, `jacobian2W!` assembles a dense n x n W,
and the factorization is O(n^3) - the 26x26 `lu!` alone costs 3.13 us of a 7.6 us
step, and it grows with the model while the algebraic block does not.

If the implicit solve were reduced to the algebraic subsystem, the Jacobian would be
k x k for k algebraic variables and the linear algebra O(k^3), with the differential
rows obtained by explicit substitution. This costs nothing in accuracy, because the
reduced solve is exact for the block.

ModelingToolkit already builds this object. With `nlstep = true` it produces an
`ODENLStepData` whose inner `NonlinearProblem` covers only the torn variables (it
reports `nunknowns = 1` for a two-state feedback loop), optionally decomposed as an
`SCCNonlinearProblem`, with an analytic Jacobian. `update_W!` has a dedicated branch
for it (`nlstep_data !== nothing`). But `nlstep_data` is a field on `ODEFunction`,
and `nlsolve_f` hands the solver `f.f1` for split problems, so a `SplitFunction`
never carries it. Connecting those two would make the implicit cost proportional to
the algebraic block rather than to the whole model, which is the point of splitting
in the first place.

niy-ati:
## Summary

SBDF (SBDF2/3/4) is a fixed-coefficient IMEX multistep method and requires a uniform step-size history. Three issues combined to make it silently run below its nominal convergence order:

- The bootstrap counter allowed the higher-order formula to be used one step before enough trusted history existed.
- The buffer history shift was gated on the bootstrap counter instead of the algorithm order, so high-order runs could skip a needed shift during ramp-up.
- No reset occurred when dt changed, so the fixed-coefficient formula was silently applied across a non-uniform step history (e.g. on a clipped final step), inflating the error by orders of magnitude.

Fixes #4329.

## Test plan

- Added an `SBDF` convergence testset measuring empirical order for pure-BDF and genuine IMEX splits, out-of-place and in-place, across SBDF2/SBDF3/SBDF4.
- Verified against both reproducers from the issue.
- Checked the ARK variant, a `tstops`-forced mid-integration dt change, `reinit!` + re-solve consistency, and in-place order-4 correctness.
- Confirmed no allocation or type-stability change relative to the unpatched baseline.

### DIFF
--- a/lib/OrdinaryDiffEqBDF/src/bdf_caches.jl
+++ b/lib/OrdinaryDiffEqBDF/src/bdf_caches.jl
@@ -89,7 +89,7 @@ end
 
 # SBDF
 
-@cache mutable struct SBDFConstantCache{rateType, N, uType} <: OrdinaryDiffEqConstantCache
+@cache mutable struct SBDFConstantCache{rateType, N, uType, dtType} <: OrdinaryDiffEqConstantCache
     cnt::Int
     ark::Bool
     k2::rateType
@@ -102,9 +102,10 @@ end
     k₃::rateType
     du₁::rateType
     du₂::rateType
+    dtprev::dtType
 end
 
-@cache mutable struct SBDFCache{uType, rateType, N} <: BDFMutableCache
+@cache mutable struct SBDFCache{uType, rateType, N, dtType} <: BDFMutableCache
     cnt::Int
     ark::Bool
     u::uType
@@ -119,6 +120,7 @@ end
     k₃::rateType
     du₁::rateType
     du₂::rateType
+    dtprev::dtType
 end
 
 function alg_cache(
@@ -143,10 +145,11 @@ function alg_cache(
     uprev2 = u
     uprev3 = u
     uprev4 = u
+    dtprev = zero(dt)
 
     return SBDFConstantCache(
         1, alg.ark, k2, nlsolver, uprev2, uprev3, uprev4, k₁, k₂, k₃, du₁,
-        du₂
+        du₂, dtprev
     )
 end
 
@@ -174,10 +177,11 @@ function alg_cache(
     uprev2 = zero(u)
     uprev3 = order >= 3 ? zero(u) : uprev2
     uprev4 = order == 4 ? zero(u) : uprev2
+    dtprev = zero(dt)
 
     return SBDFCache(
         1, alg.ark, u, uprev, fsalfirst, nlsolver, uprev2, uprev3, uprev4, k₁, k₂, k₃,
-        du₁, du₂
+        du₁, du₂, dtprev
     )
 end
 
--- a/lib/OrdinaryDiffEqBDF/src/bdf_perform_step.jl
+++ b/lib/OrdinaryDiffEqBDF/src/bdf_perform_step.jl
@@ -207,8 +207,17 @@ function perform_step!(integrator, cache::SBDFConstantCache, repeat_step = false
     alg = unwrap_alg(integrator, true)
     (; uprev2, uprev3, uprev4, du₁, du₂, k₁, k₂, k₃, nlsolver) = cache
     (; f1, f2) = integrator.f
-    cnt = cache.cnt = min(alg.order, integrator.iter + 1)
+    # cnt tracks how many past (uprev_i, k_i) points the fixed-coefficient
+    # formulas below can validly read, which is the number of completed
+    # steps -- not one more than that (that was defect 1: it ran a k-point
+    # formula with only k-1 points on record). It must also drop back to 1
+    # whenever dt changes: the BDF coefficients below are derived assuming
+    # the whole history is spaced at the current dt, so history recorded at
+    # a different dt (e.g. a step clipped to land on tspan[2]) makes the
+    # formula wrong, not just less accurate (see #4329).
+    cnt = cache.cnt = min(alg.order, integrator.iter)
     integrator.iter == 1 && !integrator.derivative_discontinuity && (cnt = cache.cnt = 1)
+    dt != cache.dtprev && (cnt = cache.cnt = 1)
     nlsolver.γ = γ = inv(γₖ[cnt])
     if cache.ark
         # Additive Runge-Kutta Method
@@ -247,11 +256,19 @@ function perform_step!(integrator, cache::SBDFConstantCache, repeat_step = false
     nlsolvefail(nlsolver) && return
     u = nlsolver.tmp + γ * z
 
-    cnt == 4 && (
+    # Shift the history every step (not gated on cnt, which only says how
+    # much of it is *usable yet* -- defect 2 was gating the shift itself on
+    # cnt, so the first step that could validly read a slot found it still
+    # holding its cache-construction value). Gated on alg.order instead,
+    # which is fixed per algorithm instance: for order < 4 (< 3), uprev4/k₃
+    # (uprev3/k₂) alias uprev2/k₁ as a memory optimization in the mutable
+    # cache's constructor, and writing through an aliased slot here would
+    # corrupt uprev2/k₁ before the unaliased lines below read them.
+    alg.order == 4 && (
         cache.uprev4 = uprev3;
         cache.k₃ = k₂
     )
-    cnt >= 3 && (
+    alg.order >= 3 && (
         cache.uprev3 = uprev2;
         cache.k₂ = k₁
     )
@@ -259,6 +276,7 @@ function perform_step!(integrator, cache::SBDFConstantCache, repeat_step = false
         cache.uprev2 = uprev;
         cache.k₁ = du₂
     )
+    cache.dtprev = dt
     cache.du₁ = f1(u, p, t + dt)
     cache.du₂ = f2(u, p, t + dt)
     OrdinaryDiffEqCore.increment_nf!(integrator.stats, 1)
@@ -290,8 +308,12 @@ function perform_step!(integrator, cache::SBDFCache, repeat_step = false)
     (; uprev2, uprev3, uprev4, k₁, k₂, k₃, du₁, du₂, nlsolver) = cache
     (; tmp, z) = nlsolver
     (; f1, f2) = integrator.f
-    cnt = cache.cnt = min(alg.order, integrator.iter + 1)
+    # See the matching branch in the constant-cache method for why cnt is
+    # min(order, iter) rather than iter + 1, and why it also resets on a
+    # dt change (#4329).
+    cnt = cache.cnt = min(alg.order, integrator.iter)
     integrator.iter == 1 && !integrator.derivative_discontinuity && (cnt = cache.cnt = 1)
+    dt != cache.dtprev && (cnt = cache.cnt = 1)
     nlsolver.γ = γ = inv(γₖ[cnt])
     # Explicit part
     if cache.ark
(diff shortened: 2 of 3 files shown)


singhharsh1708:
Checked the numbers on current master: SBDF3 and SBDF4 measure order ~1.0, and with this they go to 2.0 on the pure-BDF split and 2.99/2.98 on an IMEX one. Real improvement, and the `alg.order` gating is safe under the order<4/order<3 buffer aliasing.

SBDF4 still never reaches 4 though, and the test encodes that as expected rather than calling it out. Since #4329 is titled "do not achieve expected convergence order", worth a line in the body so `Fixes #4329` does not overclaim.

Also missing a version bump on `lib/OrdinaryDiffEqBDF/Project.toml`.

ChrisRackauckas:
Thanks! Yeah the SBDFs have had something odd for awhile, and it's good to finally have had that figured out.

PR #4373, opened 2026-08-25
Diagnostic Output Truncation

Shreyas-Ekanathan:
## Checklist

## Additional context

Fix OOM issues on large problems


lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
```diff
@@ -793,8 +793,7 @@ nonfinite_indices(u) = isfinite(u) ? Int[] : Int[1]
 
 # %g formats any real, complex values have to show themselves
 format_value(v::Real) = @sprintf("%.4g", v)
-format_value(v::Number) = string(v)
-
+format_value(v) = DiffEqBase.truncate_str(v)
```
ChrisRackauckas, line 796: needs a lower bound in the version.

Also, we should probably tell people how to write to a file for the rest of it or something?

lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
```diff
@@ -886,7 +885,7 @@ function jacobian_analysis!(msgs::Vector{String}, integrator, sym_eqs, sym_vars)
     if sym_eqs !== nothing
         for row in first(singularity_rows, 10)
             if row <= length(sym_eqs)
-                push!(msgs, "  row $row corresponds to equation: $(sym_eqs[row])") #trace rows back to symbolic eqs
+                push!(msgs, "  row $row corresponds to equation: $(DiffEqBase.truncate_str(sym_eqs[row]))") #trace rows back to symbolic eqs
```
ChrisRackauckas: should probably only truncate in the default "Standard" mode, but the enhanced logging should probably turn the whole print on.

Shreyas-Ekanathan: I mean i think that truncating equations is generally just fine since the user defines the equations? so they should have some knowledge of what is in their system

lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
Shreyas-Ekanathan: all of the debugging output will be there, we just truncate the large system stuff, whcih again should be fine since they should know their system fairly well, its just for convenience that we reiterate it for them

lib/OrdinaryDiffEqCore/src/integrators/integrator_utils.jl
ChrisRackauckas: Yes, so shouldn't they have a toggle for if they want to do it or not?

PR #4400, opened 2026-08-27
Let reinit! keep the step-size controller history

### LINKED ISSUE #4371: reinit_controller! of PIController (errold = qoldinit = 1e-4) causes a period-2 step-size oscillation
Target: OrdinaryDiffEqCore (SciML/OrdinaryDiffEq.jl). Version: OrdinaryDiffEqCore 4.13.0, Julia 1.12.6.

## Summary

`reinit_controller!(integrator, ::PIControllerCache)` sets `errold = qoldinit = 1e-4`. The PI law

    q = EEst^β1 / errold^β2 / γ,     dt_new = dt / q

with the order-2 defaults `β1 = 7/(10·2) = 0.35`, `β2 = 2/(5·2) = 0.2`, `γ = 0.9` then divides the
first accepted step's proposal by `EEst^0.35 / 1e-4^0.2 / 0.9 = EEst^0.35 / 0.142`, i.e. shrinks
it ~5x regardless of the error (clamped at `1/qmin`). The next step has a tiny `EEst`, grows
~5x (clamped at `qmax`), and the alternation decays only slowly through the `errold` memory. The
factor `qoldinit^β2` is 0.16 at order 2 but 0.69 for a DOPRI5-like `β2 = 0.04`, so low-order
methods are hit hardest.

Any code that restarts a PI-controlled integrator at a state jump — operator splitting,
multirate schemes, event handling that calls `reinit_controller!` — pays this on every restart;
for an implicit method every alternation is also a `W` refactorisation (measured on an
SDIRK2 sub-integrator restarted every sub-interval: dt changes by more than 20% on 96% of all
steps, W rebuilt on 96% of steps, 80% of the wall time).

## MWE

```julia
using OrdinaryDiffEqSDIRK, OrdinaryDiffEqCore, SciMLOperators, SparseArrays

N = 400
A = spdiagm(-1 => ones(N - 1), 0 => -2ones(N), 1 => ones(N - 1)) * N^2
u0 = [exp(-100 * (x - 0.5)^2) for x in range(0, 1, length = N)]
prob = ODEProblem(ODEFunction(MatrixOperator(A)), u0, (0.0, 0.5))
integ = init(prob, SDIRK2(); dt = 1e-2, abstol = 1e-6, reltol = 1e-4)
for k in 1:14
    step!(integ)
    println((integ.t - integ.tprev, OrdinaryDiffEqCore.get_EEst(integ)))
    k == 6 && OrdinaryDiffEqCore.reinit_controller!(integ, integ.controller_cache)
end
```

Output (dt, EEst per accepted step; `|` marks the reinit after step 6):

```
dt:   0.00171 0.000343 0.0014 0.000778 0.00153 0.00153 | 0.00209 0.000542 0.00247 0.0012 0.00285 0.00202 0.00331 0.00331
EEst: 0.92    0.013    0.33   0.057    0.19    0.12    | 0.18    0.0037   0.24    0.027  0.25    0.082   0.23    0.17
```

The same pattern appears at the start of every solve (steps 1–4), which is the intended
"first step" use of `qoldinit`; after a reinit, however, the controller already has a
converged step to continue from.

## Expected behaviour

`reinit_controller!` should restore a neutral history (`errold = 1`, or the last accepted
`EEst`) and reserve `qoldinit` for the genuine first step of a solve — or expose the choice.

singhharsh1708:
`reinit!` gains a `reinit_controller = true` keyword alongside `reinit_dae` and friends, so a mid-solve restart can keep the PI controller's converged history; the default is unchanged, `set_t!` opts out, and `DelayDiffEq` gets the same keyword. It lives on `reinit!` rather than `reinit_controller!` because the latter is public and the documented extension point for algorithm authors, so a keyword there would break downstream methods.

`reinit_controller!` reseeds `errold = qoldinit = 1e-4`, right at `t0` (Hairer's FACOLD) and wrong mid solve: on the issue's SDIRK2 problem a settled `dt` of 1.257538e-02 rings at ratios 0.20, 4.90, 0.41, 2.42 afterwards, and opting out reproduces the un-reinit'd sequence to 2.3e-6.

`reinit!` still sets `success_iter = 0`, so the step after an opt-out restart is treated as a first step for the growth cap; measured in isolation that is a one-step lag, not the oscillation, and preserving it would also flip SDIRK's first-step predictor branches, so it is left for a separate change.

`reinit_test.jl` 31/31, `disco_tests.jl` 28/28, `controllers.jl` 30/30 with the new testset.

Fixes #4371 for the mid-solve case; the first-step swing is untouched.


termi-official:
Not sure this actually fixes the root cause. Ofc this fixes the swing after reinit, but the initial swing is still present -- so the current behavior on master is the consistent one for PI.

Given your reinit just sets qold = 1, why should we keep the qoldinit in first (or not simply default it to 1)?

singhharsh1708:
Switched to defaulting `qoldinit = 1`, which covers both swings; benchmarked 1e-4/1e-2/0.5/1 work-precision normalised at 1.000/0.991/0.960/0.957, details in the description.

singhharsh1708:
Tried your suggestion and it does not hold up, so I have reverted to the reinit-only change.

Defaulting `qoldinit = 1` makes `AN5` return a wrong answer:

```
                        steps   result     error
master (1e-4)           10      1.372800   6.8e-7
qoldinit = 1            28      2.327861   9.6e-1
```

Exact is `1.3728`. More steps and a worse answer, so it is not a speed/accuracy trade. `ODEProblemLibrary.prob_ode_linear` at `reltol = 1e-6`, `nordsieck_tests.jl` catches it.

Across CI it took out 29 lanes that a comparable PR of mine does not touch: StochasticDiffEq, DelayDiffEq, Rosenbrock, Nordsieck, plus six root groups and four downstream packages. The ones I read are accuracy-at-requested-tolerance assertions, e.g. DelayDiffEq `l∞ 6.8e-5 < 3.0e-5` and Rosenbrock `err_interp 0.0115 < 0.01`.

My earlier work-precision table was the wrong metric for this. Normalising to matched accuracy hides the fact that a user asking for `1e-8` gets a larger error, and it never touched a variable-order method, which is where `AN5` breaks.

So the initial swing stays for now. Reverting does not fix it, and you are right that it leaves PI asymmetric with PID and Predictive, but the neutral-everywhere version is not safe as it stands.

termi-official:
> Defaulting qoldinit = 1 makes AN5 return a wrong answer.

Maybe I am missing something, but if we `reinit!` at exactly this point in this PR, shouldn't we run into the same problem, as we send the previous error estimate to 1 -- which should surface as the same problem?

ChrisRackauckas:
Why would `qoldinit` be 1? This is very odd.

singhharsh1708:
Both right, and I have replaced that commit. It no longer sets `errold` to 1 or touches `qoldinit` at all. It just adds `reinit_controller = false` to `reinit!` so a mid-solve restart can keep the history it already has; the default path is unchanged, and `qoldinit = 1e-4` stays as the FACOLD seed at `t0`.

@termi-official yes, exactly that. Sending `errold` to 1 hits the same problem, which is what `reinit_test.jl` caught: 7 of 8 trajectory-identity assertions failed, and it also had more sign alternations than master, 6 vs 5. Leaving `errold` alone instead reproduces the un-reinit'd `dt` sequence to 2.3e-6.

The initial swing is still there and PI is still asymmetric with PID and Predictive. I did not find a safe way to remove it, so this PR is only the mid-solve half.

Issue #4370, opened 2026-08-24
new_W_dt_cutoff not respected for linear problems


termi-official:
Target: OrdinaryDiffEqDifferentiation / OrdinaryDiffEqNonlinearSolve (SciML/OrdinaryDiffEq.jl).

Versions: OrdinaryDiffEqCore 4.13.0, OrdinaryDiffEqDifferentiation 3.7.0,
OrdinaryDiffEqNonlinearSolve 2.7.0, OrdinaryDiffEqSDIRK 2.8.2, Julia 1.12.6.

## Summary

For problems recognised by `islinearfunction` (e.g. `ODEFunction(MatrixOperator(A))`),
`do_newJW` (`OrdinaryDiffEqDifferentiation/src/derivative_utils.jl`, the `islin` branch) keeps
the current `W = M - γ dt J` factorisation while

    |1/(γ dt) / 1/(γ dt_W) − 1| ≤ new_W_dt_cutoff        (default 1//5)

and lets Newton run against the stale `W`. Newton's termination `η · ndz < κ`
(`OrdinaryDiffEqNonlinearSolve/src/nlsolve.jl`) uses an `η` carried over from the previous
solve. On a linear problem with a fresh `W` one iteration is exact, so `η` stays tiny and the
first iteration against a stale `W` is accepted as converged as well: measured 2.00 Newton
iterations per 2-stage SDIRK2 step over 13 523 consecutive steps whose `W` was up to 20% stale.
With larger drift Newton iterates a few times but still terminates early. The stage error this
leaves behind is not seen by the embedded error estimate, so adaptive solves neither reject nor
shrink the step.

## MWE

Fixed steps alternating between `dt` and `1.15 dt` (inside the cutoff), 1D heat equation with
a `MatrixOperator` right-hand side, compared to `exp(tA) u0`:

```julia
using OrdinaryDiffEqSDIRK, OrdinaryDiffEqNonlinearSolve, SciMLOperators, LinearAlgebra, SparseArrays

N = 400
A = spdiagm(-1 => ones(N - 1), 0 => -2ones(N), 1 => ones(N - 1)) * N^2
u0 = [exp(-100 * (x - 0.5)^2) for x in range(0, 1, length = N)]
prob = ODEProblem(ODEFunction(MatrixOperator(A)), u0, (0.0, 2.0))

function run(cutoff; dts = (1.0e-2, 1.15e-2), n = 100, kw...)
[...]

Issue #3933, opened 2026-07-17
Silently wrong results: implicit solvers + SciMLOperator Jacobian (e.g. SplitODEProblem with MatrixOperator) + any concrete-A linsolve


ChrisRackauckas-Claude:
## Summary

Any implicit/IMEX solver that goes through `NLNewton` with a `WOperator` whose Jacobian is an
`AbstractSciMLOperator` (e.g. a `SplitODEProblem` with a `MatrixOperator` implicit part, or a plain
`ODEProblem(ODEFunction(MatrixOperator(A)))`) **silently returns garbage with `retcode: Success`**
when the linear solver needs a concrete matrix (`LUFactorization()`, `KLUFactorization()`, ...).

Mechanism: `W._concrete_form` is built once at `init` with `gamma = 0` (dt not yet chosen), giving a
matrix whose diagonal is `-Inf` (`J - M/0`). It is **never updated afterwards** — `calc_W!` skips
`jacobian2W!` whenever `W.J isa AbstractSciMLOperator`, and `convert(::Type{AbstractMatrix},
W::WOperator{true})` (SciMLOperators) returns the stale `_concrete_form` without recomputing.
LinearSolve factorizes the `Inf`-diagonal matrix, so every Newton correction is `dz = W \ b ≈ 0`
(Newton "converges" instantly at the predictor) and the ESDIRK smoothed error estimate
`est = W \ tmp ≈ 0`, so **every step is accepted and dt grows to the maximum**. The integration
finishes in ~2 steps with an unbounded error and `Success`.

The Krylov path (`KrylovJL_GMRES()` or the operator-default) is unaffected because it uses the lazy
`WOperator` `mul!`, which reads the current `gamma`.

## Minimal reproduction (released packages, Julia 1.12)

```julia
using OrdinaryDiffEqSDIRK, LinearSolve, SciMLOperators, LinearAlgebra

A = [-2.0 1.0; 1.0 -2.0]      # implicit part
B = [0.0 0.5; -0.5 0.0]       # explicit part
u0 = [1.0, 0.5]
f2!(du, u, p, t) = mul!(du, B, u)
prob = SplitODEProblem(MatrixOperator(A), f2!, u0, (0.0, 1.0))
uref = exp((A + B) * 1.0) * u0

sol = solve(prob, KenCarp4(linsolve = LUFactorization()); abstol = 1e-8, reltol = 1e-8)
sol.retcode                                # Success
[...]

PR #4401, opened 2026-08-27
Rebuild the linear W whenever gamma dt changes on fixed steps

singhharsh1708:
On fixed steps the linear-operator path kept a stale `W = M - γ dt J` whenever `dt` drifted by less than `new_W_dt_cutoff`, so `adaptive = false` solves reused `W` across step-size changes; the cutoff is now zero unless the solve is adaptive. Comparing `γ dt` rather than forcing a rebuild halves the factorisations, because the two stages of a step share one `W`, and `njacs` stays 0 on the linear path.
On the issue's problem with `dt` alternating against `1.15 dt`, exact `W` converges at second order (0.0159, 0.00443, 0.00117 over three halvings) while the default cutoff stalled near 6.5e-3; after the change both cutoffs give identical results.
The test steps a `MatrixOperator` problem with alternating fixed steps, in `stale_w_linear_operator_tests.jl`, which already covers the sibling stale-`W` bug #3933. The adaptive half of #4370, where a carried-over `η` lets Newton stop early against a stale `W`, is separate.
Fixes the fixed-step half of #4370.
With `LHLFactorization` the W stays split and LHL absorbs each new γdt from `W.gamma` on every solve, but inside the cutoff `W_γdt` was left at the old value, so the Newton step's `dz` rescale in `compute_step!` corrected for a staleness that was not there; the cutoff is now zero for a split W on adaptive steps as well, which keeps `W_γdt` equal to γdt and drops the rescale. The new test checks cutoff 0.2 and 0 give identical results with LHL, adaptive and fixed-step; on master the end states differ.

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/Project.toml
+++ b/lib/OrdinaryDiffEqDifferentiation/Project.toml
@@ -1,5 +1,5 @@
 name = "OrdinaryDiffEqDifferentiation"
-version = "3.11.1"
+version = "3.11.2"
 uuid = "4302a76b-040a-498a-8c04-15b101fed76b"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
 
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
This should get updated for LHL

singhharsh1708:
Updated: with LHL the cutoff is now zero on adaptive steps too. Correction to what I wrote first: LHL already read the new γdt from `W.gamma` each solve; what lagged inside the cutoff was `W_γdt`, so the `dz` rescale in `compute_step!` was correcting for a staleness that was not there. Test added, fails on master.

PR #4404, opened 2026-08-27
Store the embedded error weights as the difference

### LINKED ISSUE #4403: Six adaptive methods cannot integrate u' = 1: embedded error estimate is not a difference
Six adaptive methods cannot integrate `u' = 1`.

```julia
using OrdinaryDiffEqLowOrderRK, OrdinaryDiffEqHighOrderRK, OrdinaryDiffEqSSPRK, OrdinaryDiffEqTsit5

prob = ODEProblem((u, p, t) -> one(u), 0.0, (0.0, 1.0))   # exact solution u(t) = t
for alg in (Tsit5(), DP5(), Alshina2(), Alshina3(), FRK65(), SIR54(), PFRK87(), SSPRK932())
    s = solve(prob, alg; abstol = 1e-6, reltol = 1e-6, maxiters = 100000)
    @show alg, s.retcode, s.stats.naccept
end
```

```
Tsit5      Success    5 steps      err 2.2e-16
DP5        Success    2 steps
Alshina2   MaxIters   99993 steps
Alshina3   MaxIters   99991 steps
FRK65      MaxIters   99985 steps
SIR54      MaxIters   99984 steps
PFRK87     MaxIters   99982 steps
SSPRK932   MaxIters   99991 steps
```

The step size collapses to about `1e-12` on the first step and never recovers. On `u' = -u` with `abstol = reltol = 1e-8`, `Alshina2` runs to `maxiters` at 999992 steps and returns `0.622` against an exact `0.368`.

## Cause

The error estimate is not `u - û`. A consistent embedded pair satisfies `sum(btilde) == 0`, since both methods integrate `u' = 1` exactly and the difference must cancel. Measured:

| method | sum(b) | sum(btilde) |
|---|---|---|
| Alshina2 | 1.0 | 1.0 |
| Alshina3 | 1.0 | 0.444444 |
| SIR54 | 1.0 | 7.854297 |

`Alshina2` is the clearest. `low_order_rk_perform_step.jl`:

```julia
u = uprev + dt * (b1 * k1 + b2 * k2)
utilde = dt * (b1tilde * k1)
```

with `b1 = 0.25`, `b2 = 0.75`, `b1tilde = 1.0`. So `utilde` is the embedded Euler step itself rather than the difference from the main method, and on `u' = 1` it evaluates to `dt` instead of `0`. `Alshina3` has the same shape (`utilde = dt * (b2tilde * k2)`).

`FRK65`, `SIR54` and `PFRK87` do build a full `dt * sum(btilde_i * k_i)`, so their structure is right and the stored coefficients are the problem. `SIR54`'s `7.854` is neither a difference (0) nor a set of embedded weights (1), so I do not think it is simply `b̂` stored in place of `b - b̂` there.

## Note on a candidate fix

For `Alshina2`, `utilde = dt * ((b1 - b1tilde) * k1 + b2 * k2)` turns `u' = 1` from `MaxIters` into `Success` in 2 steps, and `u' = -u` at `1e-8` into `Success` in 6177 steps with error `1.65e-9`. That is consistent with `b - b̂` for an Euler embedding, but I have not checked it against Alshina's paper, and I would rather not guess coefficients for the other five without the sources. Happy to do the work if someone can point me at the intended embedded pairs.

Separate from #4362, which is about methods that never set `EEst` at all. These do set it, from the wrong vector.

singhharsh1708:
Store the embedded error weights of Alshina2, FRK65 and PFRK87 as the difference `b - bhat` in the tableau, so the estimate sums to zero and scales like dt^(p+1); the step code is back to a plain `Σ btilde_i k_i`.
The old weights were the embedded solution's own, so the estimate was O(dt) and about 50x too large: Alshina2 needed 925740 steps at reltol 1e-6 on `prob_ode_linear` and returned no states at 1e-7; it now takes 688 and 2162.
PFRK87's differences are exact rationals from its published weights and it gains a `β13tilde = β13` term; Alshina2 gains `b2tilde`; FRK65's `β4`, `β5`, `β6` are frequency-fitted per step, so those three differences are formed in the step and only its constant weights moved into the tableau.
The Alshina2 regression test now solves at 1e-7, because its old 1e-6 bar was only reachable through that bug.
Fixes #4403 for these three; Alshina3 still needs its own tableau.


ChrisRackauckas:
No, define the weights so that they are already the difference. That is more numerically stable.

singhharsh1708:
Done: the tableaus now hold `b - bhat` (exact rationals for PFRK87 and FRK65, via `big` since two of them overflow Int64), and the step is back to `sum(btilde_i k_i)`. FRK65 keeps three runtime differences because `β4`..`β6` are fitted from `(ω dt)^2` each step.

PR #4421, opened 2026-08-28
Initialise FIRK dtprev so composite members can take a first step

singhharsh1708:
Initialise `dtprev = one(dt)` in the ten FIRK cache constructors so a Radau member of an AutoSwitch composite can take its first step; `alg_cache` receives `dt = 0` on adaptive solves and the only repair sat behind `integrator.iter == 1`, which is false when the member first activates mid solve. BDF already initialises it this way.

With `dtprev = 0` the dense extrapolant divides by it, the stage guesses come out `NaN`, Newton never converges, and `AutoVern7(RadauIIA5())` on ROBER hangs at 653241 steps with zero stiff steps accepted; it now takes 53. That hang is why #4420's own test passes `stiffalgfirst = true`.

`AdaptiveRadau` also needed its controller and order lookups to resolve the active sub-cache rather than the `CompositeCache`, a crash #4420 does not cover.

Standalone FIRK is unchanged (orders and step counts bit-identical); `ode_firk_tests.jl` 118/118. `AutoVern7(RadauIIA9)` is intermittently non-deterministic in a composite, before and after this change alike, roughly one run in fifteen, while standalone `RadauIIA9` is deterministic; that is pre-existing and needs its own issue, so the new test covers RadauIIA3, RadauIIA5 and AdaptiveRadau.

Stacked on #4420, which must merge first; together they finish #4364.

### DIFF
--- a/lib/OrdinaryDiffEqFIRK/src/alg_utils.jl
+++ b/lib/OrdinaryDiffEqFIRK/src/alg_utils.jl
@@ -21,6 +21,14 @@ alg_adaptive_order(alg::RadauIIA9) = 5
 
 get_current_alg_order(alg::AdaptiveRadau, cache) = cache.num_stages * 2 - 1
 get_current_adaptive_order(alg::AdaptiveRadau, cache) = cache.num_stages
+current_num_stages(cache) = cache.num_stages
+function current_num_stages(cache::CompositeCache)
+    return _eval_index(current_num_stages, cache.caches, cache.current)::Int
+end
+get_current_alg_order(alg::AdaptiveRadau, cache::CompositeCache) =
+    2 * current_num_stages(cache) - 1
+get_current_adaptive_order(alg::AdaptiveRadau, cache::CompositeCache) =
+    current_num_stages(cache)
 
 function has_stiff_interpolation(::Union{RadauIIA3, RadauIIA5, RadauIIA9, AdaptiveRadau})
     return true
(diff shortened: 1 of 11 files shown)


lib/OrdinaryDiffEqFIRK/src/controllers.jl
```diff
@@ -1,10 +1,15 @@
+function current_firk_cache(integrator)
+    cache = integrator.cache
+    return cache isa CompositeCache ? @inbounds(cache.caches[cache.current]) : cache
```
ChrisRackauckas: this won't be type stable.

singhharsh1708: Reworked: the controller body now runs through `_eval_index` on the active cache (same pattern as the alg lookups), and the order lookups do the same with `::Int`; `@inferred` check added. Stacked on #4420.

lib/OrdinaryDiffEqFIRK/src/alg_utils.jl
```diff
@@ -21,6 +21,10 @@ alg_adaptive_order(alg::RadauIIA9) = 5
 
 get_current_alg_order(alg::AdaptiveRadau, cache) = cache.num_stages * 2 - 1
 get_current_adaptive_order(alg::AdaptiveRadau, cache) = cache.num_stages
+get_current_alg_order(alg::AdaptiveRadau, cache::CompositeCache) =
+    _eval_index(c -> get_current_alg_order(alg, c), cache.caches, cache.current)::Int
```
ChrisRackauckas: no closure

singhharsh1708: Closures gone: the order lookups go through `current_num_stages(cache)` as a function reference, and the controller bodies are dispatched with their arguments through `_eval_index(f, caches, i, args...)`, a trailing-args extension of the Core helper in #4420.

PR #4430, opened 2026-08-29
Give each threaded complex stage its own JVP operator

singhharsh1708:
`AdaptiveRadau(threading = true)` with a matrix-free Jacobian is nondeterministic once the method has two complex stage pairs: ten solves of a 200-point heat equation at `min_order = max_order = 9` with `KrylovJL_GMRES()` gave ten different answers, spread 1.4e-3 apart at tolerance 1e-7, with one Julia thread they agree exactly.
Every `ComplexWOperator` is built from the same `W1.jacvec`, so the stage solves that `@threaded` runs in parallel all push through one `JVPCache` and race on its pushforward buffers and its product tally; the threaded `nf` came out 566 where the serial solve counts 163.
When threading is on and the Jacobian is a `JVPCache`, `build_complex_W` now gives each complex stage its own operator, `firk_new_J!` moves those operators with the real one, and the existing per-stage drain picks up their tallies, so ten threaded solves now agree to the bit and report the serial `nf`.
The new test checks the operators are distinct, that repeated threaded solves agree, and that the threaded and serial `nf` match; the FIRK Krylov file passes on four threads.

### DIFF
--- a/lib/OrdinaryDiffEqFIRK/Project.toml
+++ b/lib/OrdinaryDiffEqFIRK/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqFIRK"
 uuid = "5960d6e9-dd7a-4743-88e7-cf307b64f125"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "2.8.3"
+version = "2.8.4"
 
 [deps]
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
needs a rebase

singhharsh1708:
done

PR #4448, opened 2026-08-31
Copy the W prototype so integrators stop sharing it

singhharsh1708:
`build_J_W`'s `W_prototype` branch returns `W = f.W_prototype; J = f.jac_prototype` while the branch below it deepcopy's, so every integrator built from one `ODEFunction` holds the same two operator objects; the pair is now copied together, which keeps a `W_prototype` that references `jac_prototype` pointing at its own copy.
No wrong result demonstrates on the factorization path, because every Newton solve rewrites W before using it; the sharing is still a hazard for any path that reads W's contents outside a fresh update, and it made two integrators' behaviour depend on each other's update order.
Copying the pair widens what inference can prove about `W` and `J`, which made JET's typo mode report `is_varying_mm` in `do_newJW` as possibly undefined on the Julia 1 QA lane; the variable can never actually be undefined, since it is assigned and read under the same `isdae` test, so it is now initialised before the branch.
The new test builds two integrators from one function, asserts the operators are distinct, and steps both interleaved to their correct solutions.

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/Project.toml
+++ b/lib/OrdinaryDiffEqDifferentiation/Project.toml
@@ -1,5 +1,5 @@
 name = "OrdinaryDiffEqDifferentiation"
-version = "3.11.4"
+version = "3.11.5"
 uuid = "4302a76b-040a-498a-8c04-15b101fed76b"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "Yingbo Ma <mayingbo5@gmail.com>"]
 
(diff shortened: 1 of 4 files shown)


singhharsh1708:
Closing. `OrdinaryDiffEqAMF` builds its own `jac_prototype` and `W_prototype` operators and relies on the integrator updating those same objects, so copying them breaks it: its downgrade lane fails here with a `setfield!` type mismatch on the `InvertibleOperator` factorization and passes on master. The sharing this was meant to remove is load-bearing, and I never had a case where it produced a wrong answer.

Issue #1566, opened 2022-01-11
CuSparse Completion for ODEs


ChrisRackauckas:
I was trying to write a tutorial on using CUDA and CuSparse for PDEs but found that CuSparse was a bit too incomplete. The setup was fine:

```julia
using OrdinaryDiffEq, LinearAlgebra, SparseArrays, BenchmarkTools, LinearSolve

const N = 32
const xyd_brusselator = range(0,stop=1,length=N)
brusselator_f(x, y, t) = (((x-0.3)^2 + (y-0.6)^2) <= 0.1^2) * (t >= 1.1) * 5.
limit(a, N) = a == N+1 ? 1 : a == 0 ? N : a
kernel_u! = let N=N, xyd=xyd_brusselator, dx=step(xyd_brusselator)
  @inline function (du, u, A, B, α, II, I, t)
    i, j = Tuple(I)
    x = xyd[I[1]]
    y = xyd[I[2]]
    ip1 = limit(i+1, N); im1 = limit(i-1, N)
    jp1 = limit(j+1, N); jm1 = limit(j-1, N)
    du[II[i,j,1]] = α*(u[II[im1,j,1]] + u[II[ip1,j,1]] + u[II[i,jp1,1]] + u[II[i,jm1,1]] - 4u[II[i,j,1]]) +
    B + u[II[i,j,1]]^2*u[II[i,j,2]] - (A + 1)*u[II[i,j,1]] + brusselator_f(x, y, t)
  end
end
kernel_v! = let N=N, xyd=xyd_brusselator, dx=step(xyd_brusselator)
  @inline function (du, u, A, B, α, II, I, t)
    i, j = Tuple(I)
    ip1 = limit(i+1, N)
    im1 = limit(i-1, N)
    jp1 = limit(j+1, N)
    jm1 = limit(j-1, N)
    du[II[i,j,2]] = α*(u[II[im1,j,2]] + u[II[ip1,j,2]] + u[II[i,jp1,2]] + u[II[i,jm1,2]] - 4u[II[i,j,2]]) +
    A*u[II[i,j,1]] - u[II[i,j,1]]^2*u[II[i,j,2]]
  end
end
brusselator_2d = let N=N, xyd=xyd_brusselator, dx=step(xyd_brusselator)
  function (du, u, p, t)
    @inbounds begin
      ii1 = N^2
      ii2 = ii1+N^2
      ii3 = ii2+2(N^2)
      A = p[1]
[...]

PR #4452, opened 2026-09-01
Build W through the allocating path for a sparse GPU Jacobian with a scalar mass matrix

AJ0070:
`jacobian2W!` forms `W = J - M/dtgamma`. When the mass matrix is scalar (`I`, the default), it writes `W`'s diagonal in place, either by scalar indexing or by broadcasting into `@view(W[idxs])`. Neither works for a `CuSparseMatrix`: the view falls back to `getindex(W, i, j)`, so every implicit solver given a sparse GPU `jac_prototype` threw `Scalar indexing is disallowed` on the first step.

The allocating path for this already exists and is what `dae_jacobian2W!` uses, but in `jacobian2W!` it sits in an `elseif` that a scalar mass matrix never reaches. This adds the guard ahead of the diagonal write, using the same `copyto!(W, ...)` form. The CPU branches are untouched, and the dense `Matrix` method is not affected.

This unblocks the Brusselator from #1566. On a T4 that example now solves with a `CuSparseMatrixCSR` jac_prototype under Rosenbrock23, Rosenbrock23 with `KrylovJL_GMRES`, and FBDF, matching the CPU reference to about 1e-10. With this branch disabled all three fail with scalar indexing, so the change is required rather than incidental. That example also needs its RHS to broadcast over a device-resident index set, which is a property of the script rather than of the solver, and is noted on the issue.

Tests: a CPU unit test that the allocating formula agrees with the diagonal write, and that CPU sparse and dense still take the old branch; plus `test/gpu/sparse_jac_prototype.jl`, which `gpu_group()` picks up automatically. That GPU test passes with the fix and errors on all four solver configurations without it.

`jacobian2W!` is a hot path. The new condition is a branch ahead of the CPU code and should fold away for non-GPU arrays, but it is worth a second opinion.

AI Disclosure: Used Opus 5

### DIFF
--- a/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
+++ b/lib/OrdinaryDiffEqDifferentiation/src/derivative_utils.jl
@@ -645,7 +645,15 @@ function jacobian2W!(
         @boundscheck axes(mass_matrix) == axes(W) || _throwWMerror(W, mass_matrix)
     @inbounds begin
         invdtgamma = inv(dtgamma)
-        if _is_scalar_massmatrix(mass_matrix)
+        if _is_scalar_massmatrix(mass_matrix) && _use_allocating_sparse_W_path(W)
+            # A sparse GPU `W` can be neither scalar-indexed nor broadcast into on its
+            # diagonal, so build the shifted matrix and copy it in. Without this the
+            # branch below reaches `@view(W[idxs])`, which falls back to
+            # `getindex(W, i, j)` and scalar-indexes the device storage.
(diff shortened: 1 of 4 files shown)


singhharsh1708:
Does this hold when the prototype has a structural zero on the diagonal? `J + (λ * invdtgamma) * I` fills those in, so the sum has more nnz than the preallocated `W` and `copyto!` has nowhere to put them.

singhharsh1708 (approved):
Looks good to me, `copyto!` carries the structure so the filled-in diagonal is fine.

AJ0070:
@ChrisRackauckas

PR #4455, opened 2026-09-01
Carry the QNDF2 difference array across steps

### LINKED ISSUE #4332: QNDF2 and QBDF2 reject many more steps than expected
From Claude Opus 5 🤖 :

## Summary

`QNDF2` and `QBDF2` reject between 60% and 190% as many steps as they accept, on
problems as simple as scalar exponential decay. `QNDF1`, variable-order `QNDF` and
`FBDF` are unaffected on the same problems (0.1% to 14%).

The cause is that the local error estimate depends strongly on the ratio between the
current and the previous step size, rather than on the local error alone. There is no
step size at which the estimate settles near 1, so the controller oscillates between
"grow the step by the maximum factor" and "reject".

`QBDF2` is affected because it is an alias for `QNDF2` with `kappa = 0`
(`src/algorithms.jl:568`). `QNDF1`/`QBDF1` use a different code path and are correct.

## Environment

- Julia 1.12.7
- OrdinaryDiffEqBDF v2.4.3

---

## Reproducer

```julia
using OrdinaryDiffEqBDF, OrdinaryDiffEqNonlinearSolve

prob1 = ODEProblem((u, p, t) -> -u, 1.0, (0.0, 10.0))
prob2 = ODEProblem(
    (du, u, p, t) -> (du[1] = -1000u[1] + u[2]; du[2] = u[1] - u[2]; nothing),
    [1.0, 0.0], (0.0, 1.0)
)

for prob in (prob1, prob2), alg in (QNDF1(), QNDF2(), QBDF2(), QNDF())
    st = solve(prob, alg; reltol = 1e-8, abstol = 1e-8).stats
    println("$alg: accept=$(st.naccept) reject=$(st.nreject)")
end
```

Measured, with the final-time error against `FBDF` at `1e-13`:

| method | problem | tol | err | accepted | rejected | reject % |
|---|---|---|---|---|---|---|
| `QNDF1` | `u' = -u` | 1e-6 | 6.66e-6 | 1046 | 7 | 0.7% |
| `QNDF1` | `u' = -u` | 1e-8 | 7.11e-7 | 10352 | 7 | 0.1% |
| `QNDF2` | `u' = -u` | 1e-6 | 2.38e-6 | 401 | 308 | **76.8%** |
| `QNDF2` | `u' = -u` | 1e-8 | 3.60e-8 | 4211 | 6427 | **152.6%** |
| `QBDF2` | `u' = -u` | 1e-6 | 2.94e-6 | 560 | 442 | **78.9%** |
| `QBDF2` | `u' = -u` | 1e-8 | 8.71e-8 | 5992 | 7807 | **130.3%** |
| `QNDF` | `u' = -u` | 1e-6 | 1.14e-7 | 71 | 10 | 14.1% |
| `QNDF` | `u' = -u` | 1e-8 | 1.05e-9 | 129 | 7 | 5.4% |
| `QNDF1` | 2x2 linear stiff | 1e-8 | 9.30e-7 | 8905 | 8 | 0.1% |
| `QNDF2` | 2x2 linear stiff | 1e-8 | 3.97e-8 | 3415 | 2637 | **77.2%** |
| `QBDF2` | 2x2 linear stiff | 1e-8 | 1.16e-7 | 5491 | 10456 | **190.4%** |
| `QNDF` | 2x2 linear stiff | 1e-8 | 1.10e-8 | 168 | 11 | 6.5% |

The extra work does not buy accuracy. At `reltol = abstol = 1e-8` on the scalar
problem, `QNDF2` spends 10638 step attempts and delivers `3.6e-8`, while
variable-order `QNDF` spends 136 and delivers `1.05e-9`. Across the tolerance range
the achieved error is 2 to 20 times larger than the requested tolerance, and the
factor moves erratically (0.5, 0.8, 2.4, 3.0, 3.6, 20, 1.8 for tolerances `1e-4`
through `1e-10`).

## The error estimate tracks the step-size ratio

Instrumenting `perform_step!` to log `(ρ = dt/dtₙ₋₁, EEst)` for every step attempt on
a 25-state nonlinear ODE, at `reltol = abstol = 1e-8`:

| ρ | step attempts | median EEst | rejected |
|---|---|---|---|
| [0.0, 0.5) | 20 | 1.097 | 70.0% |
| [0.5, 0.9) | 87 | 0.258 | 6.9% |
| **[0.9, 1.1)** | **785** | **0.003** | **0.1%** |
| [1.1, 2.0) | 405 | 3.221 | 71.6% |
| [2.0, inf) | 1235 | 46.558 | **99.8%** |

At an unchanged step size the estimate claims the error is about 300 times below the
tolerance. The controller then grows the step by its maximum factor. The grown step
reports an estimate near 46 and is rejected. The step size collapses back to the
previous value and the cycle repeats. The first step attempts show the loop:

```
(ρ,     dtₙ₋₁/dtₙ₋₂, EEst)
(9.278, 1.0,        0.00435)
(10.0,  9.278,      7.856)     <- rejected
(3.211, 9.278,      0.7564)
(1.0,   3.211,      0.3223)
(1.585, 1.0,        0.1877)
(2.077, 1.585,      5.936)     <- rejected
```

There is no step size where the estimate settles near 1, so the controller cannot
converge.

## Scope: only the standalone fixed-order implementations

`QNDF1`/`QNDF2` (and their `QBDF1`/`QBDF2` aliases) are standalone implementations
with their own caches and `perform_step!` methods. There are no standalone order 3,
4 or 5 variants: `QNDF` takes `max_order`, and `QBDF` is `QNDF` with `kappa = 0`
(`src/algorithms.jl:575`), so every higher order goes through the variable-order code
path. All measurements below use the scalar problem at tolerances `1e-4` to `1e-10`,
with the row values taken at `1e-8`.

| method | reject % | work-precision slope | steps | err |
|---|---|---|---|---|
| `QNDF1()` | 0.1% | -0.96 | 10352 | 7.11e-7 |
| `QBDF1()` | 0.1% | -1.04 | 13092 | 6.97e-7 |
| **`QNDF2()`** | **152.6%** | **-1.69** | 4211 | 3.60e-8 |
| **`QBDF2()`** | **130.3%** | **-1.79** | 5992 | 8.71e-8 |
| `ABDF2()` | 0.2% | -2.01 | 941 | 1.49e-7 |
| `QNDF(max_order=Val{1}())` | 0.0% | -1.01 | 11596 | 3.49e-7 |
| **`QNDF(max_order=Val{2}())`** | **0.0%** | **-2.30** | **855** | 9.22e-9 |
| `QNDF(max_order=Val{3}())` | 0.8% | -2.90 | 264 | 1.52e-8 |
| `QNDF(max_order=Val{4}())` | 1.9% | -4.73 | 160 | 3.02e-9 |
| `QNDF(max_order=Val{5}())` | 5.4% | -6.95 | 129 | 1.05e-9 |
| `QBDF(max_order=Val{2}())` | 0.2% | -2.14 | 1068 | 5.53e-8 |
| `FBDF()` | 0.0% | -6.04 | 187 | 4.59e-10 |

The decisive row is the pair at order 2. The same order-2 NDF, reached through the
variable-order path, takes **855 steps with no rejections** and is 4 times more
accurate, where `QNDF2()` takes **4211 steps and rejects 6427 attempts**. `ABDF2`, an
independent order-2 BDF implementation, needs 941 steps and reports a clean slope of
-2.01, which confirms that about 900 steps is the right cost at this accuracy.

So the affected surface is exactly `QNDF2`, `QBDF2`, and to a much smaller degree
`QNDF1`/`QBDF1`. Every order reached through the variable-order path is unaffected.

**Workaround for users:** replace `QNDF2()` with `QNDF(max_order = Val{2}())`, or use
`ABDF2()`.

Note that the slopes for the capped variable-order runs are work-precision slopes
rather than literal convergence orders, because those runs still adapt the order below
the cap over a small number of steps. The comparison between the two order-2 rows is
the meaningful part.

## The formula is order 2; the adaptive path does not deliver it

The formula itself is correct. With `adaptive = false` on the scalar problem, the
measured orders are exactly nominal:

| method | errors at dt = 10*2^-k, k = 5..9 | orders |
|---|---|---|
| `QNDF1` | 5.78e-5, 2.57e-5, 1.20e-5, 5.80e-6, 2.85e-6 | 1.17, 1.10, 1.05, 1.03 |
| `QNDF2` | 1.73e-6, 4.71e-7, 1.18e-7, 2.92e-8, 7.27e-9 | 1.88, **2.00**, **2.01**, **2.01** |
| `QBDF2` | 1.22e-5, 2.67e-6, 6.21e-7, 1.50e-7, 3.68e-8 | 2.19, 2.10, 2.05, 2.03 |

Adaptively it is not delivered. Measuring the trajectory max-norm error against total
work (accepted plus rejected steps) over tolerances from `1e-4` to `1e-10`, and
reporting the slope between each consecutive pair of tolerances:

| method | consecutive work-precision slopes |
|---|---|
| **`QNDF2`** | **-1.19, -1.78, -1.10, -1.13, -1.36, -0.81** |
| **`QBDF2`** | **-1.44, -1.38, -1.92, -0.07, -1.50, -3.60** |
| `QNDF(max_order=Val{2}())` | -2.60, -3.01, -1.50, -2.34, -1.82, -2.00 |
| `ABDF2` | -2.29, -2.05, -2.03, -2.02, -2.01, **-2.01** |
| `QNDF1` | -1.05, -1.04, -1.01, -1.00, -1.00, **-1.00** |

`ABDF2` (fixed order 2, adaptive) settles on -2.01 and `QNDF1` (fixed order 1) on
-1.00, both monotonically, which shows the metric is sound for fixed-order methods.
Against those references, `QNDF2` sits near -1.2 and `QBDF2` swings between -0.07 and
-3.60. An order-2 formula is delivering closer to first-order work-precision, and
erratically.

The scatter in the `QNDF(max_order=Val{2}())` row is not a defect. That method holds
`dt` exactly constant on about 75% of its steps, against `ABDF2`'s 3%, and its
achieved error relative to the requested tolerance is non-monotone (0.84, 0.18, 0.92,
4.61 for `1e-6` through `1e-9`) where `ABDF2`'s is a smooth progression (3.05, 6.74,
14.9, 32.3). A consecutive-pair slope divides a log-error difference by a log-work
difference, so a non-monotone error term scatters the slope without saying anything
about the order. `QNDF2`, by contrast, is not scattering around -2; it sits near -1.2.

`QNDF` holds its step size because it sets `qsteady_min = 1, qsteady_max = 2`, while
the generic default is `[1, 1]`, which is no deadband
(`OrdinaryDiffEqCore/src/alg_utils.jl:833` and `:853`). The deadband applies to the
proposed step-size change factor `q`, where `dt_new = dt / q`
(`controllers.jl:689`), so it suppresses shrink requests up to 2x while every growth
request is applied. Whether that is what makes the achieved error non-monotone is a
hypothesis; it is not tested here, and note that err/tol is a global error over many
steps divided by a local per-step tolerance, so the numeric range of the deadband
does not bound it.

Two cautions on this metric. The error is a max-norm over the trajectory rather than
an endpoint value, because the signed endpoint error crosses zero within this
tolerance range for several of these methods and the cancellation distorts a log-log
fit. And it is meaningless for variable-order methods, where the order changes along
the sweep; `FBDF` and `QNDF(max_order=Val{5}())` fit slopes between -3.7 and -11.4
here, sometimes steeper than their own maximum order.

The achieved accuracy is also erratic. The raw sequence for `QNDF2` (accepted steps,
endpoint error) is:

```
(48, 5.3e-5) (125, 7.6e-6) (401, 2.4e-6) (1295, 3.0e-7)
(4211, 3.6e-8) (12648, 2.0e-8) (42134, 1.8e-10)
```

Each tolerance decade multiplies the step count by about 3.2, which for order 2 should
divide the error by about 10 each time. The observed factors are 7.0, 3.2, 8.0, 8.3,
1.8 and 111. So a user gets both a lower effective order and an error that does not
track the requested tolerance.

### 1. The rescaling on the hot path (`bdf_perform_step.jl:542` and `:654`)

`QNDF2` re-derives its backward differences from the raw solution history on every
step. When the previous two step sizes differ, it corrects for the change with an
ad-hoc first-order scaling:

```julia
ρ₁ = dt / dtₙ₋₁
ρ₂ = dt / dtₙ₋₂
D[1] = uprev - uprev2
D[1] = D[1] * ρ₁
D[2] = D[1] - ((uprev2 - uprev3) * ρ₂)
```

The error estimate is then a **third** difference built by subtracting these:

```julia
D2[1] = u - uprev
D2[2] = D2[1] - D[1]
D2[3] = D2[2] - D[2]
utilde = (κ * γ₂ + inv(k + 1)) * D2[3]
```

Any O(h^2) inaccuracy in the rescaled `D[2]` therefore swamps the true O(h^3) signal.
Branch instrumentation on the run above shows this path is taken on 2505 of 2533 step
attempts (the uniform-step branch is taken 26 times, the start-up branch twice).

### 2. An in-place 2x2 matrix product (same functions)

The branch that does use the exact rescaling overwrites `D[1]` before reading it:

```julia
if ρ != 1
    R!(k, ρ, cache)
    R .= R * U
    D[1] = D[1] * R[1, 1] + D[2] * R[2, 1]
    D[2] = D[1] * R[1, 2] + D[2] * R[2, 2]   # <- uses the NEW D[1]
end
```

This is present in both the out-of-place and the in-place version. It is
unambiguously wrong, but it fires only when `dtₙ₋₁ == dtₙ₋₂ && dt != dtₙ₋₁`, which
was 24 of 2533 step attempts. Correcting it alone gives bit-identical results, so it
is not the cause of the thrash.

## What variable-order QNDF does differently

`perform_step!` for `QNDFConstantCache`/`QNDFCache` (`bdf_perform_step.jl:774`) keeps
the difference array as state and transforms it exactly:

```julia
if cache.consfailcnt > 0
    # Deep copy to avoid aliasing: D[i] and prevD[i] must not share arrays
    D[i] = copy(cache.prevD[i])
end
if dt != dtprev || cache.prevorder != k
    ρ = dt / dtprev
    calc_R!(R, ρ, k)
    D_new = [sum(D[i] * RU[i, j] for i in 1:k) for j in 1:k]
    D[j] = D_new[j]
end
```

Three differences from `QNDF2`:

- It carries `D` across steps instead of re-deriving it from the solution history.
- It applies the exact `R*U` transformation on any change of step size, not only when
  the previous two steps were equal.
- It builds the result in a temporary, so there is no aliasing.
- It restores `prevD` after a failed step.

## What does not work as a fix

Substituting the exact `R*U` rescaling into the hot branch makes the solver diverge
(`maxiters` reached at 597330 accepted and 402670 rejected steps, final error 0.425).
That is expected in hindsight: the stored history `uprev`, `uprev2`, `uprev3` is not
uniformly spaced precisely when `dtₙ₋₁ != dtₙ₋₂`, so a single-ratio transformation
does not apply to differences formed from it.

A correct fix likely means giving `QNDF2` the same incremental difference array that
variable-order `QNDF` maintains, rather than re-deriving it each step. That is a
rewrite of the difference handling, not a local patch.

## Note

I have evidence for the mechanism (the ρ to EEst correlation, and the branch counts)
but no working fix, so the exact contribution of each code site is not proven by
repair.

singhharsh1708:
`QNDF2` rebuilt its backward differences from `uprev`, `uprev2` and `uprev3` on every step and corrected a change of step size with a first-order scaling, so an O(h^2) error sat inside a quantity whose signal is O(h^3) and the local error estimate tracked `dt/dtₙ₋₁` instead of the error. There was no step size at which the estimate settled near 1, so the controller alternated between growing by the maximum factor and rejecting.

`D` now stays at the step size its differences were formed with, a step that changes `dt` scales them through the exact `R * U` transform into locals, and a successful step commits the new differences incrementally, which is what variable-order `QNDF` already does. Reading `D` and writing locals also removes the aliasing in `D[2] = D[1] * R[1, 2] + D[2] * R[2, 2]`, which used the `D[1]` assigned on the line above; that was inert because `(R * U)[1, 2]` is zero for `k = 2`.

On `u' = -u` at `1e-8`, `QNDF2` goes from 4211 accepted and 6427 rejected to 765 and 2. On the issue's stiff 2x2 at `1e-8`, `QBDF2` goes from 5491 and 10456 to 923 and 23. On Robertson at `1e-8`, `QNDF2` goes from 5874 and 6115 to 1505 and 11, and the mass matrix form from 6214 and 11144 to 1504 and 9. Fixed-step convergence is unchanged at 1.88, 2.00, 2.01, 2.01.

The regression test asserts `QNDF2` and `QBDF2` reject fewer than a tenth of the steps they accept on `u' = -u` at `1e-8`; on master they reject more than they accept.

This still reproduced on master after #4328, which was suggested as the fix on the issue. Fixes #4332.

### DIFF
--- a/lib/OrdinaryDiffEqBDF/Project.toml
+++ b/lib/OrdinaryDiffEqBDF/Project.toml
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqBDF"
 uuid = "6ad6398a-0878-4a85-9266-38940aa047c8"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "2.4.6"
+version = "2.4.7"
 
 [deps]
 FastBroadcast = "7034ab61-46d4-4ed7-9d0f-46aef9175898"
--- a/lib/OrdinaryDiffEqBDF/src/bdf_caches.jl
+++ b/lib/OrdinaryDiffEqBDF/src/bdf_caches.jl
@@ -307,6 +307,7 @@ end
     uprev3::uType
     fsalfirst::rateType
     D::coefType1
+    Dtmp::coefType1
     D2::coefType2
     R::coefType
     U::coefType
@@ -359,12 +360,15 @@ function alg_cache(
     fsalfirst = zero(rate_prototype)
 
     D = Array{typeof(u)}(undef, 1, 2)
+    Dtmp = Array{typeof(u)}(undef, 1, 2)
     D2 = Array{typeof(u)}(undef, 1, 3)
     R = fill(zero(t), 2, 2)
     U = fill(zero(t), 2, 2)
 
     D[1] = zero(u)
     D[2] = zero(u)
+    Dtmp[1] = zero(u)
+    Dtmp[2] = zero(u)
     D2[1] = zero(u)
     D2[2] = zero(u)
     D2[3] = zero(u)
@@ -380,7 +384,7 @@ function alg_cache(
     dtₙ₋₂ = zero(dt)
 
     return QNDF2Cache(
-        uprev2, uprev3, fsalfirst, D, D2, R, U, atmp,
+        uprev2, uprev3, fsalfirst, D, Dtmp, D2, R, U, atmp,
         utilde, nlsolver, dtₙ₋₁, dtₙ₋₂, alg.step_limiter!
     )
 end
--- a/lib/OrdinaryDiffEqBDF/src/bdf_perform_step.jl
+++ b/lib/OrdinaryDiffEqBDF/src/bdf_perform_step.jl
@@ -547,41 +547,35 @@ function perform_step!(integrator, cache::QNDF2ConstantCache, repeat_step = fals
         κ = zero(alg.kappa)
         γ₁ = Int64(1) // 1
         γ₂ = Int64(1) // 1
-    elseif dtₙ₋₁ != dtₙ₋₂
-        κ = alg.kappa
-        γ₁ = Int64(1) // 1
-        γ₂ = Int64(1) // 1 + Int64(1) // 2
-        ρ₁ = dt / dtₙ₋₁
-        ρ₂ = dt / dtₙ₋₂
-        D[1] = uprev - uprev2
-        D[1] = D[1] * ρ₁
-        D[2] = D[1] - ((uprev2 - uprev3) * ρ₂)
     else
         κ = alg.kappa
         γ₁ = Int64(1) // 1
         γ₂ = Int64(1) // 1 + Int64(1) // 2
-        ρ = dt / dtₙ₋₁
-        # backward diff
-        D[1] = uprev - uprev2
-        D[2] = D[1] - (uprev2 - uprev3)
-        if ρ != 1
-            R!(k, ρ, cache)
-            R .= R * U
-            D[1] = D[1] * R[1, 1] + D[2] * R[2, 1]
-            D[2] = D[1] * R[1, 2] + D[2] * R[2, 2]
-        end
+    end
+
+    # `D` stays at the step size its differences were formed with, so a change of
+    # `dt` scales them through `R * U` instead of rebuilding them from the
+    # solution history, and a rejected attempt leaves `D` untouched.
+    if cnt > 2 && dt != dtₙ₋₁
+        R!(k, dt / dtₙ₋₁, cache)
+        R .= R * U
+        d₁ = D[1] * R[1, 1] + D[2] * R[2, 1]
+        d₂ = D[1] * R[1, 2] + D[2] * R[2, 2]
+    else
+        d₁ = D[1]
+        d₂ = D[2]
     end
 
     β₀ = inv((1 - κ) * γ₂)
     α₀ = 1
 
-    u₀ = uprev + D[1] + D[2]
-    ϕ = (γ₁ * D[1] + γ₂ * D[2]) * β₀
+    u₀ = uprev + d₁ + d₂
+    ϕ = (γ₁ * d₁ + γ₂ * d₂) * β₀
 
     markfirststage!(nlsolver)
 
     # initial guess
-    nlsolver.z = uprev + sum(D)
+    nlsolver.z = u₀
 
     mass_matrix = f.mass_matrix
 
@@ -612,8 +606,8 @@ function perform_step!(integrator, cache::QNDF2ConstantCache, repeat_step = fals
             OrdinaryDiffEqCore.set_EEst!(integrator, integrator.opts.internalnorm(atmp, t))
         else
             D2[1] = u - uprev
-            D2[2] = D2[1] - D[1]
-            D2[3] = D2[2] - D[2]
+            D2[2] = D2[1] - d₁
+            D2[3] = D2[2] - d₂
             utilde = (κ * γ₂ + inv(k + 1)) * D2[3]
             atmp = calculate_residuals(
                 utilde, uprev, u, integrator.opts.abstol,
@@ -627,6 +621,9 @@ function perform_step!(integrator, cache::QNDF2ConstantCache, repeat_step = fals
         return
     end
 
+    Δ = u - uprev
+    cache.D[1] = Δ
+    cache.D[2] = integrator.success_iter == 0 ? zero(Δ) : Δ - d₁
     cache.uprev3 = uprev2
     cache.uprev2 = uprev
     cache.dtₙ₋₂ = dtₙ₋₁
@@ -651,7 +648,7 @@ end
 
 function perform_step!(integrator, cache::QNDF2Cache, repeat_step = false)
     (; t, dt, uprev, u, f, p) = integrator
-    (; uprev2, uprev3, dtₙ₋₁, dtₙ₋₂, D, D2, R, U, utilde, atmp, nlsolver) = cache
+    (; uprev2, uprev3, dtₙ₋₁, dtₙ₋₂, D, Dtmp, D2, R, U, utilde, atmp, nlsolver) = cache
     (; z, tmp, ztmp) = nlsolver
     alg = unwrap_alg(integrator, true)
     cnt = integrator.iter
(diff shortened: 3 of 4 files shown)


ChrisRackauckas:
Supersceded and merged.

PR #4456, opened 2026-09-02
DAE interpolation after Rosenbrock solve: still some problems

Ickaser:
## Checklist

## Additional context

#4411 seems to improve the situation, but the tests added there were still not exhaustive enough: in my downstream use case (in LyoPronto.jl), I am still getting interpolations like the following:

plot_20

where the interpolation should go smoothly down to ~0 rather than having a step change.

This PR adds some dense interpolation tests, which currently fail for the in-place DAE with mass matrix (but pass for the out-of-place DAE and ODE).


ChrisRackauckas:
Fixed in #4466

PR #1581, opened 2022-01-22
Simplify no jac config logic

ChrisRackauckas:
Before:

```julia
using OrdinaryDiffEq, ParameterizedFunctions, BenchmarkTools
using LinearSolve

f = @ode_def Orego begin
  dy1 = p1*(y2+y1*(1-p2*y1-y2))
  dy2 = (y3-(1+y1)*y2)/p1
  dy3 = p3*(y1-y3)
end p1 p2 p3

p = [77.27,8.375e-6,0.161]
prob = ODEProblem(f,[1.0,2.0,3.0],(0.0,30.0),p)
solve(prob,Rosenbrock23(),abstol=1e-6,reltol=1e-6)

@btime solve(prob,Rosenbrock23(),abstol=1e-6,reltol=1e-6)
# 1.162 ms (14840 allocations: 1019.75 KiB)

@btime solve(prob,Rosenbrock23(linsolve=SimpleLUFactorization()),abstol=1e-6,reltol=1e-6)
# 1.248 ms (10431 allocations: 936.92 KiB)

@btime solve(prob,Rosenbrock23(linsolve=GenericLUFactorization()),abstol=1e-6,reltol=1e-6)
# 1.171 ms (14840 allocations: 1019.75 KiB)

@btime solve(prob,Rosenbrock23(linsolve=RFLUFactorization(),chunk_size = Val{3}()),abstol=1e-6,reltol=1e-6)
# 1.240 ms (10519 allocations: 951.98 KiB)


const k1=.35e0
const k2=.266e2
const k3=.123e5
const k4=.86e-3
const k5=.82e-3
const k6=.15e5
const k7=.13e-3
const k8=.24e5
const k9=.165e5
const k10=.9e4
const k11=.22e-1
const k12=.12e5
const k13=.188e1
const k14=.163e5
const k15=.48e7
const k16=.35e-3
const k17=.175e-1
const k18=.1e9
const k19=.444e12
const k20=.124e4
const k21=.21e1
const k22=.578e1
const k23=.474e-1
const k24=.178e4
const k25=.312e1

function pollu(dy,y,p,t)
 r1  = k1 *y[1]
 r2  = k2 *y[2]*y[4]
 r3  = k3 *y[5]*y[2]
 r4  = k4 *y[7]
 r5  = k5 *y[7]
 r6  = k6 *y[7]*y[6]
 r7  = k7 *y[9]
 r8  = k8 *y[9]*y[6]
 r9  = k9 *y[11]*y[2]
[...]

PR #4469, opened 2026-09-06
Resolve symbolic idxs in the ODE and DDE integrator interpolants

ChrisRackauckas-Claude:
## What changed and why

`integrator(t; idxs = sym)` only accepted integer component indices, so it threw
`ArgumentError: invalid index` for any symbolic index and could never resolve an observed
equation. `sol(t; idxs = sym)` has resolved both for a long time; this closes that gap in
the integrator interface, which came up in
https://github.com/SciML/ModelingToolkit.jl/issues/5006.

`(integrator::ODEIntegrator)(t; idxs)` and `(integrator::DDEIntegrator)(t; idxs)` now route
symbolic `idxs` to `SciMLBase.symbolic_interpolation`, which evaluates the current step's
interpolant and applies the symbolic getter to it. Integer, `nothing`, and in-place calls
are untouched and still go straight to `current_interpolant`.

`StochasticDiffEq` needs no change of its own — `SDEIntegrator` is an alias for
`ODEIntegrator{<:SDEAlgTypes}`, so it is covered by the `ODEIntegrator` method here.

`HistoryODEIntegrator` is deliberately **not** routed. It implements none of the
`SymbolicIndexingInterface` methods a `DEIntegrator` needs — it has no `p` field, so
`parameter_values` throws — which means symbolic indexing against it is unsupported both
before and after this PR. Making that work is a feature addition, not part of closing this
gap, so there is a comment in the test file saying so rather than a silent omission.

Depends on `SciMLBase.symbolic_interpolation` and `SciMLBase.has_symbolic_idxs`, both added
in https://github.com/SciML/SciMLBase.jl/pull/1581 and
https://github.com/SciML/SciMLBase.jl/pull/1585. The predicate lives in SciMLBase rather
than here because Sundials and ODEInterfaceDiffEq need the same decision and should not
reach into an OrdinaryDiffEqCore internal to get it. `symbolic_interpolation` shipped in
SciMLBase 3.52.0, but `has_symbolic_idxs` did not make that merge, so the compat floor in
OrdinaryDiffEqCore and DelayDiffEq is `3.53` and CI fails with
`UndefVarError: has_symbolic_idxs not defined` until #1585 is registered.

## Verification

New test files `test/Integrators_I/symbolic_interpolation_tests.jl` and
`lib/DelayDiffEq/test/interface/symbolic_interpolation.jl`. Both use a `SymbolCache`, whose
`Expr` indexes are observed quantities, so they cover the observed path without adding
ModelingToolkit to either test target.

Before (this branch's test file against unmodified `lib/`):

```
Test Summary:                                 | Pass  Fail  Error  Total  Time
Symbolic idxs in the current-step interpolant |    3     1      8     12  8.6s
ERROR: LoadError: Some tests did not pass: 3 passed, 1 failed, 8 errored, 0 broken.
```

After, and the same file for DelayDiffEq, both run against SciMLBase #1581:

```
DelayDiffEq    |    9      9  12.9s
OrdinaryDiffEq |   12     12   4.6s
```

End to end against the ModelingToolkit system from the issue, with `w ~ x + y + z`
observed — before, on OrdinaryDiffEq master:

```
FAIL  integrator(t; idxs = x)            ArgumentError: invalid index: x(t) of type SymbolicUtils...
FAIL  integrator(t; idxs = w)  # observed ArgumentError: invalid index: w(t) of type SymbolicUtils...
FAIL  integrator(ts; idxs = [x, w])      ArgumentError: invalid index: x(t) of type SymbolicUtils...
```

after:

```
PASS  integrator(t; idxs = x)
PASS  integrator(t; idxs = w)  # observed
PASS  integrator(ts; idxs = [x, w])
```

Also run locally: `GROUP=Integrators_I` (output below), Runic and `typos` on every file
touched.

```
Reinit Tests                            |   31     31  36.3s
Events Tests                            |   57     57   4m32.5s
Alg Events Tests                        |  177    177  14m18.0s
Discrete Callback Dual Tests            |    4      4   9.1s
Callback Allocation Tests               |    1      1   9.0s
Iterator Tests                          |   15     15  12.9s
Integrator Interface Tests              |   36     36  17.7s
Integrator Symbolic Interpolation Tests |   12     12   8.0s
Error Check Tests                       |   10  1 broken  11  26.8s
Event Detection Tests                   |   23     23   4m05.5s
Event Repetition Detection Tests        |    6      6   7.3s
Multi-VCC Mask Tests                    |   28     28  28.3s
Step Limiter Tests                      |  347    347  25m19.4s
Discontinuity Detection Tests           |   28     28   4m10.4s

     Testing OrdinaryDiffEq tests passed
```

(The single `broken` in Error Check Tests is a pre-existing `@test_broken` on master, not
introduced here. `SciMLBase` was `Pkg.develop`ed to the branch above for this run, since
the compat bound below requires a version that is not yet registered.)

## Things a reviewer should push back on

- A symbolic index evaluates the whole-state interpolant and then projects, while an
  integer index evaluates the interpolant for that component alone. The two agree to
  rounding, not bitwise, so the tests compare those two paths with `≈`. There is a comment
  in the test saying so.
- `has_symbolic_idxs` treats `Symbol` and `Expr` as symbolic, since that is what
  `SymbolicIndexingInterface.symbolic_type` reports. `integrator(t; idxs = :x)` previously
  threw and now resolves through the symbol cache. That is the intended widening, but it
  is a behavior change for anyone who was relying on the error.
- The in-place form `integrator(val, t; idxs)` is deliberately left alone: writing a
  symbolic projection into a caller-supplied buffer has no obvious shape contract. It
  still errors for symbolic `idxs`.

## Not verified

Other test groups, the GPU group, and downstream packages. This touches only the
`idxs`-dispatch of the integrator call, so the blast radius is the interpolation path that
`GROUP=Integrators_I` and `GROUP=InterfaceI` cover; CI runs the rest.

---

https://claude.ai/code/session_019wKpyZHy9u4QhR4ePiVmam


ChrisRackauckas-Claude:
CI is red across every job here for a single reason, and it is the expected one for a stacked PR rather than a defect in the change. From the Documentation job log:

```
ERROR: Unsatisfiable requirements detected for package SciMLBase [0bca4576]
```

This PR calls `SciMLBase.symbolic_interpolation` and raises its `SciMLBase` compat floor to `3.52`, which is not registered yet. Every job fails at resolution before running a line of test code.

The unblocking order is https://github.com/SciML/SciMLBase.jl/pull/1581 merged and registered as 3.52.0, then this. I will re-run CI here once that tag exists.

Locally, with SciMLBase `Pkg.develop`ed to that branch, `GROUP=Integrators_I` passes in full, including the new `Integrator Symbolic Interpolation Tests` at 12/12.

https://claude.ai/code/session_019wKpyZHy9u4QhR4ePiVmam

PR #4471, opened 2026-09-06
Accept a stage predictor on the Newton nonlinear solvers

### LINKED ISSUE #4331: Generalize `extrapolant/predictor` interface for next state prediction
It would be nice if I could provide my own say `extrapolant!(upred, u, p, t, dt)` instead of just the options `:constant` or `:linear` for the first guess in nonlinear implicit Newton solvers (or other iterative methods).

singhharsh1708:
`NLNewton` and `NonlinearSolveAlg` now take `predictor`, a callable giving the initial guess for the state at each implicit stage: `predictor(uprev, p, t, dt)` out of place and `predictor(upred, uprev, p, t, dt)` in place, with `t` the stage time and `dt` the time from `uprev` to it. It is applied once per stage in `nlsolve!` and mapped back to the stage unknown the way `compute_ustep` maps it forward, so every method that solves through these (SDIRK and ESDIRK, BDF, PDIRK, IMEX multistep, the implicit StochasticDiffEq solvers) gets it, and the SDIRK-only version is gone. A `DAEProblem` stage solve rejects it with an `ArgumentError`.

The new test runs ImplicitEuler, TRBDF2 and FBDF with both solvers, in and out of place: all 36 pass, and 12 fail with the call removed from `nlsolve!`.

On `u' = -50u` at 1e-8, passing the exact solution as the guess takes Kvaerno5 from 662 to 516 Newton iterations and in-place FBDF from 323 to 244, with the same accepted and rejected steps. A first-order Euler guess on Lotka-Volterra is slower than the methods' own predictors (TRBDF2 48057 to 55299, FBDF 8726 to 26165), so this is for guesses that know something about the problem rather than a better default.

Closes #4331.


ChrisRackauckas:
Why just SDIRK? This should generally be for any implicit method? And it should be done in the newton setup not in the algorithm, given all newton's can do this?

singhharsh1708:
Moved it into `NLNewton` and `NonlinearSolveAlg`, applied once per stage in `nlsolve!`, so any implicit method solving through them gets it; the SDIRK-specific version is gone.

PR #4489, opened 2026-09-09
Add optional MOOSE time_filter to NordsieckBDF

ChrisRackauckas-Claude:
## Summary
- Adds `NordsieckBDF(time_filter = true)` using the same MOOSE candidates as FBDF (order `k+1` filter + BDF3-Stab at order 3), after the Nordsieck corrector.
- Commits the filtered solution through an updated `acor` so `nordsieck_complete!` keeps the history array consistent; keeps a short rolling `(t, u)` history beside the Nordsieck array for the divided-difference filters.
- Default remains `false`. Identity mass matrix required when enabled. Tests cover accuracy, stiff problems, mass-matrix rejection, `max_order`, and fixed-step oscillatory behavior.

## Sharing / QNDF
FBDF and NordsieckBDF now share the low-level MOOSE primitives in `bdf_utils.jl` (`backdiff!`, `_filt_*`, `_fbdf_filter_weight`, `_fbdf_select_filter!`). The step wiring is method-specific: FBDF already owns `(t, u)` history, while Nordsieck needs the parallel history + `acor` commit path.

QNDF does **not** share that perform_step/controller path (it advances a backward-difference `D` array with NDF κ). Giving QNDF the same switch is reasonable for API uniformity later, but it is not a free enable — it needs its own history (or reconstruction from `D`) and the filter weight should be adapted for NDF leading coefficients rather than FBDF's fixed BDF coeffs. Adaptive benchmarks so far do not show a clear win for the filter, so I would leave QNDF for a follow-up unless we want opt-in parity now.

## Test plan
- [x] `nordsieck_time_filter_tests.jl` (iip/oop accuracy, stiff ROBER, mass-matrix rejection, max_order, fixed-step oscillator)
- [x] Existing `nordsieck_tests.jl`
- [ ] CI Core / OrdinaryDiffEqBDF
- [ ] Optional: adaptive WPD ROBER/Orego/HIRES for `NordsieckBDF` ± `time_filter` (local runs: fixed-step oscillatory helps; adaptive stiff suite is mixed / often more `nf`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lib/OrdinaryDiffEqBDF/src/alg_utils.jl
+++ b/lib/OrdinaryDiffEqBDF/src/alg_utils.jl
@@ -52,8 +52,12 @@ has_stiff_interpolation(::Union{QNDF, FBDF, DFBDF}) = true
 alg_order(alg::NordsieckBDF) = 1  # dummy: the running order lives in the cache
 alg_order(alg::DNordsieckBDF) = 1
 isadaptive(alg::DNordsieckBDF) = true
-get_current_alg_order(alg::NordsieckBDFAlgs, cache) = cache.order
-get_current_adaptive_order(alg::NordsieckBDFAlgs, cache) = cache.order
+get_current_alg_order(alg::NordsieckBDFAlgs, cache) =
+    hasproperty(cache, :filter_order) && cache.filter_order > 0 ? cache.filter_order :
+    cache.order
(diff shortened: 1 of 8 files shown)


ChrisRackauckas:
Users with fixed time step should just use FBDF

PR #4559, opened 2026-09-13
Fix NordsieckBDF trial-step interpolation for delay equations

svenzenker:
`MethodOfSteps(NordsieckBDF())` can return success while integrating against stale dense-history data. On the analytic DDE `u'(t) = -exp(-0.2) * u(t-0.2)`, with history `exp(-t)`, the unpatched solver has about 1.8e-3 solution error and 1.6e-2 internal history error despite `reltol=1e-9`, `abstol=1e-10`.

DelayDiffEq copies the trial step's interpolation coefficients immediately after `perform_step!`, before the acceptance controller runs. NordsieckBDF currently only publishes its corrected adaptive-step coefficients in `step_accept_controller!`. Consequently, DelayDiffEq associates the previous polynomial with the new interval and feeds it into subsequent delayed RHS evaluations.

This change publishes the corrected trial polynomial (`zn[j] + l[j] * acor`) at the end of successful adaptive `perform_step!` calls, for both mutable and constant caches. It does not commit or otherwise mutate `zn`: acceptance still owns the history update, and rejection can still undo the prediction. Fixed-step behavior and the acceptance-time polynomial refresh are retained.

The regression test uses an analytic DDE with and without a continuous callback, for in-place and out-of-place RHS functions. It checks the returned solution, the separate internal history solution, and event timing; the initial step also forces rejection. The test is registered in DelayDiffEq's Integrators group.

Validation on Julia 1.13.0:

- New DDE regression: 20/20 checks pass. Against the project's unpatched BDF 2.4.9 environment, 10/20 fail, including solution/history accuracy and callback timing.
- Existing `OrdinaryDiffEqBDF/test/nordsieck_tests.jl`: 134/134 checks pass, covering ODE and DAE behavior.
- External four-state tube DDE, 13 parameter cases: all state, pressure, flow and event agreement gates pass against Tsit5. Baseline maximum sampled state error decreases from 0.448036529 ml to 3.40819743e-5 ml; accepted steps decrease from 14,114 to 4,115, and rejected steps from 2,223 to 412. The results match the earlier diagnostic history-synchronization workaround.
- Two tube cases retain small conservation failures (approximately 1.57e-5 and 5.25e-6 ml against a 1e-6 ml gate); this PR does not claim to resolve those separate residual errors.
- `git diff --check` passes. JuliaFormatter was run, with unrelated formatting changes excluded from the patch.

The complete monorepo test suite and within-step delay fixed-point iteration were not validated locally; the regression uses steps shorter than the delay.

***Created using OpenAI Codex***


ChrisRackauckas-Claude:
Superseded by #4575, which keeps the same diagnosis but generates the corrected trial interpolant on demand in `_ode_addsteps!` rather than publishing it eagerly in `perform_step!`. Two consequences: evaluation-only callers (`integrator(t)`, saved-solution interpolation) keep seeing the committed polynomial instead of trial data, and the `lag < dt` fixed-point path is fixed too — DelayDiffEq re-invokes `perform_step!` per iteration, so `nordsieck_predict!` now skips re-shifting while `zn` is already in the predicted state (measured error ~0.1 with the eager publish vs ~8e-9 with the guard + on-demand interpolant). Kept your authorship credit via Co-Authored-By on the new commit. Thanks for identifying this!

ChrisRackauckas:
I merged the corrected version.

PR #4569, opened 2026-09-15
Cache SDC tableau construction

singhharsh1708:
Cache SDC tableaus per configuration, so the BigFloat coefficients are built once per element type, node count, node type, quadrature and sweeper instead of on every `init`.
Building them was most of the cost of a small solve: 2.1 ms and 4.4 MiB per `init` with 5 nodes.
A 2-unknown solve now takes 0.015 ms instead of 0.60 ms with 3 nodes and 0.033 ms instead of 5.9 ms with 8 nodes; a repeat lookup is 125 ns and 704 bytes.
Each call gets its own copy of the cached arrays, since a solver's tableau can be modified in place (the convergence negative controls do), and the cache sits behind a lock so threaded ensembles can fill it.
Follows up the BigFloat comment in the review of #4208.


lib/OrdinaryDiffEqSDC/src/sdc_tableaus.jl
```diff
@@ -279,15 +279,34 @@ function sdc_qdelta(
...
+Every coefficient array the sweep needs, in the element type `T`. The coefficients are
+built once per configuration and each call returns its own copy of them.
 """
 function SDCTableau(
         ::Type{T}, M::Int, node_type::SDCNodes.T, quad_type::SDCQuadrature.T,
         sweeper::SDCSweeper.T
     ) where {T}
+    key = (T, M, node_type, quad_type, sweeper)
+    tab = lock(SDCTableauCacheLock) do
+        get!(SDCTableauCache, key) do
```
ChrisRackauckas, line 299: don't you only need to lock to generate?

singhharsh1708: Dict rehashes on insert, so an unlocked lookup can read the table mid-move. It is 25ns of a 427ns cache hit, once per cache build.

PR #4582, opened 2026-09-17
Read a scalar mass matrix instead of its missing diagonal

singhharsh1708:
Explicit-first-stage SDIRK methods take a scalar mass matrix again: the FSAL conversion added in #4560 reads `diag(M)`, which neither `λ*I` nor `ScalarOperator(λ)` has, so every solve with one threw `MethodError: no method matching diag(...)`.
`_mmdiag` now reads the scalar through `_is_scalar_massmatrix` and `_scalar_massmatrix_λ`, the pair `build_J_W` already uses, at the start of the step for the FSAL it consumes and at the end for the FSAL it produces, which is what a `ScalarOperator` carrying an `update_func` needs.
`lib/OrdinaryDiffEqDifferentiation/test/scalar_operator_massmatrix_tests.jl`, added with #3916, goes from 5 passed and 1 error to 9 passed. It fails the same way on current versions, and it is what the red `downgrade-sublibraries (lib/OrdinaryDiffEqDifferentiation)` lane reports.
SDIRK's own DAE tests pass 76 of 76, and SDIRK's compat now requires Differentiation 3.12.4, the first release with the four-argument `_scalar_massmatrix_λ`.


lib/OrdinaryDiffEqSDIRK/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "OrdinaryDiffEqSDIRK"
 uuid = "2d112036-d095-4a1e-ab9a-08536f3ecdbf"
 authors = ["ParamThakkar123 <paramthakkar864@gmail.com>"]
-version = "2.9.6"
+version = "2.9.7"
```
ChrisRackauckas, line 4: missing a lower bound bump

singhharsh1708: Raised to 3.12.4, where the four-argument method lands.

